PR #372, opened 2026-07-04
Harden ACP offering for marketplace listing

DROOdotFOO:
Two fixes so the Xochi Cross-Chain Transfer offering lists cleanly on Virtuals ACP:

- **`required_funds: true -> false`** in `offering_metadata`. Leftover from the escrow/FundTransfer model. In the relay model the transfer moves off-escrow via the buyer's own signed intent, so raxol never takes the buyer's funds -- the offering requires no fund deposit (only the storefront fee is paid). Maps to the marketplace "Require Funds" toggle = off.
- **Accept `signed_intent` as a JSON string OR a map.** The marketplace field builder may deliver the nested bundle as a JSON string (flat-field schemas) rather than an object. `Offering.decode_signed_intent/1` normalizes both; `valid_requirement?` and the `Settler` use it, so the offering works regardless of how the marketplace serializes the field.

## Test status
- raxol_acp: 10 properties / 661 tests / 0 failures; `--warnings-as-errors` + format clean
- New: JSON-string `signed_intent` accepted (`valid_requirement?` + Settler relay), malformed/partial rejected; `required_funds` assertions updated

## Manual testing
- [x] Offline suite green; example still emits a conforming requirement
- [ ] List the job on Virtuals with `signed_intent` as a flat/string field and confirm a real buyer submission relays

### DIFF
--- a/packages/raxol_acp/lib/raxol/acp/xochi/offering.ex
+++ b/packages/raxol_acp/lib/raxol/acp/xochi/offering.ex
@@ -60,9 +60,10 @@ defmodule Raxol.ACP.Xochi.Offering do
       display_name: "Xochi Cross-Chain Transfer",
       description:
         "Cross-chain stablecoin settlement storefront. The buyer signs a Xochi " <>
-          "intent, the storefront relays it and returns the settlement tx hashes; " <>
-          "the buyer escrows only the storefront fee (a plain job, no fund hook).",
-      required_funds: true,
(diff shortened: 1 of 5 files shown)


DROOdotFOO:
Reverting: proportional pricing on Virtuals requires require_funds=true, so required_funds must stay true (master already has it). The signed_intent string-acceptance isn't needed since the field builder supports nested objects for requirements.

PR #423, opened 2026-07-10
Verify ZKSAR attestations before they buy trust

### LINKED ISSUE #337: Wire ZKSAR attestation verification into the trust pipeline
## Summary

`Raxol.Payments.Zksar.verify/2` / `verify_batch/2` (added in the origin-pull/signing hardening pass) perform real attestation signature recovery and fail closed, but they have **zero production callers**. The live trust pipeline still credits **unverified, self-asserted** attestations:

- `Raxol.Payments.Router.maybe_compute_trust_score/1` (`router.ex:127`) calls `Zksar.TrustScore.aggregate(attestations)` directly on caller-supplied maps.
- `Raxol.Payments.PrivacyTier` gates tier escalation/requirements on a caller-settable `:valid` boolean (`privacy_tier.ex:114`).

So a caller (or prompt-injected agent assembling Router opts) can pass `attestations: [%{type: :non_membership, valid: true}, %{type: :compliance, valid: true}]` — no `signature`, no `issuer` — and obtain max trust score → `:sovereign`/`:shielded` tier → lowest fee tier. The signature verification added in the branch never executes.

Found during the `xochi-origin-pull-signing-hardening` security audit (finding A1).

## Why this is deferred (pinned)

ZKSAR is not yet validated end-to-end (the attestation issuance/oracle side hasn't been tested). The immediate goal is to verify Raxol can settle **stealth** and **public** addresses across our chains, which does not depend on ZKSAR. Wiring verification in now would fail-close the attestation path before there's a tested issuer to produce real proofs. Tracking here so it isn't lost.

## Recommended fix (when ZKSAR is ready)

1. In `Router`, verify attestations before they buy any trust, on **every** path (computed-score AND explicit-trust_score, since on the explicit path a fake `valid: true` proof could satisfy a tier requirement and dodge a downgrade):

   ```elixir
   defp maybe_compute_trust_score(opts) do
     opts = verify_attestations(opts)   # replace :attestations with the verified subset
     # ... existing score logic, now over verified proofs ...
   end

   defp verify_attestations(opts) do
     case Keyword.get(opts, :attestations) do
       list when is_list(list) and list != [] ->
         {verified, _rejected} = Zksar.verify_batch(list, allowed_issuers: configured_issuers())
         Keyword.put(opts, :attestations, verified)
       _ -> opts
     end
   end
   ```

   `verify_batch` already returns `{verified, errors}` and fails closed when `:allowed_issuers` is empty (`:issuer_required`).

2. Pin trusted issuers via config (`config :raxol_payments, :zksar_allowed_issuers, ["0x..."]`); unset = no attestation buys trust.

3. Stop `PrivacyTier` trusting a caller-supplied `:valid`; only verification should set it. Document that `PrivacyTier` consumes pre-verified attestations (the `verify/2` output already carries `valid: true`).

4. Add fail-closed regression tests: unsigned/self-asserted attestations → trust score 0 → `x402`/`public`; attestations from a non-allowlisted issuer → 0.

5. (Related, lower severity) bind a per-use nonce / intent id and a chain id into `attestation_digest/1` so a valid attestation can't be replayed across settlements until expiry; enforce low-`s` in `recover_signer`.

## Refs

- `packages/raxol_payments/lib/raxol/payments/zksar.ex` — `verify/2`, `verify_batch/2`, `attestation_digest/1`
- `packages/raxol_payments/lib/raxol/payments/router.ex` — `maybe_compute_trust_score/1`
- `packages/raxol_payments/lib/raxol/payments/privacy_tier.ex` — `extract_attestation_types/1`

Severity: high (once ZKSAR is in production use); currently latent because attestations aren't issued yet.

DROOdotFOO:
## What

Closes #337. Wires `Raxol.Payments.Zksar.verify_batch/2` into the live trust
pipeline so attestations are cryptographically verified against a pinned issuer
allowlist BEFORE they buy any trust. Previously the verification code existed but
had **zero production callers**, so `Router` credited unverified, self-asserted
attestations.

## The vulnerability (HIGH)

A caller -- or a prompt-injected agent assembling `Router` opts -- could pass:

    Router.select(attestations: [%{type: :compliance, valid: true},
                                 %{type: :non_membership, valid: true}])

with no signature and no issuer, and obtain max trust score -> `:sovereign` /
`:shielded` tier -> the lowest fee tier. `Router.maybe_compute_trust_score/1`
called `TrustScore.aggregate/1` directly on caller maps, and `PrivacyTier` gated
tiers on a caller-settable `:valid` boolean. The signature verification added in
the origin-pull hardening pass never executed.

## Why it's safe to wire now (the deferral was stale)

The issue was pinned "deferred until a tested ZKSAR issuer exists." Verified
against the code, that rationale no longer holds: `Zksar.verify_batch` has zero
production callers, every live/E2E/property flow resolves trust via an explicit
`trust_score` (never attestations), and only self-asserted *test* fixtures
exercised the attestation path. Wiring verification is therefore pure hardening
with no functional regression -- and since no issuer is pinned by default, the
fail-closed default is strictly safer than crediting fakes.

## The fix (fail closed)

- `Router.maybe_compute_trust_score/1` now verifies attestations up front and
  replaces `:attestations` with only the verified subset -- so nothing unverified
  reaches EITHER the score aggregation OR the `PrivacyTier` tier-requirement
  check. This runs on every path (computed-score and explicit-`trust_score`).
- Issuers are pinned via `config :raxol_payments, :zksar_allowed_issuers, [...]`.
  Unset (the default) => no attestation can be verified => none buys trust.
- Callers now supply *raw, signed* proofs (`type_code`/`signature`/`issuer`/...);
  `verify_batch` requires the signature recover to an allowlisted issuer.
- `PrivacyTier` documents that it consumes pre-verified attestations (Router is
  the verification choke point); a self-asserted `%{valid: true}` map is not a
  proof.

An explicit `:trust_score` integer is a separate, caller-owned hint (clamped),
unchanged by this PR.

## Tests (RED-proven; real ExSecp256k1 signatures, no mocks)

`router_test.exs` reworked (now `async: false` -- it sets the allowlist app env):

- Positive: real signed proofs from the allowlisted issuer compute a score and
  route to `:xochi`; a verified compliance+non_membership pair satisfies the
  sovereign requirement; a verified proof that misses a tier's required types
  still forces a downgrade.
- Fail closed (each was RED before the fix): a self-asserted proof (no signature)
  buys nothing (`:x402` / score 0 / `:public`); a proof from a non-allowlisted
  issuer buys nothing; no configured allowlist means no attestation buys trust;
  unverified attestations are dropped, not credited.

`privacy_tier_test` / `trust_score_test` are unchanged -- they exercise the
pre-verified contract (feeding verified-shaped maps directly), which is exactly
what Router now guarantees.

## Manual testing

- [x] `cd packages/raxol_payments && MIX_ENV=test mix test` -> 1013 passed, 0
      failures.
- [x] Scoped `mix format` + `--check-formatted` on changed files -> clean.
- [x] `MIX_ENV=test mix compile --force` -> 0 warnings from the changed files.
- [x] Confirmed the self-asserted attestation fixtures fail RED before the fix,
      then pass GREEN after (fail-closed).

## Not in this PR (issue point 5, lower severity)

- Binding a per-use nonce / chain id / intent id into `attestation_digest/1` to
  prevent cross-settlement replay of a valid attestation until expiry. This
  changes the signed message and must be coordinated with the issuer
  (cross-repo), so it is a separate change.
- Enforcing low-`s` canonicalization in `recover_signer`.

## CI note

Root CI does not build `raxol_payments` -- the package suite
(`cd packages/raxol_payments && MIX_ENV=test mix test`) is the gate, and it is
green.

### DIFF
--- a/packages/raxol_payments/lib/raxol/payments/privacy_tier.ex
+++ b/packages/raxol_payments/lib/raxol/payments/privacy_tier.ex
@@ -6,6 +6,16 @@ defmodule Raxol.Payments.PrivacyTier do
   unlock deeper privacy at lower fees. Users can always opt into less
   privacy than their score allows.
 
+  ## Attestations are pre-verified
+
+  The `:attestations` this module reads for tier-requirement gating must already
+  be **verified** -- the output of `Raxol.Payments.Zksar.verify/2` (which carries
+  `valid: true`). `Raxol.Payments.Router` is the verification choke point: it runs
+  `Zksar.verify_batch/2` against the operator-pinned issuer allowlist and passes
+  only the verified subset here. Do not call `from_trust_score/2` with raw,
+  caller-supplied attestations expecting them to be trusted; a self-asserted
+  `%{valid: true}` map is not a proof.
+
   ## Tiers
 
   | Score | Tier      | Fee (bps) | Settlement |
@@ -101,6 +111,10 @@ defmodule Raxol.Payments.PrivacyTier do
 
   # -- Private --
 
+  # Consumes PRE-VERIFIED attestations (Zksar.verify output; Router verifies
+  # before calling here). The `valid: true` filter is a guard, not the trust
+  # decision -- a raw caller-supplied `valid: true` is meaningless here because it
+  # never passed through the Router's `Zksar.verify_batch/2`. See the moduledoc.
   defp extract_attestation_types(opts) do
     case Keyword.get(opts, :attestations) do
       nil ->
--- a/packages/raxol_payments/lib/raxol/payments/router.ex
+++ b/packages/raxol_payments/lib/raxol/payments/router.ex
@@ -34,7 +34,24 @@ defmodule Raxol.Payments.Router do
   - `:privacy` -- `:public`, `:stealth`, `:shielded`, or `:auto` (default: `:auto`)
   - `:protocol` -- force a specific protocol (overrides routing)
   - `:trust_score` -- trust score for privacy tier resolution (used by `settlement_for/1`)
-  - `:attestations` -- list of verified ZKSAR proofs (computes trust score if `:trust_score` absent)
+  - `:attestations` -- list of raw ZKSAR proofs (`type_code`/`signature`/`issuer`/...).
+    They are verified here (see below) before buying any trust; unverified proofs
+    are dropped. Computes the trust score when `:trust_score` is absent.
+
+  ## Attestation trust (fail closed)
+
+  Callers supply *raw, signed* attestations; the router verifies each via
+  `Raxol.Payments.Zksar.verify_batch/2` against the operator-pinned issuer
+  allowlist BEFORE any proof counts toward the trust score or a privacy-tier
+  requirement. This runs on every path (computed-score and explicit-`trust_score`),
+  so a self-asserted `%{valid: true}` map with no signature -- or a proof from a
+  non-allowlisted issuer -- buys nothing.
+
+      config :raxol_payments, :zksar_allowed_issuers, ["0xoracle..."]
+
+  When the allowlist is unset (the default), no attestation can be verified, so
+  none buys trust. An explicit `:trust_score` integer is a separate, caller-owned
+  hint and is clamped, not attestation-derived.
   """
   @spec select(keyword()) :: atom()
   def select(opts \\ []) do
@@ -122,6 +139,11 @@ defmodule Raxol.Payments.Router do
   @max_trust_score 100
 
   defp maybe_compute_trust_score(opts) do
+    # Verify attestations up front, on every path: the verified subset replaces
+    # the caller-supplied list so nothing unverified reaches the score
+    # aggregation OR the PrivacyTier tier-requirement check downstream.
+    opts = replace_with_verified_attestations(opts)
+
     case {Keyword.get(opts, :trust_score), Keyword.get(opts, :attestations)} do
       {nil, attestations} when is_list(attestations) and attestations != [] ->
         score = Zksar.TrustScore.aggregate(attestations)
@@ -133,6 +155,39 @@ defmodule Raxol.Payments.Router do
       _ ->
         opts
     end
+  end
+
+  # Replace `:attestations` with only the proofs that recover to an allowlisted
+  # issuer. A self-asserted map (no signature / not a raw proof) or a proof from
+  # an unpinned issuer verifies to nothing, so it cannot buy trust -- fail closed.
+  defp replace_with_verified_attestations(opts) do
+    case Keyword.get(opts, :attestations) do
+      list when is_list(list) and list != [] ->
+        Keyword.put(opts, :attestations, verify_attestations(list))
+
+      _ ->
+        opts
+    end
+  end
+
+  defp verify_attestations(attestations) do
+    case allowed_issuers() do
+      [] ->
+        # No trusted-issuer allowlist configured: trust cannot be established, so
+        # no attestation is credited (rather than trusting a caller's `valid`).
+        []
+
+      issuers ->
+        {verified, _errors} = Zksar.verify_batch(attestations, allowed_issuers: issuers)
+        verified
+    end
+  end
+
+  defp allowed_issuers do
+    :raxol_payments
+    |> Application.get_env(:zksar_allowed_issuers, [])
+    |> List.wrap()
+    |> Enum.filter(&(is_binary(&1) and &1 != ""))
   end
 
   # A trust score is a non-negative 0..100 value, but callers can pass any
--- a/packages/raxol_payments/test/raxol/payments/router_test.exs
+++ b/packages/raxol_payments/test/raxol/payments/router_test.exs
@@ -1,7 +1,73 @@
 defmodule Raxol.Payments.RouterTest do
-  use ExUnit.Case, async: true
+  # async: false -- the attestation tests set the :zksar_allowed_issuers app env.
+  use ExUnit.Case, async: false
 
   alias Raxol.Payments.Router
+  alias Raxol.Payments.Zksar
+
+  # anvil account 0/1 keys -- real secp256k1 scalars. account 0's address is the
+  # trusted attestation issuer; account 1 signs "wrong issuer" proofs.
+  @issuer_key Base.decode16!(
+                "AC0974BEC39A17E36BA4A6B4D238FF944BACB478CBED5EFCAE784D7BF4F2FF80",
+                case: :mixed
+              )
+  @rogue_key Base.decode16!(
+               "59C6995E998F97A5A0044966F0945389DC9E86DAE88C7A8412F4603B6B78690D",
+               case: :mixed
+             )
+
+  @issuer (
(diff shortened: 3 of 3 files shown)


DROOdotFOO:
Waiting until completed on the Xochi end and we wire providers.

DROOdotFOO (commented):

Scope: the ZKSAR verify -> trust-score -> privacy-tier/fee pipeline wired by this PR
(`Router`, `Zksar`, `Zksar.TrustScore`, `PrivacyTier`). Closing the attestation-forgery
path (raw self-asserted `%{valid: true}` maps buying trust) is a real, correct improvement
and the fail-closed default is sound. The findings below are the paths that survive it. All
three "HIGH/MEDIUM" claims were reproduced empirically against the PR branch.

### HIGH

- **[New Hire / Saboteur] HIGH** `router.ex:152-153` + `privacy_tier.ex:134` -- The new
  verification choke point is bypassable by the sibling `:trust_score` channel, which the
  PR's own threat model treats as caller/prompt-injection controlled. Passing
  `Router.settlement_for(trust_score: 100)` with **no** `:attestations` returns
  `:sovereign` / `:shielded` (lowest 15bps fee) -- reproduced. The attestation-requirement
  gate (`sovereign` needs `compliance` + `non_membership`) is silently skipped because
  `extract_attestation_types/1` returns `nil` when `:attestations` is absent, and
  `maybe_downgrade_for_attestations(tier, nil)` returns the tier unchanged. So the hard path
  (forge a signed attestation) is now closed while the trivial path (assert `trust_score: 100`,
  omit attestations) buys the same max trust AND skips the type-requirement downgrade.
  Scenario: a prompt-injected agent assembling `Router` opts sets `trust_score: 100` and gets
  sovereign fees + shielded settlement with zero proofs. Fix: under the hostile-caller model,
  an explicit `:trust_score` must not exceed the attestation-derived ceiling (or tier
  requirements must apply to score-only paths, i.e. treat absent attestations as the empty
  set for gating, not as "no gate").

- **[Security Auditor / Saboteur] HIGH** `zksar.ex:116-136` (no subject binding) +
  `schemas.ex:157-164` -- A verified attestation is bound to nothing the caller must possess.
  `verify/2` never checks `proof.subject` against the paying wallet/recipient/session, and the
  signed digest (`attestation_digest/1`) carries no per-use nonce/intent id (PR defers nonce;
  subject binding is not even acknowledged). Attestations are transmitted to Xochi inside
  `SettleRequest` (`schemas.ex:157`), i.e. observable on the wire. Scenario: agent B replays
  agent A's valid, far-future-expiry `non_membership`+`compliance` proofs and inherits A's
  trust/fee tier until expiry, across sessions and recipients. This is a replayable-attestation
  bypass. Fix: bind the signed digest to `subject` == authenticated caller (and, with the
  deferred nonce work, to intent/chain), and reject proofs whose subject is not the caller.

### MEDIUM

- **[Saboteur] MEDIUM** `zksar.ex:145-153` + `trust_score.ex:44-53` -- `verify_batch/2` and
  `aggregate/1` do not de-duplicate, so the diminishing-returns curve is defeated by copies.
  Each duplicate of one valid signed proof recovers to the allowlisted issuer and verifies
  independently. Reproduced: an honest `compliance`+`non_membership` pair aggregates to 43
  (`:stealth`); duplicating each proof once aggregates to 75 (`:sovereign`) with both required
  types present, so no downgrade -- a holder of two legitimate attestations jumps two tiers to
  the lowest fee / deepest privacy. Fix: de-dupe verified proofs by (type, subject, signature)
  before aggregation. Signature malleability (low-`s` not enforced, PR-acknowledged) is a
  second dedup-evasion vector for the same attestation.

- **[Saboteur / Security Auditor] MEDIUM** `zksar.ex:247-253` (`check_structure`) +
  `zksar.ex:181` (`attestation_digest/1`) + `router.ex:181` -- Verification does not fail
  closed on a malformed proof; it raises. `check_structure/1` only validates
  subject/issuer/signature, but `attestation_digest/1` also requires `:payload` and
  `:issued_at`. A proof with a future `expires_at` and an allowlisted `issuer` but no
  `:payload` passes structure/expiry/issuer, then `check_signature -> attestation_digest`
  raises `FunctionClauseError` (reproduced). Nothing in `Router` rescues it, so
  `Router.select/1` crashes on caller-controlled input -- a DoS, and a "fail-open by
  exception" shape rather than a clean `{:error, :malformed}`. Fix: validate `:payload`/
  `:issued_at`/`:type_code` presence in `check_structure` (return `:malformed`); optionally
  guard the batch reduce.

- **[Security Auditor] MEDIUM** `router.ex:181` + `zksar.ex:19-27` -- Verification-failure
  telemetry is absent and the module's own precondition is violated. `verify_batch` returns
  `{verified, _errors}` and the errors are discarded, so an operator cannot distinguish a
  forgery attempt from every real Xochi attestation being rejected. That distinction matters
  precisely because the module warns (lines 19-27) the digest scheme is provisional and "must
  still be reconciled with Xochi ... before enabling signature checks against production data"
  -- yet this PR enables production signature checks (`verify_signature` defaults true) against
  that unconfirmed digest. Fail-closed means no forgery, but if the operator pins the real
  Xochi issuer and the digest scheme differs, all genuine attestations silently drop to public
  tier with no signal. Fix: emit `[:raxol, :payments, :zksar, :verify_failed]` telemetry with
  the reason, and gate production enablement on a confirmed on-chain vector (or keep it
  `verify_signature: false` structural-only until reconciled).

### LOW

- **[New Hire] LOW** `privacy_tier.ex:114-132` + moduledoc -- "Verified" is a bare
  `valid: true` field on a plain map, not a distinct struct/type. `PrivacyTier.from_trust_score/2`
  is public and its safety against a caller-built `%{type: :compliance, valid: true}` rests
  entirely on the moduledoc convention "Router is the choke point," not on the type system.
  Any code path that reaches `from_trust_score` without going through `Router` re-opens the
  original bug. Fix: make verified proofs a tagged struct (e.g. `%Zksar.Verified{}`) so the
  compiler, not prose, enforces "pre-verified."

### Cross-persona overlaps

- Replay/binding: Saboteur (reuse) and Security Auditor (domain separation) converge on the
  same root at `zksar.ex` -- verified proofs bound to neither caller nor per-use nonce.
  Promoted to HIGH.
- "Not observably fail-closed": the dropped `_errors` (telemetry MEDIUM) and the
  crash-on-malformed (DoS MEDIUM) share the theme that failure is silent or explosive rather
  than a clean, logged rejection.

### Verdict: BLOCK

The attestation-forgery fix is genuine and should land, but two caller-controlled paths
survive it under the PR's own hostile-caller threat model: an explicit `trust_score: 100`
buys sovereign tier while skipping the attestation-requirement gate entirely (HIGH, reproduced),
and verified attestations are bound to neither caller nor nonce, making any observed proof
replayable until expiry (HIGH). Address the `trust_score` ceiling + gate-on-score-only paths
and add verification-failure telemetry before merge; track subject/nonce binding and proof
de-duplication with the already-deferred replay work.

PR #439, opened 2026-07-11
Fix ex_doc hexdoc links + bump postgrex (CVE-2026-58225)

DROOdotFOO:
The 11 docs added to `extras:` in #436 link out to files that exist in the repo but are not ex_doc extras (ADRs, plugin guides, CLAUDE.md, examples/README), so `mix hex.publish docs` warned "references file X but it does not exist" and would have shipped broken links to hexdocs.

Pointed those links at absolute GitHub URLs (no cascade of adding ADRs/CLAUDE.md to the doc set), and fully-qualified `Config.set/2` -> `Raxol.Config.set/2` so ex_doc resolves it.

Fixed links: PACKAGES.md (CLAUDE.md#hex-publishing), MCP.md + TELEGRAM.md (../adr/0012, ../adr/0014), PLUGIN_SDK.md (../plugins/{GUIDE,PLUGIN_TEMPLATES,TESTING}.md), CUSTOM_COMPONENTS.md (../../examples/README.md), CONFIG.md (Config.set/2).

Root cause of these slipping past #436: my `mix docs` verification used `timeout ... mix docs`, but `timeout` isn't installed on macOS (`command not found`), so the check never ran. Re-verified by running `mix docs` directly -> exit 0, zero reference/link warnings.

## Manual testing
- [x] `mix docs` (run directly) -> builds clean, no reference/link/undefined warnings


DROOdotFOO:
Added: bump `postgrex` 0.22.2 -> 0.22.3 (+ transitive `db_connection` 2.10.1 -> 2.10.2), which clears the LOW advisory EEF-CVE-2026-58225 (notification DoS). `~> 0.22.0` already allowed it, so mix.exs is unchanged. postgrex is `optional: true, runtime: false` here (Ecto disabled at runtime), so real exposure was ~nil regardless.

PR #440, opened 2026-07-11
Spec-correct flex layout engine, CSS text layout, overflow — with a characterization safety net

merklebonsai:
## What

Makes the layout engine honest: a CSS-flexbox-spec (§9.7) flexible-length solver on the live path, CSS text-layout primitives, unification of the layout dialects, overflow/scroll-anchoring, and fixes for five renderer bugs that were hidden behind each other — pinned by a characterization test net so none of them can silently return.

## Highlights

**Layout engine**
- `FlexItem` + `Flexbox.Solver`: spec-§9.7 resolve-flexible-lengths (pre-freeze, clamp-freeze-redistribute, exact integer accounting via largest-remainder) replaces the single-pass distributor that lost cells to `div` truncation and couldn't express `flex: 1`
- Automatic minimum size (`min-width: auto` = min-content): text floors at its longest word, dividers stop robbing fixed-width siblings
- `{:pct, n}` dimensions, `margin: :auto` (main-axis centering), min/max constraints, exact-sum justify spacing
- Literal `:row`/`:column` elements run on the flex engine through a compat map preserving their dialect (gap 1, `:start` alignment, no shrink); `Containers` module deleted
- Dead code removed: legacy `:flexbox` API, second distribution algorithm, third renderer layout stack (~2,100 LOC)

**Text layout**
- `white_space` (5 CSS modes), `text_overflow: :ellipsis` (grapheme-safe), `line_clamp` (Overflow L4), `text_wrap: :pretty` (Knuth–Plass with orphan suppression)
- Markdown inline formatting renders as styled spans instead of literal `*markers*`

**Overflow**
- `overflow:` container property stamps clip bounds the renderer enforces for every element type; `Viewport` gets `overflow_anchor` bottom-follow (terminal-log idiom)

**Bug fixes** (each with a regression test)
- Multiline text emitted literal `\n` buffer cells — content escaped its layout box
- Renderer painted a border on every box without a `:border` key while layout treated absent as `:none` — phantom frames through child cells
- Charts measured 0×0 (type-rewrite existed in layout but not measurement) and wrapped in dimensionless full-bleed boxes; same-row same-color cell runs merged across gaps (bars melted together)
- View DSL tables never rendered a single cell (five stacked shape mismatches)
- `render_diff` emitted `{:write}` before its `{:move}` — every write landed at the previous cursor position
- Terminal resize: SIGWINCH never reached the runtime, and the dispatcher swallowed resize events before apps saw them

**Verification**
- Characterization net pins engine geometry across the dialects; safety-net suite replays ANSI through the reference terminal emulator and diffs grids
- Visual snapshot harness re-pointed to the live engine (snapshots had been blank buffers since inception)
- Suite: 4,528/4,530 — both failures are pre-existing timing flakes (`metrics_collector_test.exs:150` sleep-based; parser-perf property under full-suite load, passes isolated)
- Reviewed pre-submission by three independent automated reviewers; all confirmed findings addressed (DSL alias normalization, boolean `wrap`, measure/layout basis parity, `MinContent` padding order, anchor coverage)

## Notes for review

- `docs/core/LAYOUT.md` is the property reference + intentional-CSS-divergence list (e.g. `flex: 1` is terminal-pragmatic: zeroes the automatic minimum so columns equalize)
- Reviewers suggested this could split into layout-core / renderer-fixes / resize / examples PRs — happy to split if preferred; kept together because the characterization net and demo reworks are what validate the engine changes
- Known follow-ups (documented in LAYOUT.md "Future work"): span-aware wrapping, `:fill` as `{:pct, 100}` alias

https://claude.ai/code/session_0193ra3VnoB1yLM8mqRJVwpw


DROOdotFOO:
> +10,697
> -3,415
> # Lines changed: 10697 additions & 3415 deletions

Description of GIF   @merklebonsai

PR #442, opened 2026-07-12
Dev-experience: auto-port endpoint, agent supervisor double-start, compile warnings

merklebonsai:
Part of decomposing the cross-terminal work into granular, independently-reviewable PRs. This one is fully standalone — touches only `config/dev.exs` and `packages/raxol_agent/**`, no overlap with the layout/terminal PRs, mergeable in any order.

## Changes

**Dev web endpoint auto-picks the first free port from 4000.** A co-resident service on 4000 (e.g. a local LiteLLM) made Ranch fail `:eaddrinuse` and take the whole app down at boot — so `mix run` of any example wouldn't start. `RAXOL_DEV_PORT` still pins an explicit port; absent it, the endpoint probes upward from 4000 for the first bindable port and logs which one it chose. The probe binds all interfaces with no `SO_REUSEADDR`, matching the endpoint's own bind, so it reads free/busy truthfully instead of false-positiving on a partially-bound port.

**Agent examples no longer crash on an already-started supervisor.** The main `raxol` app starts `Raxol.Agent.Supervisor` whenever the module is compiled in, and `raxol_agent` depends on `raxol` — so booting `raxol_agent` as the top app (any package example) started the supervisor twice and died with `:already_started` before the demo ran. `RaxolAgent.Application` now only owns the subtree when nobody else has started it, and the named supervisor's `start_link` is idempotent so the example/test scripts that boot it manually get the running subtree back instead of a crash.

**Two agent-path compile warnings cleared.** Pin `ct_len` in the Lumo GCM ciphertext/tag split (an unpinned var in `binary-size(...)` matches loosely and warns); drop a spurious `@impl true` on the react_agent example's `available_actions/0`, which isn't a behaviour callback.

## Verification

- `react_agent.exs` and `agent_team.exs` run to completion from the package.
- `zero_system.exs` boots from the repo root with the auto-port.
- raxol_agent package suite green; Lumo `crypto_test` 17/17.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -30,9 +30,42 @@ config :raxol, Raxol.Repo,
   show_sensitive_data_on_connection_error: true,
   pool_size: 10
 
-# Dev endpoint for Tidewave MCP (localhost:4000/tidewave/mcp)
+# Dev endpoint for Tidewave MCP (localhost:<port>/tidewave/mcp).
+# RAXOL_DEV_PORT pins an explicit port; otherwise probe upward from 4000
+# for the first free one, so a co-resident service on 4000 (e.g. a local
+# LiteLLM) doesn't take the whole app down with :eaddrinuse. The probe
+# binds all interfaces with no SO_REUSEADDR to match how the endpoint
+# itself binds -- a truthful free/busy read, not a false positive.
+resolve_dev_port = fn ->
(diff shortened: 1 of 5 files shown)


DROOdotFOO (approved):
LGTM now u can merge @merklebonsai

PR #443, opened 2026-07-12
Terminal driver hardening: OSC 11 detection, noshell stdin, total CSI parsing, frame geometry

merklebonsai:
Part of decomposing the cross-terminal work into granular, independently-reviewable PRs. Touches only `packages/raxol_terminal/**` and `lib/raxol/core/runtime/rendering/backends.ex` — disjoint from the layout and dev-experience PRs, mergeable in any order.

Hardens the terminal driver end to end. Several of these were latent bugs that a real terminal answering escape-sequence queries could trigger at startup; found and fixed via PTY-based reproduction (expect harnesses answering OSC 11 / DA / focus / mouse reports).

## Changes (each its own commit)

**Detect terminal background via OSC 11; handle SIGWINCH resize.** The driver queries the terminal background colour (OSC 11, DA probe as the unsupported-terminal sentinel), strips the reply from the input stream before key parsing, and publishes it via `persistent_term` + a `:terminal_background` event for higher layers (e.g. theming) to consume. Also installs a SIGWINCH `gen_event` handler — prim_tty delivers the signal straight to `user_drv`, bypassing the reader we trace. `InputBuffer` holds an incomplete `ESC ]` (OSC) chunk until its BEL/ST terminator so a split colour reply is never mis-parsed as keys.

**Write the OSC 11 query only after the stdin reader is active; fail-safe.** Writing it earlier raced the prim_tty `tty => true` reinit; the whole query/scan/reply path now degrades to "no background detected" on any exception instead of crashing the driver or blocking input.

**Start `user_drv` shell-less.** `start_shell` with an MFA spawns a real shell process; ours returned immediately, so `user_drv` dropped into the job-control prompt for the whole session with all stdin still copied to it. A terminal answering the OSC 11/DA query then fed its reply into that prompt, killing stdio and the node before the first frame. `initial_shell: :noshell` spawns no shell at all; since `shell_started` stays false, `user_drv` never issues the one-time prim_tty read kick, so the driver sends `{:read, :infinity}` to the reader itself.

**Make CSI parsing total; consume unmapped sequences instead of leaking them.** `parse_csi_params` only matched digits/`;`/final bytes, so a CSI private marker (`>`, `<`, `?`) raised `FunctionClauseError` — and terminals volunteer secondary-DA reports (`CSI > Pp;Pv;Pc c`) unprompted, killing the driver right after the first frame. Unknown bytes are now skipped while scanning for the final byte, and `dispatch_raw_input` wraps the parser fail-safe. A follow-up makes well-formed-but-unmapped sequences consume whole (previously the recovery path leaked their bytes as fake keystrokes — one full-frame repaint per byte, plus the leaked `c` toggling UI state).

**Own frame geometry: CRLF row separators and DECAWM off.** With prim_tty in `output=raw` (the noshell path), nothing cooks the frame's bare-LF row joins into CRLF anymore, so full-width rows autowrapped *and* line-fed — every frame rendered double-spaced. Backends normalize LF to CRLF before writing (harmless under cooked — `\r\r\n` is still one advance) and the driver disables autowrap for the session, restoring it on cleanup.

## Verification

- PTY harnesses (BEL/ST OSC replies, kitty DA, focus-in, stray mouse SGR, single-write bursts): render + nav + `q` clean exit, no crashes, single-spaced output confirmed with a real VT100 emulator.
- `input_parser` + driver suites: 118 passed. SSH rendering property suite: 20 passed.

### DIFF
--- a/lib/raxol/core/runtime/rendering/backends.ex
+++ b/lib/raxol/core/runtime/rendering/backends.ex
@@ -29,7 +29,7 @@ defmodule Raxol.Core.Runtime.Rendering.Backends do
     )
 
     # Move cursor to top-left and clear screen before each frame
-    frame = "\e[H\e[2J" <> output_string
+    frame = normalize_frame("\e[H\e[2J" <> output_string)
 
     if state.sync_output do
       IO.write("\e[?2026h")
@@ -167,7 +167,7 @@ defmodule Raxol.Core.Runtime.Rendering.Backends do
     output_string = Raxol.Terminal.Renderer.render(renderer)
 
     # Home cursor and clear screen before each frame, matching render_to_terminal
-    frame = "\e[H\e[2J" <> output_string
+    frame = normalize_frame("\e[H\e[2J" <> output_string)
 
     write_output(state.io_writer, frame, state.sync_output)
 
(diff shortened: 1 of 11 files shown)


DROOdotFOO (approved):
LGTM @merklebonsai

PR #448, opened 2026-07-12
Flex-spec convergence: §9.7 solver, min-content, dialect unification, characterization net

merklebonsai:
The layout-engine foundation for the cross-terminal rendering work. Base for the render-fixes, dimming, and modal PRs.

Replaces the ad-hoc layout path with a CSS-flexbox-spec §9.7 resolve-flexible-lengths solver and a clean `FlexItem` contract, and unifies the two flex-like dialects the engine dispatched:

- **spec-9.7 flex solver** (`Flexbox.Solver`) + `FlexItem` resolved-item contract; largest-remainder integer accounting so justify/align spacing sums exactly.
- **automatic min-content minimum** (`MinContent`) — overflow clips instead of content vanishing; terminal-pragmatic `flex: 1` zeroes the automatic minimum so equal columns actually equalize (documented divergence).
- **literal `:row`/`:column` dialect unified onto the flex engine** via a compat map (preserves the old gap/justify/align defaults).
- **text-layout modes**: `white_space`, `text_overflow` ellipsis, `line_clamp`, `text_wrap: :pretty` (Knuth-Plass).
- **overflow property** + `Viewport` scroll-anchoring (`:auto` follow-bottom / `:none`).
- **cross-terminal characterization net**: an ANSI-stream-to-grid harness pinning current geometry as a regression tripwire.

Deletes the dead layout stacks (`Containers`, `Distributor`, `Core.Renderer.Layout`, `View.Layout.Grid`) with no remaining callers. See `docs/core/LAYOUT.md` for the property reference and intentional divergences.

Passes the new FATE golden-render gate. Layout/cross-terminal suites green.


merklebonsai:
Decomposed into atomic PRs per review request. Mergeable now (no deps): #456 (box sizing + render-diff fix) and #457 (text-layout modes). The remaining flex primitives — characterization net, §9.7 solver+gap+min-content, align-content/dead-stack cleanup, row/column compat, overflow+anchor — stay stacked on those and will open as their dependencies land. The branches are built and green (they reconstruct this PR's tree exactly).

PR #450, opened 2026-07-12
Per-widget render fixes (charts, borders, tables, markdown, multiline) + examples audit

merklebonsai:
**Depends on #448 (flex-convergence) — review/merge after it.** Until #448 lands, this PR's diff includes its commits.

Per-widget rendering fixes surfaced by the flex rework, plus an examples/playground audit under the new capabilities:

- **charts**: fix height under-allocation (measure rewrites chart types to `:box`); wrapper box hugs the chart region instead of filling available space.
- **phantom borders**: renderer border default now `:none`, matching the layout engine (absent border no longer paints a frame through children).
- **multiline text**: `\n`-split lines render (never emit a newline cell that resets the cursor to column 0); layout reserves the measured row count.
- **DSL tables** actually render (layout + measurement + normalize path).
- **markdown**: styled spans instead of literal `*_`` markers; list-item wrap respects prefix width.
- **row/column compat**: honor top-level `gap`/`justify`/`align`.
- examples + playground demos updated under the new layout/text capabilities.

Suites green (1600+ tests); FATE golden gate passes.


merklebonsai:
Retracting — this bundles many independent per-widget fixes (chart height/hug, phantom borders, multiline text, DSL tables, markdown spans, gap/justify compat) plus examples/playground audits. Will re-open as atomic PRs (one primitive each), matching the flex-solver decomposition, once the flex chain lands in master (these sit on the wired engine). Branch preserved on the fork.

PR #451, opened 2026-07-12
H-K salience colour solver + semantic theme + truecolor SGR

merklebonsai:
**Depends on #448 (flex-convergence) — review/merge after it.**

The colour foundation for adaptive theming and modal dimming:

- **Helmholtz-Kohlrausch salience solver** (`Theming.Salience`): OKLCH, two-lobe apparent-lightness model; palettes expressed as `(hue, chroma, tier)` seeds with lightness *solved* against a ground so every tier reads perceptually level. Darcula fixtures pin it byte-exact.
- **semantic salience theme** — grounds on the terminal background when detected (soft dependency; falls back to a reference ground otherwise).
- **truecolor SGR for `{r,g,b}` tuple cell colours** — the ANSI stage previously had no clause for plain integer tuples, so such cells emitted no colour escape.

Theming fixtures + renderer tests green.

### DIFF
--- a/lib/raxol/ui/theming/salience.ex
+++ b/lib/raxol/ui/theming/salience.ex
@@ -1,0 +1,272 @@
+defmodule Raxol.UI.Theming.Salience do
+  @moduledoc """
+  Salience-tier color solver: OKLCH colors leveled through a two-lobe
+  Helmholtz-Kohlrausch apparent-lightness model.
+
+  Instead of hand-picking lightness per color, a palette is expressed as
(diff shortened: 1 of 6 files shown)


merklebonsai:
Rebased directly onto master — this is actually independent of the flex layout work (salience solver + semantic theme + truecolor SGR renderer, no flex files). No longer blocked; ready to merge on its own. (Dimming #452 depends on this; modal #453 depends on that.)

PR #459, opened 2026-07-12
CSS-flexbox §9.7 solver + FlexItem + min-content + gap fix

merklebonsai:
Atomic slice of the flex-convergence layout work (decomposing the closed #448). Off master, no dependencies — box-sizing (#456) has merged.

The CSS-flexbox core:
- **§9.7 resolve-flexible-lengths solver** (`Flexbox.Solver`) + `FlexItem` resolved-item contract; largest-remainder integer accounting so justify/align spacing sums exactly.
- **automatic min-content minimum** (`MinContent`) — overflow clips instead of content vanishing; terminal-pragmatic `flex: 1` zeroes the automatic minimum so equal columns actually equalize (documented divergence).
- **flex gap under-allocation fix** — the calculator summed children without the inter-item gap while the positioner laid them out with it; measured main size now includes the gap.

Wires the solver into the live layout path. The dependent flex PRs (row/column compat, overflow, align-content, dead-stack deletion) stack on this.


merklebonsai:
Split into 3 atomic PRs per review: solver+FlexItem modules (ready), min-content+justify (draft), wire-into-engine+gap fix (draft). Each green at its level.

PR #461, opened 2026-07-12
Min-content measurement + exact justify spacing

merklebonsai:
Split of #459. **Depends on the solver-modules PR — review after it.**

- `MinContent` — min-content width measurement.
- `Flexbox.Positioner` — exact justify/align spacing (largest-remainder, sums exactly to the container).


merklebonsai:
Rebased onto master now that solver-modules (#460) is merged. Ready. (Wire #462 depends on this.)

PR #509, opened 2026-07-13
Fix OSC 9 clipboard mis-route; add OSC 9;4 progress and OSC 22 pointer shape

merklebonsai:
## OSC 9 routing fix

`OSCHandler` routed OSC 9 to `Clipboard.handle_9`. That's wrong: OSC 9 is the
iTerm2/growl desktop notification sequence, not clipboard. OSC 52 is the
correct clipboard sequence and was already implemented separately and
untouched by this PR.

**Investigation / back-compat**: grepped the repo for any caller or test
depending on OSC-9-as-clipboard (`handle_9`, direct `OSCHandler.handle(_, 9,
_)` calls, fixtures). Found none — `executor.ex` is the only real caller and
it dispatches by parsed command code with no OSC-9-specific handling. The
test fixture at `test/fixtures/ansi_sequences.exs` already labels OSC 9 as
`notification`, not clipboard, confirming the routing was a bug rather than
an intentional (if nonstandard) extension. So this removes the OSC-9 →
clipboard path outright instead of keeping it for back-compat.

## New ambient OSC support

- **OSC 9** (plain payload): desktop notification, stored on
  `emulator.notification`.
- **OSC 9;4;<state>;<progress>**: taskbar/dock progress (ConEmu-origin, now
  also supported by Ghostty, iTerm2, Windows Terminal, WezTerm). `state` is
  0=remove, 1=set, 2=error, 3=indeterminate, 4=warning; `progress` is 0-100
  (defaults to 0 if omitted). Stored on `emulator.progress` as
  `%{state:, value:}`. Malformed input (bad state code, out-of-range or
  non-numeric progress) returns `{:error, :invalid_progress, emulator}`
  rather than crashing.
- **OSC 22;<shape>**: pointer/cursor shape (CSS-style keyword, e.g.
  `default`, `text`, `pointer`, `wait`, `crosshair`). Stored on
  `emulator.pointer_shape`. Empty payload returns
  `{:error, :invalid_pointer_shape, emulator}`.

Emit-side API added to `Raxol.Terminal.AdvancedFeatures` (alongside the
existing OSC 8 hyperlink emitters): `notify/1`, `report_progress/2`,
`clear_progress/0`, `set_pointer_shape/1` — all pure functions returning the
escape sequence, matching `create_hyperlink/3`'s style.

`Raxol.Terminal.Emulator` gains three new struct fields (`notification`,
`progress`, `pointer_shape`, all defaulting to `nil`) alongside the existing
`current_hyperlink` field, following the same pattern.

## Tests

- `packages/raxol_terminal/test/raxol/terminal/commands/osc_handler_test.exs`
  (new): parse tests for OSC 9 notification, OSC 9;4 progress (all five
  states, malformed/out-of-range/non-numeric input), OSC 22 pointer shape
  (including empty-payload rejection), and an OSC 52 clipboard regression
  suite that also asserts OSC 9 no longer touches clipboard state.
- `packages/raxol_terminal/test/raxol/terminal/advanced_features_test.exs`
  (new): emit tests asserting exact byte sequences for `notify/1`,
  `report_progress/2`, `clear_progress/0`, `set_pointer_shape/1`.

## Test plan

- [x] `MIX_ENV=test mix compile --warnings-as-errors` (exit 0)
- [x] `mix format --check-formatted` (exit 0)
- [x] `TMPDIR=/tmp SKIP_TERMBOX2_TESTS=true MIX_ENV=test mix test` on the new
      OSC/advanced-features test files: 30 passed
- [x] Broader terminal `commands/` + renderer/output-manager suite for
      regressions: 366 passed, 5 excluded, 0 failed

### DIFF
--- a/lib/raxol/playground/catalog.ex
+++ b/lib/raxol/playground/catalog.ex
@@ -326,6 +326,16 @@ defmodule Raxol.Playground.Catalog do
       tags: ["effects", "focus", "ring", "accessibility"],
       code_snippet: ~s'FocusRing.render(content, FocusRing.init(style: :solid))'
     },
+    %{
+      name: "OSC Ambient",
+      module: Demos.OscAmbientDemo,
+      category: :effects,
+      description:
+        "Host-terminal desktop notification, taskbar progress, and pointer shape",
+      complexity: :intermediate,
+      tags: ["effects", "osc", "notification", "progress", "pointer"],
+      code_snippet: ~s'IO.write(AdvancedFeatures.report_progress(:set, 42))'
+    },
     # --- REPL & VFS ---
     %{
       name: "Virtual FS",
--- a/lib/raxol/playground/demos/osc_ambient_demo.ex
+++ b/lib/raxol/playground/demos/osc_ambient_demo.ex
@@ -1,0 +1,189 @@
+defmodule Raxol.Playground.Demos.OscAmbientDemo do
+  @moduledoc """
+  Playground demo: OSC 9 desktop notification, OSC 9;4 taskbar/dock
+  progress, and OSC 22 pointer shape.
+
+  These are host-terminal features, not something Raxol renders itself.
+  Ghostty, iTerm2, WezTerm, and Windows Terminal draw the dock/taskbar
+  progress bar from OSC 9;4; most other terminals (including many Linux
+  emulators) silently ignore all three sequences. This demo always emits
+  them and trusts the terminal to honor or drop them -- there is no
+  capability detection here.
+  """
+  use Raxol.Core.Runtime.Application
+
+  alias Raxol.Playground.DemoHelpers
+  alias Raxol.Terminal.AdvancedFeatures
+
+  @tick_ms 200
+  @step 4
+  @max_progress 100
+  @bar_width 30
+  @pointer_shapes ["default", "text", "pointer", "wait", "crosshair"]
+
+  @impl true
+  def init(_context) do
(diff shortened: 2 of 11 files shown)


merklebonsai:
Added a playground demo (\`lib/raxol/playground/demos/osc_ambient_demo.ex\`, catalog entry "OSC Ambient" under `:effects`) that exercises the AdvancedFeatures API this PR adds:

- **[space]** start/pause a fake task; while running, each tick reports progress to the host taskbar/dock via `AdvancedFeatures.report_progress/2` (OSC 9;4)
- on completion, fires a desktop notification via `AdvancedFeatures.notify/1` (OSC 9)
- **[c]** cycles pointer shapes via `AdvancedFeatures.set_pointer_shape/1` (OSC 22)
- **[e]/[w]** flip the taskbar state to error/warning to show the colour change
- **[r]** resets and clears progress via `AdvancedFeatures.clear_progress/0`

The demo text is upfront that these are host-terminal features (Ghostty/iTerm2/WezTerm/Windows Terminal support 9;4 taskbar progress; most others ignore it) and does no capability detection.

Emission is wired through `Directive.spawn_task` so `update/2` stays pure and the escape sequences are only ever written by the runtime's executor, not during unit tests.

Tests: 24 new (`test/raxol/playground/demos/osc_ambient_demo_test.exs`), plus catalog/app/sidebar-scroll widget-count assertions bumped 30 → 31. Full `test/raxol/playground/` suite: 315/315 green. `mix compile --warnings-as-errors` and `mix format --check-formatted` both clean.

PR #521, opened 2026-07-13
A box's background fills its interior, not its border ring

merklebonsai:
Stacked on #520 (shares its background handling). Merge that first.

## The current behaviour is backwards

`BorderRenderer.render_box_borders/6` iterates **only the ring**:

```elixir
for i <- 0..(width-1), j <- 0..(height-1),
    i == 0 or i == width-1 or j == 0 or j == height-1
```

…and hands those ring cells the box's background. The **interior is never painted at all**. So today:

```elixir
box(border: :single, bg: :blue)   # a blue OUTLINE around a hollow middle
```

which is about the opposite of what it reads like.

## Fix

The background fills the **interior**; the ring is painted only when asked for explicitly, via `:border_bg`.

| | result |
|---|---|
| `box(border: :single)` | border only, fully transparent |
| `box(border: :single, bg: :blue)` | blue interior, **unpainted border** |
| `box(border: :single, bg: :blue, border_bg: :red)` | blue interior, red border |
| `box(bg: :blue)` (no border) | whole rect filled (unchanged) |

Besides being the obvious reading of "a box with a background", this is what a **transparent terminal** wants: a border that floats over the desktop with only its *content* backed, rather than a rectangle of paint tracing the frame. (Requested from exactly that use case.)

A box with no room for an interior (2×2 and smaller) emits no fill cells and still draws its border.

Tests: interior filled / ring unpainted; no background paints nothing; `border_bg` paints the ring independently; `border_bg` alone paints only the ring; degenerate box emits no fill. 5/5. Regression sweep (ui + cross_terminal + playground): 1719/1719.


merklebonsai:
Added `background_clip`, so the caller picks which side the border cell belongs to.

A cell is the smallest unit of paint — the border glyph fills a whole cell, so the background can't stop halfway through it. The cell is either *inside* the background or *outside* it. That's the only real choice, and CSS already named it:

| `background_clip:` | ring | interior | reads as |
|---|---|---|---|
| `:padding_box` *(default)* | unpainted | filled | outline **floats** on the desktop, content backed |
| `:border_box` | filled | filled | solid **panel**, glyph drawn on the fill |

An explicit `border_bg:` still wins over both (`ring=:red, interior=:blue`).

9/9 tests; regression sweep 1725/1725.

merklebonsai:
Third commit: **an unpainted background now shows what is beneath it, rather than erasing it.**

Filling the box exposed the deeper bug. Cells don't composite — writing one *replaces* what was there — so a cell with `bg: nil` didn't leave the background beneath it alone, it **cleared** it. A button drawn on a filled modal punched a hole through the modal and showed the desktop through itself.

`nil` was silently meaning two different things depending on where it landed:
- at the top of the tree → "the terminal shows through" ✓
- over a filled parent → "erase the parent's fill" ✗

It has to mean one thing — **show what is beneath** — and both cases fall out of it: over a filled parent that's the parent's fill; at the top there's nothing beneath, so it stays unpainted and the terminal shows through.

Applied at both composition points (`CellManager.merge_cells/2` and the cells→buffer write). A painted background still overrides, unchanged.

4 new tests; sweep 2605/2605.

DROOdotFOO:
CI infra unblocked: your `Setup & Cache` failure wasn't your code -- it was upstream `mix.lock` drift (`phoenix_live_view 1.2.6 -> 1.2.7`, security advisory on 1.2.6) hitting the strict `mix deps.get --check-locked`. Master already fixed this class in #522 (adds a `|| mix deps.get` fallback + refreshes the lock). I merged master into this branch, so `Setup & Cache` is green now.

The only failures left are golden/snapshot mismatches, which are the intended result of this PR -- filling a box's interior changes its rendered cells, so the committed render references no longer match:

- `Raxol.Examples.ButtonTest` -- visual + normal/disabled snapshots
- `Raxol.FATEGoldenTest` -- render corpus

To clear them, regenerate the references, eyeball the diff to confirm the new output is what you intend, then commit:

```
mix raxol.fate --gen        # FATE corpus
# then update the ButtonTest snapshots and commit both
```

DROOdotFOO (approved):

PR #524, opened 2026-07-13
Collapse dialog_surface to one box; paint it with a theme surface colour

merklebonsai:
Addresses @DROOdotFOO's review on #523. Stacked on #521 — it's what makes the collapse possible.

## The outer box was scaffolding for a bug we deleted

`dialog_surface/4` nested a bordered box inside an unbordered, fully-filled box of the same footprint. Its own docstring explained why:

> *"a border only paints its ring, so without the outer fill the dimmed flow content behind it would show through the gaps. Outer is emitted first, inner (its child) after, so **last-write-wins** cell composition lets the border/content win…"*

Both premises are now false:

- a box's background fills its **whole footprint, border included** (#521, `background_clip: :border_box`)
- an unpainted cell **inherits what's beneath** rather than erasing it (#521, invariant 3)

So the outer box, the doubled `width`/`height`, and the composition-order reliance are all dead weight. **One bordered box with a background is the entire surface.**

Verified: `children: 0` (the nested box is gone), and every cell — ring *and* interior — carries the surface colour.

## The fill colour was the literal ADR-0029 flags

It defaulted to `:black`. A dialog *is* opaque on purpose — the dimmed content behind it must not read through — so painting it is right. Painting it **black** is not: the terminal owns the canvas, and that's a black slab on a light terminal.

The colour now comes from the theme, and the theme gains a `:surface` colour to name the thing:

```elixir
surface: "#1E1E1E"   # a raised opaque surface -- a dialog, a panel
```

This is a genuinely different concept from `:background`, and worth the extra key: **`:background` is only the theme's *assumption* about the terminal's own canvas, and is never painted. A `:surface` is painted, on purpose.** Conflating them is what produced the original bug.

Falls back to the default surface when no theme is available (rendering outside a started application).

## Also fixed

The `background_clip` table in **#521**'s description was stale — written before the `:border_box` default landed, so it still claimed `box(border, bg: :blue)` leaves the border unpainted. Corrected in place.

Sweep: 2609/2609 (modal overlay + headless suites green).


merklebonsai:
Rebased onto master (which now has ADR-0029) and folded the ADR amendment in, since
this is the PR that introduces the colour it describes.

ADR-0029 said *"the terminal owns the canvas; the application colours the content"* —
and then left no colour for a thing that is opaque **on purpose**. A modal is exactly
that: it sits over dimmed content which must not read through it. With no `:surface`
to ask for, it reached for `:background` — the one colour the ADR says is never
painted. So the theme now names both:

| theme colour  | what it is                                              | painted? |
| ------------- | ------------------------------------------------------- | -------- |
| `:background` | this theme's assumption about the terminal's own canvas | **no**   |
| `:surface`    | a raised opaque thing — a dialog, a panel               | **yes**  |

Which is the ADR's own thesis, one layer up from where it was looking: *one name doing
two jobs.* `:background` was doing both, and it was found by reading the invariants
against the code they were supposed to describe. That is the intended use, so it's now
the last entry in the ADR's own list, and RENDERING.md's "where to put things" table
gained a row.

`modal_test.exs` is updated in the same commit as the collapse — it was pinning the
two-box shape (`assert surface.border == :none`, `assert [inner] = surface.children`),
which is the scaffolding this removes.

DROOdotFOO (approved):

PR #547, opened 2026-07-15
feat(harness): U1.5 — close the keystone (session_id + journal sink + id-authority + turn vocab)

merklebonsai:
## What

Closes the harness keystone: durable agent events are now journaled with the **journal as the single id authority**, `session_id` flows end-to-end for a live agent session, and turns carry a minimal loop vocabulary. Builds on U1 (emit seam), U2a (journal), U3 (commands) — all on master.

Three jobs:

- **(a) session_id flows.** `Raxol.Agent.Session` mints a stable, filename-safe `session_id` and threads it through `Lifecycle.start_link` into the Dispatcher state (the initializer already read the opt). A running agent's fold sites now emit with a non-nil `session_id`.
- **(b) Journal-backed durable sink; journal owns `Event.id`.** `Raxol.Agent.EmitBridge` becomes the durable-tier id authority. For every **durable** event it does **append-before-publish**: `Journal.append/2` FIRST → take the returned `offset` → stamp `Contract.Event.id = offset` → THEN `SessionStreamer.emit`. **Ephemeral** deltas are never journaled and carry the *last durable offset* so they never masquerade as offsets. The journal (`FileStore`) opens lazily on the first durable event under `RAXOL_SESSIONS_DIR`/`~/.raxol/sessions`, persists to disk, and is closed on `terminate/2`. Durable payloads are sanitized (new `Contract.sanitize_payload/1`) before append so non-JSON terms (message tuples) never crash the writer. Main `raxol` still emits only neutral maps to `EmitBus` — **no back-dep on `Raxol.Agent.*`**.
- **(c) Minimal loop vocabulary.** An inbound agent message = one turn. The Dispatcher mints a `turn_id`, emits durable `turn_started` before the fold and `turn_completed` after (`error` on failure), and stamps that `turn_id` onto every event in the turn (U6's `expected_turn_id` CAS depends on this). `EmitBridge` maps the new neutral types additively.

`Raxol.Agent.SessionStreamer` is now supervised by `Raxol.Agent.Supervisor` (it previously had no home).

## Why

- **Kills the dual-id landmine.** The contract says `Event.id == journal offset`, but the bridge stamped ids from a local counter — live id and journal offset diverged the moment a durable event was journaled. The journal is now the sole id authority: one identity, zero divergence.
- **Wires the durable sink** behind the emit seam on the `raxol_agent` side, respecting the hard back-dep constraint.
- **Makes replay-as-truth real:** replaying the journal reproduces exactly the ids the live tail carried.

## Durable-id authority (append-before-publish)

Located in `Raxol.Agent.EmitBridge.handle_manager_info/2` (durable clause): append to `Journal.FileStore`, take the offset, `map_event` with that offset, emit. Ephemeral clause emits with `last_offset`, no append.

## Turn boundary

`Raxol.Core.Runtime.Events.Dispatcher.dispatch_agent_turn/2` — the agent-message cast/call handlers route here; a `turn_id` is added to `State`, `emit/4` stamps it, and the fold is bracketed.

## Acceptance tests (`packages/raxol_agent/test/raxol/agent/keystone_close_test.exs`)

1. **Dual-id regression:** live `Agent.Session` turn → durable live-tail ids == journal offsets == replayed ids (1..N).
2. **Survives BEAM kill:** durable trace close/reopen — ids intact, monotonic, ephemeral delta not journaled.
3. **Turn brackets:** `turn_started{turn_id}` … items … `turn_completed{turn_id}` with one stable `turn_id`; errored turn emits `:error`.
4. **session_id non-nil** on emitted events for a live agent session.

Full `raxol_agent` suite (973) green across seeds; main-raxol dispatcher suites green; `mix compile --warnings-as-errors` clean both packages.

## TODO / follow-ups

- v0 turn boundary limitation: `turn_completed` fires when the synchronous fold returns, so async-streaming agents emit their ephemeral `item_delta`s *after* `turn_completed` (same `turn_id`). Tightening the boundary to the last async chunk is a later unit.
- `test/test_helper.exs` now redirects journals to a throwaway tmp dir for the whole run so tests never touch `~/.raxol/sessions`.

### DIFF
--- a/lib/raxol/core/runtime/emit_bus.ex
+++ b/lib/raxol/core/runtime/emit_bus.ex
@@ -4,7 +4,7 @@ defmodule Raxol.Core.Runtime.EmitBus do
 
   This is the **keystone seam** for the harness event stream. The Dispatcher
   publishes one neutral event map at each of its two model-fold sites
-  (`process_app_update/3` and `process_command_result/2`); any process that
+  (`process_app_update/3` and `process_command_result/3`); any process that
   subscribes by `session_id` receives those maps as `{:emit_bus, session_id,
   event}` messages.
 
--- a/lib/raxol/core/runtime/events/dispatcher.ex
+++ b/lib/raxol/core/runtime/events/dispatcher.ex
@@ -42,7 +42,12 @@ defmodule Raxol.Core.Runtime.Events.Dispatcher do
               # Harness keystone: when set, both model-fold sites publish a
               # neutral event to Raxol.Core.Runtime.EmitBus keyed by this id.
               # nil (terminal/plain apps) makes emit/2 a no-op.
-              session_id: nil
+              session_id: nil,
+              # Harness loop vocabulary: the turn currently in flight. Minted at
+              # the agent-message dispatch (a prompt = a turn) and stamped onto
+              # every event emitted while it is set, so all items within one turn
+              # share a stable turn_id (U6's expected_turn_id CAS depends on it).
+              turn_id: nil
   end
 
   # BaseManager provides start_link/1 and start_link/2 automatically
@@ -412,7 +417,7 @@ defmodule Raxol.Core.Runtime.Events.Dispatcher do
       "[Dispatcher] handle_cast :dispatch agent_message: #{inspect(msg)}"
     )
 
-    with_dispatch_span(fn -> dispatch_raw_message(msg, state) end)
+    with_dispatch_span(fn -> dispatch_agent_turn(msg, state) end)
   end
 
   @impl true
@@ -425,7 +430,7 @@ defmodule Raxol.Core.Runtime.Events.Dispatcher do
     )
 
     # Agent messages go directly to update/2, bypassing event/plugin pipeline
-    with_dispatch_span(fn -> dispatch_raw_message(msg, state) end)
+    with_dispatch_span(fn -> dispatch_agent_turn(msg, state) end)
   end
 
   @impl true
@@ -593,13 +598,42 @@ defmodule Raxol.Core.Runtime.Events.Dispatcher do
   # agent Contract on the raxol_agent side.
   defp emit(%State{session_id: nil}, _type, _tier, _payload), do: :ok
 
-  defp emit(%State{session_id: session_id}, type, tier, payload) do
+  defp emit(%State{session_id: session_id, turn_id: turn_id}, type, tier, payload) do
     session_id
-    |> Raxol.Core.Runtime.EmitBus.build(type, tier, payload)
+    |> Raxol.Core.Runtime.EmitBus.build(type, tier, payload, turn_id: turn_id)
     |> Raxol.Core.Runtime.EmitBus.publish()
 
     :ok
   end
+
+  # --- Harness loop vocabulary: turn brackets ----------------------------
+  # An inbound agent message is one turn. We mint a turn_id, emit a durable
+  # `:turn_started` before the fold, run the model fold (whose own durable
+  # `:app_update` emit now carries this turn_id), then bracket with a durable
+  # `:turn_completed` on success or `:error` on failure. The turn_id persists
+  # in state so any async `:command_result` deltas that stream out of the
+  # turn's commands carry the same turn_id.
+  #
+  # NOTE (v0 limitation): turn_completed is emitted when the synchronous fold
+  # returns. Agents that stream via async commands will emit their ephemeral
+  # item_deltas after turn_completed (same turn_id). Tightening the boundary to
+  # the last async chunk is a later unit — see the module TODO.
+  defp dispatch_agent_turn(msg, state) do
+    turn_state = %{state | turn_id: mint_turn_id()}
+    emit(turn_state, :turn_started, :durable, %{message: msg})
+
+    case process_app_update(turn_state, msg, msg) do
+      {:ok, new_state, _commands} ->
+        emit(new_state, :turn_completed, :durable, %{})
+        {:noreply, new_state}
+
+      {:error, reason} ->
+        emit(turn_state, :error, :durable, %{reason: reason})
+        {:noreply, turn_state}
+    end
+  end
+
+  defp mint_turn_id, do: "turn-#{System.unique_integer([:positive])}"
 
   defp log_command_failure(:error, label, reason, context) do
     Raxol.Core.Runtime.Log.error_with_stacktrace(label, reason, nil, context)
@@ -624,7 +658,7 @@ defmodule Raxol.Core.Runtime.Events.Dispatcher do
     )
 
     with_dispatch_span(fn ->
-      msg |> dispatch_raw_message(state) |> to_call_reply()
+      msg |> dispatch_agent_turn(state) |> to_call_reply()
     end)
   end
 
--- a/lib/raxol/core/runtime/events/dispatcher.ex
+++ b/lib/raxol/core/runtime/events/dispatcher.ex
@@ -47,6 +47,8 @@ defmodule Raxol.Core.Runtime.Events.Dispatcher do
               # the agent-message dispatch (a prompt = a turn) and stamped onto
               # every event emitted while it is set, so all items within one turn
               # share a stable turn_id (U6's expected_turn_id CAS depends on it).
+              # Cleared back to nil when the turn's closing bracket
+              # (turn_completed / error) is emitted — see dispatch_agent_turn/2.
               turn_id: nil
   end
 
@@ -598,7 +600,12 @@ defmodule Raxol.Core.Runtime.Events.Dispatcher do
   # agent Contract on the raxol_agent side.
   defp emit(%State{session_id: nil}, _type, _tier, _payload), do: :ok
 
-  defp emit(%State{session_id: session_id, turn_id: turn_id}, type, tier, payload) do
+  defp emit(
+         %State{session_id: session_id, turn_id: turn_id},
+         type,
+         tier,
+         payload
+       ) do
     session_id
     |> Raxol.Core.Runtime.EmitBus.build(type, tier, payload, turn_id: turn_id)
     |> Raxol.Core.Runtime.EmitBus.publish()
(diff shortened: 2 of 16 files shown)


DROOdotFOO:
## Review — the concept is right, two failure paths reopen the exact landmine

Read through the whole thing. The core idea (journal = single id authority, append-before-publish, `Event.id == journal offset`) is clean and the happy path is well-proven — the dual-id test asserting `live_ids == replay_ids == 1..N` is exactly the property that matters, and sanitizing the payload once for both the record and the live event (so they're byte-identical, not just id-equal) is the right call.

Two source-verified concerns before this lands. Both live on paths the current tests don't exercise.

### 1. Append-failure path re-introduces the dual-id divergence this PR exists to kill

`FileStore.Writer.handle_call({:append, ...})` (`writer.ex:112-131`) deliberately does **not** advance `state.offset` on a write failure — the comment there says "the offset is NOT advanced ... the caller can retry once space frees up."

But `EmitBridge.append_durable/2` doesn't retry. On `{:error, _}` it fabricates `state.last_offset + 1`, emits the event live (unjournaled), and moves on:

```
A appends OK   -> Writer offset 5, live id 5
B append fails -> fabricated id 6, NOT journaled, Writer still at 5
C appends OK   -> Writer returns 6, live id 6
```

Live tail now carries **two** durable events at id 6; replay has only C at offset 6. Live id != journal offset != replayed id — silently, exactly the divergence the unit is closing, relocated to the disk-failure path.

The transient-open variant is worse than a duplicate. If `ensure_journal` keeps failing and then succeeds mid-stream, the Writer resumes from an empty dir at offset 1 while the bridge has already emitted fabricated ids `1..N`, so live ids go `1,2,3,1` — non-monotonic *and* duplicate, contradicting the `# so ids stay monotonic` comment on that fallback.

Also: the `{:error, _reason}` reason is swallowed — no telemetry, no log — so a disk-full corruption is invisible.

Suggest: on append failure, don't fabricate an offset. Drop/park the event or fail the turn (surface the reason). Fabricating an id that the Writer will provably reuse is the one thing this contract can't tolerate.

### 2. `turn_id` is set-once, never cleared -> stale + cross-turn misattribution

`dispatch_agent_turn/2` mints a `turn_id` onto `turn_state`, and it persists into `new_state` on `{:noreply, new_state}`. Nothing ever resets it to `nil` (there's no `turn_completed -> turn_id: nil`). Two consequences:

- **Inter-turn emits inherit a stale id.** A regular event fold between agent turns (`do_dispatch_event` -> `process_app_update` -> `emit(:app_update, :durable)`) carries the *previous* turn's `turn_id` instead of `nil`. Those events aren't part of any turn.

- **Cross-turn contamination (the one that bites U6).** `process_command_result/2` (`dispatcher.ex:541`) stamps async `:command_result` deltas with `state.turn_id` read *at emit time*. The Dispatcher is a single mailbox, so if turn N+1's `agent_message` is processed before turn N's async delta lands, that delta gets stamped with **N+1's** turn_id. A U6 CAS keyed on `expected_turn_id` would then accept turn N's leftover stream as turn N+1's — defeating the guard the turn_id exists for.

This is more than the cosmetic "deltas arrive after turn_completed" noted in the TODO. The v0 boundary being loose is fine; the id being read late from mutable shared state is what makes it a correctness issue for the CAS consumer.

Suggest, minimally: reset `turn_id` to `nil` after `turn_completed`/`error`, and snapshot the turn_id into the async command's own context at dispatch time so a late delta carries its *originating* turn's id, not whatever turn happens to be live when it lands.

### Why the suite stays green

Both concerns sit on the async/streaming and disk-failure paths. `EchoAgent` and the `:ping` test both return `{model, []}` — no async commands, no streaming — so `process_command_result` never fires and the happy-path append never fails. The identity property genuinely holds for synchronous, single-message turns; it's the streaming agent and the full disk that expose these.

Neither blocks the concept — flagging so the append-failure semantics and the turn_id lifecycle get nailed down before U6 builds its CAS on top of this.

DROOdotFOO:
## Merge bar

Concept and happy path are approved — synchronous single-message turns get exactly-once identity (live == journal == replay). What's left is the failure/streaming edges. Here's the line I'd draw for merge.

### Must fix in this PR (blocking)

1. **Fail closed on journal append/open failure — do not fabricate an offset.**
   `EmitBridge.append_durable/2` currently falls back to `last_offset + 1` on `{:error, _}`. Since `Writer` does not advance its offset on failure (`writer.ex:112-131`), that fabricated id is one the Writer will reuse on the next successful append -> duplicate live id, live != replay. This reopens the exact invariant the unit closes. Drop/park the event or fail the turn, and surface the reason (no silent `{:error, _reason}`).

2. **Reset `turn_id` to `nil` after `turn_completed` / `error`.**
   One line. Stops inter-turn emits (regular event folds between turns) from inheriting the previous turn's id. Without it, events that belong to no turn carry a stale turn_id.

### Acceptable to defer (must be tracked, not prose TODO)

3. **Async cross-turn contamination.** `process_command_result` reads `state.turn_id` at emit time, so a late turn-N delta can be stamped with turn N+1's id. Deferrable only if it becomes a **filed issue marked blocking for U6** — U6's `expected_turn_id` CAS is unsound on top of the current behavior, so this cannot merely live in the module TODO. Fix shape: snapshot turn_id into the async command's context at dispatch so a delta carries its originating turn's id.

4. **Tighter turn boundary** (turn_completed after last async chunk vs after the sync fold). Genuine v0 limitation, fine to defer as noted.

### Definition of done for approval

- [ ] Append/open failure fails closed (finding 1), reason surfaced
- [ ] `turn_id` reset on turn end (finding 2)
- [ ] Issue filed for async contamination, labeled blocking-for-U6 (finding 3)
- [ ] A streaming-agent test (async command + two overlapping turns) exists — either as a regression test for the fix, or `@tag`-skipped against the tracked issue, so the gap is visible in the suite rather than invisible behind synchronous-only fixtures

Do 1 + 2 + file 3 and this is an approve. If the upcoming fixes already cover 1, it may just need a rebase.

merklebonsai:
Thanks for the line-draw — all four points addressed. Note the review predates `0ffe817e`, which had already landed findings 1+2; finding 3 turned out surgical enough to fix in this PR (commit `4c8cedfc`), so no deferral issue was needed.

- [x] **1. Append/open failure fails closed, reason surfaced** — in `0ffe817e`. `EmitBridge.append_durable/2` no longer fabricates `last_offset + 1`: a failed durable append DROPS the event from the live tail (`last_offset` untouched) and emits a loud **ephemeral** `:error` event with `reason: :journal_append_failed | :journal_open_failed`, `original_type`, and detail, plus a `Logger.error`. On `{:writer_down, _}` the stale handle is dropped so the next durable event lazily reopens the journal and resumes from the on-disk offset — ids stay collision-free. Regression test: "append failure is a hard gate (no phantom ids)" in `keystone_close_test.exs` (kills the writer, asserts no fabricated id, offset unchanged, replay reproduces exactly the journaled events).

- [x] **2. `turn_id` reset on turn end** — in `0ffe817e`. `dispatch_agent_turn/2` nils `turn_id` on **both** arms (`turn_completed` and `error`), after the bracket event. Covered by "turn_id does not bleed past turn_completed" in `keystone_close_test.exs` (a subscription tick after a finished turn emits `app_update` with `turn_id: nil`).

- [x] **3. Async cross-turn contamination — fixed in this PR** (`4c8cedfc`) using exactly the suggested shape: `build_command_context/1` snapshots the minting turn's `turn_id` into the command context at dispatch; the raxol_agent executors (`Async`, `Shell`, `SendAgent`) echo it back as `{:command_result, msg, %{turn_id: ...}}`; a tagged Dispatcher clause attributes the emit — and any follow-up commands minted by that result's fold — to the *originating* turn. The untagged 2-tuple stays valid (attributed `nil` between turns — honest "no turn", never a stale id), and apps still receive the unchanged `{:command_result, msg}` shape in `update/2`. No issue filed since the fix landed here; happy to file one anyway if you want a paper trail.

- [x] **4. Streaming-agent test (async command + two overlapping turns) — green regression test**, not skipped: `packages/raxol_agent/test/raxol/agent/streaming_turn_attribution_test.exs`. A streaming agent dispatches an async command in turn N whose task parks until released; turn N+1 starts **and completes**; the release then produces an `item_delta` asserted to carry turn N's id — not N+1's, not nil. Plus two dispatcher-level tests in `dispatcher_emit_test.exs` (tagged delta keeps its originating id across an overlapping turn; untagged stays nil).

Finding 4 (turn_completed after the last async chunk) stays deferred as agreed — the module TODO stands, and with 3 fixed a late chunk now at least carries correct attribution while the boundary is loose.

Verification: `mix compile --warnings-as-errors` clean on main + raxol_agent; full raxol_agent suite 979 passed; main `test/raxol/core/runtime/events/` 126 passed.

DROOdotFOO:
Verified all four against the current diff — this is an approve. Each fix is real and, on two of them, more careful than what I sketched.

**1. Append/open failure fails closed** — confirmed in `EmitBridge`. The durable clause returns `{:noreply, state}` on `append_durable/2` error with `last_offset` genuinely untouched, no `map_event` with a fabricated id, and a loud ephemeral `:error` (`tier: :ephemeral`, `id: state.last_offset` so it can't masquerade as a fresh offset) plus `Logger.error`. Nice touch I didn't ask for: `drop_dead_journal/2` distinguishes `{:writer_down, _}` (drop the handle so the next durable event lazily reopens and resumes from the on-disk offset) from `:enospc` (keep the live handle for a later retry). That's the correct split.

**2. `turn_id` reset on turn end** — confirmed. `dispatch_agent_turn/2` nils `turn_id` on both arms (`{:noreply, %{new_state | turn_id: nil}}` and the error arm), after the closing bracket emits.

**3. Async cross-turn attribution** — confirmed, exactly the snapshot shape. `build_command_context/1` captures `turn_id: state.turn_id` at dispatch; executors echo `{:command_result, msg, %{turn_id: ...}}`; the tagged Dispatcher clause attributes the emit (and follow-up commands from that fold) via `process_command_result/3` using the *originating* turn, while the untagged 2-tuple honestly falls back to `state.turn_id` (nil between turns). And the snapshot is scoped to the emit/context — `process_command_commands` threads it without persisting it back into GenServer state. Good.

**4. Streaming regression test** — confirmed green, not skipped. `streaming_turn_attribution_test.exs` parks an async task in turn N, starts *and completes* turn N+1, then releases; asserts the late `item_delta` carries `turn_n` and `refute`s `turn_n1`. That's the precise contamination scenario reproduced as a guard — this is the one that matters for U6.

One caveat on my end: I verified by reading the PR diff via the API, not by running the suite (the branch commits aren't in my local tree). Trusting your CI numbers (raxol_agent 979, dispatcher 126) for the run itself.

No deferral issue needed given 3 landed here. LGTM — ship it.

PR #553, opened 2026-07-15
Harness UI: byte-capture oracle harness (PaintAuthority, fail-closed O1, immutable-prefix O2) [TB]

merklebonsai:
Unit **TB** of the harness-UI lane (`docs/proposals/in-flight/harness-ui-roadmap.md` §2): the byte-capture test harness — the oracle every renderer property (T2b/T2c) will be judged by.

## What
- **`Raxol.UI.Rendering.PaintAuthority`** — the production behaviour (append_sealed / repaint_footer / keyframe_footer / with_cursor / resize / region_top) + `Dialect` module as the single home of cursor-save/region-set byte constants (T2d's dialect switch = one line).
- **`CaptureAuthority`** — test double recording origin-tagged `%Emit{origin, bytes, seq}` at the call site: "which path emitted this byte" is fact, not inference. Raises on nested `with_cursor` (single DECSC slot).
- **Two oracles** (built additively on `test/support/cross_terminal/`): **O1** — a *fail-closed* positional scanner: a stateful row walk over a three-way token split (modeled / provably-ignorable / **unverifiable → failure**), so unknown row-affecting tokens (IL/DL/ED-partial/RIS/…) can never silently pass; **O2** — emulator replay with `history/2` (scrollback ++ rows-above-footer) and the **immutable-prefix assertion**: seal-once = history-so-far is a byte-identical prefix of final history — one property catching footer-bleed, stray CUP, `\e[2J`, and resize re-wrap.
- **`BuggyAuthority`** — 12 hand-written violating streams (absolute + relative footer-bleed, VPA bleed, sealed-row and sealed-blank rewrite, full clear, IL-shift, ED-partial, nested/unbalanced cursor save, doubled region re-set, history-touching ctrl-l) with a **mutation-completeness table test**: every invariant INV-1..6 has a kill stream the oracle must catch.
- High-water sealed-row tracking derived from the authority's own emit records (exact, never content-heuristic) + `assert_seal_newline_terminated/1` for T2b's per-run discipline.

## Evidence
184 tests green (+1 `:pending_te` skip that becomes TE's regression net on merge); every oracle upgrade carries its recorded red-run against the pre-fix oracle. compile --warnings-as-errors / format / credo --strict clean.

## Review trail
Opus adversarial review (REQUEST CHANGES: absolute-CUP-only scanner = false-pass on relative-movement bleed) → stateful row walk + red-run streams → re-review **approve**; grok-4.5 invariants advisory (unmodeled-token enumeration) → the fail-closed three-way split closing the token class structurally; TE coordination: the emulator scrollback-feed gap this unit empirically pinned promoted TE from optional to required.

Merge note: **merge AFTER the TE PR** — the spike file is TE's regression net: 3 of its assertions are expected-red standalone (documented in the commit body) and go green once TE is in the base (verified 24/24 via scratch merge against TE's actual implementation).


DROOdotFOO:
Holding this one — CI is red on Unit / macOS / Windows uniformly (not a portability quirk), and the root cause is a **merge-order dependency**, not a defect in this PR.

Three tests in `test/harness/tb_scrollback_spike_test.exs` (lines ~52, ~78, ~135) assert the emulator feeds evicted rows into `scrollback_buffer` on a top-anchored / full-screen region scroll:

- `scroll_up: top-anchored region scroll feeds scrollback, oldest-first` — got `[]`, expected `["TOPROW", "SECOND"]`
- `top-anchored eviction is preserved, not discarded` — `get_scrollback` empty
- `oracle-level: eviction keeps sealed rows comparable via history/3` — missing the evicted `"line one"`

That behavior lives on `feat/harness-ui-TE` (the emulator scrollback-feed fix), which isn't merged and has no open PR. This PR's diff touches only `paint_authority.ex` + test support — no emulator change — so against current master the rows blank out and the asserts are deterministically red. The file's own moduledoc says as much ("standalone against origin/master they are expected-red ... once both are merged"). Looks like these were un-skipped on the bet that TE precedes TB in the merge order, but TE hasn't landed.

The other 4795 tests pass; the two `[]`-asserting tests (interior-region, alternate-screen) are green on master today.

To unblock, either:
1. **Preferred** — open/land `feat/harness-ui-TE` first (or fold its emulator change into this branch), so the base actually exhibits the asserted behavior; then rebase this.
2. **Ship TB alone** — restore `@tag :pending_te` on the three tests (`:52`, `:78`, `:135`) so they're excluded until TE lands.

Left open pending your call on the ordering. Everything else in the harness-UI batch (#547, #548, #549, #550, #551, #552) is merged.

PR #557, opened 2026-07-15
feat(harness): TH — ToolCall.Hook pipeline (before_call chokepoint for U7/U8)

merklebonsai:
## What

Introduces `Raxol.Agent.ToolCall.Hook` — an ordered, registerable interception pipeline around LLM tool-call execution — and wires it as a `before_call` chokepoint into `ToolConverter.dispatch_tool_call/3`. No spend logic, no approval logic: just the seam.

## Why

Upcoming harness units **U7 (SpendGate: reserve-budget-before-call)** and **U8 (BlastRadiusGate: approval before write tools)** both assume a hookable interception point immediately before a tool runs. None existed. Building U7 first would fuse the seam and its first consumer into one untestable unit. TH ships the seam ALONE so U7/U8 register into it rather than each re-wrapping the react loop.

## Hook contract

```elixir
defmodule MyGate do
  @behaviour Raxol.Agent.ToolCall.Hook

  @impl true
  def before_call(call, _context) do
    # call = %{action: module, name: String.t, params: map, call_id: term}
    if allowed?(call), do: {:cont, call}, else: {:halt, :budget_exceeded}
  end
end
```

- `before_call(call, context)` → `{:cont, call}` (allow; the call may be transformed — later hooks and the final execution see the modified call) or `{:halt, reason}` (veto).
- `after_call(call, result, context)` (optional) may transform the result.
- Registered as an ordered list under `:tool_call_hooks` in the agent context (alongside `:tool_authorizer`). Hooks run in declared order; first `{:halt, _}` short-circuits.

### Chokepoint

Wired at `packages/raxol_agent/lib/raxol/agent/action/tool_converter.ex` in `dispatch_tool_call/3` — the single point all three tool-call paths (`Stream` react loop, `Strategy.ReAct`, `Actions.Vfs` helper) funnel through. **Zero registered hooks takes the fast path — behavior byte-identical to before.**

### Veto visibility (no new event plumbing)

A veto returns `{:error, {:vetoed, reason}}`, which flows back through the **existing** tool-error path: the react loop emits `{:tool_result, %{name: name, result: {:error, {:vetoed, reason}}}}` and feeds the LLM a `[Tool error for ...]` message, so the loop continues and the model can react.

### Failure containment (documented decision: raise = veto-with-reason)

- A hook that raises in `before_call/2` → `{:halt, {:hook_raised, mod, message}}` → `{:error, {:vetoed, {:hook_raised, ...}}}`.
- A non-`{:cont,_}`/`{:halt,_}` return → `{:halt, {:invalid_hook_return, mod, value}}`.
- A raise in `after_call/3` is contained; the result passes through unchanged.
- The react loop never crashes on a misbehaving hook.

## Accepts (all green)

- No-op observer sees every call with parsed args pre-execution (order + content asserted).
- Two hooks in declared order; the second sees the first's transformation.
- Vetoing hook → typed error, the Action's `run/2` is **never** invoked (probe traces), loop continues.
- Zero hooks → dispatch identical to the direct `Action.call/2`.
- Raising hook → contained as a tool error; loop does not crash.

14 new tests in `test/raxol/agent/tool_call/hook_test.exs` (unit + `dispatch_tool_call/3` integration + `Stream.react/2` end-to-end). Full `raxol_agent` suite: **993 passed, 0 failures**. `mix compile --warnings-as-errors` clean.

## Relationship to existing hook machinery

Mirrors the established **`Raxol.Agent.CommandHook`** idiom (module-based ordered hooks, first denial short-circuits, optional post callback) — but on the *tool-call* subject rather than *Directive* execution (`Async`/`Shell`/`SendAgent` from `update/2`). Different subject and lifecycle, so a parallel behaviour rather than an extension. Complements **`Raxol.Agent.ToolPolicy`** (`:tool_authorizer`), the boolean allow/deny gate on the same dispatch path that runs *before* this pipeline and cannot transform calls: use ToolPolicy for static policy, a `ToolCall.Hook` for stateful gates (reservations, approvals) and call rewriting. The **`Raxol.Payments.SpendingHook`** (a CommandHook) gates Pay *directives*; U7's SpendGate will register here to gate *tool calls* pre-execution.

### DIFF
--- a/packages/raxol_agent/lib/raxol/agent/action/tool_converter.ex
+++ b/packages/raxol_agent/lib/raxol/agent/action/tool_converter.ex
@@ -42,7 +42,10 @@ defmodule Raxol.Agent.Action.ToolConverter do
   Dispatch order: find action -> parse arguments -> validate limits ->
   `:tool_authorizer` (see `Raxol.Agent.ToolPolicy`) -> `:tool_call_hooks`
   pipeline (see `Raxol.Agent.ToolCall.Hook`) -> `Action.call/2`. A hook veto
-  returns `{:error, {:vetoed, reason}}` without invoking the Action.
+  returns `{:error, {:vetoed, reason}}` without invoking the Action. If a hook
+  rewrites the call's action to a different module, the authorizer is re-run
+  against that module before execution (a denial returns `{:error,
+  {:tool_denied, ...}}` and the rewritten Action never runs).
 
   Returns `{:ok, result}` or `{:error, reason}`.
   """
@@ -62,10 +65,12 @@ defmodule Raxol.Agent.Action.ToolConverter do
     end
   end
 
-  # Execution chokepoint: every LLM tool call passes through here. The
-  # `:tool_call_hooks` pipeline (Raxol.Agent.ToolCall.Hook) runs immediately
-  # before the Action -- hooks may transform the call or veto it. Zero
-  # registered hooks takes the fast path (behavior unchanged).
+  # Execution chokepoint: every LLM-initiated tool call passes through here
+  # (the internal Pipeline/Direct/run_action* paths call Action.call/2 directly
+  # but are agent-internal, never LLM-reachable). The `:tool_call_hooks`
+  # pipeline (Raxol.Agent.ToolCall.Hook) runs immediately before the Action --
+  # hooks may transform the call or veto it. Zero registered hooks takes the
+  # fast path (behavior unchanged).
   defp run_hooked(module, name, params, tool_call, context) do
     case Hook.from_context(context) do
       [] ->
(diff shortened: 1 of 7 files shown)


DROOdotFOO (commented):

Three-persona pass (Saboteur / New Hire / Security Auditor). Findings verified against branch code.

### HIGH — same-action param transform bypasses re-authorization (TOCTOU)
`tool_converter.ex` — `dispatch_tool_call/3` runs `authorize_tool(module, params, context)` against the **original** params, then a `before_call` hook may transform `params` while keeping the same `:action`. `reauthorize_if_transformed(same, same, _params, _context) -> :ok` skips re-auth on any same-action change; only an action-module swap re-runs the authorizer.

Scenario: a param-inspecting authorizer (the future U8 fund-movement/spend gate) clears `transfer(amount: 50)`. A hook rewrites `%{call | params: %{amount: 5_000_000}}`. Same action → no re-auth → the fund-mover executes at 5M. The moduledoc advertises the re-auth seam as protection against "smuggling past policy," but the guard keys on action identity, not on the `(action, params)` pair actually executed — incomplete for exactly the amount-escalation case U8 builds on. (Hooks are operator-trusted, so this is defense-in-depth against a buggy/hostile hook, but the seam claims a guarantee it doesn't provide.)

### MEDIUM
- **Rewritten action not checked against the `action_modules` allowlist** — `reauthorize_if_transformed` on a module swap calls only `authorize_tool` (sensitive-flag), never verifying the new module is in the declared toolset. A hook can rewrite `:action` to any non-sensitive module exporting `call/2` and execute a tool outside the exposed set.
- **Veto/failure reasons `inspect`-ed into LLM context + logs unredacted** — `hook.ex` `{:hook_raised, hook, Exception.message(error)}` / `{:invalid_hook_return, ...}` surface via `"[Tool error for #{name}]: #{inspect(reason)}"`. A gate raising `"insufficient balance for wallet 0x… key=…"` lands that detail in the model prompt and logs.
- **`after_call` runs forward, not reverse** — `hook.ex` `run_after` folds left-to-right like `before_call`; conventional middleware unwinds after-hooks in reverse, so with `[Audit, Redact]` the audit observes the result before Redact's after-transform. Not documented.
- **Strict return-shape silently converts a benign transform into a veto** — a hook returning `{:cont, map_without_call_id}` falls to `{:halt, {:invalid_hook_return, ...}}` → the tool is blocked, indistinguishable from an intentional deny.

### LOW
- `invoke_before` catches `:exit` and surfaces it as `{:hook_exited, ...}` → veto, masking a downed-process infra outage as an ordinary tool denial.
- The `Actions.Vfs` / `raxol_payments` moduledoc examples still show `{:ok, result} = dispatch_tool_call(...)`; a veto now returns `{:error, {:vetoed, ...}}` → the documented idiom `MatchError`s.

### CI trust
All checks green, but CI tests only the transformed-**action** re-auth path (`ActionRewriter → SensitiveProbe`). There is **no** test for the transformed-**params**-same-action path — the HIGH lives entirely in that untested gap.

*Automated adversarial review (Saboteur / New Hire / Security Auditor personas). Reasoning, not a merge block — your call.*

PR #559, opened 2026-07-15
feat(harness): MS — model snapshot contract (persistent slice, manifest, redaction)

merklebonsai:
## What

The **model snapshot contract**: a declared, JSON-safe, restorable projection of a TEA model. Two modules in `raxol_agent`, plus tests:

- **`Raxol.Agent.Snapshot.Persist`** — a protocol. An app declares its durable slice with `@derive {Persist, persist: [...], redact: [...]}`. No declaration → an `Any` fallback does a conservative auto-scan.
- **`Raxol.Agent.Snapshot`** — the codec:
  - `dump/1` → `{:ok, %{v: 1, module, data, dropped, redacted}}` — a versioned, JSON-round-trippable envelope.
  - `load/2` → rebuilds the model with persistent fields restored and everything else at struct/`init` defaults.
  - `persistent_slice/1` = `load(dump(m))` — the durable projection.

Pure: no GenServers, no file IO. **U9 (Checkpoint) owns writing envelopes into a session's `snapshots/` dir** — this unit is the contract + codec only.

## Why (U9's foundation)

U9 stores checkpoints as `{journal_offset, model_snapshot}`. But arbitrary TEA models hold PIDs, Ports, function refs, ETS refs, and secrets — none JSON-safe, none faithfully restorable. Without a declared persistence boundary, U9 "will snapshot something that either can't be written or can't faithfully resume." MS defines that boundary **now**, standalone, so U9 builds on a tested contract rather than inventing serialization mid-checkpoint.

## Contract semantics

- **Declared slice.** `persist:` is the durable allowlist; any field not listed is *explicitly excluded*. Omit it (`:auto`) for auto-scan: plain data survives, non-serializable values are dropped-with-manifest.
- **Loud manifest, never silent mangling.** Every field that did not reach `data` is accounted for: PIDs / Ports / refs / functions / tuples / undeclared nested structs land in `dropped` with a reason; secrets land in `redacted`. A reviewer reading the manifest sees exactly what the slice omits and why.
- **Redaction.** `redact:` fields — plus a name heuristic (`api_key`, `password`, `token`, …) as defence-in-depth — never reach the snapshot data. Tested with a field literally named `api_key`.
- **Restore property.** `load(dump(m)) == persistent_slice(m)` — equality on the *declared slice*, explicitly NOT whole-term equality. Dropped/redacted fields come back at their struct defaults.
- **Versioned envelope.** `v: 1`; loading an unknown future `v` is a typed `{:error, {:unsupported_version, v}}`, not a crash — this artifact outlives code versions.

### Protocol, not behaviour

TEA models are frequently bare maps with no owning module (`init/1` commonly returns `%{}`). A behaviour can only attach to a module; a protocol dispatches on the *value* and, via `@fallback_to_any`, defines conservative default behaviour for a map that has no module to host a callback. Impl resolution probes the `Persist.<Mod>` module directly rather than through the *consolidated* dispatch table, so a struct that `@derive`s after its defining app is consolidated (any downstream app, and every test fixture) is seen, not silently treated as undeclared.

## Accepts (18 tests, 1 property)

- Model with PID + secret + plain data: dump → JSON-safe (Jason round-trip), PID in `dropped`, secret in `redacted` and absent from data/JSON; load → persistent slice equal, dropped fields at defaults.
- Property: generated nested plain-data models survive dump → JSON → load with slice equality.
- Struct-in-struct: declared inner recurses; undeclared inner → dropped-with-manifest, loudly.
- Versioned envelope: `v` present; unknown future `v` → typed error, not crash.
- `load(dump(m))` idempotence: double round-trip stable (struct + bare-map).

## Verify

`MIX_ENV=test mix compile --warnings-as-errors` clean; full `raxol_agent` suite green (997 passed, 5 properties). No lock committed.

### DIFF
--- a/packages/raxol_agent/lib/raxol/agent/snapshot.ex
+++ b/packages/raxol_agent/lib/raxol/agent/snapshot.ex
@@ -1,0 +1,367 @@
+defmodule Raxol.Agent.Snapshot do
+  @moduledoc """
+  Durable serialization codec for a TEA model's **persistent slice**.
+
+  This is the JSON-safe, restorable counterpart to `Raxol.Debug.Snapshot`
+  (which captures whole in-memory `{message, before, after}` cycles for the
+  time-travel debugger). Where that snapshot lives and dies in memory, this one
+  is written to disk — it is the contract a later unit (U9 — Checkpoint) builds
+  `{journal_offset, model_snapshot}` records on. This module is the **contract +
+  codec only**: it is pure, with no GenServers and no file IO. U9 owns writing
+  envelopes into a session's `snapshots/` directory.
+
+  ## The three operations
+
+    * `dump/1` — model → a versioned, JSON-round-trippable envelope.
+    * `load/2` — envelope → the restored model (persistent fields restored,
+      everything else at struct/`init` defaults).
+    * `persistent_slice/1` — `load(dump(model))`; the durable projection of a
+      model, i.e. what survives a checkpoint round-trip.
+
+  ## The envelope
+
+      %{
+        v: 1,                       # schema version — this artifact outlives code
+        module: "Elixir.MyModel",   # struct module, or nil for a bare-map model
+        data: <json-safe encoding>, # the persistent slice
+        dropped: [%{"path" => [...], "reason" => "..."}, ...],
+        redacted: [%{"path" => [...]}, ...]
+      }
+
+  The envelope is intentionally **loud about exclusions**: every field that did
+  not make it into `data` is accounted for in `dropped` (non-persistable or
+  undeclared) or `redacted` (a secret). Nothing is silently mangled — a reviewer
+  reading the manifest sees exactly what the persistent slice omits and why.
+
+  ## The restore property
+
+  `load(dump(model))` equals the model's *persistent slice*, NOT the whole model:
+
+      load(dump(model)) == persistent_slice(model)
+
+  Equality holds on the declared slice only. Non-persistable fields (PIDs,
+  Ports, refs, functions), undeclared fields, and secrets come back at their
+  struct defaults — by design, not by accident.
+
+  ## Encoding grammar
+
+  Every persisted value maps to a self-describing, JSON-safe term:
+
+    * `nil`, booleans, numbers, binaries — encoded bare.
+    * other atoms — `%{"$a" => "name"}` (tagged strings; restored with
+      `String.to_existing_atom/1`, falling back to the raw string if the atom
+      no longer exists — safe, never a leak, never a crash).
+    * lists — JSON arrays, **all-or-nothing**: a list holding a non-serializable
+      element is dropped whole (positional partial-drop would silently reindex).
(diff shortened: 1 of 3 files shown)


DROOdotFOO (commented):

Three-persona pass. Findings reproduced by exercising `snapshot.ex` + `snapshot/persist.ex` against Jason.

### HIGH — non-UTF-8 binary silently defeats the "JSON-safe envelope" guarantee
`snapshot.ex` — `encode(v, _path) when is_binary(v) -> {:keep, v, [], []}` keeps every binary bare with no UTF-8 validity check. `dump/1` returns `{:ok, envelope}` (nothing dropped/redacted), but the envelope is not JSON-encodable:
```
dump(%{buf: <<0xFF,0xFE,0x00>>}) => {:ok, env}, dropped=[], redacted=[]
Jason.encode!(env)             => ** invalid byte 0xFF in <<255,254,0>>
```
Any model field holding packed bytes / a terminal buffer / crypto material passes `dump/1` clean, then U9 crashes at write time — the exact "can't be written" failure MS exists to prevent. The `dump/1` docstring ("JSON-safe envelope … only failure is a top-level non-persistable value") is false. Fix: `String.valid?/1`-gate binaries, drop-with-manifest or base64-tag.

### MEDIUM
- **Atom keys/values downgrade to strings on cross-release restore** — `safe_atom/1` falls back to the raw string when `to_existing_atom` fails. Proven across two BEAM runtimes: `dump(%{k: :v})` restored in a fresh VM → `%{"k" => "v"}`. Breaks `load(dump(m)) == slice(m)` — the module's central property — exactly on the release-N→N+1 durability path the snapshot exists for. Invisible because tests are `async: true` on one node where atoms stay interned.
- **Name-based redaction misses common secret spellings** — `@redact_segment`/`@redact_suffix` require exact words; under `:auto` scan (the documented default) `creds`, `privkey`, `pwd`, `seed`, `apisecret`, `ssn` all persist in cleartext to the JSON (reproduced `leaked_to_json=true` for each). Tests only assert the words that DO match → false confidence.
- **Name heuristic overrides the explicit `persist:` allowlist** — `encode_fields/4` tests `sensitive?(k)` before persist-membership, so a redaction-name match wins even when the field was explicitly allow-listed. `@derive {Persist, persist: [:client_secret_version]}` → silently redacted, restores `nil`. A caller cannot force-persist a field whose name trips the heuristic.

### LOW
- `coerce/2`'s `struct(module, fields)` path skips the `persist_impl?/1` check the `$s` decode path enforces — inconsistent tamper hardening when `load(envelope, SomeModule)` is called on an untrusted envelope.
- The redaction word-list and the "heuristic beats persist" rule live only in code comments, undiscoverable at the `@derive`/`Persist` declaration site.

### CI trust
All checks green, but: `async: true` single-node hides the atom-drift MEDIUM; redaction tests assert only matching words (abbreviation gaps unasserted); no test feeds a non-UTF-8 binary → the HIGH is never exercised. Green validates happy paths, not the durability/leak edges this contract exists to guarantee.

PR #561, opened 2026-07-15
Harness UI: overlay picker primitive + list scorer [T14]

merklebonsai:
Unit **T14** of the harness-UI lane (`docs/proposals/in-flight/harness-ui-roadmap.md` §2): the overlay picker primitive + list scorer — one fzf-shape component behind every "pick one of N" in the harness (sessions, runs, tool calls, palette, file mentions), per AD-U3.

## What
- **`Raxol.UI.ListScorer.rank(items, query, key_fn, opts)`** — pure fuzzy scorer, fzf-style Smith-Waterman-lite (adjacency / word-boundary / camelCase / position bonuses), case-insensitive, stable sort. Grapheme-based and Unicode-safe; returned match positions are grapheme indices safe for TextMeasure column mapping. Deliberately NOT `Raxol.Search.Fuzzy` (that's buffer-cell search — a different category). Properties: rank is a permutation-subset of input (no invented/dup items); adding a matching char never grows the result set (monotone narrowing).
- **Bench**: 10,000-item corpus, query narrows in **~11-12ms/keystroke** under the 16ms target (three profiling rounds: hoisted bonus computation, an ASCII fast-path replacing `String.graphemes`, list-walk DP). The pathological all-match case (~48-50ms) is measured + documented, not asserted — degenerate for real corpora (tens–low-thousands, fuzzy filtering makes it rare).
- **`Raxol.UI.Components.Harness.Picker`** — prompt + ranked windowed list + optional async preview. Type-to-filter, Up/Down + clamp, Enter selects, Esc cancels. **Stale-preview cancellation** (the fzf#3134 race, double-guarded): `Task.shutdown(:brutal_kill)` + ref-matching drop of any superseded result, so a late preview from a prior selection can never land in state.
- Match-position highlight via style attrs (no raw ANSI); all display width via TextMeasure.

## RED fixed in review
Type-to-filter was **dead on the native terminal driver** — the printable clause required a `modifiers:` field that only the test helper sets; `event_translator` emits boolean `shift:/ctrl:/alt:`, `input_parser` emits bare `%{key: char}`. Fixed defensively (a `text_input?/1` predicate reading all three encodings — ctrl/alt = shortcut, plain/shift = text). Red-run recorded (3 failures on raw driver shapes); regression tests now drive `event_translator`- and `input_parser`-shaped events directly, not just the helper. (The shared normalization is being extracted as unit T27; this fixes inline meanwhile — same input-shape gap that hit the composer.)

## Evidence
47 tests green (2 properties + 45; bench excluded by default); compile --warnings-as-errors / credo --strict / format clean.

## Review trail
Opus review (REQUEST CHANGES: typing dead on the real driver, masked by the test helper's event shape) → defensive fix + raw-shape regression tests → the `preview_fn` cancel-safety contract documented.

### DIFF
--- a/lib/raxol/search/fuzzy.ex
+++ b/lib/raxol/search/fuzzy.ex
@@ -23,6 +23,10 @@ defmodule Raxol.Search.Fuzzy do
 
       # Regex search
       results = Fuzzy.search(buffer, ~r/h.llo/, :regex)
+
+  Note: list-item ranking (picker/palette/session filtering) lives in
+  `Raxol.UI.ListScorer`, a separate module -- this one searches buffer
+  cells, a different problem category.
   """
 
   require Logger
--- a/lib/raxol/ui/components/harness/picker.ex
+++ b/lib/raxol/ui/components/harness/picker.ex
@@ -1,0 +1,578 @@
+defmodule Raxol.UI.Components.Harness.Picker do
+  @moduledoc """
+  Overlay picker primitive (AD-U3): prompt + ranked list + async
+  cancelable preview -- the fzf-shape that serves every "pick one of N"
+  (sessions, runs, tool-calls, command palette, file mentions). One
+  primitive, many projections (T15's job); this unit builds the
+  component itself.
+
+  ## Substrate honesty
+
+  This component renders in **ordinary buffer mode** -- a ranked list
+  laid out in a normal flow column, no `AbsoluteLayer`/`CellDim`
+  full-viewport cover. The overlay-over-inline SUBSTRATE question (does
+  summoning this picker as a screen overlay mean a scoped full-viewport
+  cover via terminal save/restore, or a footer-anchored expansion?) is
+  deferred to the D-PA/T0 verdict and lands with T15/T24, which mount
+  this component inside whichever substrate they choose. Nothing here
+  assumes a particular host; `render/2` just needs a `context` with
+  `:available_width` (and optionally `:available_height`, used to size
+  the visible window) the way any other component does.
+
+  ## Filtering
+
+  Uses `Raxol.UI.ListScorer.rank/4` (the list-item fuzzy scorer) against
+  `key_fn.(item)` for every item on every query change -- **not**
+  `Raxol.Search.Fuzzy`, which searches buffer cells, a different problem.
+
+  ## Props
+
+    * `:items` -- required, the full unfiltered list.
+    * `:key_fn` -- required, `item -> String.t()`; derives both the
+      search key and the rendered label for each item.
+    * `:on_select` -- message tag emitted on Enter, default `:select`.
+      Emits `{:component_event, id, {on_select, item}}` for the
+      currently selected item; a no-op if the ranked list is empty.
+    * `:on_cancel` -- message tag emitted on Escape, default `:cancel`.
+      Emits `{:component_event, id, on_cancel}`.
+    * `:preview_fn` -- optional, `item -> {:ok, content} | {:error,
+      reason}`, run via `Task.async/1` for the currently selected item.
+      When absent, no preview pane renders at all. **Must be
+      side-effect-free / cancel-safe:** on any selection change the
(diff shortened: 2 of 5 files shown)


DROOdotFOO (commented):
## Adversarial review — Verdict: CONCERNS (lowest risk of the batch)

Three-persona pass. Runtime-confirmed against the pinned PR commit. Bounds handling and no-ReDoS are genuinely solid — no HIGH.

### MEDIUM
- **[Saboteur / Security] Unbounded work + allocation per keystroke** — `list_scorer.ex` `align/3` builds Q rows of K floats (+ parent list + `static_bonus_list`) with nothing bounding K (label length) or item count. The cheap `subsequence?/2` gate only helps non-matches. A picker fed attacker-influenced labels (file paths, session names, tool-call args) where a short query is a subsequence of every long label forces the full O(Q·K) DP + backtrack per item per keystroke; `refresh_preview -> Task.async` is likewise uncapped. The pathological bench measures ~50ms for 10k *short* all-match labels; unbounded K scales further. Add a label-length clamp / item cap.
- **[New Hire] Two "fuzzy" scorers, no back-reference** — `Raxol.UI.ListScorer` (new) vs `Raxol.Search.Fuzzy` (existing, and the ReDoS-capable one — it `Regex.compile!`s input; ListScorer does not). Different categories per the moduledoc, but `Search.Fuzzy` has no reciprocal pointer and a newcomer grepping "fuzzy" lands on two divergent implementations. Add a cross-reference + one-line "which to reach for."

### LOW
- `picker.ex` `update/2` map catch-all → `merge_props` `Map.merge(state, props)`. `%Event{}` is a map, so an Event routed to `update/2` corrupts state instead of being ignored. Legal-props contract undocumented/unenforced.
- Negative-score gap accumulation coupled to the `@neg_inf/2` sentinel: at ~5e8 accumulated gap a genuine subsequence match is reclassified `:no_match` and silently dropped (contradicts the "subsequence kept" contract). Practically unreachable (~500MB label) but a latent soundness coupling.
- Magic scoring constants (`@match_base 10.0` etc.) individually commented but relative calibration unexplained and unpinned by any exact-ordering test → retuning risks silent ranking regressions. Moduledoc claims `positions` are safe as TextMeasure display columns; grapheme index != display column for double-width CJK (doc nit — Picker highlights by grapheme chunk, which is correct).
- Empty-label item (`key_fn -> ""`) renders a blank selectable, un-highlightable row.

### Cleared (on record)
No ReDoS — the query is never compiled to a regex. Bounds are handled: `update_query` resets cursor/scroll on filter, `move_cursor` guards `[]`, render windowing self-heals stale `scroll_top`. The "index out of bounds when the list shrinks" attack does not land.

### CI trust
47 green tests use short ASCII/CJK labels + small lists; nothing exercises long labels, empty-label render, or the negative-score boundary, and the two perf tests are `:bench`/`:slow` (excluded by default). Green = "correct on benign small input," not "bounded under adversarial input."

merklebonsai:
Review-round fixes pushed (`3f7a13d2`), addressing all findings:
- **MEDIUM unbounded O(Q·K)**: `@max_score_graphemes 1024` + `@max_key_bytes 4096` — raw key sliced (UTF-8-safe) *before* grapheme split (the split itself was the cost); full label preserved in `key:` so highlighting is unaffected. Adversarial test: 1M-grapheme label ranks in ~0.7ms (<16ms), still matches.
- **update/2 props allowlist** (upgraded per cross-review): `Map.take` against an explicit allowlist before merge — closes both the mis-routed `%Event{}` and the arbitrary-map `update(%{cursor: 99})` ScrollWindow-clamp bypass.
- **empty-label**: dim `(no label)` placeholder so the row stays selectable/visible.
- **Search.Fuzzy** cross-reference note added.
- Confirmed (cross-review): the negative-score `@neg_inf` sentinel is unreachable given the clamp (gap ≪ 5e8) — no separate fix. 52 tests green.

PR #556, opened 2026-07-15
test(harness): invariant suite — permanent guards for the id/turn/storage contract

merklebonsai:
## What

Property-based invariant suite at `packages/raxol_agent/test/invariants/` — the permanent guards for the harness contract. Purpose: not "is today's behavior correct" but "these properties cannot be silently broken by a future refactor." Spec: `docs/proposals/in-flight/harness-invariants.md` (born from the #547 review convergence + grok-4.5 ideation, brief 16).

- **Tier 1 (running now):** I1 id-authority under injected failure/crash (dense ids, no fabrication, HEAD ≤ journal), I2 ephemeral/durable wall + post-stamp byte-identity, I3 journal-before-publish observability, I4 turn attribution under generated overlapping async turns, I5 recovery fuzz (byte-cut, corrupt interior/HEAD, mid-UTF-8, empty/zero-length segments), I6 rotation continuity + no-delete, I7 single-writer succession, I8 HEAD discipline under brutal kills, I9 deep contract-only-grows (snapshot fixture + golden v1.0.0 journal corpus), I10 immediate-sync durability under real `:kill`.
- **Fault harness** (`support/fault_journal.ex`): 5 named fault sites with per-run fire counters (`assert_all_fired!` — a dead injector fails the suite), seed-reproducible schedules, independent oracles (raw file read + own decoder; subscriber process, never publisher return values).
- **Tier 2:** 9 `@tag :pending_unit` skeletons for U4/U5/U6/U7/U8/U9 — visible in the suite, activate with their units.
- **Mutation checklist:** 13 `@tag :mutation` negative controls, each anchored to the exact production line to mutate (anchor existence asserted so the checklist can't rot).

## The suite found 3 real bugs on day one (fixed here)

1. **Reader silent-gap:** a missing/truncated interior segment replayed as `{:ok, records}` with a silent id gap — fabricated continuity. Now: any id gap ⇒ `:damaged` + alarm.
2. **Decodable torn tail corrupts the next write:** a crash cut exactly after a closing `}` (no newline frame) was accepted as a record and left on disk; the next append concatenated onto the same line. Now frame-strict: an unterminated final line is torn even when it decodes.
3. **Open failure kills the caller:** an unwritable session dir made `Writer.init` raise through the link, so `FileStore.open`'s `{:error, _}` contract and EmitBridge's documented `:journal_open_failed` fail-closed arm were unreachable for real filesystem failures. Now: non-raising layout pre-flight in `FileStore.open/2`.

## Verification

[...]

PR #562, opened 2026-07-15
Harness UI: journal-fold projection (events → blocks, recovery, identity) [T7]

merklebonsai:
Unit **T7** of the harness-UI lane (`docs/proposals/in-flight/harness-ui-roadmap.md` §2): the journal-fold projection — folds a durable event stream into the ordered block list that every harness surface renders, and defines the identity that reattach/resume rests on.

## What
- **`Raxol.Harness.Projection.project(session_or_events, opts)`** — durable events → ordered, turn-grouped `Block` list; ephemeral `item_delta` → a live-tail state (never a durable block). Retains raw events per block (`refold/1` for D-PA policy (B) re-emission and retroactive opaque-kind recognition). `tip/1` is conversational **by construction** (blocks are built only from `:loop` events; meta/untyped partitioned out before build).
- **Recovery policy table** — 7 corruption classes, each recovered-not-crashed and each firing `[:raxol, :harness, :projection, :recovered]` telemetry: duplicate id (idempotent), out-of-order id (dropped loud), orphan `item_completed` (recovered block), missing `turn_started` (synthetic turn), late `item_delta` after seal (dropped, never touches the sealed block/tail), unknown `item_type` (opaque, `raw_kind` preserved), untyped/meta record (never a block, never the tip — the Dormammu/FI-12 guard). A block can carry multiple recovery reasons.
- **Two identity keys** (split per the reattach architecture): **`transcript_identity/1`** — event_refs-keyed, excludes `fold_defaults` and strips `recovered_reasons`: the "is this the same conversation?" key for T18 reattach-consistency (invariant to per-surface fold preference + recovery re-annotation, so the restoration diff never churns). **`identity/1`** — `{transcript blocks, fold_defaults}`: the T13a regression-freeze key (a fold-default change is a wanted reviewed diff). Both drop a block's mutable `fold`, so a UI toggle perturbs neither.
- **Duration fix**: merged `tool_use`+`tool_result` produces the full span (first-started → last-completed), not the first pair; a spanless orphan keeps `duration_ms: nil` (unknown), never a misleading 0.

## Evidence
42 tests (35 examples + 7 properties). Determinism (re-run / ephemeral-shuffle / meta-noise invariance), generalized replay-from-any-offset (P-DET-04, 120 runs), and the **assembly property P-ASM** — split a journal at any durable boundary, project both halves independently, concat: `transcript_identity` equals the whole-journal projection (proves two surfaces attaching at different offsets converge — the reattach-resume contract). Anti-stub: every assertion anchors on real fixture payload strings. Six `.t7blocks.json` frozen snapshots (separate namespace from TF's Identity-pinned `.blocks.json`, untouched) with a drift tripwire.

## Load-bearing invariant (documented for T18)
Reattach converges only when replay is in journal offset order (never a peer's live tail — transient out-of-order drops differ); ids strictly monotonic, offset canonical. T18's restoration diff = an event_refs-keyed diff with recovered-reasons masked, never a position-only diff, never a direct tail diff.

## Review trail
Opus review: PASS (determinism, both identity leak-guards, full recovery table with telemetry, tip-by-construction, duration merge — all correct, non-stub-passable). Two yellows → offset property generalized + T7 snapshot tripwire added; longcat conceptual advisory → the transcript_identity/identity split + the assembly property (the reattach foundation).

### DIFF
--- a/lib/raxol/harness/projection.ex
+++ b/lib/raxol/harness/projection.ex
@@ -1,0 +1,305 @@
+defmodule Raxol.Harness.Projection do
+  @moduledoc """
+  The journal-fold projection: durable events -> an ordered block list;
+  ephemeral `item_delta` -> a live-tail state, never a durable block.
+
+  Roadmap unit T7 (journal-fold projection). Test design covers P-DET/
+  P-TIER/P-FOLD properties, N-ADV/N-DORM/N-SEAL/N-FWD negatives, and the
+  recovery policy table documented inline across this module,
+  `Raxol.Harness.Projection.Recovery`, and
+  `Raxol.Harness.Projection.BlockBuilder`.
+
+  ## Pipeline
+
+  `project/2` accepts either a `Raxol.Harness.Fixture.Session` (loaded
+  from a `.jsonl` fixture) or a plain list of event-shaped maps (the
+  property-test generators build these directly, matching the fixture
+  wire shape -- string-keyed payloads, atom top-level fields):
+
+    1. `Raxol.Harness.Projection.Recovery.filter_ids/1` -- global
+       id-monotonic recovery: duplicate/out-of-order (N-ADV-02/03) are
+       dropped; a forward gap (a dense-id journal missing one or more
+       ids, e.g. 1, 2, 5 with 3 and 4 never arriving) is accepted --
+       soft-render, survivor blocks are never withheld -- but sets
+       `damaged?: true` (hard-mark) on the returned tuple, threaded onto
+       the `%__MODULE__{}` struct's `damaged` field. Mirrors
+       `Raxol.Agent.Journal`'s own `status/1 :: :damaged` contract: the
+       journal fails closed on interior corruption upstream, so a gap
+       reaching this filter means events were lost somewhere between
+       journal and here -- diagnose and mark, don't withhold, so a
+       downstream consumer (T18) can refuse a gapped tail as canonical.
+    2. `Recovery.partition_families/1` -- `:loop` vs `:meta` vs
+       untyped/unrecognised (N-DORM-04, including a non-atom `:type` on
+       an otherwise-loop record); only `:loop` events ever become blocks.
+    3. `Recovery.bucket_by_turn/1` -- groups by `turn_id` in
+       first-seen order (N-ADV-05), independent of raw arrival order.
+    4. `Raxol.Harness.Projection.BlockBuilder.build_turn/2` per turn --
+       folds items into blocks, drops late deltas (N-SEAL), caps the
+       live-tail delta buffer for an item that never completes, flags
+       orphans/unknown kinds recovered (N-ADV-04, N-FWD-01), merges
+       well-formed `tool_use`+`tool_result` pairs into one `:tool_call`
+       block and fixes its duration span (the STATE-note bug: Block's
+       own `duration_from_timestamps/1` only sees the first
+       started/completed pair in a merged list).
+
+  Every recovered condition -- duplicate, out-of-order, forward gap,
+  orphan, late delta, capped delta buffer, missing turn_started, untyped
+  record, unknown item_type -- emits
+  `[:raxol, :harness, :projection, :recovered]` telemetry via
+  `Recovery.emit/2`. Recovery is never silent; the resulting
+  `durable_block_list` is always deterministic (same input, same
+  output, no ambient state).
+
+  ## The `damaged` field
+
+  `true` iff `Recovery.filter_ids/1` detected at least one forward id
+  gap in the input. This is projection-level METADATA, not transcript
+  content: it is not part of either `transcript_identity/1` or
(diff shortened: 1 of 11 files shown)


DROOdotFOO (commented):

Three-persona pass. Reviewed at PR head in an isolated worktree; id-gap and tail-collision findings run against the actual code.

### HIGH — torn/partial journal silently fabricates continuity; id gaps neither dropped, diagnosed, nor marked `:damaged`
`recovery.ex` `classify_id/6` handles exactly two anomalies — duplicate (`MapSet.member?`) and out-of-order (`id < max_id`). A **forward gap** (ids 1, 2, 5 with 3, 4 lost to a torn write) hits the `true ->` branch and is accepted clean. Verified:
```
Recovery.filter_ids([%{id: 1}, %{id: 2}, %{id: 5}])
=> accepted ids [1,2,5], diagnostics []   # gap detected? false
```
This contradicts the module's own contract (`projection.ex` "Recovery is never silent"; `recovery.ex` "Recovery here is never silent") and diverges from the sibling `Raxol.Agent.Journal`, where interior loss → `:damaged` + hard alarm + damaged content never returned. T7 takes the same interior-loss shape and emits a normal, undamaged transcript — the id-gap ⇒ `:damaged`+alarm guarantee the just-merged invariant suite (#556) exists to enforce. Add a `id > max_id + 1` gap diagnostic (and `:damaged` marking) to the `true ->` branch.

### MEDIUM
- **Cross-turn tail identity collision drops a live item** — `block_builder.ex` `build_tail` keys by raw `item_id`; `projection.ex` merges per-turn tails with `Map.merge`. `item_id` isn't unique across turns, so turn B's incomplete `"i1"` overwrites turn A's live-tail preview with no diagnostic. Ephemeral, but real cross-identity data loss; craftable by reusing `item_id`s.
- **Unbounded allocation from an unsealed item's deltas** — `fold_event(:item_delta, …)` drops late deltas only after the item is in `acc.completed`. An item that emits `item_started` + an unbounded run of `item_delta` with no `item_completed` never seals; every chunk is prepended to `acc.deltas[item_id]` and surfaced in the tail, uncapped.
- **`source_events` retains the entire id-recovered journal (incl. ephemeral deltas) for the struct's life** — `projection.ex` stores `source_events: id_ok` for `refold/2`, but P-DET-05 proves deltas never affect identity → retaining them is provably unnecessary for refold while making the projection O(full journal) resident. Monotonic growth on a long session.

### LOW / New Hire
- Every module + both test files anchor design authority on `docs/proposals/in-flight/harness-ui-roadmap.md` and `.../harness-ui-testing/06-projection.md` — **neither path exists** in the tree. A newcomer can't locate the spec naming every property/recovery rule the code claims to implement.
- Two-site coupling of `[:recovered, :recovered_reasons]` (merged in `block_builder`, independently `Map.drop`ped in `projection.ex` for `transcript_identity`) — hand-synced; a third recovery key leaks into the reattach identity.
- Naming inversion: `identity/1` (freeze key) is the *superset*; `transcript_identity/1` is the subset — the shorter name is the broader concept.
- Public `project/2` accepts `[map()]` but dispatches on atom `:type`; a caller passing string `"item_started"` hits the `_other` catch-all and is silently discarded.

### Credit
No atom exhaustion — unknown item-type strings pass through unchanged and fold to `:opaque` via `normalize_kind`; no `String.to_atom` on untrusted input.

### CI trust
Tests are thorough on modeled cases (dup id, out-of-order, orphan, late-delta, interleaved turns) with real content + 6 golden snapshots — but there is **no test feeding a forward id gap**, so green here does not evidence torn-journal safety. That's exactly where the HIGH lives.

merklebonsai:
Review-round fixes pushed (`4b0dd816`):
- **HIGH forward id-gap** → `recovery.ex` `classify_id` gains an `id > max_id + 1` arm: accepts the event (soft-render) but emits `:forward_id_gap` telemetry AND sets a new `damaged` field on the projection (hard-mark), mirroring `Raxol.Agent.Journal`. First id of any `filter_ids` is exempt so a valid suffix-replay/reattach isn't misdiagnosed. `damaged` excluded from both identity keys. New `N-FWD-GAP` tests: the `[1,2,5]` gap (event_id 5 + telemetry + damaged), the contiguous false-positive guard, and the suffix-replay guard — plus the adversarial fixture's real id-7→8 gap now asserts too. (This closes the divergence from the #556 invariant suite.)
- **cross-turn tail collision**: `build_tail` keys by `{turn_id, item_id}` (was raw item_id) — two turns' unsealed `i1` both survive.
- **unbounded delta buffer**: sliding window of the newest 200 chunks/item + a `:delta_buffer_capped` diagnostic.
- **ephemeral in source_events**: refold retains durable-only; identity unchanged (P-DET-05).
- **non-atom :type** now emits `:untyped_record` instead of silent drop (parallel to the id-gap contract breach).
- doc anchors repointed; recovery-meta-keys de-duplicated to one accessor. 117 tests green.

PR #560, opened 2026-07-15
Harness UI: prominence attribute (H-K resolution, ground-aware fade, opt-in legibility clamp) [T8]

merklebonsai:
Unit **T8** of the harness-UI lane (`docs/proposals/in-flight/harness-ui-roadmap.md` §2): the prominence attribute — H-K salience resolution that turns a `0.0..1.0` prominence into a ground-aware color, the mechanism behind the attention-tiered transcript (the category-empty moat).

## What
- **`Raxol.UI.Harness.Prominence.resolve(hex, prominence, opts)`** — default is a **pure ground-aware fade, no floor**: the salience gradient. Context text is *meant* to recede; "faded but not lost" = legible on promotion (T9 raises a focused/approval block to prominence 1.0 = identity = the seed color), not readable-at-a-glance.
- **Opt-in legibility floor** (`legibility_floor: true`, default false): engages a WCAG-contrast clamp against the terminal ground. T9 sets it true ONLY for acting/interactive tiers (current-turn, needs-input, composer) where blind-reading risk lives; context history fades free. This is a deliberate correction to the suite-design doc's "floor everywhere," which fought the gradient.
- **Two latent bugs fixed** (found by the suite design, recorded red-runs): **F1** — the fade now threads the actual OSC-11-detected ground, so on light themes contrast *decreases* as prominence drops (a hardcoded-dark reference made it increase — the Claude-Code light-theme failure class, latent in our own DiffViewer). **F2** — a legibility clamp distinct from the prominence floor (the 0.4 floor guarded the input multiplier, not output contrast; a dim source at 0.4 could be unreadable).
- The clamp bisects along the single fade line (apparent-lightness and chroma move together), so its `t=1.0` ceiling is the **true full-chroma prominence:1.0 color**, never a reduced-chroma proxy; best-effort + documented when even full strength misses the floor.
- `block.ex` threads `prominence`/`ground`/`legibility_floor` through render context — strict no-op at prominence ≥ 1.0 or absent (T4's tests byte-identical).
- Documented: the floor is truecolor-only; 256-color survival is deferred to a redistributed `tiers_for(ground, :color256)` ladder (T9 must not assume the clamp survives quantization).

## Evidence
[...]

PR #563, opened 2026-07-15
Harness UI: streaming markdown body (provisional close, seal parse, GFM tables) [T26]

merklebonsai:
Unit **T26** of the harness-UI lane (`docs/proposals/in-flight/harness-ui-roadmap.md` §2): the streaming markdown message body — renders markdown for message/reasoning blocks without the incomplete-construct flash the cohort research flagged (P2).

## What
- **`Raxol.UI.Components.Harness.MarkdownBody`**: two modes. `:sealed` = full parse via `MarkdownRenderer`. `:streaming` = `provisional_close` (the remend pattern) then parse — incomplete constructs at the live tail are provisionally closed FOR RENDERING ONLY, source never mutated, so a partial `**bold` or unclosed ``` never shows raw markers mid-stream.
- **provisional_close is O(n), single-pass** (rewritten from an O(n²)/Θ(n³)-streaming first cut): one left-to-right scan builds an open-construct stack (O(1)/grapheme — killed the per-char full-stack map), one tail-walk strips trailing content-free openers and closes the rest (no re-scan-per-strip). >256KB safety-belt degradation.
- Dangling zero-content openers are *stripped*, not closed (`****` would leak the markers back out); a chain of nested empty openers collapses in the same walk.
- **GFM tables** (additive to MarkdownRenderer's builtin path): header+separator lookahead renders an aligned table; narrow widths shrink proportionally to a 3-column floor (never zero-width collapse — the glow failure); below the floor cells clip with ellipsis.
- `block.ex` opt-in: `context[:markdown]` (default false) routes `:message`/`:reasoning` through MarkdownBody — byte-identical when omitted (T4's tests untouched-green).

## Evidence
48 markdown_body + 40 markdown_renderer + 40 block tests green. Prefix-property: all 17 delta-boundary prefixes of the markdown-stream golden + char-granular (~500) + a 200-run byte-cut property — zero raw-marker leaks; sealed render = full parse. Fuzz: never-crash/never-hang across binary/prefix/pathological-nesting/adversarial-unicode. **Large-input fuzz (N-MDFUZZ-05)**: 20k brackets = 6.5ms, 50k = 8.9ms, mixed = 131ms, cumulative-streaming (provisional_close on every prefix 1..2000, the Θ(n³) real-world case) = 149ms — all far under the 2s bound.

## Review trail
Opus review: APPROVE with two yellows → the O(n²) single-pass rewrite (concrete algorithm from a grok-4.5 tests-invariants advisory) + large-input fuzz lock; doc-overclaim corrections (table "scrollable" → clip/overflow; premature-frame behavior).

**Merge note: merge this BEFORE the T8 PR (#560)** — both touch block.ex render context; this changed `build_render/2→/3`, T8 rebases onto that signature (trivial union — each reads its own context key).

### DIFF
--- a/lib/raxol/ui/components/harness/block.ex
+++ b/lib/raxol/ui/components/harness/block.ex
@@ -88,24 +88,32 @@ defmodule Raxol.UI.Components.Harness.Block do
   `exit_code`, `duration_ms`, and `cost` are all `nil`; otherwise it renders
   only the fields that are present.
 
-  ## Prominence (T8)
+  ## Prominence
 
   `context[:prominence]` (`0.0..1.0`) resolves the header/content/outcome
-  text colors through `Raxol.UI.Harness.Prominence` -- the H-K salience
-  solver, ground-aware. `context[:ground]` overrides the ground lightness
-  (default: OSC-11-detected, see
-  `Raxol.UI.Theming.SalienceTheme.detect_ground/0`).
+  text colours through `Raxol.UI.Harness.Prominence` -- a salience solver
+  that fades a colour toward the background as prominence drops.
+  `context[:ground]` overrides the background lightness (default:
+  terminal-detected, see `Raxol.UI.Theming.SalienceTheme.detect_ground/0`).
   `context[:legibility_floor]` (default `false`) is threaded through to
-  `Prominence.resolve/3`: the default is the **pure salience-gradient
-  fade** (context text recedes, legible on promotion); T9 sets it `true`
-  only for acting / interactive tiers where blind-reading risk lives (see
-  the `Prominence` moduledoc's "Two modes"). **Default is neutral**: when
-  `:prominence` is absent from `context`, or is `1.0`, no style is touched
-  -- the render is byte-identical to a pre-T8 render (no `:fg` added to any
-  style map). This is deliberate (roadmap unit T8's regression guard,
-  SAL-P-06): existing callers that never pass `:prominence` see zero change.
+  `Prominence.resolve/3`: the default is a pure fade (context text recedes,
+  becoming legible again as it is promoted); set it `true` for interactive
+  tiers where a minimum legibility must be preserved (see the `Prominence`
+  moduledoc's "Two modes").
+
+  When `context[:markdown]` is enabled, the Markdown body is faded to the
+  same resolved colour as the header, so the whole block dims together.
+
+  **Default is neutral**: when `:prominence` is absent from `context`, or
+  is `1.0`, no style is touched -- the render is byte-identical to a render
+  without prominence (no `:fg` added to any style map), so existing callers
+  that never pass `:prominence` see zero change.
+
+  The colour is resolved once per `render/2` call and threaded into every
+  branch, so a multi-line body never re-runs the solver per line.
   """
 
+  alias Raxol.UI.Components.Harness.MarkdownBody
   alias Raxol.UI.Harness.Prominence
   alias Raxol.UI.TextLayout
   alias Raxol.UI.TextMeasure
@@ -343,6 +351,19 @@ defmodule Raxol.UI.Components.Harness.Block do
   @doc """
   Renders `block` as a plain view map. `context[:width]` sets the wrap/
   truncation budget (defaults to `Raxol.Core.Defaults.terminal_width/0`).
+
+  `context[:markdown]` (default `false`, additive/opt-in) routes a
+  `:message`/`:reasoning` block's text content through
+  `Raxol.UI.Components.Harness.MarkdownBody` instead of the plain
+  line-split body: `:sealed` mode while the block is `:sealed`, `:streaming`
+  (provisional-close) while it is still `:live`. Every other kind, and
+  every block when the option is omitted, renders exactly as before.
+
+  `context[:prominence]` (see the moduledoc's "Prominence" section) fades
+  the header, content, and outcome to one resolved colour. A Markdown body
+  fades in lockstep -- its text nodes carry the same colour as the header,
+  so a faded header never sits above a bright body.
+
   Never raises: any unexpected internal shape falls back to a one-line
   placeholder rather than crashing the caller.
   """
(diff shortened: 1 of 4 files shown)


DROOdotFOO (commented):

Three-persona pass. HIGH/MEDIUM findings reproduced against a faithful standalone copy of both pure functions.

### HIGH — the "never leaks raw markers" invariant is false for nested emphasis at the streaming tail
`markdown_body.ex` `provisional_close/1` (LIFO closer stack) and `markdown_renderer.ex` `builtin_segments/1` (flat non-nesting regex `@builtin_pattern`) model incompatible grammars. `provisional_close` closes nested constructs CommonMark-style (italic-then-bold ⇒ appends `***`), but the builtin renderer requires ≥1 non-marker char between each delimiter pair, so the provisionally-closed string never round-trips. Reproduced on natural LLM prose:
```
"note: *very **important"  -> provisional_close -> "...important***"
                           -> RENDERS: "note: very important**"   (literal ** leak)
"text _a __b"              -> RENDERS: "text a b__"                (leaks _)
"a *b `c"                  -> leaks a literal backtick
```
Falsifies P-MD-01 and the stated goal. Uncaught because the prefix/property tests only walk prefixes of one golden fixture that never nests italic-then-bold at a live tail. Root cause: `provisional_close` assumes a nesting-capable downstream parser; the builtin one isn't.

### MEDIUM
- **Mid-grapheme UTF-8 cut corrupts already-committed content** — `to_text/1` + `latin1_fallback/1`: when a delta boundary lands mid multi-byte grapheme, `String.valid?/1` fails and the **entire buffer** is reinterpreted Latin-1. Streaming `café ☕ done` cut inside `☕` renders `cafÃ© â` — committed `café` mojibakes for that frame, reverts next delta. Violates the P-MD-05 monotonicity claim; any CJK/emoji/accented message flickers. The byte-cut property only asserts no fence leak, so it misses this.
- **The 256KB "safety belt" guards the wrong operation** — `@provisional_close_byte_cap` short-circuits only `provisional_close` (the cheap O(n) scan). `render_with_builtin` re-splits and re-parses the **whole** buffer every delta with no size guard → O(N²) cumulative CPU on a large message streamed in small deltas, unbounded by the cap. False DoS assurance.

### LOW
- Greedy table-row absorption: a `|`-containing line immediately after a table (no blank separator) is swallowed as a ragged row; two adjacent prose lines can be promoted to a table.
- Mixed fences (` ``` ` vs `~~~`) treated as interchangeable toggles → miscounts mixed-fence input.

### New Hire
- The "merge before #560, trivial union" note is inaccurate — both PRs refactor the same private `content_lines_view` to *conflicting arities* (#563 `(block, width, context)`, #560 `(block, context)`, plus #560 bumps `header_view`/`outcome_row_view`). Ordering is real (shared `block.ex`), but the second merger eats a manual reconcile regardless of order.
- Task-ID / AI meta-narration pervades comments + moduledocs ("T26 wire-in", "N-MDFUZZ-05", "the glow failure"), against the repo style rule.
- **No control-char/ANSI sanitization of untrusted LLM markdown** — `to_text/1` normalizes only UTF-8 validity; ESC (0x1b) and C0 controls pass into `Components.text(content:)`, so raw ANSI/OSC in LLM output can reach the terminal renderer (cursor moves, screen clear, title set). The repo's own rule is "never embed raw ANSI in `text()`." Pre-existing pipeline gap, but this PR newly routes untrusted content into it. (No HTML/XSS — output is terminal cells; link URLs render inert.)

### Complexity-claim audit (passes)
`provisional_close` is genuinely single-pass O(n) per call; the regexes are ReDoS-clear. The `Θ(n³)` in the PR body refers to cumulative re-parse-per-delta, inherent to re-render streaming, not this function.

### CI trust
Green = "the one golden `markdown-stream.jsonl` doc's prefixes don't leak and random byte-cuts don't crash." It does not fuzz arbitrary markdown structure, so the HIGH (nested-emphasis leak), the mid-grapheme corruption, and the DoS all pass green while the stated invariants are broken.

merklebonsai:
Review-round fixes pushed (`cafe98ab`):
- **HIGH nested-emphasis leak** — `unwind` used to emit a LIFO closer for *every* open frame (`***`/`___` unparseable by the flat builtin). Now only the outermost surviving construct gets a real closer; any un-closed inner opener has just its own token **erased** (new `{index,count}` erasure via `apply_erasures`, plumbed so fenced lines stay untouched). Matches all 3 review repros exactly (`note: *very **important` → `note: *very important*`). New StreamData property generates arbitrarily-deep genuinely-nested never-closed emphasis chains, asserts zero marker leak through the **builtin path directly** at every prefix cut (200 runs × 5 seeds).
- **MEDIUM mojibake** — whole-buffer Latin-1 fallback replaced with longest-valid-UTF-8-prefix retention; trailing incomplete bytes held for next delta (café/CJK/emoji/ZWJ prefix tests).
- **MEDIUM render cap** — `render_via` now honors the 256KB cap (plain-text fallback above it), not just the scan.
- **SECURITY** — `to_text` strips C0 (except \n/\t), DEL, C1 on both UTF-8 paths (SGR/OSC/C0/C1/DEL tests).
- **LOW** table absorption requires leading `|` + matching header cell count; mixed fences tracked by marker identity (a `~~~` closes only on `~~~`), renderer gained `~~~`.
- Docs de-jargoned; backslash-escape noted as known limitation.
945-test UI sweep green; 152 targeted (5 properties) × 5 seeds.

merklebonsai:
Two follow-ups pushed (`eac96bc3`, rebased onto master):

**1. Render-cap CI flake → deterministic.** The wall-clock `:timer.tc within budget` assert is out of the default suite. The cap describe now asserts the *algorithmic* fact: above `@render_byte_cap` → `render` returns the un-parsed fallback (structural match on a single text node + raw `**` markers survive verbatim = parser never ran); below the cap → parser ran (markers consumed). The timing check stays as a `@tag :slow` perf guard (CI-excluded). Default run: 75 passed, 1 excluded.

**2. Rebased onto T8 (#560 merged first) + prominence-fade union.** block.ex three-helper conflict resolved to the union: `build_render/3` resolves the fade colour **once** (memoized) and threads it into header/content/outcome; new `fade_view/2` stamps the resolved `:fg` onto every text node of the `MarkdownBody.render` result — **markdown bodies now fade in lockstep with the header** (Drew's bright-body-under-faded-header gap, closed here since #563 owns the markdown routing). `nil` fg (absent/1.0 prominence) = byte-identical neutral path. New `markdown body fades with prominence` describe (3 tests, green): faded body carries the same `:fg` as the header; 1.0 → byte-identical to no-fade; plain body fades too. `prominence.ex` untouched (its fixes are the T8b follow-up, #568).

177 tests green on the merged tree; block regression intact.

PR #564, opened 2026-07-15
Harness UI: canonical input-event normalization [T27]

merklebonsai:
Unit **T27** of the harness-UI lane: canonical harness input-event normalization — extracted because two components (T11 composer, T14 picker) independently hit the same gap at build time.

## Why
The terminal driver emits key events in **three incompatible shapes**, and harness components had no single place to reconcile them — the composer (merged) reads only `data.modifiers`, which is **dead on the real terminal driver**; the picker hand-rolled a defensive `text_input?/1`. Same bug class twice → extract the seam (R2 interface unit).

| Source | shape | modifiers |
|---|---|---|
| `event_translator.ex` (termbox) | `%{char/key, shift/ctrl/alt/meta: bool}` | boolean fields, no list |
| `input_parser.ex` (ANSI) | `%{key: :char, char: "a"}` (bool mods only when true) | omitted-when-false |
| `Event.key_event/3` (API) | `%{key: "a", modifiers: [atoms]}` | list |

## What
- **`Raxol.UI.Harness.InputEvent.normalize/1`** → one canonical form: `%{kind: :char|:key|:paste|:other, char, key, text, mods: %{ctrl,alt,shift}, raw}`. `char`/`key` are mutually-exclusive lanes; `mods` always has all three booleans regardless of which upstream shape omitted which (the actual bug); `kind: :other` is a total fallback so normalize never raises.
- Helpers: `text?/1` (printable, no ctrl/alt — shift-only is text), `shortcut?/1` (ctrl/alt held), `printable_char/1` (grapheme, never truncates emoji/ZWJ), `key/1` (special-key atom).
- Accepts a full `%Event{}` or a bare data/paste map.

## Evidence
43 tests (40 unit + 3 properties): every one of the three shapes + paste + Event.key_event canonicalizes correctly; text?/shortcut?/printable_char across shift-only-is-text / ctrl-A-is-shortcut / emoji-grapheme-is-text; special keys map consistently; **totality** (200 runs, incl. arbitrary terms — never raises) and **idempotence** (raw round-trips losslessly). compile --warnings-as-errors / credo --strict / format clean.

## Follow-up
T11 composer and T14 picker retire their inline fixes for `InputEvent.normalize/1` in a cheap follow-up; T12 (keybinds) and T13a build on it directly. `TextInput`'s own `data[:modifiers] || []` read is the same class, noted.

### DIFF
--- a/lib/raxol/ui/components/harness/composer.ex
+++ b/lib/raxol/ui/components/harness/composer.ex
@@ -37,10 +37,28 @@ defmodule Raxol.UI.Components.Harness.Composer do
   no history yet. The modifier branches activate automatically when F1b
   (kitty keyboard protocol) lands -- zero logic change here.
 
-  Also known: printable characters arriving with compound modifiers (e.g.
-  `[:shift, :alt]`) fall through to MultiLineInput delegation and are
-  dropped today -- canonical modifier normalization belongs to F1a input
-  canon, not this unit.
+  ## Input-shape normalization
+
+  `handle_event/3` normalizes every incoming event through
+  `Raxol.UI.Harness.InputEvent.normalize/1` before deciding anything.
+  Before T27's review round this component matched key events by pattern
+  (`%{key: :enter, modifiers: modifiers}`), which only exists on
+  `Event.key_event/3`'s test-API shape -- the real terminal driver's two
+  shapes (`event_translator.ex`'s boolean `shift:`/`ctrl:`/... fields,
+  `input_parser.ex`'s optional fields) have no `:modifiers` key at all, so
+  the pattern never matched a REAL keypress. Enter never submitted,
+  printable characters never inserted, and everything silently fell
+  through to raw `MultiLineInput` delegation for actual terminal input --
+  this unit only ever worked in tests that construct `Event.key_event/3`
+  directly. `InputEvent.normalize/1` erases that shape difference: Enter
+  from a real termbox driver, a real ANSI parser, or `Event.key_event/3`
+  all normalize to the same `kind: :key, key: :enter` and reach the same
+  `handle_enter_key/2`.
+
+  Printable characters arriving with compound modifiers (e.g. Shift+Alt)
+  are `InputEvent.shortcut?/1` (alt held) and fall through to
+  `MultiLineInput` delegation, same as before -- canonical compound-
+  modifier handling belongs to F1a input canon, not this unit.
 
   ## Bracketed paste
 
@@ -80,6 +98,7 @@ defmodule Raxol.UI.Components.Harness.Composer do
   alias Raxol.Core.Events.Event
   alias Raxol.UI.Components.Input.MultiLineInput
   alias Raxol.UI.Components.Input.MultiLineInput.TextHelper
+  alias Raxol.UI.Harness.InputEvent
   alias Raxol.UI.StyleHelper
   alias Raxol.UI.TextMeasure
   alias Raxol.View.Components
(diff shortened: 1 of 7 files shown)


DROOdotFOO (commented):

Three-persona pass. The module itself is total, well-`@spec`'d, has catch-all clauses, and does no `String.to_atom` on input (no atom-exhaustion vector). The problems are claim-vs-reality and that it's unwired.

### HIGH — "cross-shape agreement" is false in reality, and the property test is built so it can't catch it
`input_event.ex` moduledoc + `input_event_test.exs` headline property. The property hand-constructs the three shapes as synthetic maps — for a `:key` payload it uses `%{char: nil, key: :enter}` as the "translator shape." But the **real** `EventTranslator.translate/1` only maps keycodes 65/66/67/68/265/266 (arrows + F1/F2); everything else → `key: :unknown`. Empirically:
```
enter (kc=13)  => norm.key=:unknown
tab   (kc=9)   => norm.key=:unknown
escape(kc=27)  => norm.key=:unknown
left  (kc=68)  => norm.key=:left     # only the 6 mapped codes work
```
So the same logical Enter normalizes to `:enter` via the ANSI parser but `:unknown` via the native termbox driver — the exact divergence the invariant claims to forbid. The property never drives `EventTranslator`/`InputParser` for `:key` rows, and the table tests conspicuously omit translator enter/backspace/tab/escape — every case that diverges. False assurance on the PR's central claim.

### MEDIUM
- **The normalizer is dead code** — grep for `Harness.InputEvent` / `InputEvent.normalize` across `lib` + packages returns zero call sites outside the module + test. The moduledoc says composer/picker/`text_input` "should converge on" it, but this PR changes only 2 files. `composer.ex` still matches only the `key_event/3` shape and has no clause for the native termbox boolean-mods shape (native enter is `:unknown` with no `modifiers` key), so the F1 divergence is live in the real component and this PR doesn't mitigate it — it adds a parallel unused abstraction.
- **`state` (`:pressed`/`:released`/`:repeat`) silently discarded** — `key_result` never reads `data.state` though `Event.key_event/3` models all three. A `:released` or `:repeat` for `"a"` normalizes identically to a press (`kind: :char`, `printable_char == "a"`) → any component receiving press+release double-inserts. Not mentioned in the moduledoc.

### LOW
- **No content validation** — `extract_char` returns any non-empty binary unchanged. `%{char: "\e[2J"}` normalizes to `kind: :char`, `text?/1` true, `printable_char/1` returns it verbatim; the documented migration then `insert`s an ANSI sequence into the buffer. Paste is forwarded with no bounds (upstream bracketed-paste-without-end turns "the rest of the buffer" into `text`). The stated single choke point performs no grapheme/printability check.
- `extract_char` ambiguity: a special key mis-encoded as a string (`"enter"`, `"tab"`) is silently inserted as literal text instead of dispatched as a key. Convention (atom = special, binary = char) is real but undocumented at this fallback.

### CI note
The "3 pending" was stale — `statusCheckRollup` shows all 25 checks completed (23 success, 2 skipped). CI is green precisely *because* the tests never drive the real `EventTranslator` for divergent keys and the normalizer has no production call site to break — consistent with the HIGH + dead-code MEDIUM.

merklebonsai:
Review-round fixes pushed (`ced9442f`). Drew's 'the property tests a fiction' finding led to a bigger discovery: **the termbox EventTranslator's keycode table was itself fiction.**

- **HIGH — cross-shape agreement made real + a pre-existing keycode bug fixed.** The old translator used `65/66/67/68` for arrows (actually ANSI CSI final bytes) and `265/266` for F1/F2 (ncurses `KEY_F()` values) — **neither is a real `TB_KEY_*` constant**, so every special key from real termbox was mis-decoded. Verified against vendored `termbox2.h`: arrows = `0xffff-{18..21}` (65517-65514), F1/F2 = `0xffff-{0,1}` (65535/65534), ENTER/TAB/ESC = 0x0d/0x09/0x1b. Rewrote the full table (nav, F1-F12, back-tab); Enter/Tab/Esc/Backspace resolve from char_code **or** key_code so a control byte can't leak as `char:`. Only consumer is `driver.ex` (its EventTranslator tests updated to real codes, green). **Scope note:** this fixes native special-key decoding repo-wide, beyond the original normalization scope — flagging for reviewer attention on the input path.
- The old cross-shape property hand-built maps; rewrote it to drive real `EventTranslator.translate`/`InputParser.parse` across the full ctrl/alt/shift power set. Full agreement is honestly **scoped** in the moduledoc: meta is unrepresentable over raw ANSI; Enter/Tab/Esc/Backspace have no modifier-carrying ANSI encoding (Shift+Tab excepted via CSI `Z`).
- **composer WIRED to normalize** (was dead code — it matched only the `key_event/3` `modifiers:` shape, so Enter never submitted and chars never inserted for real terminal input). Tests drive real driver events into `handle_event/3`.
- **state** (:pressed/:released/:repeat) carried; release no longer double-inserts. **content validation**: extract_char requires one grapheme, no C0/DEL; paste stripped of controls.
66 + 46 tests green.

PR #567, opened 2026-07-15
runtime: graceful shutdown reaches driver terminate (trap_exit + driver-first) [T28a]

merklebonsai:
Unit **T28a** of the harness-UI lane: **graceful shutdown now reaches the driver's `terminate/2`** — fixes a real pre-existing bug found + confirmed by four independent reviewers during T2d. On graceful stop, `Lifecycle` orphaned the driver, so terminal teardown (mode reset, scroll-region reset, stty restore) was never emitted — leaving the terminal dirty on every exit, in every environment.

## The bug (Facet 1 — the in-VM race)
`Lifecycle` start-links every child from `init/1` but never trapped exits. `handle_cast(:shutdown)` stopped the rendering engine *before* the driver; the engine's `:shutdown` exit propagated through the untrapped link and could kill `Lifecycle` before the driver-stop line ran → the driver's `terminate/2` orphaned. Measured ~50% miss under load.

## Fix
- `Process.flag(:trap_exit, true)` in `init/1`.
- `handle_info({:EXIT, pid, reason})` turns a child exit into an ordinary message; non-`:normal` → `{:stop, reason, state}`, preserving the old "die together" contract (a supervisor above sees the identical reason and restarts identically — verified).
- **All child teardown moved into `terminate/2`** as the single source of truth, with **driver-first ordering** (driver emits reset bytes → engine → … → plugin_manager), reached identically whether triggered by our own cast, an external exit, or a genuine child crash.

## Verified NOT to regress OTP semantics
Four reviewers (Opus deep-trace + grok triad) converged: die-together and supervisor-restart are byte-identical; only a `:normal` link exit is swallowed, which never killed a non-trapping process anyway. A crashing child now incurs bounded driver-first teardown *before* the crash propagates upward (documented behavioral addition).

## Tests
308 `test/raxol/core/runtime` pass, stable across 6+ seeds. The **crash-propagation test** (the #1 risk, previously untested) proves a genuine child crash (`:boom`) still dies-together, propagates the reason verbatim, **and** runs driver teardown on the crash path — confirmed **red when the abnormal-EXIT clause is neutered**. Plus: parent-exit-brings-Lifecycle-down-gracefully vs non-parent-`:normal`-peer-does-not (two tests), `plugin_manager` driver-first ordering, and `terminate/2` idempotency over nil/dead pids. The 30-iteration determinism sweep proves the ~50% race is structurally gone.

## Scope
This is **Facet 1 only**. The SIGTERM-handler path (Facet 2) had a self-deadlock (`handle_event` runs inside `:erl_signal_server`, `stop_and_await` blocks it while `terminate/2`'s `:gen_event.delete_handler` waits on that same callback) — split out to **T28b**, reworked, with the deadlock fixed + one-per-VM guard + `:on_teardown` callback (a library must not hijack SIGTERM by default). `trap_os_signals` is entirely removed from this PR.

Unblocks T2d's `Raxol.stop` teardown and gives T2a/T2b/T2c a driver whose teardown actually runs.

## Review trail
Opus deep-trace + grok triad (tests/invariants, code-quality, conceptual), all four converging → the split decision (ship Facet 1, rework Facet 2) + the four triad-found fixes.

### DIFF
--- a/lib/raxol/core/runtime/lifecycle.ex
+++ b/lib/raxol/core/runtime/lifecycle.ex
@@ -12,6 +12,26 @@ defmodule Raxol.Core.Runtime.Lifecycle do
   `dispatcher_ready` and `plugin_manager_ready` flags are set. In dev mode,
   a CodeReloader is started to watch for source changes and trigger re-renders.
 
+  ## Shutdown & exit trapping
+
+  Every child (dispatcher, driver, rendering engine, plugin manager) is
+  `start_link`'d from `init/1`, so all are LINKED to this process, and this
+  process traps exits. `terminate/2` is the single source of truth for
+  shutdown ordering: it stops children driver-FIRST so the Terminal Driver's
+  own `terminate/2` (mode reset, scroll-region reset, stty restore) runs to
+  completion before anything under it is torn down. This runs identically for
+  every stop trigger: our own `:shutdown` cast, an external abnormal exit
+  signal, or a genuine child crash.
+
+  Trapping exits preserves the prior die-together + supervisor-restart
+  contract (an abnormal child/peer exit still stops this process with the
+  same reason, propagated verbatim), while a `:normal` peer exit is ignored
+  exactly as an untrapped process would have dropped it. One behavioral
+  addition: a genuine child CRASH now incurs the full driver-first teardown
+  BEFORE the crash reason propagates upward (bounded latency:
+  `GenServer.stop` over the surviving children) -- teardown-on-crash is
+  intentional, not a correctness break.
+
   ## Sub-modules
   - `Lifecycle.Initializer` -- component startup sequence
   - `Lifecycle.Shutdown`    -- stop_process, cleanup, registry management
@@ -113,6 +133,8 @@ defmodule Raxol.Core.Runtime.Lifecycle do
 
   @impl GenServer
   def init(options) when is_list(options) do
+    trap_child_exits!()
+
     app_module = Keyword.fetch!(options, :app_module)
 
     Log.info_with_context(
@@ -124,6 +146,10 @@ defmodule Raxol.Core.Runtime.Lifecycle do
       |> maybe_start_time_travel()
       |> maybe_start_cycle_profiler()
 
+    finish_init(app_module, options)
+  end
+
+  defp finish_init(app_module, options) do
     case Initializer.initialize_all(app_module, options) do
       {:ok, registry_table, pm_pid, initialized_model, dispatcher_pid,
        driver_pid, rendering_engine_pid} ->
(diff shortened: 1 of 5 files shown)


merklebonsai:
CI fix pushed (`6c598bac`). The Windows red was a **test bug, not a shutdown bug** — the determinism sweep asserted `driver_at < engine_at` on monotonic timestamps, and Windows' coarse (~15ms) clock collapsed the two microsecond-apart teardown events to the identical value, breaking a strict `<` while the logs proved the order was correct.

Fixed to be **sequence-position-based, no clock**: the recording doubles now send bare tags (`:driver_terminated`/`:engine_terminated`), and all three ordering tests assert on mailbox arrival order (`next_teardown_event() == :driver` then `:engine`). Platform-independent by construction — `Shutdown.stop_process` is a synchronous `GenServer.stop`, so the driver's `terminate/2` completes and enqueues its tag before the engine is even asked to stop. Red-if-broken confirmed: reversing the stop order fails all three ordering tests. 308 `test/raxol/core/runtime` green across 4 seeds.

The other red (`TpPtyTest` signal-delivery on Linux) is the pre-existing merged-TP flake (#550), unrelated to this PR — a rerun should clear it.

DROOdotFOO:
Pushed `3c1698ba` for the remaining Windows red — a **pre-existing test-isolation flake**, not a shutdown issue.

`block_test.exs`'s `attach_recovered_handler` attaches to `[:raxol, :harness, :projection, :recovered]`, but that event name is **shared** by two emitters: `Harness.Block` (metadata `%{kind:, reason:}`) and `Harness.Projection.Recovery` (metadata `%{reason:, event_id:}`). Telemetry handlers fire on every global emission of the name, so a concurrent `async: true` test emitting a `Projection.Recovery` event landed `%{reason: :missing_turn_started, event_id: 2}` in this test's mailbox first; `assert_received` matched it and `meta.kind` blew up. Windows' scheduling made that interleaving reproducible.

Fix filters the handler to Block's own recovery (`is_map_key(metadata, :kind)`), dropping foreign `Projection.Recovery` events. Local run: 40/40 green. Unrelated to the T28a shutdown change.

DROOdotFOO:
Superseded the test-only filter with a **source-level fix** (`942f13ef`, squashed over the earlier `3c1698ba`).

Rather than filter the foreign event out in the test, the two emitters now use **distinct event names** so they can't collide at all:
- `Harness.Block` recovery -> `[:raxol, :harness, :block, :recovered]` (metadata `%{kind:, reason:}`)
- `Harness.Projection.Recovery` -> unchanged `[:raxol, :harness, :projection, :recovered]` (metadata `%{reason:, event_id:}`)

Projection.Recovery and its `t7_projection_test` are untouched. Note the collision only materializes in the CI **merge commit** (branch + master) — `Projection.Recovery` lives on master, not on this branch — which is why it reproduced in CI but not in a branch-only run. Local block + projection tests green.

Issue #566, opened 2026-07-15
Session-scope CommandRegistryTable — concurrent same-module sessions share and delete one ETS table


merklebonsai:
## Problem

`Raxol.Core.Runtime.Lifecycle` (`lifecycle/initializer.ex:97-98`) names the per-session command registry table `Module.concat(CommandRegistryTable, app_module)` — a `:named_table` keyed by **app_module, not session_id**. `CompilerState.ensure_table` is idempotent by name, so a second session of the same agent module **adopts** the first's table, and `Lifecycle` teardown's `cleanup_registry_table` calls `:ets.delete` on that shared table.

## Reproduces (confirmed during SS #558 re-review)

Two supervised sessions, same `EchoAgent`, session_ids `crt-a` / `crt-b`: both resolve the identical table `CommandRegistryTable.…EchoAgent` (`shared=true`). After `stop_session("crt-a")`, session B stays alive but `:ets.info(table)` becomes `:undefined`. Any subsequent command/directive execution in session B that touches the registry breaks. (`EchoAgent` issues no directives, so only its plain model-fold survives — an agent that uses directives would fail.)

## Why now

Pre-existing in `Lifecycle`, but **SS (#558) makes concurrent same-module sessions a first-class case** (per-session supervised trees via `start_session/2`). This turns a latent bug into a reachable data-loss / crash path for that use case.

## Fix shape

Key the registry table by `session_id` (or `{app_module, session_id}`) instead of `app_module` alone, so each session owns its own table and teardown of one never affects another. Touches main `raxol` (`Lifecycle`), so a separate PR from SS.

## Scope note

Not blocking single-session or distinct-module concurrent use (the common cases today). Blocks safe concurrent **same-module** sessions. Ref: PR #558 re-review.

PR #569, opened 2026-07-15
docs(harness): freeze contracts constitution + research corpus

merklebonsai:
## What

The ratified contract surface for the agent harness, frozen for parallel red-suite authoring. **Review focus: `harness-freeze-contracts.md`** — permanent test suites get authored against it; the other 30 files are supporting research/evidence.

## The constitution (harness-freeze-contracts.md)

Three frozen surfaces, each with positive + negative contours and named dead-injectors:
- **JS-FREEZE** — journal record schema: one offset law, derived tip (never stored), `branch_id` + per-branch tip + closure rule, session lineage (typed-edge DAG at session level, journals linear forever, NC-12: no CRDT journal ever), `$blob`/`$redacted` value markers, GC low-watermark, annotation/schedule kinds.
- **U11-CONTRACT** — meta event family: grow-only type registry, two-point taint lattice (no laundering), `refs` session-scoping (permanent), actor stamped at producer seam, model/params fingerprint + precedence law, speculation/approval/policy records.
- **U12-CONTRACT** — probe runner: pure build/interpret probes, reserve-before-call token budget, bounded parking, cache-riding = byte-identical prefix.

Governing law: **the contract only grows** — no rename/repurpose/optional-to-required, ever; the UI fork never needs a lockstep update.

## Provenance

Survived: 2 adversarial longcat audits (F1-F9 fixed, 9 open questions ruled), 2 independent ideation passes, a 93-find community-ecosystem survey (verdict: zero uncaught day-one gaps), and a 5-model quorum on the 3 contested rulings. The 20-item fold-before-reds revision is ratified under the event-sourcing principle: every offset states the complete fact of what happened there.

## What this unblocks

The full red-suite fan-out (U4/U5/U6/U7/U8/U9/U10/U11/U12/U21-R) — red PRs will follow tonight, each self-contained with positive + negative contours.

### DIFF
--- a/docs/proposals/in-flight/README.md
+++ b/docs/proposals/in-flight/README.md
@@ -1,0 +1,68 @@
+# In-flight harness docs — index and glossary
+
+One page, scannable. If you're a red-suite author (or reviewer) landing in
+this directory cold: start with `harness-freeze-contracts.md`, then use the
+tables below to find where any tag you hit (`AD-6a`, `FI-8`, `AF-3`, `I3`,
+`N-JS6`, `R1`...) actually lives.
+
+## File → purpose
+
+| doc | purpose |
+|---|---|
+| `harness-freeze-contracts.md` | **The constitution.** Frozen shapes (journal record kinds, meta/taint contract, probe runner interface) that red suites are authored *against*. Everything else in this directory is either an input to it (research) or a consumer of it (protocol/roadmap). |
+| `harness-roadmap.md` | The autonomous-units dependency chart (DAG) — what U1, U4, U9, U12... are, their ordering, and the false-parallel warnings the freeze resolves. |
+| `harness-invariants.md` | I1–I10 (the permanent Tier-1/Tier-2 guards) + the meta-invariants (fired-counters, seed-reproducibility, oracle independence — the rules every red suite itself must obey). |
+| `harness-spec-protocol.md` | The shared typed contract (`Event` envelope, §3 meta table) both backend and frontend conform to — the "zod schema" of the split. |
+| `harness-spec-backend.md` | The headless core spec (process topology, journal, probe swarm, control/meta layers). §4's Ecto/DETS/Oban sub-decision is superseded by D1/D2 — see the banner in that file. |
+| `harness-spec-frontend.md` | The detachable-UI spec — pure protocol subscribers, holds nothing durable. |
+| `harness-synthesis.md` | Cohort-research synthesis (Phase 5–6) — a primary source of `AD-*` (architecture-disposition) tags. |
+| `harness-storage-research.md` | Round-2 cohort research (journal/storage/commands) — a primary source of `AD-9..15` and `FI-*` (finding-item) tags. |
+| `harness-storage-foundations.md` | Research pass hunting for storage-shaping features that become structural rewrites if added post-freeze. Feeds ratified fixes into `harness-freeze-contracts.md`. |
+| `harness-future-foundations-ideation.md` | Ideation pass over the *next* reservation tier (R1–R6 = RESERVE-NOW; the `F1`–`F18` candidate table in §3.1 is a **separate** numbering from `harness-freeze-contracts.md`'s `AF-*` audit-finding tags — see the ID-prefix map below). |
+| `harness-community-gaps.md` | Gap-pattern synthesis: ~93 community fixes to popular agent harnesses, clustered and mapped against the freeze. Research/input, not itself frozen. |
+| `harness-yolo-safe-research.md` | YOLO-safe-by-construction research/design — the auto-approve predicate (`effect_class`/`egress`) frozen in `harness-freeze-contracts.md` §5.2 traces back here. |
+| `harness-design.md` | Design record — the seven concepts + architecture narrative. Companion to `harness-synthesis.md` and `harness-baseline-features.md`. |
+| `harness-baseline-features.md` | Reference list of "table stakes" features (the floor, not the differentiators). |
+| `harness-facts-two-perspectives.md` | Raw fact extraction (operator + systems vantage) over `harness-research/01–09`, zero processing — an input to `harness-design.md` and `harness-synthesis.md`. |
+| `harness-cohort-research.md` | Phase 1–2 (frame + priors) for the first cohort-research round, written before research returned. |
+| `harness-research/` | Raw per-topic research briefs (01–09, plus later additions) underlying `harness-facts-two-perspectives.md` and `harness-synthesis.md`. |
+
+Non-`harness-*` files in this directory (`palette-inventory.md`, `t0-runbook.md`,
+`t0-verdict-schema.md`) belong to the UI/rendering lane, not this constitution;
+out of scope for this index.
+
+## ID-prefix → owning doc
+
+| prefix | meaning | owning doc |
+|---|---|---|
+| `AD-<n>` (e.g. `AD-6a`) | architecture disposition | `harness-synthesis.md` + `harness-storage-research.md` (sublabels like `AD-3a`/`AD-3b`/`AD-6a` are glossed inline at first use in `harness-freeze-contracts.md`, but the parent `AD-3`/`AD-6` are defined in `harness-synthesis.md`) |
+| `FI-<n>` (e.g. `FI-8`) | finding-item | `harness-storage-research.md` |
+| `AF-<n>` (e.g. `AF-3`) | adversarial-audit finding (longcat/review) | `harness-freeze-contracts.md` — **do not confuse with `F<n>` below.** The audit trail behind these (5-model / 93-find run) lives in the PR review threads, not in a committed corpus doc. |
+| `F<n>` (e.g. `F5`, `F9`) | future-foundations reservation-candidate tag | `harness-future-foundations-ideation.md` §3.1/§3.2 (the verdict table + detail sections). **Distinct namespace from `AF-<n>` above** — `harness-freeze-contracts.md` uses bare `F<n>` for only two things now: (1) two genuine future-foundations citations (the `annotation`/`schedule` kind rows, `F14`/`F9`); (2) references to the separately-named `f2-action-registry.md` doc ("the F2 `Raxol.Action` draft") in §5.2, a third, unrelated `F2` — every other bare `F<n>` that used to appear in `harness-freeze-contracts.md` was an audit-finding collision, now renamed to `AF-<n>`. |
+| `NC-<n>` (e.g. `NC-12`) | frozen negative constraint | `harness-freeze-contracts.md` |
+| `OQ-<n>` (e.g. `OQ-JS1`) | open question, ruled | `harness-freeze-contracts.md` (§1.5/§2.5/§3.5 "Ruled" sections) |
+| `P-<n>` (e.g. `P-JS4`) | positive-contour property | `harness-freeze-contracts.md` |
+| `N-<n>` (e.g. `N-JS6`) | negative-contour required failure | `harness-freeze-contracts.md` |
+| `meta-inv <n>` | meta-invariant (fired-counter, seed-reproducibility, oracle independence, ...) | `harness-invariants.md` (defined under "Meta-invariants"); cited throughout `harness-freeze-contracts.md`'s red-suite obligations |
+| `I<n>` (e.g. `I1`–`I10`) | permanent Tier-1/Tier-2 invariant | `harness-invariants.md` |
+| `U<n>` / `U<n>-R` (e.g. `U9`, `U9-R`) | roadmap unit / its red suite | unit definitions in `harness-roadmap.md`; the red suites themselves are test code (not a doc), one git branch per unit (e.g. `red/u9-checkpoint`) |
+| `R1`–`R6` / `R-tags` | future-foundations RESERVE-NOW candidates | `harness-future-foundations-ideation.md` (R1 = session-lineage-as-typed-edges, §3.1) |
+
+## Acronyms (expanded once)
+
+- **CAS** — content-addressed store. Snapshots/blobs are written to
+  `<session>/snapshots/<sha256>.json` / `<session>/blobs/<sha256>`, named by
+  the hash of their own bytes (dedupe + integrity for free).
+- **SpendGate** — the reserve-before-call budget gate (`Ledger.try_spend`)
+  that every provider call must pass through; `AD-6a` names its frozen
+  discipline.
+- **TEA** — The Elm Architecture, Raxol's canonical `init/1` + `update/2` +
+  `view/1` application model (see root `CLAUDE.md`).
+
+## Provenance note
+
+The PR review that produced the `AF-*` rename (5-model / 93-find adversarial
+audit) is real, but the individual findings were never committed to this
+corpus as a standalone doc — the audit trail lives in the review threads
+themselves, not here. This index intentionally does not fabricate a doc for
+it.
--- a/docs/proposals/in-flight/agent-lane-response-to-ui-intersection.md
+++ b/docs/proposals/in-flight/agent-lane-response-to-ui-intersection.md
@@ -1,0 +1,201 @@
+# Agent-Lane Response to the Harness-UI Intersection Audit (2026-07-17)
+
+**Status**: RULED (agent-lane side) — items marked *co-design* need UI-lane sign-off.
+**Responds to**: `harness-ui-agent-lane-intersection.md` (UI lane, 2026-07-17).
+**Author lane**: harness-agent (protocol / ACP / storage / reattach; #613 confinement
+seam; #569 constitution).
+
+## Sync channel
+
+The two lanes coordinate on exactly two surfaces, and only these two:
+
+1. the shared `lane-docs/` scratchpad directory (where the UI lane's audit lives and
+   where a copy of this response sits next to it), and
+2. the `docs/harness-freeze-constitution` branch, `docs/proposals/in-flight/` (this
+   file's canonical home).
+
+There is **no live cross-session channel** — both sides should check both surfaces.
+**UI lane: please reply in the same `lane-docs/` dir** (a round-trip file thread) on the
+three items that need your sign-off: (a) the TermText `:sgr_keep` mode co-design in §B —
+you own the allowlist byte-semantics and the #608 `sanitize == identity`-for-safe-SGR
+seal invariant; (b) the `surface.ex` merge-order coordination in §C.1; (c) the shared
+de-flake sweep in §C.3.
+
+---
+
+## A. Rulings on the six §5 protocol questions
+
+### Q1 — context_pct denominator: additive `context` field on `turn_completed.usage`
+
+**Ruling: field, not meta event.** `turn_completed.usage` gains an optional
+`context: %{used: non_neg_integer(), budget: pos_integer() | nil}`, producer-stamped
+(the backend knows both the consumed prompt tokens and the model's window; the UI never
+guesses a denominator). `budget: nil` is legal — providers that don't disclose a window
+stay honest and the strip keeps rendering `—`, exactly the producer-decision pin the UI
+already made. Justification: this is per-turn producer telemetry with a natural home on
+the turn rollup; a separate meta event buys nothing at turn granularity and costs a new
+type. Additive-only compliant: fields only ever get ADDED optional-with-default
+(freeze-contract §0 growth rule); `schema_version` minor bump. A *mid-turn*
+context-growth meta event stays a later additive option if streaming-granularity demand
+materializes — not ruled in, not ruled out.
+
+### Q2 — Evidence schema (U21/T19): the payload IS `refs`; resolve locally; summary shape reserved
+
+**Ruling: no verbatim map, no new blob.** U21's evidence is already typed:
+`turn_completed{final: true, refs: [offset]}` — journal offsets naming the evidence
+records (tool results / verification outputs), validated by `Raxol.Agent.DoneGate`
+(exist / evidence-class / same-turn / postdates-last-mutation / not-mutation-echo; first
+violation wins; `refs: []` ⇒ `:evidence_required`). T19's render contract is therefore:
+**resolve each ref against the UI's own journal projection** — the UI already holds
+every one of those records; "done because X" is a local lookup, not a payload.
+Absence renders explicitly ("no evidence") off `refs: []` / the
+`[:raxol, :agent, :done_gate, :ungated_done]` signal — which T19 already planned.
+If a no-lookup display affordance proves needed, the reserved additive shape is
+`evidence_summary: [%{ref: offset, class: String.t(), label: String.t()}]`, bounded
+(≤ 16 entries, label ≤ 120 chars), landing **with the U21-I refs-citation seam (after
+U8)** — do not build it before the citation seam exists; DoneGate is observe-only until
+then and the shape would freeze against a producer that doesn't exist yet.
+
+### Q3 — Cross-surface last-seen: client-local v1 stands; v2 shape ruled now as a reattach extension
+
+**Ruling: T17 v1 ships client-local, unchanged. The v2 shape is ruled now** (so the
+divider design can't paint itself into a corner) **and rides the reattach extension**
+— this is squarely the durable-session moat: per-operator read position must survive
+the client, so it lives in the journal like everything else that survives the client.
+Concrete v2 shape, all additive on the existing `_meta["raxol.io"]` rider surface of
+`acp-reattach-design.md` §3.1:
+
+- **Report** (client → agent): optional `"lastSeen": n` in `_meta["raxol.io"]` on
+  `_raxol/session.detach` and on the attach request itself (late report of the previous
+  incarnation's position). One rider namespace, one parser — same rule as
+  `fromOffset`/`historyPolicy`/`capability`.
+- **Persist**: the Writer appends a grow-only record kind `"marker.last_seen"`, payload
+  `%{"actorId" => grant.actor["id"], "surface" => ctx.surface, "offset" => n}`.
+  Writer-only append (single-publisher law holds); readers that don't know the kind
+  skip it (§1.2 skip-unknown fold) — grandfather-safe by construction. Actor identity
+  comes from the `%Grant{}` (CDI-3), never peer-asserted.
+- **Return** (agent → client): `LoadSessionResponse._meta["raxol.io"]["lastSeen"] = n | nil`
+  — the requesting actor's own high-water read position (fold of its
+  `marker.last_seen` records), next to the already-ruled `"highWatermark"`.
+
+The unread divider then derives as `lastSeen + 1 .. h` on any surface. Not scheduled —
+this rides the reattach implementation wave, not before it.
+
+### Q4 — View-descriptor vocabulary (T23): agree, with two pins
+
+**Ruling: accepted as proposed.** UI lane owns the bounded declarative vocabulary AND
+its validator; never eval, never code. Agent lane owns transport only — the U11 meta
+family / `_raxol` frame is a carrier of an opaque payload. Two pins from the transport
+side: (1) descriptors travel as meta-event payload under the frozen grow-only
+discipline — an unknown descriptor version/element is skip-not-error at every reader
+(N-U11.6 class); (2) the transport will **never** interpret, normalize, or "helpfully
+fix" descriptor contents — a validation failure is the renderer's to degrade (render as
+inert text), never a transport error. That keeps the trust boundary in exactly one
+place: the UI validator.
+
+### Q5 — U14 C2 projections: not ours to schedule; defer to V
+
+**Defer.** U14 is not on master and the agent lane does not own its timeline — it is a
+pending harness backlog item (#27 on the problems backlog). What we can confirm: nothing
+in the agent lane blocks it, and nothing in-flight on our side collides with its likely
+shape. Timeline is V's call; raise it in the next planning pass. T22/T23 stay
+transitively parked until then, per your own matrix.
+
+### Q6 — Stall-detector hidden-channel signals: yes, and they will be observable-only
+
+**Ruling: telemetry/journal events, never control.** The agent lane does plan
+retry/resample-class signals (the DoneGate verdicts already ship this pattern:
+`ungated_done` / `rejected_evidence` telemetry). The standing rule for all of them:
+self-correction signals are emitted as **observable events** — durable meta-family
+journal events where replay/evidence value warrants (grow-only kinds, skip-unknown), or
+telemetry where it doesn't — and are **never a control channel**: the agent does not
+act on its own emitted signals, and nothing in the agent lane will ever consume the UI
+detector's verdict. Your detector stays a pure observer and can upgrade its evidence
+quality from these events for free, which is exactly the "wire telemetry, not control"
+line your doc drew. We will name event kinds with the UI lane before the first one
+lands so the detector can subscribe from day one.
+
+---
+
+## B. #613 TermText contract ruling — two contracts, one seam (the hot item)
+
+**Verified against source first** (this ruling is grounded, not assumed):
+
+- `ContentGuard.sanitize_line/1`
+  (`lib/raxol/ui/rendering/paint_authority/content_guard.ex`, origin/master) is an
+  **SGR-allowlist**: `CSI … m` passes **verbatim** (`match_sgr`), every other ESC-led
+  sequence has *only the ESC byte stripped* — printable residue survives as the
+  deliberate **"visible-honest"** record (`\e[2J` → literal `[2J`); C0 stripped except
+  `\t\r\n`; DEL stripped. #608 tests `sanitize == identity` for safe-SGR as a seal-seam
+  invariant.
+- `ViewText.sanitize` (`lib/raxol/harness/surface/view_text.ex:221`) and
+  `MarkdownBody.to_text → sanitize_controls`
+  (`lib/raxol/ui/components/harness/markdown_body.ex:266`) are **strip-all**: every C0
+  (incl. ESC) except `\t` (+`\n` where line-relevant), DEL, and (MarkdownBody) C1
+  removed; no SGR survives.
+- `Raxol.Core.Boundary.TermText.sanitize/2` (#613,
+  `packages/raxol_core/lib/raxol/core/boundary/term_text.ex`) is **strip-all** and
+  removes ESC sequences **as a unit** (CSI body and all — no residue).
+
+**The ruling:**
+
+1. **These are two distinct contracts and both are correct for their seams.**
+   Strip-all is the `text()` / View-DSL boundary contract. The SGR-allowlist
+   "visible-honest" contract is the sealed/inline paint-path contract, where content
+   legitimately carries the renderer's own styling vocabulary and where silent deletion
+   would falsify the sealed history. **ContentGuard is NOT flattened into strip-all.
+   Ever.** Doing so would break the #608 seal invariant and erase the visible-honest
+   audit property.
+2. **TermText grows an additive `mode:` option** — `:strip` (default; byte-for-byte the
+   current behavior, so every existing migration row and the shared vectors are
+   untouched) and `:sgr_keep` (allowlist: preserve safe SGR color/style; strip
+   cursor-movement CSI, OSC, DCS/APC/PM/SOS, and every other dangerous sequence).
+3. **`:sgr_keep` is a CO-DESIGN with the UI lane, not a unilateral agent-lane change**
+   — this is a follow-up on #613, not part of it. The UI lane owns the allowlist
+   byte-semantics and the #608 `sanitize == identity`-for-safe-SGR seal invariant.
+   One decision point is theirs to pin explicitly: **residue semantics for rejected
+   sequences** — ContentGuard today strips only the leading ESC (visible-honest
+   residue), while TermText `:strip` removes the whole sequence; `:sgr_keep` must pick
+   one, on the record, because it *is* the observable contract of the sealed history.
+4. **The UI corpus is absorbed into the shared vectors**: #607 fence-label injections,
+   #608 residue classes, and the G11 wrap/ESC cases enter
+   `packages/raxol_core/test/support/boundary_vectors/` as the `:sgr_keep` conformance
+   set (a new vector file scoped to the mode; the existing `term_text_vectors.json`
+   stays the `:strip` set). ContentGuard then binds to the `:sgr_keep` vectors —
+   whether it delegates to TermText or remains its own impl bound to the same vectors
+   is the co-design's call; the vectors, not the module identity, are the contract.
+
+---
+
+## C. Merge-order coordination (§6)
+
+1. **`surface.ex` churn — agreed, and we'll sequence behind you.** Agent-lane T13b-prep
+   work that touches Surface wiring will not open against `surface.ex` until the UI's
+   pending #610 merges; we take the rebase, not you (you've absorbed three
+   conflict rounds this week already). Standing rule going forward: whichever lane has
+   a `surface.ex` PR already open holds the right of way; the other lane rebases.
+   Flag `surface.ex`-touching PRs in the `lane-docs/` thread when opened.
+2. **#569 additive-only freeze enforced by your byte-goldens — agreed and welcomed.**
+   The byte-golden CI is exactly the enforcement mechanism the constitution wants:
+   any event-schema drift fails LOUDLY before it reaches a live session. Everything in
+   §A above was ruled additive-only specifically so it passes that net (Q1's `context`
+   field, Q3's rider keys and record kind, Q6's event kinds — all optional-with-default
+   or grow-only vocabulary).
+3. **Shared de-flake sweep — agreed; one shared PR.** MetricsCollector env failure,
+   PlatformGraphics global-TERM mutation, and the renderer-SGR trio are repo-wide and
+   hit both lanes' CI; splitting them by lane would just double the review overhead.
+   Proposal: one jointly-reviewed PR, agent lane drives the branch, UI lane reviews the
+   renderer-SGR trio (it's your bytes). Confirm in the `lane-docs/` thread and we'll
+   open it.
+
+---
+
+## References
+
+- `harness-ui-agent-lane-intersection.md` (UI lane audit, `lane-docs/` + this dir's peer)
+- `confinement-seam-proposal.md` / `confinement-migration-plan.md` (#613; the migration
+  plan now carries the two-contract caution from §B)
+- `acp-reattach-design.md` (attach rider `_meta["raxol.io"]`, offset law, Writer) — Q3
+- `harness-freeze-contracts.md` (§0 growth rule; U11-CONTRACT; CONVERSATIONAL closure) — Q1/Q4/Q6
+- `Raxol.Agent.DoneGate` + `u21_evidence_done_red_test.exs` (#570) — Q2
+- `content_guard.ex` / `view_text.ex` / `markdown_body.ex` (origin/master) — §B ground truth
--- a/docs/proposals/in-flight/confinement-migration-plan.md
+++ b/docs/proposals/in-flight/confinement-migration-plan.md
@@ -1,0 +1,274 @@
+# Migration Plan: Centralizing Boundary Confinement onto `Boundary.Path.confine/3` + `Boundary.TermText.sanitize/2`
+
+**Status**: PROPOSED — migration half of the confinement seam. The contract half is
+[confinement-seam-proposal.md](./confinement-seam-proposal.md) (do not re-litigate the
+contract, must-reject tables, or dependency option (b) here — all ratified there).
+Both functions are **implemented** in `packages/raxol_core` (PR #613,
+`lib/raxol/core/boundary/{path,term_text}.ex` + shared conformance vectors at
+`test/support/boundary_vectors/`). This document is the site-by-site plan for
+V + Drew to approve and assign. No code lands from this doc directly.
+
+**Date**: 2026-07-17 · **Origin**: PR #569 thread 2 ("same gap, four patches"),
+V ruled "centralize it"; grounded in the 2026-07-17 full-tree discovery pass over the
+`confinement` worktree (main `lib/` + all 14 `packages/*`).
+
+---
+
+## Context
+
+`raxol_core` now owns both boundary functions and is the **publish root** (zero raxol
+deps, published first in the Hex order). Every in-repo candidate package already reaches
+it — directly (`raxol_terminal`, `raxol_symphony`, `raxol_speech`, main `raxol`) or
+transitively (`raxol_agent` via `:raxol`/`:raxol_mcp`). The single package that cannot
+take the dep is the external **Agent Client Protocol** package
+(`raxol_agent_client_protocol`, deliberately zero-raxol-dep, jason-only), which under
+ratified option (b) keeps its own `FsSandbox` copy bound to the shared vectors.
+
+Discovery found the live gap is **four hand-rolled lexical-only path checks** (no
+realpath/symlink re-check) plus **zero call sites for `TermText.sanitize/2`** — the
+entire harness-transcript UI renders live LLM/tool bytes into `Components.text()`
+unsanitized, relying on the "never embed raw ANSI" convention upstream.
+
+Three of the four seam-proposal seed sites are **not in this worktree**: the Agent
+Client Protocol `FsSandbox` is external; the PA-6 CAS `$blob`/`snapshot_ref` deref does
+not exist yet (harness storage here is jsonl fixtures, no content-addressed store); the
+#586 untrusted `:erl_tar.extract` path is absent (the only `:erl_tar.extract` in-tree
+operates on a self-created trusted backup). Their rows below are ordering slots, not
+immediate work.
+
+---
+
+## Decision
+
+### 1. Triage
+
+#### (A) Confirmed migration sites — 17 rows
+
+**Path-confinement (4 in-repo + 1 external):**
+
+| # | Site | Gap |
+| - | ---- | --- |
+| P1 | `packages/raxol_agent/lib/raxol/agent/actions/fs.ex:145` (`resolve/1`) | Lexical prefix check only (`Path.expand` + `starts_with?(cwd <> "/")`); LLM-supplied `path` for ReadFile/ListDir/FileStat; a symlink inside cwd pointing outside is not caught |
+| P2 | `packages/raxol_symphony/lib/raxol/symphony/path_safety.ex:56` (`validate_inside_root/2`, `workspace_path/2` at :37) | Pure lexical `inside?/2`; MCP-supplied workspace identifiers; **feeds `File.rm_rf!` at `workspace.ex:93–96`** — a symlinked workspace dir could let rm_rf escape root. Migrating PathSafety fixes the workspace.ex sink transitively (one migration, two discovered rows) |
+| P3 | `packages/raxol_terminal/lib/raxol/terminal/input/file_drop_handler.ex:524` (`validate_directories/2`) | Naive `String.starts_with?(file.path, allowed_dir)` — **missing trailing-`/` boundary** (`/foo/barbaz` passes for allowed `/foo/bar`), no normalization, and the allowlist is **not re-applied after** its own `resolve_symlinks_recursive/2` (depth 10, at :408). Untrusted drag-and-drop URIs |
+| P4 | `packages/raxol_agent/lib/raxol/agent/skills/store.ex:395` (`read_view/2` + `safe_relative?/1` at :408) | Rejects absolute + `..` segments lexically, no realpath; skill dirs can originate from downloaded/imported skills, so a planted symlink escapes |
+| P5 | Agent Client Protocol package (external repo) — `Raxol.AgentClientProtocol.Client.FsSandbox.resolve/2` | **No code change** (option b). Task = copy `path_reject_vectors.json` + `path_accept_vectors.json` VERBATIM into its test tree, assert `FsSandbox.resolve/2` agrees (map its Error `data.reason` onto the vector `expect` atom; SKIP vectors carrying `ref_format` — the ref-shape gate is `confine/3`-only), and add the duplicate-marker moduledoc line |
+
+**TermText-sanitization (12, all main `raxol` lib, UI lane):**
+
+| # | Site | Untrusted bytes |
+| - | ---- | --------------- |
+| T1 | `lib/raxol/ui/components/markdown_renderer.ex:60` (boundary) — covers text nodes :86, headings :94/107/118, earmark pre :171–172, link href :271, builtin fence body :379 | `markdown_text` prop: LLM/assistant message content. **This is the #607 family.** Sanitizing once at :60 transitively covers `harness/message_block.ex:58` and `playground/demo_helpers.ex` `markdown/2` |
+| T2 | `lib/raxol/ui/components/harness/tool_result_block.ex:156` | Tool/subprocess stdout — explicitly modeled as tainted (`provenance.trust: :tainted`); the prime injection vector |
+| T3 | `lib/raxol/ui/components/harness/reasoning_block.ex:108` (+ collapsed summary :98, `content_lines/1` :84) | LLM chain-of-thought content |
+| T4 | `lib/raxol/ui/components/harness/diff_viewer.ex:466` (`highlight/3`; + `state.path` at :273, fallback token :477) | On-disk file content + agent-proposed edit + file path; `language: nil` path passes raw bytes straight to `text()` |
+| T5 | `lib/raxol/ui/components/code_block.ex:28` (+ raw fallback :30) | Fence/code content; `strip_html_tags` removes HTML only — **ESC survives both paths** |
+| T6 | `lib/raxol/ui/syntax_highlighter.ex:114` (nil-language + no-Makeup fallback, also :76) | Raw source lines become token `text` — defense-in-depth behind T4/T5 (token producer, not a direct emitter) |
+| T7 | `lib/raxol/ui/components/harness/tool_call_block.ex:84` (name) + :113 (binary-args `truncate` path) | LLM-chosen tool name/args; map-args `inspect/1` path is already safe |
+| T8 | `lib/raxol/ui/components/harness/approval_prompt.ex:166` (`action_description`) + :239 (`option_label`) | Agent-proposed action text in a **security-critical approval UI** — spoofing matters most here |
+| T9 | `lib/raxol/ui/components/harness/blast_radius_preview.ex:228` (items) + :210 (group label) | Side-effect paths/identifiers in an irreversible-action warning surface |
+| T10 | `lib/raxol/ui/components/harness/advisory_feed.ex:93` (`entry_line/1` at :99) | `entry.source` / `entry.text` (sensor/agent/retrieval-sourced) |
+| T11 | `lib/raxol/ui/components/harness/toast.ex:60` | `state.message` (agent/external event text) |
+| T12 | `lib/raxol/ui/components/harness/composer.ex:516` (queued steer banner) + MLI value render path (:492) | Pasted input echo — bracketed-paste can carry ESC/ANSI; classic local-echo injection |
+
+#### (B) Trusted / internal — explicitly NOT migrating
+
+Considered and dismissed; listed so a reviewer sees they were weighed:
+
+| Site | One-line reason |
+| ---- | --------------- |
+| `packages/raxol_agent/lib/raxol/agent/curator.ex:232` (`:erl_tar.extract`) | Archive is a self-created backup written by the same module from its own root — trusted; not the #586 seed. Revisit only if backups become importable |
+| `lib/raxol/commands/file_system.ex:313` (VFS `resolve_path`/`normalize_path`) | Pure in-memory map-backed VFS, zero real-filesystem syscalls — `confine/3`'s realpath is meaningless; `..` is contained by construction (stack-pop can't escape `/`). Revisit if a real-FS passthrough is ever added |
+| `packages/raxol_core/lib/raxol/core/runtime/plugins/loader.ex:179/277` (`Code.compile_file`) | Arbitrary code load, not path traversal — `confine/3` is the wrong primitive; out of scope for this migration |
+| `lib/raxol/config/loader.ex:264/376` | Loading a config file the user explicitly names is a feature; there is no root to confine under |
+| `lib/raxol/themes.ex:210` | Theme-identifier-as-path has no root/base anchor; the confined variant is `style/colors/persistence.ex` (see C) |
+| `lib/raxol/headless.ex:331` | CLI-argument file compile; trusted operator input, no boundary |
+| `lib/raxol/plugins/examples/file_browser_plugin.ex:82` | A file browser's explicit purpose is unconfined real-FS traversal; example/demo code |
+| `packages/raxol_terminal` renderer/emulator/parser, `ui/components/terminal.ex`, `core/renderer/*`, `protocols/*`, `style/colors/*` | Trusted framework ANSI **producers** — the framework's own output generation, not an untrusted-content boundary |
+| Map/JSON "sanitize" false positives (`raxol_agent/contract.ex sanitize_payload`, `raxol_mcp/context_tree.ex sanitize_tree`, `raxol_payments/req/auto_pay.ex sanitize_error`) | Payload-shape sanitization, unrelated to either boundary |
+
+#### (C) Uncertain — needs owner confirmation
+
+| Site | Question | Ask |
+| ---- | -------- | --- |
+| `lib/raxol/core/runtime/plugins/api.ex:135` (`plugin_data_dir/1`) | Can `plugin_id` ever be third-party/attacker-influenced, or is it always a validated module-ish id? | If external: `confine(base_path, plugin_id, ref_format: ~r/^[\w.-]+$/)`. Owner: core runtime |
+| `lib/raxol/style/colors/persistence.ex:49/101/221/223` (theme name → path) | Is `theme_name` ever user/UI-supplied, or always internal? Textbook confine shape (root = themes dir) | Cheap hardening; owner: UI lane |
+| `lib/raxol/ui/components/harness/rules_panel.ex:89/97` | Can rules be agent-proposed, or operator-config only? | If agent-proposed: sanitize `when_clause`/`then_clause`. Owner: harness-ui |
+| `lib/raxol/ui/components/harness/worktracks_panel.ex:99` | Do worktrack names derive from branch/task names (external)? | Owner: harness-ui |
+| `lib/raxol/ui/components/harness/error_block.ex:72/81` | Fault payloads can embed tool output; only the `is_binary` branch leaks (inspect path is safe). Cheap to fold into the T-wave if owner agrees | Owner: harness-ui |
+| `lib/raxol/repl/evaluator.ex:113/142` (captured stdout of evaluated code) | Data source, not the render site — locate the actual `text()` emission for REPL results and sanitize there | Owner: REPL |
+| `packages/raxol_speech/lib/raxol/speech/tts/sanitize.ex` (`strip_control_chars/1`) | `TermText.sanitize/2` is a strict superset, and raxol_speech already has the direct dep — but the target is a TTS argv, not a terminal, and the vectors README scopes term_text to "raxol_core only (UI/terminal lane)". Reuse is **opportunistic, not required** | Owner: speech surface |
+
+### 2. Migration rows (confirmed sites)
+
+#### P1 — `raxol_agent` `Actions.Fs.resolve/1` (fs.ex:145)
+
+- **Current**: `abs = Path.expand(path, working_dir())`; accept iff `abs == cwd or String.starts_with?(abs, cwd <> "/")`; else `{:error, :outside_cwd}`. Then `File.ls`/`File.read`/`File.stat`.
+- **Target call**: `Raxol.Core.Boundary.Path.confine(working_dir(), path)`
+- **Diff sketch**:
+  ```elixir
+  # before
+  abs = Path.expand(path, cwd)
+  if abs == cwd or String.starts_with?(abs, cwd <> "/"), do: {:ok, abs}, else: {:error, :outside_cwd}
+  # after
+  case Raxol.Core.Boundary.Path.confine(cwd, path) do
+    {:ok, real} -> {:ok, real}
+    {:error, _reason} -> {:error, :outside_cwd}   # keep the public reason stable
+  end
+  ```
+- **Rejection mapping**: collapse `:path_traversal | :symlink_escape | :too_many_symlinks` → existing `:outside_cwd` (the LLM-facing tool error contract is unchanged; optionally log the granular reason).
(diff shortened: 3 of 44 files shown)


DROOdotFOO:
## Adversarial review — `harness-freeze-contracts.md`

Rigorous freeze, and the journal claims check out against landed code (framing, additive `kind`/`branch_id`, dense-id law `reader.ex:144-150`, `Reader.last_offset`, `snapshots/` placeholder; `N-JS7` is a correct negative control, not a live bug). Issues are in the security substrate and a stale baseline.

**Stale current-code claims (fix before reds encode them):**
- "command-result path emits nothing / silent model update — live bug" is **already fixed** — `dispatcher.ex:564-579` emits `:command_result :ephemeral`; the comment says so. Only the `session_id: nil` terminal case + missing `:render_needed` remain.
- "SessionStreamer unplumbed / Dispatcher never calls it" is **wrong** — wired via `EmitBridge` (`emit_bridge.ex:153,162,178`).
- `approval_requested` / `item_started` **don't exist yet**; only `turn_started` is landed.

**HIGH**
- **Gates read un-folded `trust`.** `OQ-U11.3` makes a taint miscount an alarm+marker, not a reject — a mis-stamped `:trusted` event reads trusted until something folds, yet YOLO clause-4 / blast-radius gate consumes that field to auto-approve. Freeze: gates admitting `TrustedLineage` MUST fold taint over `refs`, never read the stamped envelope value. (TOCTOU on the field the trifecta defense rests on.)
- **`schedule` sits outside the taint lattice.** `provenance`/`trust`/`actor` are `Event` fields; `schedule` (the autonomy/wake trigger store) is a separate *kind* with none — a wake armed from tainted context is indistinguishable from a trusted one. Give the arming decision a `refs`/provenance path now.

**MEDIUM**
- Silent swarm degradation: `submit` never fails + parking sheds → probes become no-ops while returning `{:ok}`. No backpressure signal. Freeze an observable saturation signal.
- Redaction (`§0` clause 6) vs the "complete fact" spine + I2 byte-identity — scope I2 as write-time or add a redacted-record I9 fixture.
- GC `low_watermark` can orphan a checkpoint `tip_offset` (retention deferred `OQ-JS2`); freeze "GC never truncates at/above the newest checkpoint's tip."
- Two-level probe budget (`OQ-U12.1`) partial-failure rollback unspecified (session reserved, run refused → release).
- **`N-JS7` cites `I13`, which doesn't exist** — `invariants.md` defines I1–I10; the publish-ahead concept is `I3`. Dangling ref in a must-fail negative control.

**Pattern:** the contract repeatedly picks *tolerate + mark + fold-later* over *reject synchronously* (taint→marker, saturation→park+shed, redaction→mutate). Together, a fully-green red suite can coexist with a live system not enforcing at decision time.

**Verdict: CONCERNS** — ratify after the two HIGH + the `I13`/stale-baseline fixes. Under "only grows" these are cheap now, permanent later.

<sub>Verified against landed code in `raxol_agent` + `lib/raxol/core/runtime` via three review passes.</sub>

merklebonsai:
Thanks — all accepted, fixes in flight (commit landing on this PR shortly):

- **HIGH-1 ratified as a general law**: admission decisions (gates, TrustedLineage, gc acceptance) evaluate the FOLDED truth synchronously at decision time; the stamped `trust` field is display/audit metadata only. Your pattern finding (tolerate+mark+fold-later vs reject-at-decision) is being added to §0 as the governing principle — tolerate-and-mark is for replay/audit seams, never admission seams. The in-flight U8-R suite is being updated live: its taint contour now asserts the gate folds over refs (mis-stamped :trusted with tainted refs MUST escalate) with a field-reading gate as the dead-injector.
- **HIGH-2**: `schedule` grows `armed_by` refs + arming-actor; `woken` refs include the schedule offset; a wake from tainted arming lineage folds tainted.
- **MEDIUMs**: saturation observable frozen (telemetry + status/2 distinguishes accepted-running from parked/shed; submit's ok = acceptance receipt, not execution promise); I2 scoped write-time + redacted-record I9 fixture; GC never truncates at/above newest healthy checkpoint tip; two-level budget releases session reservation on run-refusal (+ conservation red); I13→I3 fixed here and in the in-flight U4-R suite.
- **Stale-baseline**: all three corrected (command_result emit acknowledged, SessionStreamer wiring acknowledged, approval_requested/item_started marked contract-frozen-not-yet-landed).

Will re-request your look once the commit lands.

DROOdotFOO:
## Adversarial Review (automated: 3-persona, docs-adapted)

This is a large, deliberate corpus and the three frozen surfaces (JS-FREEZE / U11-CONTRACT / U12-CONTRACT) are internally coherent, self-justifying, and admirably precise about positive/negative contours and named dead-injectors. The landed-code claims spot-check as accurate (`emit_bridge.ex` `map_event/3` + `durable_record/1`, the journal modules, `contract.ex` all exist as described). The findings below are about **downstream navigability and two cross-cutting contract tensions** that the follow-on red PRs (#570-#574) will inherit — not runtime bugs.

**Verdict: CONCERNS**

Nothing here is a leaked secret or a flatly-wrong invariant (no secret/endpoint/key leakage found in a full-corpus scan). But two cross-persona items (the taint/auto-approve tension and the F-ID namespace collision) should be resolved **before the escalation/U8 and U4/U9 reds are authored against these tags**, even if this doc-only PR merges.

---

### CRITICAL

**[Security + Saboteur] `harness-freeze-contracts.md` §5.2 auto-approve predicate omits taint, contradicting the §2.1 taint algebra.**
§5.2 freezes the escalation gate as:
> "a call **escalates** (requires human approval) iff `effect_class == :irreversible_external` **OR** `egress == true`; otherwise it is **YOLO-applicable over trusted lineage**"

This predicate never mentions taint. But §2.1 taint algebra point 5 freezes:
> "Gates read taint: a tool call whose args derive from tainted content takes a **distinct confirmation path** (FI-5 verbatim)"

So a `:bounded_sandboxable`, `egress:false` call built from tainted content is **auto-approvable by §5.2's letter yet must be gated by §2.1.5**. These are two frozen, safety-critical contracts and their interaction is unspecified: does taint force escalation regardless of `effect_class`/`egress`, or does §5.2's YOLO branch win? This is the lethal-trifecta core. An implementer authoring the U8/escalation red can satisfy §5.2 while violating §2.1.5. The freeze needs one sentence stating taint is an **additional** always-on escalation input layered over §5.2 (or explicitly folding a taint term into the auto-approve predicate).

---

### HIGH

**[Saboteur + New Hire] The `F<n>` ID namespace collides across two committed files, and the constitution's F-references have no committed source.**
The constitution tags load-bearing rulings `(F2, ratify-before-impl)`, `(F4)`, `(F5 fix)`, `(F6)`, `(F7)`, `(F3)`, `(F9)` — per the PR body these are the "longcat audit" findings ("F1-F9 fixed"), but **those audits are not committed** in this PR. Meanwhile the *only* committed `F<n>` table (`harness-future-foundations-ideation.md` lines 82-99) defines an **independent** F1-F18 with clashing meanings:

| ID | constitution meaning | future-foundations meaning |
|---|---|---|
| F2 | "not-yet-landed `Raxol.Action` draft / `effect_class` / scope values" (§5.2, §2.4) | "Fuse/merge threads" |
| F4 | "settlement is internal to probe interface" (§3 Budget) | "Session transplant / export / import" |
| F5 | "bounded parking / `max_parked` / `park_timeout_ms`" (§3) | "Partial-journal sharing / redaction" |
| F6 | "`last_offset` semantics" (§1.1 Reader) | "Model-migration / counterfactual replay" |
| F7 | "single-Writer lockstep test" (N-JS6) | "Live multi-user sessions" |

A red-author resolving "(F5 fix)" against the only committed F-source lands on *partial-journal redaction* when the actual referent is *bounded probe parking*. Since #570-#574 reference these tags, please either (a) commit the audit F-findings, or (b) rename the constitution's audit-finding tags to a non-colliding prefix, or (c) drop them to prose.

---

### MEDIUM

**[Saboteur/New Hire] Dangling invariant reference `I13`.**
`harness-freeze-contracts.md:477` (N-JS7) cites "emit-ahead-of-journal, **I13**", but `harness-invariants.md` defines only **I1–I10** (the constitution's own header at line 6 says "I1–I10 + meta-invariants"). I11/I12/I13 are undefined anywhere in the corpus.

**[New Hire] Sub-lettered `AD-3a` / `AD-3b` / `AD-6a` are used as governing dispositions but never defined.**
JS-FREEZE ("Governs: … AD-10/AD-3a … AD-3b") and the Budget section ("reserve-before-call … (AD-6a)") lean on these, but `harness-storage-research.md` defines only `AD-8..15`, and `harness-synthesis.md` defines the *parent* `AD-3` ("compaction and resume are one subsystem", line 170) and `AD-6` ("blast-radius gate", line 179). The `a`/`b` sub-letters are never spelled out — a reader must infer AD-3a=checkpoint, AD-3b=compaction, AD-6a=reserve-before-call.

**[New Hire] No index / map / glossary for 32 files / 9811 lines.**
`docs/proposals/in-flight/` has no README and the corpus has no glossary. `CAS` (§0 clause 5, "relative CAS paths") is never expanded; `SpendGate`, and the parallel ID schemes (`AD-*`, `FI-*`, `F-*`, `NC-*`, `OQ-*`, `I*`, `meta-inv N`, `U*-R`, `R1`, `N-*`, `P-*`) are used throughout with no single resolution map. For a doc set explicitly meant to be authored-against in parallel, a one-page index (file -> purpose, ID-prefix -> owning doc) would pay for itself immediately.

**[Saboteur + Security] The `annotation`/`schedule` "any writer (reopen the single Writer)" phrasing under-specifies the mutual exclusion that NC-12 / N-JS6 depend on.**
The `annotation` kind is described as "any writer (reopen the single Writer)" and "**open-any-writer through the single Writer**" (§1.1 kind table), yet NC-12 + the offset law require a strict single-writer/single-counter, and N-JS6 is called "the sole guard against a silent second-counter bypass … the Writer … cannot enforce single-counter routing at compile time." "Any writer can reopen" vs "single writer" is a real ambiguity: the freeze never names the mutual-exclusion mechanism (a singleton `Writer` GenServer? an advisory lock on a dead-BEAM session dir?) that lets an out-of-band annotation append without racing the offset counter. An implementer could satisfy "I routed through a `Writer` module" while running two concurrent writers and still pass everything except a timing-sensitive N-JS6.

---

### LOW / nitpicks

- **[New Hire] `R1` cross-reference is silent.** `harness-freeze-contracts.md:608` says billing is "a derived fold over Ledger + lineage subtrees (walk the **R1** edges)". `R1` is defined only in `harness-future-foundations-ideation.md §3.1` with no pointer from the constitution.
- **[Saboteur] `tip_offset: pos_integer()` cannot encode `:no_tip`.** The tip is "undefined on an empty/no-conversational branch -> `:no_tip`" (§1.1-tip), and N-JS1 rejects a checkpoint whose `tip_offset` fails `conversational?`. Consequence: a checkpoint on a fresh/no-conversation branch has no legal `tip_offset`. Probably intended (can't checkpoint before any conversation), but it's not stated — worth an explicit sentence.
- **[New Hire] Superseded-doc hazard with no in-file banner.** The intro declares `harness-spec-backend.md §4` ("foldable Ecto event table (or DETS)" + "Oban = probe scheduler") **superseded** by D1/D2, but that file stays committed with no banner; a reader opening it directly gets stale storage/runner guidance. A one-line "SUPERSEDED, see harness-freeze-contracts.md" note at the top of that section would prevent it.
- **[Saboteur] Unverifiable provenance claims.** The PR body's "5-model quorum", "93-find community survey", and "2 adversarial longcat audits (F1-F9 fixed)" are not reconstructable from the committed corpus (the F-findings they cite aren't present — see HIGH above).

---

### Cross-persona overlaps (promoted)

- **Taint vs §5.2 auto-approve** — Security **and** Saboteur independently flagged the escalation predicate; promoted to **CRITICAL** (safety-critical contract contradiction on the trifecta gate that U8-R depends on).
- **`F<n>` namespace collision** — Saboteur (self-contradictory/undefined IDs) **and** New Hire (can't resolve a reference) → **HIGH**.
- **Single-writer ambiguity** — Saboteur (satisfy-the-letter loophole) **and** Security (the invariant bundle NC-12 protects) → **MEDIUM**.

Great foundational effort overall — resolving the two cross-persona items above (a sentence each) and adding a one-page ID/file index would take this from CONCERNS to CLEAN and materially de-risk the #570-#574 fan-out.

merklebonsai:
@DROOdotFOO fixes landed in `41bfbd66d` — ready for re-review.

**HIGH-1** (the load-bearing one): new **§0 clause 7** freezes the general principle you named — *tolerate-and-mark is for replay/audit seams; admission decisions (gates, gc acceptance, gate-relevant trust) MUST fold the truth synchronously at decision time; a replay-only red suite is incomplete without its decision-time counterpart.* Echoed as §2.1 taint-law point 6 (stamped `trust` = display/audit metadata, gates fold over refs), §5.2 predicate, and **N-U11.11** (dead-injector: field-reading gate passes a mis-stamped :trusted event with tainted refs). The in-flight U8-R suite carries the matching decision-time contour.

**HIGH-2**: `schedule` grows required `armed_by` refs + arming-actor; `woken` refs include the schedule offset; tainted arming lineage folds to a tainted wake.

**MEDIUM**: saturation telemetry + 'acceptance receipt not execution promise'; I2 scoped write-time + mandatory redacted-record I9 fixture; GC-never-orphans-checkpoint-tip (rejected at the :invalid_tip seam); two-level budget releases session reserve on run-refusal + conservation red; I13→I3.

**Stale-baseline**: emit-vocabulary landing-status paragraph added (approval_requested/item_started frozen-but-not-landed). Note: the command-result-path and SessionStreamer claims you flagged don't textually appear in *this* doc (grep-verified) — they're likely in the roadmap/spec-backend sibling docs; can chase separately if you want.

Your pattern finding is now the doc's governing principle, not a patch.

merklebonsai:
Superb review — the cross-persona promotion is exactly right. Fixes in flight (commit landing shortly):
- **CRITICAL taint/§5.2**: already fixed last round (§0 clause 7 decision-time-fold law + N-U11.11 field-reading-gate injector) — confirmed the U8-R suite carries the matching fold-vs-field contour.
- **HIGH F-namespace collision**: renaming all audit-finding refs to **AF-<n>** (distinct from future-foundations F-tags) with a disambiguation note; the #570-587 refs resolve correctly after.
- **MEDIUM**: AD-3a/3b/6a glossed inline; single-writer mutual-exclusion named (per-session-dir :global singleton Writer — 'any writer' = routes THROUGH the one process, never two coexisting; a second = the N-JS6 violation); **one-page in-flight/README.md index + ID-prefix→doc map + CAS/acronym glossary** added.
- **LOW**: tip_offset-on-fresh-branch sentence, R1 pointer, spec-backend §4 SUPERSEDED banner, I13→I3 (done last round).

The single-writer-mechanism and the index were the two that most de-risk parallel authoring — good catches.

merklebonsai:
@DROOdotFOO round-2 fixes landed in `26e6d48ea` — ready for re-review:

- **HIGH F-collision**: all 20 audit-finding refs renamed `F<n>`→`AF-<n>` with a disambiguation gloss at first use; the genuine future-foundations F-tags (F14 annotation, F9 schedule) and the pre-existing f2-action-registry F2 namespace verified-and-left-alone. #570-587 refs now resolve unambiguously.
- **AD sublabels** glossed inline (AD-3a=checkpoint, AD-3b=compaction=resume, AD-6a=reserve-before-call; parents in synthesis).
- **Single-writer mechanism** named as a frozen paragraph: the :global-registered per-session-dir singleton Writer GenServer (verified against the landed writer.ex); 'any writer' = routes through the one process, a second = the N-JS6 violation.
- **Index/glossary**: new `docs/proposals/in-flight/README.md` — file→purpose table, ID-prefix→owning-doc map, acronym expansions, audit-trail provenance note.
- **LOW**: no-legal-tip-on-fresh-branch sentence, R1 pointer, spec-backend §4 SUPERSEDED banner.

That closes every item from the round-2 review. The two cross-persona promotions (decision-time-fold from round 1, single-writer mechanism this round) are both now frozen laws. Combined with the in-flight red-suite fixes (#570-573), the fan-out contract should be at CLEAN.

merklebonsai:
Housekeeping pass landed (`9bf2efe36`) — addresses your provenance critique + de-risks the impl phase:
- **`harness-research/audit-trail/`** — the 2 longcat audits, the composer+longcat ruling quorum, and the grok-4.5 ideation are now committed verbatim with an index (the '5-model quorum / adversarial audits' claims are reconstructable from the repo, not just PR prose). README flags where composer and longcat genuinely disagreed (Q2 actor breadth, Q3 fingerprint breadth) — resolved by the event-sourcing 'record-at-source' principle.
- **`harness-parked.md`** — every deferred/flagged item from the red-PR authoring collected into one tracked home (restore-path hardening, U4/U7/U8/U12 contract gaps, the U21 gating-strength policy question, wiring debt, deferred @tag items, #566) so nothing's buried when reds merge.
- **roadmap synced** — merges marked, red-first fan-out section added (unit→PR map), impl critical path U5-I→U7-I→U4-I→U9-I recorded.

(One pre-existing stray code-fence at the roadmap tail was left untouched — unrelated doc bug, flagging rather than silently touching it.)

DROOdotFOO:

Scope: `docs/proposals/in-flight/harness-freeze-contracts.md` (the constitution -- 1327 lines, read in full). The other 30 files were treated as provenance per the PR's stated review focus. Three personas run independently; every finding is grounded in a section/line with a concrete consequence for the red-suite authors and the UI fork that consume these frozen shapes.

### HIGH
- [Saboteur] §0 clause 3 + §1.1 closure rule + §1.4 -- **the only-grows law does not actually protect tip stability across version skew.** Adding a `CONVERSATIONAL` member is a `schema_version` minor bump. §0.3 promises "readers built against 1.x read all later 1.y" and §1.4 claims the tip predicate "it can read forward." But the tip predicate is version-coupled: a reader/UI-fork built against 1.x lacks the newly-whitelisted type, so on a 1.y journal it selects an *older* record as tip than a 1.y reader does. §1.4 only forbids *removing* a member ("would silently move historical tips"); the symmetric additive case moves tips for older readers just as much. Consequence: reattach/resume (U4) lands at the wrong offset under version skew -- the "UI fork never needs a lockstep update" promise the freeze exists to guarantee is false for all tip-derived behavior. Needs an explicit rule (older reader refuses/degrades tip on a higher `schema_version` journal, or CONVERSATIONAL additions are barred from minor bumps).
- [Security Auditor] §0 clause 6 + §1.1 `$redacted` marker -- **the sole legal rewrite of the append-only audit log carries no authorization, actor, or approval requirement.** `$redacted` payload is `%{reason, at_ts}` only. The one operation that destroys evidence in an integrity-critical log is the *least* gated act in the whole contract, while `promote` (a knowledge rollup) mandates a human `approval_decided` and `refs != []`. There is no `redaction` meta event recording who authorized the scrub, no envelope `actor` binding, and no negative contour asserting an unauthorized redaction is rejected. Frozen as-is, this bakes an unattributable evidence-destruction primitive into a system whose entire value proposition is a tamper-evident journal. The hash/framing story (P-JS12) is covered; the authorization/provenance of the redaction *act* is entirely absent.

### MEDIUM
- [Security Auditor] §2.1 taint point 2 -- **the root of the trust boundary is explicitly NOT frozen.** "Which sources are untrusted is policy ... NOT frozen here." The derivation algebra is robust (tainted-absorbing, fold-at-decision), but entry classification at the `tool_result` producer is unfrozen per-deployment policy with *no negative contour / dead injector*. Two implementations can legally disagree on what is tainted, and the mutation suite cannot catch a mis-classification at the source -- the fail-closed lattice is only as sound as an out-of-contract policy the reds never exercise.
- [Saboteur / Security overlap] §0 clause 2 vs §2.1.1 / §2.4 -- **conflicting tolerance rules for unknown trust values.** §0.2 says readers "skip [unknown] in typed folds"; §2.1.1/§2.4 say unknown trust folds "fail-closed to `:tainted`." For a taint fold over `refs` where one ref carries a future (unknown) trust value, the general rule drops it from absorption (fail-open) while the specific rule taints the result. §2.1 states the governing carve-out, so it is resolvable -- but a red author applying §0.2 literally ships a fail-open taint fold. Needs an explicit "§2.1 trust-value handling overrides §0.2 skip." (Held at MEDIUM rather than mechanically promoted: §2.1 already states the correct rule, so the defect is a missing cross-reference, not an unresolved hole.)
- [Saboteur / New Hire] §1.1 tip predicate vs closure-rule prose -- **the closure rule is structurally impossible to satisfy for record kinds.** `conversational?` hardcodes `record.kind == "event"`, yet the closure rule says "every new record *kind* ... is tip-excluded **unless** explicitly added to CONVERSATIONAL." A non-event kind can never pass the first clause, so the "unless" is unreachable for kinds -- they are permanently excluded regardless of the whitelist. Safe direction, but it will confuse any implementer trying to determine how a future kind could ever become tip-eligible. Either the predicate or the prose is wrong.
- [New Hire] Whole doc -- **the "constitution" is not self-contained despite being the freeze target.** Load-bearing red-suite obligations `meta-inv 1/2/5/6` (fired-counter, seed-reproducible schedules, oracle independence, required generator patterns -- cited at §1.3 L567, §1.2 L487/L495, §2.2 L850, §4 L1234) are referenced repeatedly but **never enumerated in this file**; they live in `harness-invariants.md`, which is not frozen here. Same for numerous `AD-*/FI-*/I*` IDs. Permanent suites are authored against these shapes, yet a red author must chase mutable sibling docs for the meaning of the invariants they must implement. A freeze doc should inline or version-pin every obligation it makes binding.
- [New Hire / Saboteur] §3.1 cache-riding vs Bounded parking -- **prefix byte-identity is undefined for parked runs.** Cache-riding requires a prefix "byte-identical to the primary loop's *most recent* request prefix at `tip_offset`." A run can be parked up to `park_timeout_ms` before executing, by which time the primary loop has advanced. The contract never says whether byte-identity is measured against tip-at-submit or the current primary prefix. A parked run riding a stale prefix cannot cache-hit against the live primary, quietly defeating the economic premise (P-U12.3) that justifies the whole probe design. Needs a frozen rule tying the ride target to a fixed offset.
- [Saboteur] §1.1 (checkpoint same-branch tip) vs §1.3 N-JS1 -- **cross-branch tip validation has no negative contour.** §1.1 requires a checkpoint's `tip_offset` to name a conversational record on the checkpoint's *own* `branch_id`, but N-JS1's dead injector only covers a tip pointing at a non-conversational record. A checkpoint whose `tip_offset` names a conversational record on the *wrong branch* has no must-fail red, so that bug ships green.

### LOW
- [Security Auditor] §1.1 `annotation.consent_class` (`private|shareable|train`) and §2.1 `policy_amended` -- enforcement unspecified. "Sharing/export honors `consent_class`" names no enforcer and no negative contour; a `train`-class leak has no dead injector. `policy_amended` is "journaled BEFORE enforcement" but ordering against a concurrent decision-time gate fold is not pinned.
- [New Hire] §1.1 GC low-watermark + P-JS4 -- restoring from an *old* (non-newest) checkpoint whose `tip_offset` was truncated below `low_watermark` relies on a "covering checkpoint," but which checkpoint covers a truncated old checkpoint is not defined. Deferred (`@tag :gc`), but the restore formula is underspecified for that path.

### Cross-Persona Overlaps
- Trust boundary left to unfrozen policy: Security (taint entry classification, §2.1 pt 2) and New Hire (`consent_class` enforcement, §1.1) independently land on "the enforcement/classification that matters is out of the frozen contract with no negative control."
- Unknown-trust fold: Saboteur (rule contradiction, §0.2 vs §2.4) and Security (fail-open taint drop) hit the same clause from correctness and security angles. Per the promotion rule this qualifies for HIGH; held at MEDIUM because §2.1 already states the governing rule and to respect the docs-PR bias.

### Verdict: CONCERNS
Two genuine HIGH findings in a *frozen* contract -- an unattributable evidence-destruction primitive (`$redacted` with no authorization) and a tip contract that silently diverges across version skew (defeating the freeze's central UI-fork promise) -- plus several MEDIUM gaps in negative coverage and self-containment. None rise to BLOCK: no CRITICAL, HIGH count under 3 (the unknown-trust overlap was not inflated). All are fixable in-doc before the red suites are authored against them, which is exactly the window this freeze targets. Recommend addressing the two HIGH items plus the cache-riding and cross-branch-tip coverage gaps before ratification.

merklebonsai:
## Response to adversarial review — PA-1..PA-5 drafted (PROPOSED, awaiting V ratification)

Pushed to `docs/harness-freeze-constitution` as a new **§7 "PROPOSED amendments"** block. This is a **constitution one-way-door**, so nothing in §7 is finalized — it is drafted-not-applied and awaits V's ratification. Two clean doc-consistency clarifications were applied inline (they change no law).

### HIGH-1 — tip stability across version skew (the headline). Confirmed, and it's deeper than a missing symmetric clause.

**Root cause:** the tip is an `argmax`-by-offset **selection**, not an associative **fold**. Skip-unknown (§0.2) is safe for folds — a dropped record degrades the result *consistently*. For `argmax`, dropping the max element doesn't degrade the answer, it *changes* it to a wrong one. So a 1.x reader skipping a newly-whitelisted CONVERSATIONAL type on a 1.y journal selects a silently-**older** tip → U4 reattach lands wrong → the "UI fork never needs a lockstep update" promise is false for *all* tip-derived behavior. §1.4 forbade only *removal*; the additive case moves tips across skew just as much and was simply never named.

**Options weighed:**

| Option | Cost to UI fork | Keeps no-lockstep promise? | vs ratified tip law |
|---|---|---|---|
| **A. Reader-seam skew-guard (degrade tip loudly)** ✅ | Zero to read/fold/render; resume degrades *only* on an actual collision | Refines it honestly: loud-degrade replaces silent-wrong; lockstep needed only to resume at a colliding newer-minor tip | **Clean** — tip stays derived/recomputable/dual-oracle-checkable, whitelist stays the only door, guard is inert at same-version so every existing red is unchanged |
| B. CONVERSATIONAL adds → MAJOR bump | Walls 1.x readers out of a 2.0 journal entirely, even when the new type is nowhere near the tip | No — makes lockstep *required* for all conversational growth | Reverses ratified "minor bump = additive growth" (§0.3) and "adding a member is a minor bump" (§1.1); breaks AD-11 upcast-on-read |
| C. Writer stamps `tip_eligible` (version-robust) | Zero; genuinely *preserves* correct tips under skew | Yes, strongest | **Guts two ratified properties:** derived-not-stored (HEAD-lag class), and — fatally — P-JS2/P-JS8 are verified by *two independent oracles recomputing* the tip (meta-inv 6); a stamped bool collapses both oracles to one source. Also still needs a grandfather predicate fallback. |

**RECOMMENDATION: Option A**, in a precise (minimally over-broad) form. A reader degrades **iff** there exists a higher-offset same-branch record with `kind==event ∧ family==loop ∧ type ∉ its own known loop-type registry` — a type that *could* be a CONVERSATIONAL member it cannot see. If everything trailing the computed tip is known-non-conversational (checkpoint/meta/idle/`woken`), the tip is **provably correct even under skew** and the reader does not degrade. So the common case (new types not near the tail) computes the correct tip and never degrades; degradation fires exactly in the collision case, never spuriously. Option A is the only choice that keeps *every* ratified property intact while making the promise *true* instead of false, converting silent-wrong-offset into a fork-handleable `:tip_uncertain` — same fail-closed philosophy as §0.7 and the taint lattice. Drafted §0/§1.4 text + a new inert must-fail red **N-JS-skew** are in §7 PA-1.

### Other findings — dispositions

- **HIGH-2 `$redacted` unauthorized (PA-2):** confirmed. Drafted an additive gate — new grow-only `redaction` meta type binding `actor`+`authorized_by`, a producer-seam reject for a rewrite with no accompanying `redaction` event, and must-fail **N-JS-redact**. Brings the most destructive op up to `promote`'s provenance bar without an optional→required flip. Contract ruling → **flagged for ratification.**
- **§3.1 cache-riding for parked runs (PA-3):** ride target frozen to **tip-at-submit** (fixed offset), so byte-identity stays provider-free-testable; a stale target predictably cache-misses rather than being undefined. Contract ruling → **flagged.**
- **cross-branch tip contour (PA-4):** drafted must-fail **N-JS-xbranch** (`{:error, :cross_branch_tip}`, same shape as N-JS1) — closes the wrong-branch-but-conversational tip gap N-JS1/N-JS8 miss. Additive → **flagged.**
- **taint entry-classification (PA-5):** narrow draft — keep *which* sources are tainted as policy, but freeze that a classification **exists and fails closed**, + dead injector **N-U11-classify**. Closes the "no negative control at the source" gap without freezing policy content. → **flagged.**
- **§0.2 vs §2.4 unknown-trust (MEDIUM, doc-consistency):** **fixed inline** — added the §0.2 carve-out that skip-unknown is a *fold* rule; taint folds fail-closed and the tip selection state their own overriding rules.
- **closure-rule prose vs `kind=="event"` predicate (MEDIUM, doc-consistency):** **fixed inline** — CONVERSATIONAL clarified as a *type* whitelist; new *kinds* are excluded structurally, whitelist-add is unreachable for kinds (making a non-event kind tip-eligible would need a deliberate predicate amendment, not a whitelist add).
- **self-containment / meta-inv enumeration (MEDIUM):** recommend **version-pinning** `harness-invariants.md` + the `AD-*/FI-*` siblings in §0 rather than inlining all; pin-rev awaits V. Noted in §7.
- **LOW (consent_class enforcement, GC covering-checkpoint):** noted as deferred follow-ups (the GC one is already `@tag :gc` / OQ-JS2).

All §7 items are **PROPOSED — V ratifies frozen-law changes.** Flagging @merklebonsai for the ratification pass on PA-1..PA-5 and on whether the two inline clarifications should stand.

merklebonsai:
## Update — PA-1 RATIFIED (Option A) and implemented; PA-2..PA-5 still pending

V ratified **Option A — the reader-seam tip skew-guard** (2026-07-16). It's now promoted from §7-PROPOSED to **frozen law**:

- **§1.4 — "The tip skew-guard (RATIFIED by V, 2026-07-16)".** A reader computes the tip over the types it knows, then MUST degrade to `{:tip_uncertain, reason}` **iff** an unknown-to-it `family:loop` event sits at an offset **above** its computed tip on a higher-`schema_version` journal; otherwise it returns the correctly-computed tip. Minimally over-broad by construction (fires only in the collision case) and **inert at same-version**. The rule states the invariant it preserves — tip stays **derived-not-stored**, **whitelist-only door**, **P-JS2/P-JS8 dual-oracle recompute** all intact — which is exactly why Option A beat B (major-bump, reverses minor-only growth) and C (stamp `tip_eligible`, collapses the two determinism oracles to one source). Root cause is recorded: the tip is an `argmax` **selection**, not a fold, so §0.2 skip-unknown never protected it.
- **§1.3 — N-JS-skew** is now a real must-fail red (bound to the meta-inv 1/2 fired-counter convention via the existing "every fault site above" note). Dead injector: a reader that silently falls through to the older offset — so a regression toward the reattach-lands-wrong bug is caught green-on-broken.
- §0.2 carve-out cross-ref updated to point at the ratified §1.4 rule / N-JS-skew.
- The two inline doc-consistency clarifications (closure-rule prose §1.1; skip-unknown carve-out §0.2) stand.

**PA-2..PA-5 remain PROPOSED — awaiting separate V ratification** (unchanged): PA-2 `$redacted` authorization (`redaction` meta type + N-JS-redact), PA-3 cache-riding ride-target = tip-at-submit, PA-4 cross-branch-tip N-JS-xbranch, PA-5 taint entry-classification existence + N-U11-classify, plus the LOW/self-containment notes. The §7 header now flags PA-1 as ratified-and-promoted and PA-2..5 as still pending. @merklebonsai for the next ratification pass on PA-2..5.

Pushed to `docs/harness-freeze-constitution` (`ced8a16e8`). No merge.

merklebonsai:
**Transport addendum** (§1.5, V-ratified 2026-07-16): added **OQ-JS4** (RATIFIED — `{:reattach_live, session_id, record}` is the frozen U4 live-tail wire; `record` is the full durable record, grow-only), **OQ-JS5** (RULED — `attach` grows a trailing optional `opts` with `:base_dir`, env-default; done now while the arity door is cheap, purely additive), and **OQ-JS6** (OPEN — the broadcast/permission bus, AD-15 second half, under an active dual research+design pass). U4-I is unblocked for read-side + single-subscriber live tail against OQ-JS4; the bus stays deferred until OQ-JS6 ratifies. Closure law (P-JS5) + no-publish-ahead (I3/N-JS7) unchanged.

DROOdotFOO:
## Adversarial review (fresh pass) + a request to restructure toward ADRs

Ran `/adversarial-reviewer` (Saboteur / Security Auditor / New Hire) against `harness-freeze-contracts.md`. Since the doc already carries a §7 audit (PA-1..PA-5), every finding below is **net-new** — cross-checked against PA-1..PA-5 and the LOW list so this isn't re-reporting your own audit. All route cleanly through the additive §7 amendment lane as PA-6..PA-9.

### HIGH

**1. [Security] CAS pointer deref has no runtime confinement — traversal via `$blob` / `snapshot_ref`.**
Self-containment (§0.5) is enforced only by a *fixture lint scan*. The runtime deref rules (§1.1 `$blob` "deref-or-render-opaque"; `snapshot_ref: "snapshots/<sha>.json"`) never require the reader to validate the path is a bare hex sha256 confined to `<session>/blobs|snapshots`. **N-JS10 covers missing / write-ordering blob failures but not a pointer that escapes the session dir** (`{"$blob": "blobs/../../../../etc/…"}`). An `:import`ed or corrupted journal (import lineage is frozen now) feeds attacker-controlled pointers into the tolerant reader seam. No contour asserts deref rejects a non-confined / non-sha256 ref → ships green.
*Fix:* freeze a deref-confinement rule (path must match `^(blobs|snapshots)/[0-9a-f]{64}(\.json)?$`, resolved inside the session dir) + a must-fail red `N-JS-traversal`.

**2. [Security] `$redacted` on snapshot content does not remove the secret in v1.**
§1.1: redaction of a snapshot "produces a NEW snapshot file; the referencing checkpoint's hash then names the pre-redaction file, which **may be deleted under FI-7 consent**." But FI-7 is "never deleted implicitly" and GC is *deferred entirely* (OQ-JS2). So for all of v1 the pre-redaction snapshot — the secret, content-addressed, on disk — is never deleted. The "one legal rewrite" scrubs the event *payload value* but leaves the *snapshot bytes* recoverable. Distinct from PA-2 (authorization): even a fully-authorized redaction fails to actually redact snapshot content.
*Fix:* either (a) redaction of snapshot-referenced content requires synchronous deletion/overwrite of the pre-redaction CAS file (not deferred to GC), or (b) explicitly document that snapshot redaction is a no-op until GC ships, so nobody relies on it for PII/secret scrubbing.

**3. [Saboteur + Security → promoted MED→HIGH] GC truncation orphans taint lineage (`refs` / `armed_by`).**
The taint algebra folds over `refs` (§2.1.4) and `schedule.armed_by` (§1.1). GC may truncate any prefix below the watermark. A `ref`/`armed_by` offset below the watermark can't be read, and the covering checkpoint stores MS model state, **not** per-event trust stamps. The contract freezes "GC never orphans **checkpoints**" but is silent on GC never orphaning **taint lineage**, and never says whether an unreadable ref target folds fail-closed `:tainted` or reads `:trusted`. This defeats the §2.1.3 "no laundering" invariant: GC the tainted ancestor, and the descendant's fold can no longer see it. GC is deferred, but `schedule.armed_by`/`refs` freeze *now*.
*Fix:* freeze "GC never orphans a live `refs`/`armed_by` target above any surviving descendant" as a companion to the no-orphan-checkpoints law, OR freeze "unreadable ref target folds fail-closed `:tainted`."

### MEDIUM

**4. [Saboteur] §5.1 `client_msg_id` — zero negative contour + silent collision-swallow.**
§5 carries **no must-fail reds at all**. A genuinely-new command that reuses an id is deduplicated as "a live ack referencing the original turn, NOT a second durable event" — silently dropped, never journaled, no error. Also the "in-memory index rebuilt by fold" grows unbounded over a session-lifetime dedup window with no frozen bound. Needs an `N-cmd-dedup` contour + an index-bound decision.

**5. [Saboteur] `schedule` missed-wake semantics on resume are unfrozen.**
The record freezes `next_fire`/`armed` and the session enum grows `:dormant`, but nothing specifies firing when a dormant session resumes with `next_fire` in the past: fire once? fire-per-missed-interval (thundering herd)? drop? Record shape frozen; firing law that reds need is absent.

**6. [New Hire] The binding ratification authority "V" is never defined.**
"RATIFIED by V" / "V's ruling" gates frozen-vs-proposed (PA-1 promotion, OQ-JS4/JS5, eval gate, tip skew-guard) and is expanded nowhere — not in the doc, not in the README ID-map. Compounding: the README admits the AF-* audit provenance "lives in the PR review threads, not here," so the authority behind part of the doc's claims is uncommitted/unreachable.

### LOW
- A "frozen" constitution ending in four non-frozen one-way-door amendments (PA-2..PA-5) forces a red author to hold "build against §0–§6 incl. PA-1, ignore PA-2–5" as out-of-band state. Belongs in status metadata, not inside the file.
- `F2` has three live meanings (self-documented in the README after a mid-flight rename) — standing hazard.

### Verdict: **CONCERNS** (escalates to **BLOCK** if the affected surfaces are frozen-as-is)
Three HIGH mechanically implies BLOCK, but the doc changes zero code and has an additive amendment lane, so the doc merges fine. **However** — if red suites are authored against the CAS-deref seam, the redaction/snapshot path, GC×refs, or §5.1 as currently frozen, those gaps ship green. Recommend: **do not treat those four contours as frozen until amended (PA-6..PA-9).**

---

## Request: reduce the planning sprawl to match our ADR process (`docs/adr/`)

This is 41 files / ~11.5K lines in `docs/proposals/in-flight/`. Our formal decision record is one ADR = one decision (template: Status / Context / Decision / Consequences {Positive, Negative, Mitigation} / Validation / References; sequential numbering; index table in `docs/adr/README.md`). A single 1,663-line "constitution" plus a 30-file research corpus is the opposite of that, and it's why the review above needed six cross-referenced docs and a private tag glossary to navigate.

The constitution already *is* ~4 distinct architectural decisions wearing one hat. Proposed split (next free number is **0030**; 0029 is taken on disk):

| ADR | Source | Decision it captures |
|---|---|---|
| **ADR-0030 — Harness contract-evolution law** | §0 | only-grows, two-seams strictness, self-containment, decision-time-fold. Cross-cutting; the next three reference it instead of restating it. |
| **ADR-0031 — Agent journal record-kind schema & offset law** | §1 (JS-FREEZE) | kinds, single offset law, derived tip + closure rule, `branch_id`, lineage DAG, `$blob`/`$redacted`, GC low-watermark. |
| **ADR-0032 — Meta-event family + provenance/taint algebra** | §2 (U11) | meta type registry, `refs`, two-point taint lattice, actor/fingerprint precedence. |
| **ADR-0033 — Probe runner interface** | §3 + §6 | pure build/interpret, reserve-before-call, bounded parking, isolation, cache-riding, eval gate as Validation. |
| **ADR-0034 — Command idempotency + action reversibility taxonomy** | §5 | `client_msg_id`, `effect_class`/`egress`. Distinct seam → its own short ADR. |

How the current sections map into the template (this is the mechanical part):
- **Frozen shape (§X.1) → Decision.**
- **Positive/Negative contours (P-* / N-*, §X.2–3) → Validation** — and the *actual* P-*/N-* text moves into the red-suite test moduledocs where it's executable; the ADR just cites the suite. (Contours are test specs, not architecture prose.)
- **Forward-compat "Grows by / Never" (§X.4) → Consequences → Negative + Mitigation.**
- **Ruled/open (§X.5):** resolved OQs fold into Decision; genuinely-open ones (OQ-JS6) become an "Open items" line under Status or a linked issue.
- **§7 amendments:** the ADR process already models this — PA-1 (ratified) folds into ADR-0031's Decision; PA-2..PA-5 (proposed) become either follow-up ADRs or `Status: Proposed` open items with issues. No in-file pending-amendments section, no "V" ambiguity — `Status: Accepted` vs `Proposed` *is* the ratification signal.

The 30-file research corpus (`harness-research/*`, `harness-storage-*`, `-synthesis`, `-facts-*`, `-cohort-*`, `-community-gaps`, `-yolo-safe-*`, `-eval-first-*`) is **input to** the decisions, not decisions themselves. Recommend moving it to `docs/proposals/archive/harness/` (or dropping it from this PR entirely and keeping it local), cited only from each ADR's **References**. That's the "why" preserved without 10K lines of it living in the active proposals lane.

Net effect: five focused, template-conformant ADRs a cold reader can consume one at a time, the research kept as citations, and the frozen/proposed state carried by ADR status rather than by prose buried at line 1,447. Happy to do the mechanical extraction if you'd like — the content is already there, it just needs re-homing.

merklebonsai:
Drafted **PA-6..PA-10** in §7 as **PROPOSED — drafted-not-applied, awaiting V ratification** (same additive amendment lane as PA-2..PA-5; nothing folded into frozen sections). Commit cde6807ca.

- **PA-6 (HIGH 1)** — CAS deref confinement: pointers must match `^(blobs|snapshots)/[0-9a-f]{64}(\.json)?$` AND resolve strictly inside the session dir, on BOTH read paths (`$blob` deref and `snapshot_ref` restore); non-conforming refs render opaque, never deref. Red: **N-JS-traversal**.
- **PA-7 (HIGH 2)** — snapshot redaction honesty: primary variant freezes synchronous deletion of the pre-redaction CAS file inside the redaction act (argued FI-7-compatible — FI-7 forbids *implicit* deletion; redaction is the explicit, consented destruction act, and deleting the bytes is the completion of that consent); fallback variant freezes the loud documented no-op. V picks. Red: **N-JS-redact-cas**.
- **PA-8 (HIGH 3)** — GC × taint lineage, both halves: GC never orphans a live `refs`/`armed_by` target above any surviving descendant (prevention), and an unreadable ref target folds fail-closed `:tainted` (containment). Red: **N-JS-gc-taint**.
- **PA-9 (MED 4)** — `client_msg_id`: red **N-cmd-dedup** (exactly one durable event, journal-silent duplicate, dedup-marked live ack + telemetry so the collision is observable) + a frozen index bound = the readable suffix `[low_watermark, n]` (a compaction boundary was checked and rejected — AD-3b's compaction=resume=one-artifact means compaction does NOT create a fresh session artifact, so the watermark is the only boundary consistent with the rebuilt-by-fold law).
- **PA-10 (MED 5)** — missed-wake law: fire-ONCE-on-resume coalescing (no thundering herd, no silent drop), skipped count surfaced as a grow-only `missed_fires` field on the `woken` payload. Red: **N-sched-coalesce**.

**MED 6** was a doc fix, not a PA: the header now defines the ratification authority ("V" = the project owner, vsevolod.rodionov, rulings recorded inline with dates) and notes AF-* provenance lives in this PR's review threads by design.

The **2 LOWs** (PROPOSED-amendments-as-out-of-band-state; the triple-meaning `F2` hazard) are accepted as inputs to the restructure discussion — no doc change now.

The ADR restructure proposal is escalated to V for a process ruling — it changes how frozen/proposed status is carried and is not mine to decide.

DROOdotFOO (commented):

Scope: docs/contracts freeze (+11689, 41 files). Personas tuned to a frozen-contract PR.
Verified the load-bearing claims against the tree: the "landed code" it freezes on
(`Raxol.Agent.Contract.Event`, `EmitBridge`, `Journal.*`, `Records.Checkpoint`,
`Fingerprint`) all exist and the frozen shapes DO match code (the `Event` defstruct,
`checkpoint` `tip_offset: pos_integer()`, `canonical_json/1`). Good. No secrets leak
(every "API key" hit is research prose citing others' CVEs/issues, link-attributed).
Doc-voice is clean (real em-dashes, only 4 ` -- ` substitutes, no AI meta-narration in
the constitution). The findings are about what the freeze *declares done* while its own
authors flag it isn't.

### HIGH
- **[Saboteur+Security] `harness-freeze-contracts.md:85,440` vs `:1716`** -- `$redacted`
  over-promises a security property it does not deliver. §0 clause 6 and §1.1 present
  `$redacted` as THE mechanism "for secret/PII scrubbing" and "the one legal rewrite of
  history." PA-7 then admits that for **all of v1** the pre-redaction content-addressed
  snapshot file (the secret bytes, on disk) is "never deleted" because GC is deferred
  (OQ-JS2) -- "scrubbing the pointer while keeping the bytes is redaction theater." A
  frozen contract that advertises secret-scrubbing it doesn't perform is worse than none.
  The fix (PA-7 primary=synchronous delete, or at minimum PA-7's fallback warning) is
  drafted but explicitly NOT ratified/frozen. Fix: fold PA-7 into binding §1.1 before the
  freeze, so U9-R/U10-R reds aren't authored against a scrub-that-doesn't-scrub.
- **[Security+Saboteur] `harness-freeze-contracts.md:419-447` (PA-6 at :1717+)** -- the
  frozen reader seam derefs CAS pointers with no runtime path confinement. `$blob` and
  checkpoint `snapshot_ref` restore are "deref-or-render-opaque"; self-containment (§0
  clause 5) is enforced only by a fixture lint scan, never at runtime. An `:import`ed or
  corrupted journal supplying `{"$blob":"blobs/../../../../etc/passwd"}` derefs first,
  reads-then-tolerates. PA-6 drafts the regex+containment fix but it is un-ratified.
  Freezing the reader seam without confinement and then authoring U4-R against it bakes
  the traversal hole into the permanent suite. Fix: promote PA-6 to binding §1.1.
- **[Saboteur] `harness-freeze-contracts.md:3,1475`** -- freeze-vs-status contradiction.
  PR title/body call this the "frozen constitution" that unblocks the red fan-out
  "tonight," but the doc's own Status is "**freeze proposal** -- once ratified" and
  PA-2..PA-10 "**remain PROPOSED -- NOT YET FROZEN**," each "a constitution one-way-door."
  Three of those open items are the HIGH security gaps above (PA-6/7/8). Declaring the
  freeze and authoring permanent reds against §0-§6 now encodes the un-amended (vulnerable)
  shapes while the amendments that would fix them are still open. Either ratify PA-6/7/8
  first or don't call it frozen / don't launch the reds yet.

### MEDIUM
- **[Saboteur] `harness-freeze-contracts.md` §2.1 taint (pt 4) + §1.1-schedule (PA-8 at
  :1759+)** -- GC truncation can launder taint. The lattice folds over `refs`/`armed_by`,
  but the frozen no-orphan law covers only *checkpoints*, and the contract never states how
  an unreadable ref target folds. PA-8 half 2 ("unreadable `refs`/`armed_by` target folds
  fail-closed `:tainted`") is testable **today** (one branch in the fold, independent of
  GC) yet is left un-ratified. The fail-closed default is the entire point of the lattice
  (§2.1 pt 1/3); it should be frozen now, not deferred with GC. Fix: fold PA-8 half 2 into
  binding §2.1.
- **[New Hire] `README.md` (ID-prefix map) + `harness-freeze-contracts.md` §5.2 / OQ-JS4 /
  OQ-JS6** -- dead cross-references to docs not in the repo. The index and §5.2 send readers
  to `f2-action-registry.md` ("the F2 `Raxol.Action` draft"); OQ-JS4/OQ-JS6 cite
  `harness-bus-{protocol,encoding,transport}.md` and the `raxol_agent_client_protocol`
  package as "V-ratified bus specs." None of these exist in the PR tree or on master. A
  newcomer following the navigation hits nothing. Fix: commit stubs, or mark them in the
  README as external / not-yet-committed rather than as if they are siblings in this dir.

### LOW
- **[New Hire+Saboteur] AF-* provenance non-reproducible.** The constitution cites
  AF-1/3/5/6/7/9 as the authority behind frozen contract changes, but README + preamble
  state the 5-model/93-find audit trail "lives in the PR #569 review threads by design,"
  uncommitted. Review threads are ephemeral; a frozen constitution's load-bearing
  provenance shouldn't live only there. Self-acknowledged, hence LOW -- consider a committed
  digest of the AF findings.
- **[Security] `harness-freeze-contracts.md:16`** -- binds ratification authority to
  "V ... (vsevolod.rodionov)". If that is a real person's identity, confirm consent to
  publish it in a public repo (raxol.io); a handle alone would serve the authority binding.
- **[New Hire] `harness-freeze-contracts.md:729` vs `:894`** -- consistency/idiom nit: the
  Elixir signature `attach(session_id, offset, historyPolicy, opts \\ [])` uses camelCase
  `historyPolicy`, while the meta registry writes `history_policy` for the same concept.
  Non-idiomatic Elixir and internally inconsistent.

### Cross-persona overlaps
- PA-6 (`$blob` traversal) and PA-7 (`$redacted` doesn't scrub) are each flagged by both
  Saboteur (the frozen contract is wrong / over-promises) and Security Auditor (exploitable),
  so both were promoted to HIGH. The freeze-status finding (HIGH) is the umbrella: all three
  HIGH security amendments are un-ratified, yet the freeze is being declared.

### Verdict: CONCERNS
It is docs (no runtime), the gaps are self-flagged in §7, and the frozen shapes correctly
match landed code -- so not a BLOCK. But the substantive risk is real: the freeze is
declared and permanent red suites are greenlit against §0-§6 while the doc's own authors
mark three HIGH security amendments (PA-6 path traversal, PA-7 redaction-doesn't-redact,
PA-8 taint-laundering-via-GC) as PROPOSED/NOT-YET-FROZEN one-way-doors. Recommend: ratify
and fold PA-6, PA-7 (primary or fallback), and PA-8-half-2 into binding law, and fix the
dead index links, before calling this frozen or launching the red fan-out.

DROOdotFOO:
## Cross-PR threads (Harness UI series #604-#608 + #586 + #569)

Posting this here on the constitution PR since it's the common root. Two issues span
the stack that no single per-PR review can act on alone -- flagging them together so
merge order and a shared seam can be decided once.

### Thread 1 -- #586 is authored against contracts that only live in this PR

The U4-R red suite (#586) cites `harness-freeze-contracts.md` (§1.1-1.3), the P-JS/FI/AD
codes, and `harness-invariants.md` as the shapes it freezes tests against. None of those
exist in-repo -- they live only here in #569, which is still open and, by its own status
line, a "freeze proposal" with PA-2..PA-10 marked NOT YET FROZEN. So the reds are being
frozen against shapes that are explicitly still movable.

The coupling is tighter than "merge #569 first," because three of the un-ratified
amendments are the HIGH security findings on #569 itself (PA-6 path confinement, PA-7
redaction retention, PA-8). If those get fixed during ratification, the shapes the #586
reds assert against change -- and the reds silently rot.

Suggested resolution:
- Ratify PA-6 / PA-7 / PA-8 into binding law *before* merging #569, then rebase #586 onto
  the ratified shapes.
- Or gate #586 behind #569 explicitly (stack it) and hold both until the constitution is
  actually frozen, not proposed.
- Either way: fix the dead index links first (`f2-action-registry.md`, `harness-bus-*.md`,
  `raxol_agent_client_protocol` -- none resolve in-repo), so a newcomer landing on #586's
  citations can actually reach the spec.

### Thread 2 -- untrusted-input confinement is being solved per-site, three times

The same "confine external bytes at the boundary" gap shows up independently across the
series, each patched (or not) locally:

- **#607** -- fence info string now rendered via `MarkdownRenderer.lang_word/1` with no ESC
  strip; safe only because `MarkdownBody.to_text` happens to sanitize upstream. Other
  callers (playground `demo_helpers.ex`) are not covered.
- **#569 / PA-6** -- `$blob`/`snapshot_ref` CAS deref has no runtime path confinement;
  self-containment is only a fixture lint, so an imported journal with `blobs/../../etc/...`
  derefs first.
- **#586** -- the self-containment red models `:erl_tar.extract` blind-into-cwd with no
  member-path validation, encoding a tar-slip pattern the U4 implementer is invited to copy
  for received cross-host/tenant archives.

These are the same class (ESC smuggling, path traversal, tar-slip = untrusted bytes reaching
a sink without a boundary check) landing at three different sinks. Solving them one PR at a
time means the confinement contract lives in three places and drifts.

Suggested resolution: define one boundary seam in the constitution (a "received/imported
content is untrusted until confined" invariant + a single sanitize/confine helper the reader
seam, markdown renderer, and archive extractor all route through), and have the reds assert
*that* seam rather than each sink locally. This is squarely constitution-level, which is why
it belongs in #569 rather than any one consumer PR.

---
_Cross-PR synthesis from the adversarial review pass over #604-#608, #586, #569._

merklebonsai:
Per V (2026-07-17), acknowledging the two cross-PR threads:

**Thread 1 — merge order:** #586 is now explicitly stacked behind this PR (commented there). Ratify PA-6/7/8 into binding law → freeze #569 → rebase #586 onto the ratified shapes. #586 holds until then.

**Thread 2 — centralize the confinement seam:** agreed, V ruled centralize. Drafting a proposal for a single boundary-confinement module so the CAS `$blob`/`snapshot_ref` deref (PA-6), the tar extract (#586), and the terminal-output ESC strip (#607) stop being solved three times. Note there are actually *two* distinct confinements bundled here — filesystem path-traversal (PA-6 / #586 / and the ACP package's already-tested `fs_sandbox`) vs terminal-injection ESC-strip (#607); the proposal will separate them. One dependency constraint to resolve up front: the `raxol_agent_client_protocol` package is deliberately zero-raxol-dep, so a shared util in `raxol_core` can't be consumed by it — the proposal will weigh a leaf util package vs the ACP package keeping its own small copy. Will post the design for a decide-once ruling.

DROOdotFOO (commented):

Since last review: cde6807c..c09a63bf (many commits; PA-1 ratified, PA-6..PA-10 drafted, boundary-confinement seam proposed as `confinement-seam-proposal.md`).

Scope: re-verified the 3 prior HIGHs with evidence, then ran the new material (PA-6..PA-10 + the thread-2 confinement seam) through Saboteur / New Hire / Security Auditor.

- **New Hire (status legibility) — improved, not fully closed.** There is now an explicit ratification ledger: `harness-freeze-contracts.md:15-19` names V as the sole PROPOSED->frozen gate, and `:1471-1486` cleanly separates **PA-1 RATIFIED/promoted** from **PA-2..PA-10 PROPOSED — NOT YET FROZEN**. The two inline clarifications are marked "clarified per AF review #569 / change no law". A reader who reaches §7 can now tell RATIFIED from PROPOSED. (Residual nit in New Hire section below.)
- **Security Auditor (leakage) — clean.** No secrets, credentials, or internal URLs introduced in the new PA-6..PA-10 text or `confinement-seam-proposal.md`. The `0700`/`0600` perms note (`:458-459`) is a sound addition.

### HIGH

- **[UNRESOLVED — prior HIGH 1, redaction theater] Frozen §1.1 still advertises secret/PII scrubbing that v1 does not perform.** `harness-freeze-contracts.md:440` (frozen §1.1-redaction) still says `$redacted` is "for secret/PII scrubbing" and that the pre-redaction snapshot "**may be deleted** under FI-7 consent" — but GC is deferred entirely (OQ-JS2, `:696`), so for all of v1 the secret bytes are **never** deleted. The honest disclosure lives ONLY in the unratified PA-7 (`:1712-1745`), which itself still offers a **no-op FALLBACK variant** ("V picks the variant"). Net: the *frozen* contract that reds are authored against still sells scrubbing-that-doesn't-scrub, and the fix is neither ratified nor even settled on deletion. The self-contradiction is live in ratified text, ungated by the ledger. Must be reconciled (either land PA-7 deletion, or move the honest warning into frozen §1.1) before red suites cite §1.1.

- **[UNRESOLVED — prior HIGH 2, CAS deref traversal] Path confinement is still non-binding.** PA-6 (`:1674-1710`) is explicitly "Proposed frozen text (draft — contract ruling, needs ratification)" and falls under `:1475` "PA-2..PA-10 remain PROPOSED — NOT YET FROZEN". The thread-2 seam that would close it, `confinement-seam-proposal.md:3-5`, is "**PROPOSED — design-only. Blocked on (1) V's pick... (2) Drew's sign-off... Do NOT implement any site migration until both land.**" Migration row 2 (`:153`) is the PA-6 runtime deref, unimplemented. So the frozen reader tolerance still "derefs first" for `blobs/../../etc/...`; the traversal hole documented last review remains open at both the contract and the seam level.

- **[Saboteur — reds against vaporware; broadens prior HIGH 3] Even RATIFIED PA-1 constrains reader behavior that does not exist in the landed code.** The doc claims it "freezes on top of what landed... changes zero bytes of the Reader's framing logic" (`:11-13`) and cites `Journal.FileStore.Reader` as landed. But `packages/raxol_agent/lib/raxol/agent/journal/file_store/reader.ex` is 196 LOC exposing only `scan/2`, `last_offset/1`, `list_segments/1` — **no tip derivation, no loop-family/conversational whitelist, no `$blob`/`snapshot_ref` deref, no `$redacted` handling** (grep across `packages/raxol_agent/lib` returns only a one-line "minimal loop-family slice" mention in `contract.ex:6`). The RATIFIED PA-1 tip skew-guard, and the §0-§6 CAS/redaction contracts, therefore freeze shapes the implementation has not grown yet. "RATIFIED by V" now certifies reds against an unbuilt reader — the exact umbrella concern from last review, now wearing the ratification seal.

### MEDIUM

- **[Saboteur / Security Auditor — the seam's seed does not exist] The confinement seam's whole option-(b) recommendation rests on a package that is absent from the repo.** `confinement-seam-proposal.md:24-28,183` cites `raxol_agent_client_protocol`'s `FsSandbox` as "a **tested, correct** confinement... the best instance in the repo and the natural contract seed", and `:128,137-140` builds the recommended dep-resolution on "ACP keeps its ~70-LOC `FsSandbox` as a documented intentional duplicate, bound to shared conformance vectors". Evidence: `git ls-tree c09a63bf`, `master`, and `git ls-remote origin` show **no `raxol_agent_client_protocol` package, no `FsSandbox`, and no `feat/agent-client-protocol` branch** — the string exists only inside this proposal doc. The "already-tested seed", the "shared vector fixture ... mirrored verbatim into the ACP package" (`:135-136`), and `test/fixtures/boundary/confine_vectors.exs` are all unverifiable. The seam is a plausible design, but its central "reuse a proven seed / drift-guarded by shared vectors" claim over-promises against a seed that is not in the tree. Cite a real path or mark the seed as not-yet-landed.

- **[New Hire — dead index links persist] Prior dead links unresolved, plus one new.** `f2-action-registry.md` (referenced `README.md:41`, `harness-yolo-safe-research.md:471`, freeze §5.2) and `harness-bus-{protocol,encoding,transport}.md` (freeze `:717-718,752`) still do not exist anywhere in `c09a63bf` or `master`. New: `raxol_agent_client_protocol` (above) is a fourth dangling reference. A newcomer following any of these lands on nothing.

### LOW

- **[New Hire — top-of-doc status still says "freeze proposal"]** `harness-freeze-contracts.md:3` header reads Status: "**freeze proposal**", while `:21` calls this "the final pre-red revision; permanent red suites are authored against these shapes next" and §7 calls PA-1 "frozen law". The single answer to "what is frozen *right now*" (§0-§6 + PA-1) is only assembled 1471 lines in. A one-line RATIFIED / PROPOSED / PARKED table at the top of a 1896-line doc (with 41 siblings) would let a newcomer orient without reading to §7.

### Verdict: CONCERNS

Docs-only PR, nothing executes, and the new ratification ledger honestly quarantines PA-2..PA-10 as un-frozen — so this is not a BLOCK. But **none of the three security-critical fixes landed as binding**: PA-6 (CAS traversal) and PA-7 (redaction retention) are still PROPOSED, the thread-2 seam that would close PA-6 is design-only and explicitly "do not implement", and — worse — the *frozen* §1.1 text still advertises `$redacted` as secret/PII scrubbing while the secret bytes provably persist for all of v1 (the one contradiction that is live regardless of ratification). The thread-2 seam was adopted as a *decision* ("V ruled centralize it") but not as a *binding closure*, and its recommended path leans on an `FsSandbox` seed that is not in the repo. Fix the live §1.1 contradiction, land PA-6/PA-7 (or gate red-authoring on them explicitly), and either produce or de-cite the confinement seed before treating these shapes as frozen.

DROOdotFOO (commented):

Since last review: c09a63bf..f5c15994 (branch has since advanced to 6c6e81b5 `docs: 2nd confinement sweep` — reviewed through HEAD).

Scope this pass: the two new in-flight docs only. `git diff --stat c09a63bf..6c6e81b5` = `confinement-migration-plan.md` (+392) and `injection-risks-2nd-sweep.md` (+169), nothing else. The frozen constitution (§1.1), PA-6, and PA-7 are byte-for-byte untouched in this range.

Migration-plan site accuracy spot-check (read against source in-tree): P1 `actions/fs.ex:145` (lexical `Path.expand` + `starts_with?(cwd <> "/")`, returns `:outside_cwd`) — accurate. P3 `file_drop_handler.ex:~532` (`String.starts_with?(file.path, allowed_dir)`, no trailing-`/`) — accurate, the `/foo/barbaz` prefix bug is real. T2 `tool_result_block.ex:156` (joined `content` straight into `text()`) — accurate. NP1 `skills/store.ex:386` (`skill_dir/2` joins root/category/name with no gate; write+`rm_rf` side, distinct from P4's read side) — accurate. The triage is genuinely well-researched; the (B) trusted-producer exclusions and (C) needs-owner rows are defensible.

- None of the three open HIGH cleared (see below) — but two genuine improvements are worth crediting:
  - The plan correctly gates PA-6 on its own freeze: "PA-6 is proposed-not-frozen — this migration cannot land until PA-6 freezes" (§4.5, and the PA-6 checklist row). It does not migrate onto an unratified amendment. Good.
  - Per-site behavior-CHANGE flags (§Risk, ranks 1-3 + NT17 replay policy note) correctly surface that every path site becomes *stricter* and will reject previously-accepted inputs (macOS `/tmp`->`/private/tmp`, symlinked `node_modules`, `.asdf` shims). That is the right thing to have written down before assigning the work.

### HIGH

- **The plan's foundational premise is a confine primitive that is currently BLOCK'd.** #613 (`Boundary.Path.confine/3`) was reviewed BLOCK for a reproduced CRITICAL: it resolves relative symlink targets against the lexical parent, not the resolved-real parent, so it does not actually confine — and its "never escape" property test validates confine against itself with no planted symlinks, shipping green over the live bypass. This migration plan states the opposite as settled fact: confine is "**implemented**" (header), the seed is "**already correct** (leaf + ancestor symlink resolution, cycle-guarded)" (P5, §2), and cites "Property test already lives in raxol_core (`confine/3` never returns `{:ok, p}` outside resolved root)" (§Risk) as the base guarantee — i.e. it leans on the exact toothless property the #613 review flagged. Migrating P1-P4 + NP1 (5 in-repo path sites) + P5 onto this primitive **spreads the symlink-escape vuln across the monorepo** while every migrated site is advertised as "stricter / now rejects symlink escape." The NEW per-site reds (P1 "symlink inside cwd -> outside must reject", etc.) *might* catch the bug at migration time — but the plan cannot guarantee its reds hit the specific lexical-parent bypass because it does not know the bug exists.

- **Every wave gates on "#613" as a MERGE dependency, not a CORRECTNESS/fix dependency.** §4.2 directs Wave A (P1+P4, P2) "**immediately after #613 merges**"; the checklist "Blocked-by" column reads "#613" throughout. #613 can merge with its CRITICAL unfixed (its own tests are green). A newcomer executing this plan starts path migration the day #613's PR merges. REQUIRED FIX (non-optional, docs-level): change the path-wave gate from "#613" to "**#613 confine symlink-escape CRITICAL fixed + adversarially re-verified against a planted relative-symlink vector**", and strike the "already correct" / "proven seed" assertions in P5/§2. Until then the path waves must be marked HOLD, not "immediately after merges."

- **Safe (TermText) and unsafe (Path) halves share one gate — invert the ordering.** The #613 review cleared `TermText.sanitize/2` and asked to let the #607 TermText migration proceed while PAUSING the path-confinement migrations (#586, PA-6) until the confine CRITICAL is fixed. This plan structurally separates the two classes (P-rows vs T-rows) — good — but then sequences **Wave A = path (P1/P2/P4) BEFORE Wave B = TermText (T1-T17)**, and files both under the same "#613" gate. That is backwards relative to the dependency health: the T-wave rests on a verified-safe primitive and can proceed; the P-wave rests on a BLOCK'd one and must hold. Recommend: promote the TermText wave to run first/independently, and hold the entire path wave (P1-P5, NP1, plus the PA-6/#586 slots) behind the confine fix.

- **Three prior HIGH still open (unchanged this range):** (1) PA-6 CAS path confinement still PROPOSED/unratified; (2) PA-7 redaction retention still PROPOSED/unratified; (3) the redaction-theater contradiction still live in frozen §1.1. This commit is docs-planning layered on top of the same unratified security holes — it neither ratifies nor fixes any of them.

### MEDIUM

- **`injection-risks-2nd-sweep.md` surfaces a new HIGH signing-abuse finding (I1) but only as a triage note.** I1 (wallets sign provider-supplied EIP-712 verbatim; `x402.ex` uses `challenge.currency` as `verifyingContract`; `SpendGate` bounds recorded amount, not signed semantics) and I2 (default-allow `/bin/sh -c` when no `sandbox/0`) are real and correctly scoped as "distinct from the confinement seam." No objection to the doc — flagging that these are net-new HIGH-severity items now sitting in an in-flight findings file with owner lanes but no ratified disposition, i.e. the corpus keeps growing faster than V/Drew ratify. Non-blocking for this docs PR; call-out for the constitution's freeze cadence.

### Verdict: CONCERNS

Docs-only, and everything is gated behind "V + Drew approve and assign / no code lands from this doc directly," so this does not meet the BLOCK bar. But it is a hair from it: the plan asserts a BLOCK'd primitive is "already correct," gates 5+ path sites on #613 *merging* rather than being *fixed*, sequences the unsafe path wave ahead of the safe TermText wave, and — read literally — directs path migration "immediately after #613 merges." The required, minimal, docs-level fix is to (a) strike the "already correct / proven seed / property-test-as-guarantee" language, (b) re-gate the path waves on the confine symlink-escape fix + adversarial re-verification (not merge), and (c) invert Wave A/B so TermText proceeds and the path wave holds. No HIGH cleared; PA-6/PA-7 and the §1.1 redaction-theater remain open.

DROOdotFOO (commented):

Since last review: f5c15994..3827b32c

Range is two commits: the confinement migration plan (f5c15994) and the "2nd
confinement sweep" (3827b32c). Docs-only. Spot-checked the new sweep rows against
`origin/master` source: NP1 (`raxol_agent/skills/store.ex` `skill_dir/2` join at
:62-65 + `do_delete` `File.rm_rf(entry.dir)`), NT16
(`symphony/surfaces/terminal.ex` raw `issue_identifier`/`last_event` into `text()`),
and NT17 (`recording/player.ex` `IO.write(data)` at :90/:154/:215) all reproduce as
described. The sweep's adjudication is careful and correctly classified: NP1 is a
path site (traversal -> `rm_rf` outside root), NT17 is a text site that correctly
refuses a blanket strip-all in favor of a `:replay` profile (a blind `sanitize/2`
would destroy replay's whole purpose). Good, source-grounded work.

- None of the three prior HIGH items resolved (see Security Auditor below) -- this
  range is pure docs-planning, no ratification commit.
- Credit where due (new safeguard, not a resolution of a prior finding): the
  two-contract caution (§B of `agent-lane-response-to-ui-intersection.md`, mirrored
  into the migration plan) correctly prevents strip-all-migrating the
  `content_guard.ex` `sanitize_line/1` SGR-allowlist site. Flattening it to
  `TermText :strip` would have broken the #608 `sanitize == identity`-for-safe-SGR
  seal invariant. That catch actively PREVENTS an unsafe migration and is verified
  against `content_guard.ex` source. NT17's replay-profile carve-out is the same
  kind of correct restraint.

### HIGH

- **The confine primitive is still asserted "already correct" and the whole
  migration is still gated on mere #613 _merge_, not on fixed-and-vector-covered.**
  This commit did not touch the sequencing section: line ~170 still reads the seed
  `confine/3` as "already correct (leaf + ancestor symlink resolution, cycle-guarded,
  reject-before-syscall)", and every Blocked-by cell -- including the six new
  sweep rows -- is still just "#613". Per the #613 review, `Path.confine/3` shipped
  with a reproduced CRITICAL symlink-target-vs-resolved-parent escape that required a
  follow-up fix (715417b1) to close, and the shared vector set still does not cover
  that escaped class. So the migration plan (a) states as settled truth a claim that
  was false at freeze time and remains uncorrected, and (b) binds NP1/P1-P4's test
  strategy to a `path_*_vectors.json` set with a known hole for the exact bug class it
  is supposed to close. Rows should gate on "#613 merged AND confine hardened
  (715417b1) AND shared vectors extended to cover symlink-target-resolution," and each
  path row's NEW red must include the escaped class specifically, not a generic
  "symlink inside cwd -> outside" red.

- **Wave inversion still not done; and this commit adds a new path site to the wrong
  wave under the weak gate.** Wave A (path/`confine` migrations, the ones carrying
  primitive risk) still runs "immediately after #613 merges," ahead of Wave B
  (strip-all `TermText` sites, which carry zero primitive risk). The prior finding to
  run the safe TermText wave first stands unaddressed. NP1 -- the sweep's own
  self-described "highest-severity new site" and a NEW path traversal into `rm_rf` --
  rides the same raxol_agent mix.exs edit as P1/P4, i.e. Wave A, under Blocked-by
  "#613". More path surface, same merge-gate, same inversion.

- **Prior three HIGH remain unratified/live.** PA-6 (CAS path confinement) is still
  "proposed-not-frozen" -- the plan itself says "this migration cannot land until
  PA-6 freezes." PA-7 (redaction retention) did not ratify in this range. The §1.1
  redaction-theater contradiction in the frozen constitution is untouched by these
  two commits and remains live. No HIGH cleared.

### MEDIUM

- **"Injection findings held private": right principle, missing accountability
  binding.** Withholding UNPATCHED exploit specifics (SSRF / command-injection /
  signing-abuse / unsafe-deserialization) from a public branch is correct responsible
  disclosure, and the commit message frames it correctly ("the owner's call, not an
  automated push"). The gap is that there is no durable tracker binding: no GH issue
  number, no security-advisory ID, no path to the "private triage list" -- the
  pointer is literally "ask the agent-lane owner." Worse, three FALSE-POSITIVE
  dismissals in this public sweep lean on that invisible artifact to justify
  themselves: `agent/directive/executor.ex:37` `:cd` ("theater... subsumed by the
  shell finding in the injection doc"), the OSC title writers
  ("OSC-terminator-breakout hardening is in the injection doc"), and the session-id
  row ("a separate deserialization concern is held in the private triage list"). A
  reviewer of this PR cannot verify those dismissals are sound because the referenced
  doc is unreviewable and untracked. Recommendation: land a tracking stub (private GH
  security advisory or an issue number with a redacted body) so the dismissals cite a
  durable, auditable reference even while exploit detail stays private. As written,
  the frozen source-of-truth dismisses live surface by pointing at a ghost.

### LOW

- The `Co-Authored-By: Claude Fable 5` trailer on 3827b32c contradicts the repo
  convention of no co-authored-by lines. Cosmetic, not blocking.

### Verdict: CONCERNS

Docs-only, and by construction nothing lands from this doc (every row still gated on
#613 + explicit V/Drew sign-off), so this does not directly execute an unsafe action
-> not BLOCK. But the range perpetuates the two standing HIGH structural issues
(confine still "already correct" + merge-gated, path wave still ahead of the safe
TermText wave) and compounds them by adding a new high-severity path site (NP1) under
the same weak gate and binding it to a vector set with a known coverage hole for the
class that was the actual CRITICAL. The private-findings posture is the right call in
principle but lacks any durable tracker binding, and public dismissals now rest on an
unreviewable artifact. Fix the gate wording (fixed-and-vector-covered, not merged),
invert the waves, and attach a tracking reference for the held findings before this
freezes.

DROOdotFOO:
**Merge readiness: HOLD.**

Docs-only and CI-green, so it won't break the build, but it is substantively incomplete: the security-critical amendments PA-6 (CAS path confinement) and PA-7 (redaction retention) are still PROPOSED / unratified, and the confinement-migration plan still assumes the seed `confine` is "already correct" -- it needed the #613 fix, and its FsSandbox twin needed #633. Ratify PA-6/PA-7 and re-gate the migration plan on the fixed-and-vector-covered confine before merging as the frozen source of truth.


DROOdotFOO:
still need to compact all the docs and research sprawl, change PR title and summary. distill this into proper ADRs.

PR #570, opened 2026-07-15
feat(harness): U21 evidence-gated done — reds + impl (FI-6)

merklebonsai:
## Status: complete unit

---

Permanent **failing-first** red suite for **U21 "Evidence-gated done" (FI-6)**, authored *before* the implementation exists, against the roadmap disposition (`docs/proposals/in-flight/harness-roadmap.md` §U21 + FI-6) and the frozen Event vocabulary. Part of the red-first fan-out authored against docs PR #569.

## What U21 is

An agent may not declare a turn "done" on its own say-so. The done/turn-success transition (`turn_completed{final: true}`) is **gated on journaled evidence** — tool results / verification outputs (test runs, file checks) — named by `refs` (journal offsets). Each ref must:

1. **exist** — else `{:error, {:missing_ref, offset}}`;
2. **be evidence-class** — a tool result, never the agent's own `:message` self-report, never an internal `:state_change` — else `{:error, {:not_evidence, offset}}`;
3. **postdate the turn's last mutating action** — stale evidence (predating a later mutation) does not count — else `{:error, {:stale_evidence, offset}}`.

A done with no refs is `{:error, :evidence_required}`. Every reject leaves the turn open (no done event journaled).

## Contours (RED — `@moduletag :harness_red`, against the `DoneGate` skeleton)

- **Positive:** valid postdating evidence → `{:ok, turn_completed{final: true, refs: [...]}}`, carrying its evidence refs.
- **Negative:** `evidence_required` (no refs) · `stale_evidence` (evidence predates a later mutation) · `not_evidence` (ref at a `state_change` or `:message` self-report) · `missing_ref`.
- **Fold property:** for every generated journal, the gate agrees with an independent spec oracle (accept iff all refs resolve, are evidence-class, and postdate the last mutation). The generator **includes the stale-evidence ordering**, enforced by a coverage guard so the stale red is never vacuous.

## Dead-injectors (CI-green negative controls, untagged)

Each broken implementation must fail exactly its targeted red:

- **AcceptsWithoutRefs** (skips the ref check) → must fail the `evidence_required` red;
- **ExistenceOnly** (checks existence but not ordering) → must fail the `stale_evidence` red;
- **TextAsEvidence** (lets self-reported text count) → must fail the evidence-class red.

Plus a correct **Reference** satisfying every contour (proves the contours are satisfiable), generator-coverage (**meta-inv 5**: cited stale evidence present, every outcome class exercised), and **fired-counters** (**meta-inv 1**: an armed-but-never-fired injector fails the suite). Generated schedules are seed-reproducible.

## `:harness_red` discipline

- **Enabler commit:** `Raxol.Agent.DoneGate` skeleton — behaviour + verdict typespec + `:not_implemented` `gate/3`.
- **Red commit:** the suite + `:harness_red` added to the `test_helper` exclude list. The reds run **RED** against the skeleton until U21 lands; the negative controls stay **untagged and run in CI**.
- Package suite green (**1024 passed**, 40 excluded); `MIX_ENV=test mix compile --warnings-as-errors` clean; `mix.lock` drift restored.


merklebonsai:
Two policy questions surfaced while authoring — flagging for review rather than deciding silently:

1. **Gating strength**: the roadmap's U21 text reads weaker (`turn_completed` *carries* verification artifacts as data) than what this suite enforces (claim-without-evidence is **rejected**, turn does not transition). The suite implements the stronger gating reading of FI-6. If the weaker reading is intended, the reject contours relax to warnings.
2. **Zero-tool turns**: the suite requires non-empty evidence refs but doesn't pin whether a pure-reasoning turn (no mutations at all) may be 'done'. Currently vacuously satisfiable. Needs a ruling eventually; not load-bearing for the red suite.

Also noted in-suite: the `mutating?` oracle is a structural stand-in until the §5.2 `effect_class` taxonomy lands in the Action draft — rebind then.

DROOdotFOO:
# Adversarial Review — PR #570 (U21-R evidence-gated done)

**Verdict: CONCERNS**

Nice work: this is a genuinely well-structured red-first harness — the reject reasons are split into distinct typed classes, the dead-injector negative controls prove each red is non-vacuous, and the spec oracle is kept independent of the gate. The concerns below are about two gaps in *what the contract actually pins* and one determinism claim that does not hold. Because this is a test-only PR whose reds compile and whose controls run green, none is a hard blocker, but the two HIGH items materially weaken the guarantee the PR advertises ("evidence for a different turn/task id", "seed-reproducible").

Reviewed via three adversarial personas (Saboteur, New Hire, Security Auditor). Findings are cited to `test/raxol/agent/red/u21_evidence_done_red_test.exs` unless noted.

---

## HIGH

### H1 — Determinism claim is false; the anti-vacuity guards are wall-clock seeded (Saboteur + New Hire overlap)
`Gen.sample/1` does `:rand.seed(:exsss, @seed)` then `journal_and_refs_gen() |> Enum.take(n)`, with the comment *"Fixed seed so sampled schedules reproduce (meta-inv 2)"*.

That seed line is dead. `Enum.take/2` on a `StreamData` goes through `Enumerable.reduce -> StreamData.__reduce__/3`, which hard-codes its own seed:

```elixir
# deps/stream_data/lib/stream_data.ex:2620
def __reduce__(data, acc, fun) do
  reduce(data, acc, fun, new_seed(:os.timestamp()), 1, 100)   # <- os.timestamp, NOT the process :rand seed
end
```

StreamData's enumerable path never consults the process `:rand` state you set. So `Gen.sample(n)` reseeds from wall-clock on every run and is **not reproducible**. Concrete consequences:

- The generator-coverage test (meta-inv 5, `"the generator exercises every outcome class, INCLUDING cited stale evidence"`) and `"each dead injector disagrees with the spec on >= 1 generated journal"` are **non-deterministic**. They pass with high probability at n=400/300 because the `evidence_offsets` constant branch reliably manufactures stale orderings, but nothing *guarantees* it on a given run — and when one does flake, it is unreproducible by design, which is the worst failure mode for a "coverage guard."
- This violates the repo's deterministic-tests rule (no clock) and the suite's own "seed-fixed (meta-inv 2)" promise.

Fix: drive the sample through a real StreamData seed — e.g. `StreamData.__reduce__` isn't public API; use `Enum.take(StreamData.map(...), n)` won't help. Instead sample deterministically with an explicit seed via the ExUnitProperties/`StreamData` seeded API (`check all` uses ExUnit's seed) or `StreamData`'s internal seeded reduce, and delete the no-op `:rand.seed` line.

### H2 — Cross-turn evidence is accepted and never pinned (Security Auditor)
`Oracle.classify_ref/3` validates a ref by existence -> evidence-class -> ordering, but **never checks the evidence belongs to the turn being completed**:

```elixir
def classify_ref(journal, last_mut, ref) do
  case Build.resolve(journal, ref) do
    nil -> {:error, {:missing_ref, ref}}
    ev ->
      cond do
        not evidence_class?(ev) -> {:error, {:not_evidence, ref}}
        last_mut != nil and ev.id <= last_mut -> {:error, {:stale_evidence, ref}}
        true -> :ok           # no `ev.turn_id == turn_id` guard
      end
  end
end
```

Only `last_mutation/2` is turn-scoped; evidence class/ordering is journal-global. So a `:tool_result` journaled under **turn A** that merely has a higher offset than **turn B**'s last mutation satisfies done for turn B. That is exactly the "evidence for a different turn/task id" replay/spoof path the review brief calls out, and the spec oracle itself would accept it. No contour, no oracle rule, and no generator case uses differing `turn_id`s (`Build.ev` defaults `turn_id: "t"`, every contour and every generated journal is turn `"t"`), so the bypass is entirely unexercised. An adversarial agent completes the current turn by citing a passing test result recorded in an unrelated turn. Either add the same-turn constraint to the oracle + a `{:not_evidence | :foreign_turn, offset}` contour with cross-turn journals, or explicitly document that cross-turn evidence is in scope and why.

---

## MEDIUM

### M1 — Evidence is validated by class only, never by content (Security Auditor)
`evidence_class?/1` accepts any `item_type: :tool_result` regardless of `result`. `Build.evidence` defaults to `"tests: 12 passed, 0 failed"`, but nothing in the gate, oracle, or contours inspects the payload. A `:tool_result` with `result: ""`, `result: "0 passed, 500 failed"`, or the output of `echo done` is fully valid evidence. The pinned contract is therefore "a tool ran and was journaled," not "a verification passed" — an agent that emits any trivially-satisfiable tool result clears the gate. This may be an intentional U21 scope line (class vs. self-report), but the suite should pin it as an explicit non-goal (a contour with an empty/failing `:tool_result` asserted `:accept`, commented as "content is out of scope for U21") so the boundary is deliberate rather than silent.

### M2 — CI correctness rests on two test-local reimpls that share code (Saboteur + New Hire overlap)
The reds (`@moduletag :harness_red`) are excluded in `test_helper.exs`, and the production `DoneGate.gate/3` is a `:not_implemented` skeleton — so **nothing in normal CI exercises the real contract**. All green CI signal comes from `...ControlsTest`, which checks `Injector.Reference` against `Oracle`. But `Injector.Reference.gate/3` is literally `Oracle.verdict/3` wrapped in an event constructor, so the fold "gate agrees with oracle" against the reference is tautological; the only *independent* CI check of the reference is the six hand-built contours. That is thin, and the discipline depends on a human remembering to drop the `:harness_red` tag when U21 lands (no automated tripwire that the reds ever transition red->green). Consider a CI job that runs `--only harness_red` and asserts it fails today (so the day it starts passing is visible).

---

## LOW

- **L1 (New Hire, ties to H1):** `@seed {0x51, 0x21, 0x2026}` + the `:rand.seed` call + "seed-fixed / reproduce" comments are misleading dead code. Remove them or make the seeding real.
- **L2 (New Hire):** The negative-contour tests are named `"... turn stays open"`, but the gate is stateless and no turn object is asserted to remain open — the test only checks the absence of `{:ok, _}`, and `refute match?({:ok, _}, verdict)` is redundant with the preceding `shape/1 == expected` assert. Rename to reflect "no done event emitted" or add a real turn-state assertion.
- **L3 (Saboteur):** `Build.read_action/2` exists with the comment "must not gate the evidence window," but there is no named contour pinning "a read-only action after the last mutation does not move the evidence window." It is only incidentally covered by the (non-deterministic, see H1) generator. Add a hand-built contour.
- **L4 (Saboteur):** `Contours.shape/1` compares the full reject reason including the offset, so for multi-violation `refs` the real gate is forced to report the *same* offending offset as the oracle — i.e. it must iterate refs in the oracle's exact first-violation-in-refs-order. That is an evaluation-order contract; fine if intended, but it is over-specification that should be documented as part of the contract rather than an accident of `reduce_while`.

---

## Cross-persona overlaps (promoted)
- **H1** flagged independently by Saboteur (non-deterministic coverage guard) and New Hire (misleading seed comment) -> HIGH.
- **M2** flagged by Saboteur (tautological reference/oracle) and New Hire (excluded reds, manual tag flip) -> MEDIUM.

## Does the red suite pin the right thing?
It pins reject-without-refs, stale-ordering, evidence-class (self-report + state-change), and missing-ref well, with non-vacuity controls. It does **not** pin: (a) evidence belongs to the turn (H2), or (b) evidence content is validated (M1) — both of which are adversarial "claims done but supplies attacker-friendly evidence" paths. Combined with the non-deterministic anti-vacuity guards (H1), the suite proves less than its description claims. Address H1 + H2 before U21 lands.

merklebonsai:
Excellent adversarial pass — all findings accepted, fix commit in flight (additive test-only, reds stay red): turn-scoped evidence (H2 cross-turn spoof), real seed + CI tripwire (H1/M2), content-non-goal pin (M1), read-window + naming nits. These are exactly the safety edges a red suite must plant teeth on before the impl is built against the frozen shape; landing them now. Will re-request review when the commit lands.

merklebonsai:
@DROOdotFOO review fixes landed in `d12c8b9ca` — ready for re-review:
- **H2 turn-scoped evidence**: Oracle.classify_ref now checks ev.turn_id == done-turn (existence→class→**same-turn**→ordering); new {:foreign_turn, offset} reject + a cross-turn-spoof contour + SkipsTurnCheck dead-injector; generator grew a :foreign_evidence step so the property + coverage guard exercise it. Closes the cross-turn 'done by citing another turn's passing test' path.
- **H1 real seeding**: found empirically that :rand.seed before StreamData.take was dead (Enumerable reseeds from os timestamp); real fix derives an independent per-index seed via __call__ + seed_s — byte-identical, diverse, cross-process reproducible. The seed theater is now real.
- **M1**: :evidence_content_not_validated contour (failing :tool_result still accepts), documented as an intentional U21 scope boundary (class-ran, not content-passed — a future unit may gate content).
- **M2**: CI tripwire — untagged test asserting 0/N contours pass against the real DoneGate today, so red→green transition at U21-I is visible; noted the Reference==Oracle tautology, real teeth are the hand-authored contours + injectors.
- **L1-L4**: seed folded into H1; negative tests renamed to 'no turn_completed{final:true}' + dropped redundant refute; read-action-doesnt-gate contour; first-violation-order contract documented.
Reds stay red (10 tagged fail), controls green incl. tripwire, full suite 1114 green. (Pre-existing format debt on the file is non-blocking — CI Format Check passes, scope excludes this test path; happy to clean in a follow-up if you want it formatted.)

merklebonsai:
@DROOdotFOO Status: #570 has advanced from red-suite to **complete unit** — the U21-R reds now run GREEN against a real `DoneGate.gate/3` impl (un-tagged from `:harness_red`), plus two adversarial review rounds folded in.

Impl + fixes landed:
- `gate/3`: turn-scoped evidence gate (exists / evidence-class / same-turn H2 / postdates-last-mutation), first-violation-in-refs-order.
- **nil-turn spoof closed** — a done with nil claiming turn is rejected `:unturned_done` (was: `nil==nil` collapsed same-turn + mutation window).
- **mutation-by-type per decision-time-fold law** — derives mutating-ness from `effect_class`, not a producer-stamped bool; the bool survives only as a fail-safe *narrowing* tiebreak.
- **string-key tolerance** — Reader yields `Jason.decode`d string-keyed maps; accessors read either shape (`to_existing_atom`, no atom-leak).
- **session_id** derives from the claiming turn, not `List.first(journal)` (foreign/GC-dropped head).
- oracle-independent boundary control added (literal verdicts, no shared predicate).

1131 passed, 0 failures. Ready for full-unit review.

DROOdotFOO:

Scope: `packages/raxol_agent/lib/raxol/agent/done_gate.ex` (298), the U21 red suite (948), and the `:harness_red` test_helper change. Read in full; findings grounded against the real event producer (`Raxol.Agent.Contract`) and a caller search.

### CRITICAL
- **[Saboteur + Security] `done_gate.ex:130-138` (`mutating?/1` + `effect_bearing?/1`) — the staleness invariant is inoperative on real journals.** `mutating?/1` only recognizes a mutation when the `:tool_use` payload carries a non-nil `effect_class` (or an explicit `mutating: true`). The real producer never stamps either field: `contract.ex:159-165` emits `:tool_use` as `%{item_type: :tool_use, name, arguments, call_id}`, and `grep` across `packages/raxol_agent/lib` finds **zero** occurrences of `effect_class` or `mutating:` outside `done_gate.ex` itself. Consequence: on every journal produced by `Contract.pump`/`EmitBridge`, `last_mutation/2` returns `nil`, so the stale-evidence check at `done_gate.ex:117` (`last_mut != nil and ...`) can never fire. The central U21 guarantee — "evidence must postdate the last mutating action" — silently accepts *arbitrarily stale* evidence in production. The 948-line suite is green only because the `Build.mutation/1` test helper synthesizes `effect_class: :reversible_local`, a field production code never sets. This is precisely the false-negative starvation the review brief warns about: the gate cannot see a mutation, so nothing is ever stale. Failure scenario: agent runs tests (pass) at offset 5, then edits code at offset 9 (a `:tool_use` with no `effect_class`), then cites ref 5 as evidence — gate accepts `done`, though the passing run predates the later edit.

### HIGH
- **[Saboteur + Security] Evidence-class check cannot distinguish a verification result from a mutation's own echo (`done_gate.ex:113-114`, class = `:item_completed`/`:tool_result`).** In the real loop every tool call, including a *mutating* one, emits a `:tool_result` (`contract.ex:169-174`) at a later offset than its `:tool_use`. That result is evidence-class and postdates the (would-be) mutation, so an agent can satisfy `done` by citing the echo of its own `fs_write` — no test run or independent file check ever needs to have happened. The moduledoc claims the gate "pins that a verification tool RAN," but it pins only that *some* tool_result was journaled; `run_tests` and `fs_write` produce class-identical results. "An agent may not declare done on its own say-so" is defeated: it declares done citing the byproduct of its own mutation. Gate can be tricked into marking done without real verification evidence.
- **[Saboteur + Security] `DoneGate` is never invoked — the "security boundary" is unenforced.** A caller search (`grep DoneGate` across `packages/` + `lib/`) finds zero references to `gate/3` outside `done_gate.ex` and its test. The real `turn_completed{final: true}` is still emitted unconditionally by `contract.ex:187-195` on `{:done, ...}`, with no evidence check. The `mutating?/1` comment calls the fold "the security boundary (§0 decision-time-fold law)," yet as shipped it guards nothing — it is an unreferenced pure function with a large suite. A landed gate that no code path calls provides no production protection; the PR does not disclose that wiring is deferred.

### MEDIUM
- **[New Hire] Undocumented cross-reference jargon with no legend.** `done_gate.ex` moduledoc and `mutating?/1` comment, plus the suite and test names, are dense with `H1`/`H2`/`L1`-`L4`/`M1`/`M2`/`§0 decision-time-fold law`/`meta-inv 1`/`meta-inv 5`/`U21-R2 #1-#5`. None resolve to an in-repo key. A new maintainer cannot tell what "H2" or "§0" mean or where they are defined. This also trips the repo's own recorded rule against dev-jargon/AI-meta-narration in comments (feedback_no_dev_jargon_in_comments).
- **[New Hire] Comments narrate deleted code.** The `sample/1` `@seed` comment (test ~159-182) recounts two prior broken implementations (`:rand.seed(:exsss, ...)` "decorative", `seeded/2` "insufficient") that no longer exist, and the M2 tripwire block (test ~602-613) documents a *removed* feature. Reading current behavior means wading through archaeology of code that isn't there.

### LOW
- **[Saboteur] Quadratic scan per gate call.** `resolve/2` (`done_gate.ex:151`) does `Enum.find` over the whole journal for each ref, and `last_mutation/2` re-scans the whole journal; on a long session journal with several refs this is O(refs x journal). Small in practice (per-turn slices), but the accessors also re-walk on every field read.
- **[New Hire] Magic constants unexplained.** `9_999` sentinel done id (test) and `@seed 0x51212026` carry no rationale beyond "a real fixed seed."
- **[Security] Verified safe:** `atomize/1` (`done_gate.ex:...`) correctly uses `String.to_existing_atom` with an `ArgumentError` rescue, so JSON-replayed enum strings cannot exhaust the atom table. No action.

### Cross-Persona Overlaps
- The CRITICAL (staleness dead on real data), and both HIGHs (mutation-echo as evidence; gate unwired) were each independently reached by the Saboteur and the Security Auditor — same-issue overlap, promoted one level accordingly. Together they mean the feature does not actually gate `done` on real journals by any of its three declared mechanisms (ordering, evidence-source, enforcement).

### Verdict: BLOCK
Top finding: the mutation predicate keys on `effect_class`/`mutating` fields that no real producer emits, so `last_mutation` is always `nil` and the stale-evidence gate never fires in production — U21's core invariant is inert against real journals, and the gate is additionally never wired into the done path.

merklebonsai:
## Review round 3 — processed (commit 0fb9bbf42)

### CRITICAL — staleness inoperative on real journals: FIXED
Confirmed: `Contract.pump` emits `:tool_use` as `%{item_type, name, arguments, call_id}` with no effect stamping, so `last_mutation/2` was always nil in production and the stale check never fired. The fix follows the frozen contract rather than patching the symptom: `harness-freeze-contracts.md` §5.2 freezes an all-effect-bearing taxonomy (`:reversible_local | :bounded_sandboxable | :irreversible_external`) and rules effect enforcement *structural, compiled in our own tree — never self-reported*. So `mutating?/1` is now fail-safe: **every completed `:tool_use` is a mutation**; absent `effect_class` = unclassified = mutating; no stamped flag can remove a call from the set. The sole refinement seam out of the mutation set is `classified_effect_free?/1`, constantly false until the F2 `Raxol.Action` draft lands a structural class.

Consequence honestly taken: the suite's own `mutating: false`-trusting Oracle and the L3 "read doesn't gate" contour encoded exactly the `destructiveHint`-is-a-lie class §5.2 forbids. They were **inverted, not weakened**: `:tool_use_after_evidence_gates` now expects `{:stale_evidence, 3}`, and the R2 #5 boundary anchor gained a hand-authored literal for the bare producer shape.

Regressions from the REAL producer as requested: new `U21RealProducerRegressionTest` pumps a run-tests → edit → done stream through `Contract.pump` + `SessionStreamer` and gates the actual journal. It pins (a) the regression premise (producer stamps neither `effect_class` nor `mutating` — fails loudly if the producer grows stamping), (b) your exact scenario: citing the pre-edit test run yields `{:stale_evidence, 3}`, (c) the fs_write echo yields `{:mutation_echo, 5}`, (d) full fail-closed: no offset in a real v0 journal is acceptable evidence. **9 tests fail against the pre-fix gate; 29/29 green post-fix.**

### HIGH — mutation's own echo as evidence: FIXED (new check 5)
Real, and under the fail-safe default it was the whole remaining hole (the only result that can postdate the last mutation is its own echo). New check, after ordering: a `:tool_result` paired to the turn's **last** mutating call is rejected as `{:mutation_echo, ref}` (additive verdict growth). Pairing is by tool name + nearest-preceding order within the turn — the only pairing signal in the frozen shape (the producer stamps `name` on both sides, no `call_id` on results). Oracle mirrors it independently; a hand-authored `:mutation_echo` contour + the real-producer regression cover it.

**Flag for a contract ruling:** combined with the fail-safe default, the gate is now *fully fail-closed on today's real journals* — every call gates, every result is some call's echo, nothing passes. That is the intended §0-clause-7 posture while unwired, but it means U21-I needs a structurally-classified verification class (post-U8/F2: is a shell-run test result a mutation echo or verification?) before wiring can make any done pass. Filed alongside the parked gating-strength ruling.

### HIGH — gate unwired: DEFERRED-WITH-RATIONALE, now disclosed
Partially documented already: the roadmap's impl-phase section orders `U21-I` after its `needs U8` edge, and `harness-parked.md` tracks both the U21 gating-strength ruling and the wiring-debt class. What was missing was disclosure in the artifact itself — the moduledoc now has a "Wiring status" section stating plainly that `gate/3` is not invoked by any production path, that `Contract.pump` still emits `turn_completed{final: true}` unconditionally, and what wiring waits on (U8 structural classes + the refs-citation seam). Not wired here on purpose: wiring without the refs seam would either be a no-op or break the frozen v0 producer behavior.

### MEDIUM — jargon legend: FIXED
Label key added to both the moduledoc and the suite header resolving H*/M*/L*, `U21-R2 #N`, `§0`/`§5.2`, `meta-inv N`, and FI-6 to their defining documents.

### MEDIUM — archaeology comments: FIXED
`@seed` comment rewritten present-tense (the determinism contract and why per-index seeding is the only shape satisfying it, without the two dead implementations); the removed-M2-tripwire tombstone block deleted (the header keeps a one-line note); the stale "skeleton returns :not_implemented" narration in the `gate/apply` helper updated.

### LOW — quadratic scan: FIXED
`gate/3` builds an offset index once; per-ref resolution is O(1). `last_mutation` was already a single pass; the echo check scans only for refs that survive the first four checks.

### LOW — magic constants: FIXED
`9_999` documented as an out-of-range sentinel (max journal is 8 records); `@seed` documented as arbitrary-but-fixed.

### LOW — `atomize/1` safe: acknowledged, no action. (Tool names are deliberately read raw, never atomized, in the new pairing path.)

Full raxol_agent suite: 1138 passed, 0 failures. `--warnings-as-errors` clean. mix.lock untouched.

DROOdotFOO:
Use the [web editor](https://github.com/DROOdotFOO/raxol/pull/570/conflicts) or the command line to resolve conflicts before continuing.

PR #571, opened 2026-07-15
feat(harness): U7 SpendGate reserve-before-call — reds + impl (AD-6a)

merklebonsai:
## Status: complete unit

---

## What U7 is

Every spend-bearing call (an LLM provider call, a paid tool) journals **`reserve → call → settle`, in that order, per call**, correlated by an opaque `cost_ref`:

- **No reserve ⇒ no call, EVER** (fail-closed). A refused reserve means the call does not happen and a typed `reserve_refused` record states why.
- **Settle records actuals.** The refund of `estimate − actual` is internal, but the `settle` record is the authoritative post-hoc fact — the refund is *derivable* from the `(reserve, settle)` pair.
- **The journal fold IS the accounting.** No side ledger of truth for the harness loop; a crash between reserve and call leaves the reserve dangling but **visible** in the fold, never silently lost.
- **Concurrent calls under one budget never over-reserve past the cap** — atomic `try_spend` semantics.

U7 is the same reserve-before-call law the payments stack enforces at the wallet boundary (`Ledger.try_spend` + `Payments.Actions.SpendGate`), generalized to the primary-loop tool/provider boundary. `raxol_agent` does **not** depend on `raxol_payments` (dependency runs the other way), so U7 replicates the atomic `try_spend` *shape* rather than importing it.

## Contours (positive, `@moduletag :harness_red`)

Drive the real `Raxol.Agent.SpendGate` (a `:not_implemented` skeleton), so they are **red until U7 lands** — the point of a red-first suite. Each asserts the *good* outcome happened, so a not-implemented gate fails them non-vacuously.

1. reserve → call → settle, in order, per call (fold asserts; `cost_ref` correlation).
2. a refused reserve makes **zero** provider-stub calls and records a typed refusal.
3. settle records the actual; refund derivable from the `(reserve, settle)` pair.
4. crash between reserve and call → the dangling reserve is **visible** in the fold.
5. concurrent calls under one budget never over-reserve past the cap (seed-reproducible schedule; seed dumped on failure).

## Dead-injectors (negative controls, run in CI, green)

meta-invariant m4, one mutation each; each must be flagged by its matching checker (proving the checkers are not green-on-broken):

- **(a) settle-only** injector (post-hoc accounting, no reserve) → fails the ORDER red.
- **(b) provider call on a refused reserve** → fails the FAIL-CLOSED red (the stub call-counter catches it).
- **(c) reserve against a stale/second counter (dual-truth)** → fails the OVER-RESERVE red.
- **(+) crash-loses-reserve** injector → fails the DANGLING-RESERVE-VISIBLE red.

Plus a well-formed-trace pass (checkers not vacuous) and an m1 fired-counter self-test (a dead site dumps the schedule, m2).

## `:harness_red` discipline

- Enabler commit first (`SpendGate` skeleton, `@callback`s, `:not_implemented`), then the red commit.
- `@moduletag :harness_red` on the positive contours; excluded in `test_helper.exs`, so **CI stays green** while they are red. Run them with `--only harness_red`.
- The negative controls carry **no tag** and run in CI.

## Validation

- `MIX_ENV=test mix compile --warnings-as-errors` — clean.
- Full `raxol_agent` suite: **1026 passed, 38 excluded** (the 5 positive contours among the excluded).
- `--only harness_red`: **0/5 passed** (all red, each on the not-implemented gate).
- `mix.lock` drift restored.


merklebonsai:
Authoring flags for review:

1. **Wire encoding deliberately unfrozen**: the freeze doc froze JS-FREEZE/U11/U12 but not U7's concrete spend-record encoding. This suite freezes the *fold semantics* (reserve→call→settle order, cost_ref correlation, fail-closed, dangling-reserve visibility) over an injected sink — binding to the real journal encoding (likely family:meta records + turn_completed rollup per freeze §2.1) is U7-I work and will get a small freeze addendum then.
2. **Crash contour is modeled, not real-kill**: asserts the durability property via granular reserve + no settle. The real SIGKILL + reopen version belongs in the durability-spine invariant suite once U7 emit is wired to Journal.FileStore (I10 discipline — no virtual-clock cheats for fsync proofs).
3. **Budget atomicity validated as observable** (exactly cap/est reservations fit under concurrency, seed-reproducible) against a local try_spend-shaped budget — raxol_agent must not depend on raxol_payments, so the real budget module arrives with U7-I.

DROOdotFOO:
## Adversarial review — U7-R SpendGate reserve-before-call red suite

Strong red-first work: the negative-control design (one-mutation dead injectors, each with a matching checker, plus the m1 fired-counter self-test and a non-vacuous well-formed-trace pass) is genuinely rigorous and rare to see. The `@moduletag :harness_red` / untagged-controls split keeps CI green while the contract stays red, exactly as intended. My concern is not the scaffolding — it's that the frozen contract does **not** actually pin the headline invariant it names, and leaves several money-leak paths untested. Since this suite *is* the permanent contract that a future U7 green implementation will be graded against, those holes matter as much as shipping logic would.

Verdict: **CONCERNS** (strong). Findings #1–#3 should be closed before any U7 implementation is allowed to turn this green, because as frozen the suite can go green on a gate that bypasses reserve-before-call.

Severity counts: CRITICAL 1, HIGH 3, MEDIUM 4, LOW 2.

---

### CRITICAL

**[Saboteur + Security Auditor] `red_spend_gate_probe.ex:139` (`reserve_before_call`) — the order checker pins emitted-record order, not real temporal ordering; a call-before-reserve gate passes green.**

`reserve_before_call/1` only inspects the *kinds* of emitted records per `cost_ref` and checks the sequence is a prefix of `[:reserve, :call, :settle]`. It never ties the real provider invocation to the reserve. The provider stub (`provider_call/2`) just bumps a `:counters` ref; the temporal position of the actual call relative to the reserve is lost. So a gate that invokes the provider FIRST and *then* journals `reserve → call → settle` in tidy order passes both `reserve_before_call == :ok` and `provider_calls == 1`. I verified this against a faithful reimplementation of the checker:

```
kinds_for c1: [:reserve, :call, :settle]
reserve_before_call: :ok
provider_calls: 1
=> call-before-reserve gate is GREEN on the order+provider checks? true
```

The whole point of U7 is "reserve BEFORE the call, fail-closed." The fail-closed contour (test 2) *is* robust because a refused reserve forces the counter to stay 0. But the success-path ordering — the more common case — is defeatable while green. To actually pin it, the provider stub must record its invocation into the same journal (with a `seq`) so the fold sees a `:call`/provider-event that must come *after* a successful `:reserve`; today the checker can't distinguish "reserved then called" from "called then wrote reserve."

---

### HIGH

**[New Hire + Saboteur] `red_spend_gate_probe.ex:113` — refund/release to the budget is never asserted; `budget_reserved/0` is dead code.**

The suite claims the refund of `estimate - actual` is "derivable from the `(reserve, settle)` pair." Test 3 asserts the reserve/settle *records* exist with `estimate: 100` / `actual: 70` but never asserts the budget's reserved amount actually dropped. `budget_reserved/0` is defined and used **nowhere** (`grep` confirms zero call sites). Consequences a green gate could exhibit undetected: (a) reserve + journal + settle-record but never refund the 30 to `context.budget` → every call permanently shrinks the cap (a slow leak); (b) settle never releases the held reservation. Neither is caught. For a gate that guards funds, "the reserved balance returns to the correct value after settle" is the core no-leak property and it is unpinned. Add an assertion on `P.budget_reserved(budget)` after settle.

**[Saboteur + Security Auditor] `u7_spend_gate_red_test.exs` — no invalid-amount coverage; `:invalid_amount` refusal is advertised but never pinned.**

`spend_gate.ex:69` lists `:invalid_amount` in the `refusal` type and the moduledoc calls out fail-closed refusals, but no test drives `reserve(ctx, "c", 0)`, a negative estimate, or a non-finite/float estimate. The payments `Ledger` this suite says it mirrors explicitly rejects non-positive/non-finite amounts (per `SpendGate`/`Ledger` in raxol_payments). As frozen, a U7 gate that happily reserves 0 or a negative estimate is green. Add a contour: non-positive/non-finite estimate ⇒ `{:error, {:reserve_refused, :invalid_amount}}` and **zero** provider calls.

**[Saboteur] `u7_spend_gate_red_test.exs:109-149` — the real gate has no frozen protocol to reserve against `context.budget`, so the atomicity contour can't actually drive a real gate.**

The positive contours pass `budget: <pid>` into `context`, but `P.try_reserve/2` (the atomic get-and-update seam) is called **only inside the dead injectors** — never on any path the real `SpendGate.around/reserve` would take (`grep` confirms `try_reserve` appears only in the injectors). Production `Raxol.Agent.SpendGate` cannot call a test-support function, and nothing in the frozen `context` shape defines *how* the real gate atomically reserves against `context.budget`. So when U7 lands, the concurrency test only exercises whatever budget protocol the implementer happens to invent, and today it pins the probe's atomicity, not the gate's. The seam needs to freeze a reserve callable (e.g. `context.try_reserve.(amount)`) that both the real gate and the injectors go through, or the atomicity guarantee is illusory.

---

### MEDIUM

**[Saboteur + New Hire] `red_spend_gate_probe.ex:181` (`reserved_within_cap`) — sums cumulative reserves, not peak concurrent reserved.**

The checker sums *every* `:reserve` estimate ever emitted and asserts `<= cap`. That is strictly stronger than "peak concurrent reserved ≤ cap." A correct gate that refunds on `actual < estimate` and re-lends the freed budget to a later call would push the cumulative sum past `cap` while never exceeding it at any instant — and this checker would falsely flag it. It only passes here because test 5 forces `actual == estimate` (comment at line 137 acknowledges this). As a frozen correctness contract, it can reject a legitimate refund-and-reuse implementation. Track peak reserved (reserve `+est`, settle `-est`) instead.

**[Security Auditor + Saboteur] `u7_spend_gate_red_test.exs` — no double-reserve / double-settle / replay coverage.**

Nothing pins that a repeated `cost_ref` reserve is rejected, or that settling the same reservation twice double-refunds the budget (money inflation), or that a reservation token can't be replayed. `@canonical_orders` would reject `[:reserve, :reserve, ...]` via `bad_order`, but no test drives it, so the behavior is unspecified for the green implementation.

**[Saboteur] `u7_spend_gate_red_test.exs:96-107` — the "crash" contour never crashes.**

The test calls `SpendGate.reserve/3` and omits settle; the comment says "the process dies before the call/settle" but no process is killed and no recovery is exercised. It's a reasonable journal-before-return proxy (reserve alone must be durable), but the label oversells it and it does not cover the real risk it names: a reservation stranded by a mid-call crash reducing future budget. Pair it with a real `spawn`/kill (or an assertion that the stranded reserve is later reclaimable).

**[Saboteur] `u7_spend_gate_red_test.exs:43` — `@seed` is captured at compile time; `U7_SEED=... mix test` has no effect on an already-compiled module.**

`@seed String.to_integer(System.get_env("U7_SEED", ...))` evaluates the env var when the module compiles, not per run. The m2 "seed overridable / dumped for reproduction" claim only holds on a clean recompile, which is a reproducibility footgun. Read the seed at runtime (e.g. in `setup` or inside the test body).

---

### LOW

**[Saboteur] `u7_spend_gate_red_test.exs:126` — `Process.sleep(:rand.uniform(3))` jitter adds nothing to the guarantee.** With a serialized atomic budget the successful count is invariant regardless of interleaving, so the sleep only varies which tasks win (which the test never asserts). Mild sleep-in-test smell; harmless but noise.

**[New Hire] `u7_spend_gate_red_test.exs:55,87` — magic amounts (`80` vs `70`) are only lightly motivated.** The `100`/`cap 300` amounts are well-explained; the differing actuals read as arbitrary. A one-word note ("under-estimate, so a refund is due") would help.

---

### What's genuinely good
- The dead-injector matrix (settle-only / call-on-refused / dual-truth / crash-loses-reserve) with a matching checker each, run untagged in CI, is a real proof the checkers aren't green-on-broken.
- The m1 fired-counter + `assert_all_fired!` self-test (lines 254-266), including the newline-safe `refute` on the message, is careful.
- The well-formed-trace pass (lines 239-252) correctly guards against vacuous controls.

The scaffolding is excellent; the gap is that the *contract's teeth* don't yet bite on the success-path ordering, the refund/no-leak path, or invalid amounts — the three things a money gate most needs pinned.

merklebonsai:
Excellent adversarial pass — all findings accepted, fix commit in flight (additive test-only, reds stay red): frozen reserve seam in context (HIGH), peak-not-cumulative reserved (M), double-reserve/settle/replay coverage (M), real-kill crash + runtime seed (M). These are exactly the safety edges a red suite must plant teeth on before the impl is built against the frozen shape; landing them now. Will re-request review when the commit lands.

merklebonsai:
@DROOdotFOO review fixes landed in `62e6651a2` — ready for re-review:
- **HIGH reserve seam**: froze `context.try_reserve.(estimate) :: {:ok, remaining} | {:error, :over_limit}` in the enabler's context spec + moduledoc (real gate MUST invoke it); concurrency contour + CallOnRefused injector now route through it, so atomicity binds the GATE not the probe.
- **peak vs cumulative**: reserved_within_cap rewritten to peak-concurrent (reserve +est, settle −(est−actual)); added a regression proving a legit refund-and-reuse trace now passes (was falsely flagged).
- **double-reserve/double-settle/replay**: DuplicateReserveInjector + DoubleSettleInjector + a direct test that reserve_before_call flags [:reserve,:reserve].
- **real crash**: spawn_monitor + Process.exit(:kill), stranded-reserve visible.
- **runtime seed**: @seed → runtime seed/0.
Reds stay red (7/7 fail on :not_implemented), 10 controls green, full suite 1030 green.

merklebonsai:
CI note (for merge triage): the red checks here are **stale-base cross-lane failures**, not this unit's work. Failing tests are all UI-lane / flaky: `test/harness/tp_pty_test.exs` (the exact test #588 is tagging `:skip_on_ci`), `test/harness/tf_fixture_test.exs` (UI-lane fixture toolchain, TF #551), and `HeadlessGoldenTest` (rendering golden). All green on `master`; this branch was cut before #588 + the UI stabilizations landed. The `raxol_agent` unit suite is green locally (impl + review-fixes). Clearing CI needs a rebase onto current master — deferring that since the stacked impl branches (u8/u12/u10) are being built on this base right now. No raxol_agent regression.

DROOdotFOO:

Scope: U7 SpendGate reserve-before-call (`spend_gate.ex`, `spend_gate/reservations.ex`, red suite, robustness suite, probe). Read in full. Three personas, each grounded below.

### HIGH
- [Saboteur + Security] `spend_gate.ex:242,295` (via `guarded_emit/2`, 358-374) — **A money-gate that silently drops audit records while keeping the irreversible charge.** `try_reserve` charges the budget (durable `Ledger.try_spend`-shaped ledger in prod), then `guarded_emit(%{kind: :reserve})` writes the journal record. In prod `emit` is durable I/O that *can* throw; `guarded_emit` swallows the raise into a `Logger.warning` and returns `:ok`. Result: the budget is charged but **no `:reserve` record exists in the fold**. The same happens on `:settle` after the CAS flip (295) — actual recorded nowhere. This directly contradicts the module's own stated invariant ("The journal fold IS the accounting", "never silently lost", moduledoc 43-46) and the inline comment at `spend_gate.ex:234` ("A dropped reserve record is loud (logged) and visible in the fold") is **factually wrong** — if the emit is what failed, the record is *not* in the fold. Failure scenario: journal write blips (disk full / DB timeout) on a charged reserve, the reserver process then dies before settle -> a charge that is invisible to any reconciliation of ledger-vs-journal, i.e. the exact "silently lost" case the gate promises to prevent. Chosen tradeoff (return usable handle vs. strand the charge) is defensible, but the "never silently lost" guarantee and the 234 comment are not honored. *(Promoted MEDIUM->HIGH: Saboteur resource/accounting + Security audit-integrity, same root.)*

### MEDIUM
- [Saboteur] `spend_gate.ex:336-346` (`safe_call`) + `around/4` 314-330 — **Brutal `:kill` leaks the ETS claim and strands the budget; the cleanup path is never wired.** `safe_call` releases the claim via `rescue`/`catch`, but `Process.exit(pid, :kill)` is untrappable — no `rescue`/`catch`/`after` runs. An agent primary loop that runs `around/4` under a per-call timeout or a supervisor `:brutal_kill` shutdown (both routine) leaves the `{scope, cost_ref}` claim in the singleton public ETS table forever *and* the budget charged with no settle. `Reservations.sweep_scope/1` (reservations.ex:73-79) exists for exactly this, but `grep` confirms **nothing in `lib/` ever calls it** and there is no supervisor child / scope-end hook. Steady-state: unbounded ETS growth + monotonic budget starvation across a long-lived session. The red crash-test (`u7_..._red_test.exs:115`) even performs a real kill but only asserts the journal shows `[:reserve]` — it never asserts claim cleanup, so the leak is untested.
- [New Hire] `reservations.ex:11-15` vs `spend_gate.ex:394-401` — **Stale doc on the exact dedup key.** `Reservations` moduledoc states "The `scope` is the frozen `context.try_reserve` closure itself ... the one field guaranteed present and unique per budget scope." But `reserve_scope/1` actually derives a *tagged tuple* preferring `:budget_id -> :scope -> :budget -> {:try_reserve, closure}`, and `spend_gate.ex` comments (376-393) explicitly warn the closure-term key is fragile. A maintainer reading `Reservations` (the module that owns the table) is told the wrong thing about how claims are keyed.

### LOW
- [Security] `reservations.ex:91-97` — `:public` ETS dedup table is VM-globally writable. Any co-resident process (a plugin, another agent) can `:ets.delete/2` a live claim (permitting a duplicate reserve for one `cost_ref` -> a second `try_reserve` charge, bounded by cap) or `:ets.insert/2` a fake claim to wedge a `cost_ref` (DoS). No access control on money-adjacent dedup state. `:protected` is incompatible with the direct-to-ETS `claim`/`release` design, so this is a noted attack-surface tradeoff, not a trivial fix.
- [New Hire] `spend_gate.ex:239` — the `:atomics` `settle_guard` is embedded in the reservation map, which is otherwise a plain serializable term. A reservation handed across nodes (swarm) or persisted/replayed carries a dead atomics ref -> double-settle protection silently voids off-node. Also `reservations.ex:129-141`: magic 100-retry / `Process.sleep(1)` start spin with no named constant.
- [Saboteur] `spend_gate.ex:16-21,286-296` — `settle/3` records `actual` but never refunds the budget (deferred), while `reserve` charges at `estimate`; the budget therefore only ever decreases, on estimates. A long session refuses calls once cumulative *estimates* hit cap even when *actuals* ran well under. Documented deferral and the safe (never-overspend) direction, but the `reserved_within_cap` checker's "refund-and-reuse" model (probe 195-233) describes behavior the real budget does not yet exhibit.

### Cross-Persona Overlaps
- Dropped-journal-record (Saboteur accounting leak + Security audit-integrity) converge on the same `guarded_emit` behavior -> promoted to HIGH.
- Claim/budget lifecycle leaks appear from two angles (Saboteur `:kill` bypass, Security public-table tampering) — both exploit that claim removal is not guaranteed; kept at their individual severities since the mechanisms differ.

### Verdict: CONCERNS
One HIGH (audit-record loss on a money gate) plus two MEDIUM. Core reserve-before-call atomicity is sound: `insert_new` dedup and the serialized `try_reserve` are each atomic and orthogonal, `:atomics.compare_exchange` correctly makes settle-once race-safe, and `valid_amount?` rejects non-positive/non-integer estimates per convention. The concerns are lifecycle/audit-integrity edges around crashes and journal-write failures, not the happy-path law.

merklebonsai:
Processed the adversarial review (all findings triaged) — pushed as 461822b47.

### Fixed

**HIGH — `guarded_emit` swallows real journal failures (Saboteur + Security).** Confirmed real: the moduledoc's "never silently lost" was not honored and the `:234` comment was factually wrong. Fixed via the *propagate-a-distinguishable-error* arm of your fix menu — **emit-before-charge was ruled out** because the frozen red suite pins the refusal contour to `[:reserve_refused]` only (`u7_spend_gate_red_test.exs:95`); write-ahead would journal a `:reserve` on every refused reserve, changing the frozen fold shape.
- `guarded_emit` now has exactly the U12 `safe_emit` narrowing: only an ephemeral `:noproc` exit (dead in-memory test bus) is tolerated; any real failure raises `Raxol.Agent.SpendGate.JournalWriteError` carrying the exact record (`kind`, `cost_ref`, `estimate`/`actual`) so the budget owner — who composes `try_reserve` and is the only party that CAN roll back the charge — reconciles instead of proceeding on silently diverged accounting.
- `:reserve`/`:call` writes (claim still held) release the claim before propagating (`emit_or_release/4`): `cost_ref` stays retryable, same contour as a raising `call_fun`.
- `settle`: CAS stays flipped + claim released (settle IS complete, replay still rejected), but the missing record now raises instead of returning a clean `:ok`.
- Regressions (verified failing pre-fix, all three): real emit failure on reserve/settle/call raises — never `{:ok, _}`/`:ok` — with a payload sufficient to reconcile (the reserve test refunds via `err.record.estimate` and asserts the budget is whole); the `:noproc` boundary is pinned on both reserve and settle.

**MEDIUM — stale `Reservations` moduledoc on the dedup key (New Hire).** Fixed: it now describes the tagged budget-identity tuple derivation (`:budget_id -> :scope -> :budget -> {:try_reserve, closure}` with the closure as documented last-resort fallback).

**MEDIUM — `:kill` leaks the claim / `sweep_scope` never wired (Saboteur).** Split disposition. The untested-leak half is fixed: a new regression performs a real `Process.exit(_, :kill)` and pins BOTH halves — the claim leaks (retry refused `:duplicate_reserve`) and `sweep_scope/1` reclaims exactly it. The wiring half is **deferred to U7-I, verified**: nothing in `lib/` drives the gate yet, so no run/session scope *exists* to end — there is no teardown site to hook. `sweep_scope`'s doc now carries an explicit NOT-YET-WIRED contract: U7-I MUST call it from scope teardown.

**LOW — magic 100/1ms spin.** Fixed: named `@table_await_retries` / `@table_await_interval_ms`.

### Deferred (verified + documented)

**LOW — `:public` ETS tamper surface (Security).** Accepted tradeoff, as you noted: `:protected` would serialize every reserve through the owner process. Now documented in the `Reservations` moduledoc as a bounded, money-adjacent surface (duplicate bounded by cap; fake claim wedges until swept); revisit at U7-I if untrusted in-VM code becomes a deployment shape.

**LOW — `:atomics` guard dead cross-node (New Hire).** Deferred with doc: the `reservation` typedoc now states the handle is node-local and non-serializable — settle where you reserved; cross-node settle is out of U7 scope.

**LOW — budget only ever decreases on estimates (Saboteur).** Already a documented deferral (moduledoc "Budget-side release on settle", U7-I); the probe's peak/refund-and-reuse checker models the FOLD semantics, which is what the red suite freezes — the real budget refund lands with the U7-I `Ledger.try_spend` wiring. No change beyond the existing doc.

Verification: `mix compile --warnings-as-errors` clean; U7 red suite + robustness suite 28/28; full raxol_agent suite 1048 passed; the three HIGH regressions confirmed red against the pre-fix gate; mix.lock untouched.

DROOdotFOO:
Use the [web editor](https://github.com/DROOdotFOO/raxol/pull/571/conflicts) or the command line to resolve conflicts before continuing.

PR #572, opened 2026-07-15
feat(harness): U5 interrupt = staged supervised kill — reds + impl (AD-12)

merklebonsai:
## Status: complete unit

---

## What U5 is

U5 "Interrupt = staged supervised kill" (AD-12) makes interrupt a bounded escalation, each stage a **durable event** (the staging is observable in the journal):

```
cooperative signal  →  bounded wait  →  OS process-group SIGKILL
```

emitted as `interrupt_signaled → interrupt_waited → interrupt_killed → turn_canceled`. Two laws the reds pin, both straight from the kill-dynamics spike (`spike-u5-kill.md`):

- **Effectiveness.** Killing the BEAM process that owns the Port buys nothing against a hostile tool — only the OS **process-group** SIGKILL (`kill -9 -<os_pid>`) works, confirmed dead out-of-band. **Never trust `:exit_status`** (a surviving grandchild forges "port closed").
- **Post-kill quiescence (no-zombie-emission law).** After the `interrupt_killed` (kill-complete) event, NO `item_delta`/`item_completed`/tool-result event for that turn may ever appear.

This PR is **red-first**: permanent failing tests authored BEFORE the implementation, so U5-I is built to satisfy them, not the reverse.

## Structure — two commits

1. **Enabler** — `Raxol.Agent.Interrupt`: the behaviour + staged-kill vocabulary with **no implementation** (`interrupt/3` raises `:not_implemented`). The compile anchor the reds bind to; U5-I fills in the body to flip them green.
2. **Red suite** — the contours + dead-injector controls + `test/support` scaffolding.

## Positive contours (`u5_interrupt_red_test.exs`, red by design)

Every test drives the unimplemented `Raxol.Agent.Interrupt` and folds the journal via an **independent raw reader** (oracle independence):

- **P1** staged sequence `signal → wait → kill → turn_canceled` journaled in order, one `turn_id`
- **P2** turn terminates `:turn_canceled` with a reason
- **P3 / P5** kill effective mid-shell-Port: a rogue tool (`trap '' TERM INT; sleep &; wait`) is group-killed, OS-confirmed dead, **zero orphaned grandchild** (spike ground truth via `ps`, not `:exit_status`)
- **P3b** mid-provider-stream interrupt (no tool Port): cancels with no trailing output
- **P4** post-kill quiescence: no output event after kill-complete

## Negative contours + dead-injectors (`u5_interrupt_controls_test.exs`, run in CI)

Each contour gets one mutant — **green on the correct `Reference`, red on the mutant** (the negative-control pattern from the invariant suite):

| dead injector | breaks | contour that must fail |
| --- | --- | --- |
| `:skip_wait` | emits `signal → kill` (no wait) | staging |
| `:trust_exit_status` | top-pid kill + trusts `:exit_status`, orphans the grandchild | effectiveness |
| `:late_result` | lets a `tool_result` through after kill-complete | quiescence |

Every fault site carries a fired-counter; a dead injector (armed-but-never-fired) fails loudly and dumps a reproducible **seed/schedule** (meta-invariants 1 & 2). The effectiveness control is `:unix_only` (real OS processes); the pure-journal controls are deterministic.

## CI discipline (`:harness_red`)

The positive suite is `@moduletag :harness_red` and added to the `test_helper` exclusion, so **CI stays green now** — 1025 passed, 0 failed, reds excluded. The controls are NOT tagged `:harness_red` and DO run (they test the tests). The U5-I implementation PR deletes the tag / implements `interrupt/3` and the reds flip on.

- Full `raxol_agent` suite — 1025 passed, 0 failed (6 reds excluded, 5 controls run green)
- Forced `--only harness_red` — 0/6 (all raise `:not_implemented`), zero leaked OS processes
- `mix.lock` drift restored

## Contours from the freeze doc I could NOT express as tests (flagged, not faked)

- **Exact grace-window latency / the `~2–4 ms` SIGKILL-to-gone number** is a policy knob, not a contract (the spike found the nice/rogue regime gap is orders of magnitude). The reds assert the wait **stage** happens, never a specific millisecond value.
- **"Kill mid-provider-stream" at the OS level** — a provider SSE stream is not an OS Port in this codebase, so there is no process group to kill. P3b expresses it at the observable-contract level (turn_canceled + no trailing output), not as an OS kill.
- **The session survives for the next prompt** (roadmap U5 accept clause) needs the `SS` Session.Supervisor tree, which is not on master; the reds stay at the Interrupt-behaviour + journal seam and leave session-liveness to a U5-I integration test.

Part of the red-first fan-out authored against docs PR #569 (freeze contracts).


merklebonsai:
Authoring flags: (1) grace-window latency is a policy knob, not contract — reds assert the wait *stage*, never milliseconds. (2) mid-provider-stream kill expressed at observable-contract level (turn_canceled + no trailing output) since SSE streams aren't OS Ports. (3) Correction to the suite's own note: SS (#558) IS merged on master — the session-liveness acceptance clause still lands as a U5-I integration test as planned, the note in the moduledoc just misstates why.

DROOdotFOO:
# Adversarial Review — PR #572 (U5-R red suite: interrupt = staged supervised kill)

**Verdict: CONCERNS** — merge-able as a red-first scaffold, but the "teeth" story is narrower than the PR body claims and two dimensions the title promises ("supervised", "escalation") are unpinned.

Genuine positive: the oracle-independence discipline is real and well-executed — `Contours.records/1` folds the raw JSONL segments with its own decoder and never touches `FileStore.read/2` or the writer offset, and I verified the writer opens segments `:append, :raw, :binary` with **no** `:delayed_write` and `append/2` is a synchronous `GenServer.call`, so the independent reader cannot race the writer. The negative-control shape (green-on-correct + red-on-mutant + fired-counter for dead injectors) is the right pattern and, where it is applied, it has teeth.

The core caveat for a red suite: the positive contours are `@moduletag :harness_red` and excluded from CI, so **the only thing running in CI is the controls file**. That makes control coverage the entire real signal — and it is incomplete.

---

## HIGH

**[Saboteur + New Hire] `interrupt_contours.ex` — 2 of 5 contour assertions have no negative control (meta-inv 4 not satisfied).**
`assert_turn_canceled!` (P2) and `assert_no_trailing_output!` (P3b, the mid-provider-stream contour) are invoked *only* from the CI-excluded positive suite (`u5_interrupt_red_test.exs`). No dead injector exercises them; `u5_interrupt_controls_test.exs` covers only staging (`:skip_wait`), quiescence (`:late_result`), and effectiveness (`:trust_exit_status`). Per the PR's own governing rule ("one mutation per contour must make its property fail"), these two contours are unproven: they run *only* against an unimplemented `Interrupt.interrupt/3` that raises, so if either assertion were vacuous or buggy, nothing in CI would catch it until U5-I lands. The body's claim of the negative-control pattern holding across "each contour" is true for 3/5. Either add `:trust_reason`/`:trailing_output` mutants, or drop the "every contour" framing.

**[Saboteur] Escalation is never conditional — the "well-behaved tool stops at signal" law is untested, and `spawn_nice` is dead code.**
The whole point of a *staged* kill is that a cooperative tool exits at the signal stage and the sequence short-circuits (no wait, no kill). Nothing pins that. `Injectors.Reference.interrupt/3` (injectors.ex ~L931-938) unconditionally emits `signal → wait → kill → turn_canceled` regardless of tool behavior; the "bounded wait" is a hard-coded event, not the *result* of an observed timeout. `KillLab.spawn_nice/1` (kill_lab.ex ~L1106) — the cooperative tool that would demonstrate short-circuit / naive-signal leakage — is defined and **never used** anywhere in the suite. So the escalation dynamic (cooperative exit avoids kill; hang forces kill) is asserted nowhere, and the moduledoc's "a well-behaved tool stops here" is prose without a contour. This is the substantive half of "staged".

---

## MEDIUM

**[Saboteur] `interrupt_contours.ex` `assert_effective!` (~L703) + `u5_interrupt_red_test.exs` P5 (~L555) — grandchild liveness is a single-shot `ps`, inconsistent with the parent poll → CI flake vector.**
`assert_effective!` polls the top pid with `KillLab.await_dead(lab.os_pid)` (up to 500 ms) but checks the grandchild with a single non-polling `KillLab.dead?(lab.child_pid)`. After `kill -9 -<pgid>` the kernel signals parent and grandchild together, but there is no ordering guarantee that the grandchild leaves the `ps` table before the parent; under load the `sleep` can still be reaping when the single check fires. This runs in CI via the `:unix_only` effectiveness control. Use `await_dead(lab.child_pid)` for symmetry with the parent check.

**[Security Auditor] Blast radius is asserted only positively — no test proves a bystander survives, and the `group_leader_safe?` guard is code without a test.**
The effectiveness contour proves the *target* subtree dies, but nothing pins the dual security property: that the interrupt kills *only* the intended group. `KillLab.group_leader_safe?/1` (kill_lab.ex ~L1217) is the sole thing standing between a mis-derived pgid and `kill -9 -<pgid>` hitting the test runner (or an unrelated group), yet no test feeds it a pid whose `pgid != pid` to confirm it declines the group-kill and falls back. A red suite about killing OS process groups should pin "spawn a bystander in a different group, run the interrupt, assert the bystander is still alive" and "a non-leader pid never triggers a group SIGKILL." Right now the safety guard is untested defense-in-depth.

---

## LOW

**[New Hire] `kill_lab.ex` `reap/1` (~L1211) — bare `rescue _ -> :ok` swallows all exceptions.** Violates the repo convention against broad rescues (CLAUDE.md: "avoid broad `rescue e ->`; list specific exception types"). A teardown that also silences programmer errors can mask a real leak of a 30s `sleep`. Narrow to the exceptions `System.cmd` can actually raise, or drop the rescue since every call is already `_ = System.cmd(...)`.

**[New Hire] `u5_interrupt_controls_test.exs` (~L269) — `pre_kill = 1 + :rand.uniform(3)` injects nondeterminism into a "deterministic" pure-journal control.** The seed is dumped on failure so it's reproducible, but the random count doesn't vary any part of the contract (quiescence forbids *any* post-kill output regardless of pre-kill volume). A fixed `pre_kill = 2` would be simpler and genuinely deterministic; the `:rand.seed/2` in `setup` otherwise buys nothing.

**[Saboteur] `assert_staging!` has no reorder mutant.** `:skip_wait` proves a *missing* stage fails, and `seq == expected` structurally covers reordering, but no injector proves a `wait`-after-`kill` transposition is caught. Cheap to add given the injector scaffold already exists.

**[Security Auditor] PID-reuse TOCTOU on the captured `os_pid`.** `group_kill/reap` act on the pid captured at spawn; if the tool exited and the pid recycled, `kill -9 -<os_pid>` could target an unrelated group. Mitigated in practice (rogue never exits within the window; the `pgid == os_pid` guard further narrows it) — noting as accepted risk, not a fix request.

---

## Cross-persona overlaps (promoted)
- **Coverage/teeth gap** flagged independently by Saboteur (does it verify the stages/escalation?) and New Hire (is the contract legible *and* exercised?) → the HIGH findings above.
- **"Only the intended subtree dies"** flagged by both Security Auditor (blast radius) and Saboteur (orphan/cascade) → the MEDIUM bystander finding.

## On vacuous-pass risk
The controls that *do* run are not vacuous — `assert_quiescent!` requires a kill fence to exist and `assert_staging!` compares the full ordered sequence, both proven red-on-mutant. The vacuity risk is confined to the two uncovered contours (P2, P3b) and to the untested escalation-conditionality, which is exactly where a red suite is supposed to have already planted its teeth. Also note the "supervised" clause of the title (session survives, supervisor restart, `trap_exit`, linked-process cascade, stage-hang timeout escalation) is entirely out of scope here — partly disclosed in the PR body (session liveness deferred to U5-I integration), but the escalation-timeout and cascade gaps are not called out.

Nothing here blocks the scaffold from landing; the asks are (1) controls for P2/P3b, (2) a cooperative-tool short-circuit contour that actually uses `spawn_nice`, (3) `await_dead` on the grandchild, (4) a bystander-survives / guard-declines test.

merklebonsai:
Excellent adversarial pass — all findings accepted, fix commit in flight (additive test-only, reds stay red): P2/P3b negative controls (HIGH), cooperative-tool short-circuit using spawn_nice (HIGH), bystander-survival + group_leader_safe guard test (M), grandchild await_dead poll (M). These are exactly the safety edges a red suite must plant teeth on before the impl is built against the frozen shape; landing them now. Will re-request review when the commit lands.

merklebonsai:
@DROOdotFOO review fixes landed in `f5d62f8ad` — ready for re-review:
- **HIGH P2/P3b controls**: TrustReason injector (wrong terminal) + TrailingOutput injector (post-kill provider-stream output) — each contour now has a red-on-mutant control.
- **HIGH escalation conditionality** (the deep one): Reference previously journaled :interrupt_signaled without a real signal, so cooperative short-circuit was untestable. Added KillLab.signal_group/1 (real group SIGTERM, same guard as group_kill), rewired Reference to send it + poll for cooperative exit before deciding to escalate; assert_short_circuit! contour (cooperative tool dies on signal → no wait/kill stage) + NaiveEscalate injector (signals but always escalates). spawn_nice now exercised for the first time — the substantive half of 'staged'.
- **MEDIUM**: grandchild await_dead poll (symmetric, flake killed); bystander-in-different-pgroup survives + group_kill declines -pgid when pgid≠pid (fed the grandchild pid).
- **LOW**: dropped bare rescue in reap/1, pre_kill fixed to 2, WaitKillTransposed staging mutant added.
Controls 11/11 (was 5), stable across 5 seeds; reds 0/6 (still red); full suite 1120 green; zero leaked OS procs confirmed via ps sweep.

merklebonsai:
CI note (for merge triage): the red checks here are **stale-base cross-lane failures**, not this unit's work. Failing tests are all UI-lane / flaky: `test/harness/tp_pty_test.exs` (the exact test #588 is tagging `:skip_on_ci`), `test/harness/tf_fixture_test.exs` (UI-lane fixture toolchain, TF #551), and `HeadlessGoldenTest` (rendering golden). All green on `master`; this branch was cut before #588 + the UI stabilizations landed. The `raxol_agent` unit suite is green locally (impl + review-fixes). Clearing CI needs a rebase onto current master — deferring that since the stacked impl branches (u8/u12/u10) are being built on this base right now. No raxol_agent regression.

DROOdotFOO:

Scope: `packages/raxol_agent/lib/raxol/agent/interrupt.ex` is the only shipping (non-test) code; the rest is red suite + injectors + `KillLab`. Findings weighted toward the production module and its OS-kill semantics.

### HIGH
- [Saboteur + New Hire] `interrupt.ex` `await_dead/2` -> `dead?/1` -> `alive?/1`, and its callers `await_dead(os_pid)` in `hard_kill/1` and `signal_ok? and await_dead(os_pid)` — the "OS-confirmed death" oracle only ever observes the **top `os_pid`**, never the process group or any descendant. `outcome.confirmed_dead?` is set `true` (and `:interrupt_killed` emitted) the moment the *parent* pid is gone. The module's whole reason to exist (law 1, moduledoc "Effectiveness") is "the grandchild survives a top-pid kill; only a group kill counts" — yet the confirmation logic makes exactly the top-pid-only observation it condemns. Failure scenario: on any host where `group_leader_safe?/1` returns false (the documented macOS/BSD pgid-derivation regression, U5-#4), `hard_kill/1` takes the fallback in `group_signal/2`, which signals only the parent + its **direct** children (`children_of/1`, depth 1) — a `sleep` grandchild of a child is never signaled, survives, and the turn still returns `killed?: true, confirmed_dead?: true` with an `:interrupt_killed` fence. The effectiveness law is silently violated while the journal claims success. The green `:unix_only` reds only exercise the happy group-kill path, so this gap ships untested.

### MEDIUM
- [Saboteur] `alive?/1` via `ps -p <pid>` — a killed-but-not-yet-reaped port child is a zombie (`STAT Z`); `ps -p` still exits 0 for it, so `alive?/1` reports it alive. BEAM reaps its own port children asynchronously, so after a successful group SIGKILL `await_dead/2` can burn the full 500 ms budget on a zombie and return `false`, yielding `confirmed_dead?: false` on a genuinely-dead process. Both a correctness lie and a CI flake vector on loaded machines. `kill -0` or a `STAT`-filtered check is the correct liveness oracle, not bare `ps -p`.
- [Saboteur] The OS kill in `hard_kill/1` executes *before* `emit(sink, kill_stage, ...)` / `emit(sink, @terminal, ...)`. `emit/4` calls the durable journal sink with no `rescue`/`try`. If that write fails (disk full, writer dead), the OS process is already dead but the journal's last record for the turn is `:interrupt_waited`, and `interrupt/3` crashes up into its caller. A journal replay sees a turn wedged mid-kill that is actually complete — reality/journal divergence with no reconciliation path.
- [Saboteur + Security] `await_dead/2` polls with `Process.sleep(10)` and one `System.cmd("ps", ...)` per tick, up to 500 ms (~50 forks), and `cooperative_exit/2` adds another `grace_ms`/10 (~30) ticks, plus the `kill` shell-outs: ~80 short-lived `sh`/`ps` OS processes per single `interrupt/3`, run **synchronously in the caller's process**. If `interrupt/3` runs inside a session/dispatcher GenServer, that process is blocked and unresponsive for `grace_ms + 500 ms`; N concurrent interrupts multiply the fork pressure with no bound or back-pressure. Prefer a monitored port or `:erlang`-level signal over a shell-out poll loop.

### LOW
- [New Hire] `kill_stage_for(_) -> @kill` maps the `:noop` (tool-less, mid-provider-stream) disposition to `:interrupt_killed`. A turn that never had an OS process journals an `interrupt_killed` "kill-complete" event — a bookkeeping white lie that a maintainer reading the journal will misread as an actual kill. Only the code comment distinguishes it.
- [New Hire] `interrupt.ex` moduledoc and inline comments are saturated with project-internal jargon: `U5-R`, `U5-I`, `AD-12`, `finding U5-#5`, `finding U5-#1/#4`, `spike gotcha #4`, `meta-inv`, `the freeze doc`. Per the repo's own documented convention (strip AI meta-narration + Phase N / ADR-NNNN from comments/docstrings), these references are opaque to anyone without the harness spec and should be reduced to the behavioral contract.
- [New Hire / Saboteur] `test/test_helper.exs` adds `:harness_red` to the CI exclusion list, but **no test file in this PR carries `@moduletag :harness_red`** (confirmed: `u5_interrupt_red_test.exs` tags only `:capture_log`). The exclusion is vestigial, and the PR body's claim that "the positive suite is `@moduletag :harness_red` and added to the exclusion" contradicts the shipped file — these real-OS-process-spawning reds actually run in CI on every push, unguarded except by `:unix_only`.
- [Security] `bsd_children_of/1` runs a full `ps -Ao pid=,ppid=` table scan on every macOS/BSD fallback (where `--ppid` is unsupported), parsing the entire process table. Unbounded by system process count; a busy host makes each fallback O(processes).

### Cross-Persona Overlaps
- The HIGH (group vs top-pid confirmation) was reached independently by the Saboteur (effectiveness-law violation on the fallback path) and the New Hire (`confirmed_dead?` name over-promises what `await_dead` verifies) — promoted to HIGH.
- The poll-loop finding is both a Saboteur (blocked supervised process) and Security (unbounded fork/DoS) concern — kept MEDIUM.

### Verdict: CONCERNS
One HIGH (the `confirmed_dead?`/`:interrupt_killed` claim is systematically weaker than the group-kill contract it advertises, and breaks outright on the documented pgid-fallback path) plus multiple MEDIUM (zombie masking, kill-before-emit journal divergence, synchronous fork-loop blocking). No CRITICAL, fewer than 3 HIGH. Recommended: verify the whole process group / grandchild is dead before setting `confirmed_dead?`/emitting `:interrupt_killed` (or downgrade the claim on the fallback path), switch liveness to `kill -0`/state-aware check, and guard `emit` around the post-kill fence.

merklebonsai:
## Review round 3 — dispositions (commit 5da73ce)

### Fixed (each with a regression that fails pre-fix)

**HIGH — confirmation observed only the top pid.** Confirmed. `confirmed_dead?`/`:interrupt_killed` are now established group-scoped or not at all:

- **Group path** (safe pgid): confirmation is a pgid-scoped liveness poll (`ps -Ao pgid=,stat=` filter) — the kill-complete fence now means *the group* was observed gone, per the effectiveness law.
- **Degraded per-pid fallback** (no safe pgid — the macOS/BSD pgid-derivation regression, or a non-leader target): the sweep now enumerates **transitive** descendants (one pid/ppid table scan, BFS) as best-effort containment, but a sweep can never prove the group is gone (a mid-sweep fork escapes), so it **never claims the kill**: new `:degraded` disposition → `:interrupt_kill_failed` fence, `killed?`/`confirmed_dead?` both false — consistent with the round-2 kill-integrity fix.
- **Cooperative short-circuit**: group-observable hosts confirm via the group oracle; a top-pid-only observation still short-circuits the escalation but reports `confirmed_dead?: false`.

Regression **P8** (`:unix_only`): a nested rogue (`bash → sh MID (traps TERM) → sh → sleep`) interrupted by its non-group-leader MID pid — pre-fix this journaled a false `:interrupt_killed` with `confirmed_dead?: true` while the grandchild `sleep` lived; now it yields `kill_failed`, both flags false. Fails pre-fix, green post-fix.

**MEDIUM — zombie masks liveness.** Confirmed. The oracle is now state-aware: `ps -o stat= -p` with STAT `Z*` counted as dead (a zombie is a killed process awaiting reap), for both the single-pid and group oracles. Regression **P9**: a manufactured zombie (`sleep 0.2 & …; exec sleep 30`) — raw `ps -p` exits 0 for it, `Interrupt.os_alive?/1` reports dead. Fails pre-fix.

**MEDIUM — kill-before-emit journal divergence.** Confirmed. The emits after the irreversible kill are now guarded: a sink failure converts to `{:error, {:sink_failure, error, outcome}}` where `outcome` carries the OS truth, instead of raising out of `interrupt/3` mid-divergence. The journal then verifiably lacks the kill fence (it never lies); the caller owns reconciliation. Regression **P10**: a sink that dies on the kill fence — pre-fix raised, now returns the error tuple and the journal holds signal+wait only. **Flag for ruling:** the full replay-reconciliation policy for a turn wedged mid-kill (re-drive vs. tombstone) is a journal-contract question above this module; the fix here is the fail-closed half (never raise, never forge).

**LOW — jargon in comments.** Confirmed against repo convention (the "strip process narration" precedent). Unit tags, finding numbers, and roadmap shorthand are stripped from `interrupt.ex`; behavioral content kept.

**LOW — bsd_children_of full-table scan.** Partially folded into the HIGH fix: child enumeration is now ONE `ps -Ao pid=,ppid=` scan walked transitively in memory (previously a scan per call, per level). The scan itself is inherent — portable BSD ps has no `--ppid` — and now runs at most twice per interrupt, only on the fallback path, which no longer claims success anyway.

### Deferred (verified, not fixed here)

**MEDIUM — synchronous fork/poll loop in the caller.** Real: `interrupt/3` blocks its caller for up to `grace_ms + 500ms` with ~1 fork per 10ms tick. Bounded per turn and currently uncalled by any production process (the loop wiring lands with the session integration), so no supervised GenServer is blocked today. Moving the kill to a monitored async worker is the right shape for that integration PR — deferring rather than speculatively designing the ownership model here.

**LOW — `:noop` maps to `:interrupt_killed`.** Verified, deferred as needs-ruling: the tool-less 4-stage bookkeeping sequence is pinned by the frozen `sequence/0` and reds P1/P3b — changing the tool-less fence to a distinct event type is a freeze-contract amendment, not a code fix. Note the journal already distinguishes it structurally, not just by comment: the kill-stage payload carries `os_pid: null` for the bookkeeping fence vs. a real pid for a kill.

**LOW — `:harness_red` exclusion vestigial.** Verified: no current file carries the tag (dropped in 5e628f8 when the reds went green — the PR body's header notes this; the "CI discipline" section below it is the preserved as-authored spec). The exclusion stays in `test_helper` as the standing failing-first authoring mechanism for the next red suite, per its own comment block. The OS-spawning reds running in CI under `:unix_only` is intended now that they are the green contract pins.

### Verification

- U5 reds + controls: 22/22 green (P8/P9/P10 each verified failing against the pre-fix module)
- Full raxol_agent suite: 1131 passed, 0 failed
- No leaked OS processes after the suite; `mix.lock` drift restored

DROOdotFOO:
Use the [web editor](https://github.com/DROOdotFOO/raxol/pull/572/conflicts) or the command line to resolve conflicts before continuing.

PR #573, opened 2026-07-15
feat(harness): U6 steer via expected_turn_id CAS — reds + impl (AD-13)

merklebonsai:
## Status: complete unit

Red suite (the spec, below) **+ implementation + adversarial review round** — reds now run **green in CI**. Original red-suite description preserved below as the authored spec.

---

## What U6 is

**Steer** = redirect a running turn with new user input *without killing it* (the sibling of U5 interrupt: interrupt is kill-now, steer is inject-at-next-boundary). The core mechanism is **`expected_turn_id` compare-and-swap** (AD-13): a steer command carries the turn it believes is running; if the actual current turn differs (the turn ended, or another steer already won the race), the steer is **rejected with a typed error** — never silently misdirected into the wrong turn. An accepted steer lands a durable event in the target turn; the model sees the steering input at the next boundary. A duplicate `client_msg_id` deduplicates (freeze-contracts §5.1).

This is the **permanent failing-first red suite** for U6, authored *before* the implementation exists. It drives `Raxol.Agent.Steer.resolve/2` + `rebuild/1` — a skeleton that raises `:not_implemented` — so every contour is red until U6 lands (drop `@moduletag :harness_red` and they go green).

## Contours covered

The contract lives once in `test/support/steer_contours.ex` (the single source of truth both the reds and the controls run):

- **accept lands** — matching `expected_turn_id` → one durable event, attributed to the target turn, carrying the steering text + `client_msg_id`
- **stale reject** — mismatched `expected_turn_id` → `{:error, {:stale_turn, expected, actual}}`
- **nothing on reject** — a rejected steer journals nothing and has zero model effect (state byte-identical)
- **racing steers** — two concurrent steers built against the same observed turn → exactly one wins the CAS, the loser gets a typed reject; schedule is **seed-reproducible**
- **duplicate dedup** — same `client_msg_id` re-delivered → `{:ok, {:duplicate, original_ref}}`, one durable event, second acked-as-duplicate
- **dedup survives a BEAM restart (§5.1, load-bearing)** — the journal is the dedup truth (session-lifetime window); the in-memory index is **rebuilt by fold** over the durable log via `rebuild/1`, so a `client_msg_id` re-delivered after a restart still deduplicates even though a different turn is running by then

## Dead injectors (negative controls)

`test/support/steer_injectors.ex` — each a broken `Raxol.Agent.Steer` that violates exactly one invariant, caught by its target contour in the **in-CI** controls (`u6_steer_controls_test.exs`):

| injector | breaks | fails |
| --- | --- | --- |
| `SkipCas` | the CAS compare | stale-reject |
| `JournalBeforeCas` | write-after-decide | nothing-on-reject |
| `DropDedup` | idempotency memory | dedup |
| `InMemoryOnlyDedup` | dedup not rebuilt from fold | **only** post-restart dedup (passes in-process dedup — the discrimination the delta-audit pinned) |

The controls also assert the **reference** implementation passes every contour (no red is vacuous), fired-counters on every fault site (a site that stops injecting fails the control — meta-inv 1), a dead-injector-detection self-test, and seed-reproducibility of the racing schedule (meta-inv 2).

## `:harness_red` discipline

- `@moduletag :harness_red` on the red suite is **excluded** in `test_helper.exs`, so CI stays green while the suite is red (6 red tests show as *excluded*, visible not hidden).
- The negative controls are **un-tagged and DO run in CI** (8 tests, all green today — they never touch the unimplemented `Steer`).
- Two commits: enabler (Steer skeleton, `@callback`s, `:not_implemented` raises) then the red suite + support.

## Verification

- Full `raxol_agent` suite: **1028 passed, 39 excluded** (6 of the excluded are these reds)
- Reds included → **6/6 fail** with `:not_implemented` (genuine red-first); controls → **8/8 pass**
- No `mix.lock` drift committed


merklebonsai:
Authoring flags: (1) racing steers expressed as seed-ordered application of the pure CAS machine (order-independent invariant asserted across seed spread) — true multi-process racing composes at U6-I with the live turn process. (2) BEAM restart modeled as rebuild/1 fold-over-log; the physical disk round-trip is already guarded by the landed I1/I10 invariants and composes at U6-I. Both boundaries documented in moduledocs. Incorporates the pinned §5.1 dedup semantics (session-lifetime window, journal-as-truth, fold rebuild) incl. the InMemoryOnlyDedup injector that fails ONLY the post-restart red.

DROOdotFOO:
## Adversarial Review (automated: 3-persona)

**Verdict: CONCERNS**

What's good: this is a genuinely well-engineered red harness. The negative controls are not theatre — they prove the checkers are non-vacuous (reference passes every contour, `u6_steer_controls_test.exs:198`), prove each dead injector is caught by its target contour (`:213`), enforce fired-counters so a silently-disarmed injector fails the control (`SteerFaults.assert_all_fired!`, `steer_faults.ex:666`), and pin seed-reproducibility of the race schedule (`:299`). The single-source-of-truth contours shared between reds and controls (`steer_contours.ex`) genuinely closes the "green lie" gap. The decision-order that matters most — dedup checked *before* CAS so a post-restart duplicate still fires against a different running turn — is pinned tightly (`assert_dedup_survives_restart`, `steer_contours.ex:583`).

The concerns are not vacuity — they are **missing safety contours**. Because U6's real implementation will be built against this frozen shape, a gap here means a broken-but-passing implementation can slip through. That is exactly the failure mode a red suite exists to prevent.

### HIGH

**[Saboteur + Security] steer_reference.ex:927-936 / steer_contours.ex:460 — ABA / CAS-token reuse is never pinned.** The reference makes ABA structurally impossible by minting a globally-unique token (`swap/1` uses `System.unique_integer([:positive])`, `steer_reference.ex:984`). But no contour *freezes* that requirement. A U6 implementation that swapped with a boolean toggle or a per-turn counter that can repeat (A→B→A) would pass all six reds: `assert_one_winner` only exercises a single A→(swap)→stale chain (`steer_contours.ex:504-519`), never a token that returns to a previously-observed value. Result: a steer built against a *stale* turn-A token could wrongly win the CAS after the turn cycled back to an equal token — the precise "silent misdirection into the wrong turn" U6 claims to prevent. Add a contour asserting the post-accept token is distinct from every previously observed token (monotonic/unique), not merely `!= cur`.

**[Security + New Hire] steer_reference.ex:932-936 — nothing forbids steering an idle / no-live-turn session (nil == nil accept).** The accept branch fires whenever `expected == cur`. `TurnState.turn_id` defaults to `nil` (`steer.ex:91`) and `Request.expected_turn_id` is enforced-present but may be `nil` (`steer.ex:69-70`). So `resolve(%TurnState{turn_id: nil}, %Request{expected_turn_id: nil})` **accepts** and lands a durable steer event attributed to a nonexistent turn. `assert_stale_reject` only ever tests a non-nil mismatch (`@other` vs `@target`, `steer_contours.ex:460-467`); there is no negative contour pinning "a steer against a session with no running turn must reject." For a mechanism whose whole safety story is "never misdirect input into the wrong turn," steering into a *nonexistent* turn is the load-bearing edge and it is unpinned.

### MEDIUM

**[Security] steer_contours.ex:552-572 — dedup keyed on `client_msg_id` alone; payload-mismatch behavior unpinned.** `assert_dedup` re-delivers a byte-identical request, and the reference dedups on `cmid` only (`steer_reference.ex:928`, `Map.has_key?(seen, cmid)`). The suite never pins what happens when a second delivery carries the *same* `client_msg_id` but *different* `text`. If `cmid` is client-controlled, this is a suppression vector: an attacker (or a buggy client) reusing a `cmid` silently drops the new steering text and acks it as a duplicate of an unrelated original. Whether the contract intends "cmid is the sole idempotency key, payload ignored" or "payload must match" is a real safety decision that this frozen suite leaves undecided.

**[Saboteur + New Hire] steer_contours.ex:497-522 — "concurrent racing steers" is sequential; real concurrency is out of scope but the naming oversells it.** `assert_one_winner` is a pure `Enum.map_reduce` threading state in a seed-chosen order (`:506-510`) — there is no concurrency, no interleaving, no TOCTOU. The moduledoc is honest that the runtime's mailbox serializes for real and this is the pure decision core, which is a legitimate scoping choice. But the test name "concurrent racing steers" and the red at `u6_steer_red_test.exs:368` claim more than they exercise: the actual read-then-write race in the session runtime (the thing that *could* let two steers both win) is untested here and has no companion contour. At minimum rename to reflect "serialized CAS ordering," and note explicitly that runtime-level concurrency is a separate (currently uncovered) obligation.

### LOW

**[New Hire] steer_injectors.ex:699-703 — the injector table lists three rows but four injectors exist.** `InMemoryOnlyDedup` (`:860`) is absent from the moduledoc table, and the module/PR prose says "three dead injectors" in several places while `SteerFaults.@sites` correctly lists four (`steer_faults.ex:632`). Documentation drift on the exact artifact whose job is precision.

**[Saboteur] steer_faults.ex:638-641 — `SteerFaults.new/1` starts an `Agent` that is never stopped.** Harmless under `async: true` with an unnamed process, but it leaks a process per control test and relies on GC rather than `start_supervised`/`on_exit`. Prefer `start_supervised` for deterministic teardown, consistent with the repo's stated pattern.

**[New Hire] steer_contours.ex — nil `client_msg_id` (no idempotency key) behavior is unpinned.** The reference only dedups when `cmid != nil` (`steer_reference.ex:928`, `:951`), so two nil-cmid steers both apply. Plausibly intended, but no contour freezes it; a future impl that dedups-on-nil or crashes-on-nil would pass the suite.

### Cross-persona overlaps (severity promoted)
- **ABA / token-reuse** — Saboteur (determinism/race) + Security (replay). Promoted to HIGH.
- **idle/nil-turn steer** — Security (unauthorized steer into nonexistent turn) + New Hire (missing edge case). Promoted to HIGH.
- **"racing" sequential naming** — Saboteur (no real race) + New Hire (name doesn't match behavior). MEDIUM.

### Bottom line
The harness *machinery* is solid and non-vacuous — no notes on the controls, fired-counters, or seed-reproducibility. The verdict is CONCERNS rather than CLEAN because the frozen contract has two safety-critical holes (ABA token-reuse, idle/nil-turn accept) that a passing-but-broken U6 implementation would sail through, plus an undecided dedup-payload semantics. These are additive test-only changes with zero runtime risk, so not a BLOCK — but the two HIGH contours should be added *before* U6 is implemented against this shape, so the frozen contract actually covers the cases it claims to protect.

merklebonsai:
Excellent adversarial pass — all findings accepted, fix commit in flight (additive test-only, reds stay red): ABA token-uniqueness contour (HIGH), nil/idle-turn reject (HIGH), dedup payload-mismatch semantics (M), 'serialized CAS' rename (M). These are exactly the safety edges a red suite must plant teeth on before the impl is built against the frozen shape; landing them now. Will re-request review when the commit lands.

merklebonsai:
@DROOdotFOO review fixes landed in `a149eca82` — ready for re-review:
- **HIGH ABA/token-reuse**: assert_token_uniqueness (post-accept token distinct from every previously-observed, not just != current) + RepeatableToken injector (A→B→A); ABA-safety law stated in the steer moduledoc.
- **HIGH idle/nil-turn**: assert_no_live_turn_reject + NilTurnAccept injector; reference now rejects is_nil(cur) after dedup, before CAS, with {:error, :no_live_turn}.
- **MEDIUM dedup payload-mismatch**: decided same cmid + different payload → {:error, :client_msg_id_reuse} (safer than silent-ack); seen now stores {ref, text} so the check survives rebuild/1 fold; + IgnorePayloadMismatch injector.
- **MEDIUM naming**: assert_one_winner → assert_serialized_cas_order + moduledoc note that runtime concurrency is a separate uncovered U6-I obligation.
- **LOW**: injector-table drift fixed (7 listed), SteerFaults uses start_supervised!, nil-cmid-not-deduped contour added.
Reds 0/10 (still red), controls 11/11, full suite 1120 green. (Branch chain verified intact: enabler→red→fix.)

merklebonsai:
@DROOdotFOO Status: #573 advanced red-suite → **complete unit** and back to ready-for-review. Steer CAS core: idempotency-before-CAS (duplicate replay only for genuinely-accepted originals — no staleness masking), nil-turn reject (your tooth honored), stale CAS reject with zero model effect, ABA-safe run-scoped swap tokens. Review round folded in — the substantive fix: `rebuild/1` now folds the **real disk shape** (string-keyed, `client_msg_id` nested in payload per frozen §5.1 — it previously assumed atom keys and would have silently rebuilt an empty dedup index from a real journal). Pinned by a physical FileStore write→scan→rebuild round-trip regression that is red against the pre-fix code. `accept/5` emits the frozen payload-nested event shape. 1131 passed, 0 failures. Known CI noise: stale-base UI-lane tests (green on master, task'd for rebase).

DROOdotFOO:

Scope: `Raxol.Agent.Steer` (`resolve/2` + `rebuild/1`, the only production code) plus its red suite / injectors / reference. Three personas, grounded in file:line.

### HIGH

- **[Saboteur] steer.ex:190-211 — the "CAS" provides no atomicity; concurrent-steer safety rests on an unshipped, untested contract.** `resolve/2` is a pure function: the "compare" (`expected != cur`) and the "swap" (`swap(cur)`) are just a read and a return on the caller-supplied struct. There is no atomic read-modify-write here. `SteerContours` moduledoc concedes this outright: the checker is a sequential `Enum.map_reduce`, and "an actual read-then-write race inside the session runtime's process ... is a SEPARATE, currently-uncovered obligation belonging to U6-I." Failure scenario: an integrator does the natural `state = fetch(ets); {res, next} = Steer.resolve(state, req); put(ets, next)` from two processes (or a `Task`) — both read the same `turn_id`, both hit the `true ->` accept clause, both win, two durable steer events land in one turn. That is exactly the "silent misdirection into the wrong turn" the module claims to prevent, and nothing shipped in this PR forecloses it. Naming/​documenting the primitive as a CAS implies an atomicity guarantee the function does not carry.
- **[Security + Saboteur] steer.ex:244, 263-280 — unbounded, client-controlled dedup state (memory-exhaustion DoS), compounded by O(n^2) log growth.** `client_msg_id` is a client-supplied `term()` memoised into `seen` for the *session lifetime* (moduledoc: "the dedup window is the session lifetime") with no cap or eviction. An adversarial/buggy client streaming unique `client_msg_id`s grows `seen` without bound; `rebuild/1` then folds the *entire* journal on every restart (O(journal)). Independently, each accept does `length(log)` (O(n)) and `log ++ [event]` (O(n)) — quadratic over a session of N steers. Two personas land on the same object (retention DoS + quadratic amortization); promoted. Idiomatic fix: prepend + a monotone counter, and a bounded/windowed dedup index.

### MEDIUM

- **[Saboteur] steer.ex:246,324 — latent ABA the swap comment waves away but does not enforce.** The safety argument is "swap tokens are never persisted as CAS targets ... on resume the `turn_id` comes from the journal's turn brackets." But `next.turn_id` *is* a swap token (`{:steered, cur, unique_integer}`) and is the most natural field an integrator persists as the turn id. `System.unique_integer/1` resets on VM restart, and the token is deterministic given a replayed accept sequence, so a persisted-then-replayed token can collide post-restart -> ABA. The invariant "MUST NOT persist a swap token as a client-visible turn id" is prose only; nothing in the type or API prevents it. `rebuild/1` setting `turn_id: nil` is not enforcement — the caller immediately overwrites it.
- **[Saboteur] steer.ex:225,279 — post-restart `log` is heterogeneous and `offset` is a positional guess, not the durable offset.** `rebuild/1` sets `log: journal` (the *entire* scanned journal, including non-steer families, as string-keyed maps). A subsequent accept computes `offset = length(log) + 1` — a position over the whole mixed journal, which is not the offset `FileStore.append` actually assigns, so the returned `ref.offset` can diverge from the real durable position. The same accept appends an atom-keyed event onto a list of string-keyed records; any later atom-path read (`hd(log).payload.text`) on an old entry crashes. The disk round-trip test only asserts `length`, never field access on pre-restart entries, so this is untested.
- **[New Hire] steer.ex:1-92, 313-324 — production docstrings are dense with unresolvable insider jargon, against the project's own convention.** "AD-13", "freeze-contracts §5.1", "JS-FREEZE §1", "§0 reader-tolerance", "meta-inv 1", "U6-I", "Drew's nil-turn tooth". These resolve to docs PR #569, not this repo, so a new maintainer cannot follow them. The repo convention (per CLAUDE.md memory) is explicitly to strip AI meta-narration and ADR-NNNN references from comments. The 12-line essay on `swap/1` is more narrative than the one-line function warrants.
- **[New Hire] test_helper.exs:1714-1716 — adds a `:harness_red` exclusion whose cited example no longer carries the tag.** The comment names "U6-R steer" as the example harness_red suite, but `u6_steer_red_test.exs` has *no* `@moduletag :harness_red` (it was dropped; `grep` confirms no red test in the package still carries it). steer.ex moduledoc line ~88 says the tag "was dropped." So this diff adds a dead exclusion documented against the exact suite that removed it — contradictory and misleading to the next reader.
- **[New Hire] steer.ex:111-123 — `TurnState.log` is documented as "the pure-model stand-in for the journal," inside a `lib/` production module.** A production struct field described as a test-model surrogate blurs whether `Steer` is production code or scaffolding, and invites callers to treat `log` as authoritative when the runtime's real durable store is elsewhere.

### LOW

- **[Security] steer.ex:293-300 — `field/2` cannot distinguish "payload key present with `nil` value" from "absent."** A legitimately-`nil` payload field silently falls through to a top-level lookup; if the two ever differ, the wrong value is read. Harmless for today's shapes, latent otherwise.
- **[New Hire] steer.ex:293,304 — generic helper names `field/2` / `get_either/2`** contravene the project's "avoid generic names" rule; `payload_field` / `fetch_atom_or_string` would say what they do.
- **[New Hire] steer.ex:234-241 (+ reference + 5 injectors) — the frozen event map is copy-pasted verbatim across 7 sites** with no shared constructor; a shape change must be made in lockstep in every copy.
- **[Security] steer.ex:240 — client `text` is unbounded and written durably** with no length guard.

### Cross-Persona Overlaps
- Resource exhaustion: Saboteur (O(n^2) log + unbounded `seen`/rebuild) and Security (client-controlled session-lifetime dedup keys) converge on `seen`/`log` -> promoted to HIGH.
- CAS integrity: Saboteur's atomicity gap (HIGH) and ABA-via-persisted-token (MEDIUM) both stem from the same root — the module ships a CAS *name* whose guarantees live in an unshipped, unenforced runtime contract.

### Verdict: CONCERNS
Two HIGH (CAS provides no atomicity as shipped/tested; unbounded client-controlled dedup state) plus several MEDIUM. Not a BLOCK — no CRITICAL, fewer than three HIGH, and the code is a well-factored pure core with a genuinely thorough negative-control harness. But the "CAS" safety property is asserted in docs and untested in code, and the dedup index is an unbounded client-controlled surface; both should be resolved (or the atomicity/retention obligations explicitly tracked to U6-I with a failing guard) before the integrating runtime lands.

merklebonsai:
## Review triage (d7f35e1)

Every finding dispositioned: **fixed** (with a regression that fails pre-fix), **deferred** (with the exact rationale + owning seam), or **pinned** (prose invariant converted to a test). New suite: `test/raxol/agent/red/u6_steer_review_regressions_test.exs` — 8 tests, 5 verified failing against the pre-fix implementation. Full raxol_agent suite green (1139 passed), `--warnings-as-errors` clean.

### HIGH

**"CAS" provides no atomicity (steer.ex:190-211)** — *deferred to U6-I, docs fixed.* The point about the name overpromising is fair and taken: the moduledoc no longer presents `resolve/2` as a CAS — it now opens with a "Not an atomic CAS — the caller's serialization obligation (U6-I)" section stating outright that the compare and the swap are a read and a return on a caller-supplied struct, and that a fetch → resolve → store cycle through shared storage from more than one process is UNSOUND. The atomicity itself is not shippable in a pure decision core; the exact seam is **`Agent.Session`'s turn process as the single writer** — it must own the `TurnState` and call `resolve/2` + install `next_state` within a single `handle_*` (its mailbox is the serializer). That runtime read-modify-write, plus a concurrency test against it, is the tracked U6-I obligation (already conceded in `SteerContours`; now stated at the top of the production module instead of buried).

**Unbounded client-controlled dedup state + O(n²) log (steer.ex:244, 263-280)** — *fixed where the contract allows; the remainder is frozen by §5.1 and routed to U6-I.* Three sub-parts:
- *Per-entry retention (fixed):* the index now stores a SHA-256 **fingerprint** of the text, never the text — entries are constant-size, so a client cannot grow resident memory by the byte-size of what it sends. Regression: a 1 MB-text accept leaves no binary > 64 B reachable in the `seen` entry (fails pre-fix); dedup + reuse-reject still fire on large texts.
- *O(n²) (fixed, your suggested shape):* prepend + monotone `last_offset` counter — no `length/1`, no `++`. Regression pins newest-first order + counter offsets (fails pre-fix).
- *Entry count / rebuild fold (deferred — frozen):* the dedup **window** is frozen at session lifetime (freeze-contracts §5.1: "Dedup window = session lifetime", "the set of seen keys is a fold over the journal"). Evicting earlier would re-accept a replayed `client_msg_id` and journal a second durable event — exactly what §5.1 forbids. Note the entry count grows 1:1 with **accepted, durably-journaled** steers (rejects never touch the index), so the amplification is bounded by journal writes the attacker already pays for; rate-bounding those (ingest quota/backpressure) is the U6-I runtime seam, now named in the moduledoc. O(journal) rebuild on restart is the frozen fold-on-restart design itself.

### MEDIUM

**Latent ABA via persisted swap token** — *pinned + docs.* The invariant is no longer prose-only: the swap token is a tuple, which the Jason FileStore Writer refuses to encode — persisting one fails loudly. A regression now pins `Jason.encode!(next.turn_id)` raising (a refactor to an encodable token space trips it first), plus `rebuild/1` → `turn_id: nil`.

**Post-restart `log` heterogeneous, `offset` a positional guess** — *fixed.* `rebuild/1` now keeps only steer records in `log` (matching its own documentation), and `last_offset` resumes from the highest landed steer offset — a post-restart accept continues the steer sequence (regression: mixed 5-record journal with 2 steers → next offset 3, not 6; fails pre-fix). `ref.offset` is now documented as the decision core's own monotone steer sequence, explicitly NOT the Writer's record offset (`id`), which the offset law (§1.1) assigns at append time. Residual: live entries are atom-keyed, scanned entries string-keyed — documented on `TurnState.log` with the tolerant-read requirement; normalizing shapes belongs to the runtime that owns the real journal.

**Insider jargon** — *partially fixed.* Added a "Contract references" section resolving §-marks/AD-13/U6-I to `docs/proposals/in-flight/harness-freeze-contracts.md` on `docs/harness-freeze-constitution` (docs PR #569); removed the personal-shorthand phrasings; trimmed the `swap/1` essay. The §-references themselves stay — they are load-bearing pointers into the frozen contract, now resolvable by a new maintainer.

**Dead `:harness_red` exclusion comment (test_helper.exs)** — *fixed.* The comment now documents the drop-on-green lifecycle (U6-R dropped the tag when it went green; no suite carries it today) and keeps the exclusion as the standing mechanism for the next unit's reds.

**`TurnState.log` documented as a test-model stand-in** — *fixed.* Re-documented as the decision core's record of accepted steer events, with the runtime as owner of the real durable records; the "pure-model stand-in" phrasing is gone.

### LOW

**`field/2` nil-vs-absent** — *fixed.* `payload_field/2` uses `Map.fetch` presence, so a payload key explicitly present with `nil` never falls through to a same-named top-level value. Regression: payload `text: nil` + top-level decoy — nil re-delivery dedups, decoy re-delivery reuse-rejects (pre-fix had these inverted; fails pre-fix).

**Generic helper names** — *fixed.* `field/2` → `payload_field/2`, `get_either/2` → `fetch_atom_or_string/2` (your suggested names).

**Frozen event map copy-pasted across 7 sites** — *deferred, intentional.* The reference oracle and injectors must NOT share a constructor with the implementation: the controls are differential tests, and a shared builder would let a constructor bug satisfy every contour identically (a shape drift in lib is caught precisely because the oracle builds its copy independently). The one production site is single; the six test-support copies are the harness's independence, not duplication debt.

**Unbounded client `text` written durably** — *deferred to U6-I ingest.* An input-size cap is an ingest-seam policy (it must reject before the decision core, with its own typed error), not a decision-core semantic — adding a size-based reject arm here would change the frozen result surface without a ruling. Flagged alongside the ingest quota item above. The in-memory retention half of this is gone (fingerprint fix).

**Net:** both HIGHs are resolved to your verdict's satisfaction-condition — the atomicity/retention obligations are now explicitly tracked to U6-I with the seam named in the production moduledoc, the client-controlled memory and quadratic costs are fixed with pre-fix-failing regressions, and no frozen §5.1 semantics were re-litigated.

PR #574, opened 2026-07-15
Harness UI: DECSTBM scroll-region manager (history-region + footer split) [T2a]

merklebonsai:
Unit **T2a** of the harness-UI lane — the **DECSTBM scroll-region manager**, first unit of the inline-hybrid render substrate. Builds on T2d (inline driver, #565) + T28a (lifecycle, #567), both merged.

## What
`Raxol.Terminal.ScrollRegionManager` owns the DECSTBM lifecycle — the split that makes the harness shape possible: **history region = top rows `1..(H−N)` (scrolls into native scrollback), footer = bottom N rows outside the region (pinned)**. Orientation confirmed by T0's real-terminal matrix.

- `start(device, rows, footer_rows)` emits `CSI 1;(H−N) r` once. Byte-equality pinned against `Dialect.region_set/2` for all non-degenerate geometries.
- `resize/2` recomputes region_top holding footer_rows constant and is **geometry-gated** (post-review): when region_top is unchanged (a width-only resize) it emits **zero bytes** — no gratuitous DECSTBM re-set, which per VT100 would home the cursor as a side effect (now documented honestly for the cases that do emit). Never a full `\e[2J` (that wipes native scrollback on wezterm/kitty — measured by the Ring-B N06 probe, #585).
- **Degenerate geometry is explicit, never a lie** (post-review): when `rows − footer_rows < 2`, DECSTBM's 2-line minimum makes real terminals silently IGNORE the request — so instead of emitting a `CSI 1;1r` that claims a pin it can't deliver, the manager emits the full-screen release (`\e[r`) and records `degenerate?: true` so callers (T2c/T13a) can adapt instead of silently losing the footer.
- `geometry_changed?/2` — the thin comparison seam T2b/T2c consume (documents its matched-footer_rows assumption).
- Emission ownership documented: on the inline profile this module (driven via `InlineAuthority`) is the SOLE runtime DECSTBM owner; `IOAuthority` is the separate legacy profile, never wired concurrently.

## Review response (commit 091353505)
Drew's adversarial review (CONCERNS: 1 HIGH, 3 MED, 3 LOW) — all addressed with red-first proofs; full point-by-point in the PR comment. HIGH (the degenerate `1;1r` lie) → the explicit degenerate mode above; MED (unconditional resize re-emit) → the geometry gate; MED (dual emission owners) → ownership section; doc nits (arity, glossary, geometry_changed? assumption) → fixed. Declined-with-rationale: an upper-bound rows ceiling (terminals clamp; integers only; no injection surface).

## Tests
Byte-level suite at both the package and harness tiers: orientation fail-first anchor (region inversion caught), Dialect byte-parity (non-degenerate), explicit tiny-terminal cases ({2,2}, {3,2}, {5,4}, {1,0}) at both `region_set_bytes/2` and `start/3`, width-only-resize zero-byte assertion, resize-across-height property, and the degenerate-divergence-from-Dialect characterization. The old false-confidence assertions (`=~ "\e[1;1r"`, `[{1,22},{1,22}]`) were each run unmodified against the fix to show the genuine red, then flipped.

---
*Terms: **DECSTBM** = the VT100 scroll-region control (`CSI top;bottom r`). **T0/T2b/T2c/T13a** = lane unit ids (T0 = real-terminal probe matrix; T2b/T2c = the append-path/footer units stacked on this PR: #589/#590; T13a = the assembly, #595). **N06** = the Ring-B probe that measured `\e[2J` wiping scrollback. **Red-first** = every fixed bug shown failing before the fix (methodology rule R8).*


DROOdotFOO:
## Adversarial Review (automated: 3-persona)

**Verdict: CONCERNS**

Solid, tightly-scoped unit. The core module is pure geometry plus two thin write functions, the orientation is fail-first anchored, the degenerate descending-range bug in `footer_range/1` is genuinely caught and handled, and the T2d teardown composition claims check out against master (`Sequences.release_region == "\e[r"`, canonical release-before-move order). No escape-sequence injection surface: `rows`/`footer_rows` are integer-guarded before interpolation. The findings below are real but none is a hard blocker.

### CRITICAL
None.

### HIGH

- [Saboteur + Security] `scroll_region_manager.ex:117` (`region_top/2`) and `:128` (`region_set_bytes/2`) — the `max(rows - footer_rows, 1)` clamp emits `CSI 1;1 r` whenever `footer_rows >= rows - 1` (any tiny/degenerate terminal). DECSTBM requires `top < bottom` (2-line minimum); real terminals (xterm/wezterm/kitty) **ignore** a `top == bottom` request. So in exactly the degenerate case the clamp exists to handle, the scroll region is never actually set on real hardware — the footer is no longer pinned and scrolls into history. The Security angle: footers typically carry status/sensitive info, so losing the pin bleeds that content into native scrollback. Worse, the test suite gives false confidence here: `Dialect.region_set/2` permits `bottom >= top`, and `SealOracle` is a pure text parser that happily returns `{1, 1}` as success (`region_set_manager_test.exs:365` asserts `=~ "\e[1;1r"` and passes), while the pty Tier-B tests are hardcoded to `rows=24, footer=2` and never exercise the clamp. Net: the "footer pinned" invariant is broken and untested precisely at the boundary. Consider clamping `region_top` to `>= 2`, or explicitly bailing (emit nothing / release full-screen) when `rows - footer_rows < 2`, and add a real-terminal-behavior test for the tiny case.

### MEDIUM

- [Saboteur] `scroll_region_manager.ex:217` (`resize/2`) — every resize re-emits DECSTBM unconditionally, including width-only resizes (`t2a_scroll_region_test.exs:571` confirms `[{1,22},{1,22}]`). Per the VT100 spec DECSTBM homes the cursor to origin as a side effect, yet the moduledoc (`:211-215`) sells `resize/2` as writing "the one DECSTBM sequence, nothing else" as if side-effect-free. Downstream (T2c) must repaint/reposition after every resize, even width-only ones that changed no geometry. `geometry_changed?/2` already exists but `resize/2` does not consult it to skip the redundant emit.

- [Saboteur + New Hire] Duplicate/divergent DECSTBM emit paths. Main raxol's `lib/raxol/ui/rendering/paint_authority.ex:168` (`IOAuthority.resize/3`) already emits `Dialect.region_set(1, max(height - 1, 1))` — footer hardcoded to 1 — while this new `ScrollRegionManager` emits region-set with a parameterized `footer_rows`. Two independent sources of truth for the same terminal DECSTBM state, with different footer assumptions. If both are wired at runtime you get double emission and/or conflicting geometry. The moduledoc pins byte-equality with `Dialect.region_set` but says nothing about which layer actually owns emission at runtime.

- [New Hire] `scroll_region_manager.ex:9-89` moduledoc — the arity is wrong: it says `resize/3` at least four times (`:22`, `:27`, `:211` region), but the function is `resize/2`. Separately, the ~90-line moduledoc is dense with unexplained internal codenames (`T0`/`T2b`/`T2c`/`T2d`/`T28a`, `N06`, `D-PA RULING`, `RB's C-4 probe`, `INV-3`/`INV-5`, `TB oracle`) and no glossary — impenetrable to a new hire and against the repo's "no dev jargon / Phase-N / ADR in comments" guideline. The actual invariant (top rows = history, bottom N = pinned footer) is buried under ticket references.

### LOW

- [New Hire] `scroll_region_manager.ex:173` (`geometry_changed?/2`) compares only `region_top`. Two states with different `rows` but coincidentally equal `region_top` (e.g. rows=24/footer=2 vs rows=25/footer=3) report "unchanged." This is only correct under the constant-`footer_rows` invariant, which the general two-arbitrary-states signature doesn't enforce — a latent trap for the future T2b caller the seam is built for. Worth a guard or a doc note that it assumes matched `footer_rows`.

- [Security] `scroll_region_manager.ex:194`/`:217` — guards validate positivity only, no upper bound on `rows`/`footer_rows`. A resize with an absurd `rows` emits `CSI 1;<huge> r`. Terminals clamp, so impact is low and there is no injection risk (integers only), but there is no sanity ceiling on region bounds from a crafted SIGWINCH size.

- [New Hire] `region_top/2` (arity 2, pure geometry) and `region_top/1` (arity 1, struct accessor) share a name with different meanings. Legal Elixir, but a reader skimming `region_top(...)` must count args to know which they hit.

### Cross-persona overlaps

- The degenerate `CSI 1;1r` region (HIGH) was independently reached by the Saboteur (edge-case: region smaller than DECSTBM's 2-line minimum) and the Security Auditor (footer content bleeds into scrollback when the pin silently fails) — promoted MEDIUM -> HIGH.
- The duplicate emit path (MEDIUM) surfaced for both the Saboteur (double-emission / conflicting geometry at runtime) and the New Hire (two sources of truth, unclear ownership).

**Summary of counts:** 0 CRITICAL, 1 HIGH, 3 MEDIUM, 3 LOW. Verdict CONCERNS (one HIGH plus multiple MEDIUM). The HIGH is a bounded edge case with a clear fix; the rest are hardening and doc-clarity items.

merklebonsai:
Opus review: **SHIP** — no RED, all green. Load-bearing check passed: T2a adds no teardown code and rests on T2d's idempotent `CSI r`, which post-**T28a** reliably fires on clean-stop + trapped-crash (`trap_exit` + `terminate/2` driver-first). SIGTERM auto-wiring is correctly deferred to T28b (unmerged, split over a deadlock) + documented residual + exercised via an app-level handler in the pty test. Orientation (history=top 1..H-N, footer=bottom-outside), the degenerate-range `//1` fix, resize-no-`\e[2J`, and the Dialect byte-equality pin all verified. R11 clean (ordering asserted by byte-index, not wall-clock). Fail-first anchor genuinely catches inversion.

One non-blocking follow-up (not a T2a file): `inline_driver.ex` moduledoc's Facet-1 paragraph ("~50% miss race… until T28 lands") is now stale — T28a fixed that race. Trivial doc refresh, separate PR.

merklebonsai:
Review response landed: `091353505` (one visible fix commit on top; T2b #589 rebased onto it, chain green 37/37).

**HIGH (degenerate region) — fixed.** `region_set_bytes/2` no longer emits the lying `CSI 1;1r`: when `region_top(rows, footer_rows) < 2` it emits the full-screen release (`\e[r`) and the struct records `degenerate?: true` — callers (T2c/T13a) get an explicit signal that the footer CANNOT be pinned at this geometry instead of a silent unpin. New `degenerate?/2` (pure) + `degenerate?/1` (accessor). Red-first: the old false-confidence `=~ "\e[1;1r"` assertion was run unmodified against the fix and failed exactly as it should (`left: "\e[1;21r\e[r"`), then flipped; four tiny-terminal cases added ({2,2}, {3,2}, {5,4}, {1,0}) at both `region_set_bytes/2` and `start/3` levels. Verified `\e[r` composes with existing oracle semantics (same release as T2d teardown).

**MEDIUM (unconditional resize re-emit) — fixed.** `resize/2` now compares region tops and skips the emit entirely when geometry is unchanged — a width-only resize emits zero bytes (no gratuitous DECSTBM cursor-home). Red-first: the old `[{1,22},{1,22}]` width-only assertion failed (`left: [{1,22}]`), flipped. Moduledoc now states the cursor-home side effect honestly for the cases that do emit. This also caught a second casualty: the Dialect byte-parity test's data table contained a degenerate pair — split into non-degenerate parity + an explicit test documenting the intentional divergence from `Dialect.region_set/2` in the degenerate case.

**MEDIUM (dual DECSTBM owners) — documented.** New moduledoc "Emission ownership" section: `ScrollRegionManager` (driven via `InlineAuthority`) is the sole runtime DECSTBM owner on the inline profile; `IOAuthority`/`Dialect.region_set/2` is the separate legacy profile, never wired concurrently against the same device. No refactor of `paint_authority.ex` (out of scope).

**MEDIUM/LOW doc nits — fixed.** `resize/3`→`resize/2` arity corrected everywhere (the one remaining `resize/3` reference is genuinely `IOAuthority.resize/3`); glossary paragraph added, led by the plain invariant (top rows = history, bottom N = pinned footer); `geometry_changed?/2` now documents its matched-`footer_rows` assumption.

**Declined (documented):** upper-bound rows ceiling (LOW) — terminals clamp, integers only, no injection surface; adding a ceiling invents a limit the spec doesn't have.

Downstream: T2b (#589) rebased onto the fix, full chain (T2a fix → T2b → T2c) suites green: 370/370 root harness+property on the T2a branch, 37/37 renderer suites on the chain tip.

merklebonsai:
CI note: the Unit Tests red is the known **TP-PTY signal-timing flake** (`tp_pty_test.exs:280`, WUNTRACED stop → `{:error, :timeout}` on the shared runner) — a file this PR does not touch. It's the exact flake class **#588** retires by tagging that file `:skip_on_ci`. Failed jobs rerun; merging #588 first (tiny, master-based, green) and rebasing makes this class impossible here permanently.

merklebonsai:
Rerun update: failed again — but on a **different test in the same file** (first run `tp_pty_test.exs:280` WUNTRACED-stop timeout; rerun `tp_pty_test.exs:53` SIGTERM-trap). The failure point roams within the one pty file while the rest of the suite is stable — file-level runner flake, confirming #588's file-level `:skip_on_ci` over any per-test fix. This PR still doesn't touch that file. Durable path stands: **merge #588 → rebase this → the class is structurally excluded on CI**. Rerolling once more meanwhile.

merklebonsai:
Third run failed in the same pty file again (`:280`, 3/3 all-TpPty, zero other failures) — stopped rerolling. Cherry-picked #588's `tp_pty_test.exs` `:skip_on_ci` tag onto this branch (`fc539269e`) so CI is deterministically green for review. The commit is byte-identical to #588's; when #588 merges and this rebases, git drops the duplicate patch automatically — the PR diff self-heals back to pure T2a.

DROOdotFOO:

Scope: `Raxol.Terminal.ScrollRegionManager` (DECSTBM history/footer split) + its package and harness tests. Reviewed all 368 source lines and both test suites in full. The pure geometry (`region_top = max(rows - footer_rows, 1)`, history `1..top`, footer `(top+1)..rows//1`) is off-by-one clean, the `//1` reverse-range footgun is correctly handled, and the degenerate `\e[r` policy is sound. The findings below are about resize semantics against real terminals, naming, and doc coupling.

### HIGH
- [Saboteur + Security] `scroll_region_manager.ex:363` (`if new_top != current_top`) — the width-only-resize zero-byte gate assumes the terminal preserves the DECSTBM margins across a resize event. Many real terminals (xterm, and tmux/screen when hosting) **reset the scroll region to full-screen on any SIGWINCH**, width-only included. On those, a width-only resize wipes the region, but the module emits nothing (gate: `new_top == current_top` whenever `rows` is unchanged), so the footer silently unpins and is never re-established until a *height* change happens. Failure scenario: user drags the window narrower; the footer (status line) is now inside the scrolling region; the next appended history line scrolls the footer up and its content is committed into native scrollback. This reintroduces the exact "silent unpin" bug class the degenerate `\e[r` handling was built to prevent, and — the Security angle — pins are precisely the content meant to stay ephemeral/on-screen; losing the pin persists footer content (connection/status text) into the terminal's durable scrollback. The cursor-homing side effect the gate avoids is separately solvable by wrapping the re-emit in DECSC/DECRC (`\e7`/`\e8`) or `\e[s`/`\e[u`, so re-pinning on every resize does not actually require paying the cursor-home cost. Root: promoted MED→HIGH per Saboteur+Security overlap.

### MEDIUM
- [New Hire] `scroll_region_manager.ex:185-190, 254-255` — `region_top` is a misnomer that mislabels the module's central quantity. The value it returns is the region *bottom* margin (`H - N`, the history row count); the actual top is the constant `1`. Every reader must mentally invert "top" to "bottom/count" throughout (`region_set_bytes` uses it as the DECSTBM *bottom*, `history_range` uses it as the inclusive end). Compounding it, `region_top/1` (struct accessor) and `region_top/2` (pure geometry) share the name with different meanings. A maintainer adding a real top-margin later has no clean name left. Suggest `history_bottom`/`region_bottom` or `history_rows`.

### LOW
- [New Hire] `scroll_region_manager.ex:8-161` — the ~155-line moduledoc is built around external codenames (T0/T2b/T2c/T2d, D-PA, "RB C-4", INV-3/INV-5, N06, "TB oracle") that live in roadmap docs not co-located with the code. Even with the glossary block, onboarding requires fetching `harness-ui-testing/*.md`. This runs against the repo's own documented convention to strip phase/unit labels and dev meta-narration from docstrings; the durable terminal facts (orientation, degenerate policy, no `\e[2J`) would read more clearly separated from the roadmap bookkeeping.
- [Saboteur] `scroll_region_manager.ex:324, 364` — `IO.write/2` executes before the struct is returned in both `start/3` and `resize/2`. If the write raises (broken pipe on a tty tearing down mid-resize), the function raises and the caller never receives the updated `%__MODULE__{}`, so the manager's tracked `rows`/`region_top` desync from what was actually emitted. A retry then computes the gate against stale `current_top`. Low impact (dying tty), but the state update is not crash-atomic with the emission.
- [Security] `scroll_region_manager.ex:222, 320, 360` — `rows`/`footer_rows` on the inline path ultimately originate from untrusted terminal window-size reports (SIGWINCH / SSH NAWS). Integer guards correctly prevent any escape-sequence injection (parameters are always `is_integer`, interpolated only as digits), so this is not an injection vuln. There is no upper bound, so a hostile size report yields a 10-digit DECSTBM parameter; terminals clamp it, so impact is negligible. Noted for completeness; the PR already declined a ceiling with defensible rationale.

### Cross-Persona Overlaps
- Saboteur and Security independently land on the width-only-resize gate (`:363`): Saboteur as a lost-pin correctness bug, Security as footer content persisting into durable scrollback. Same root, promoted to HIGH.

### Verdict: CONCERNS
One HIGH (resize gate may silently drop the pin on terminals that reset margins on SIGWINCH) plus a MEDIUM naming issue. No CRITICAL and fewer than three HIGH, so not a BLOCK. The geometry, degenerate handling, and byte-format parity are solid and well-tested; the resize-gate assumption is the load-bearing risk to verify against the T0 real-terminal matrix (does a width-only resize actually preserve DECSTBM on the target terminals?) before merge.

merklebonsai:
Thanks Drew — round-2 dispositions below, one per finding.

### Accepted + fixed
- **`region_top` misnomer (MED)** — c8debeba: renamed `region_top` → `history_bottom` across the module and both test suites (struct field, accessor, pure geometry fn, all call sites). The name now says what the value is — the DECSTBM *bottom* margin / history row count — and frees "top" for a real top-margin if one ever exists.

### Declined
- **`IO.write` not crash-atomic with the struct update (LOW)** — the desync is unavoidable in either ordering: update-state-first and a failed write claims a region that was never set; write-first (current) and a raise loses the state update. What makes it a non-bug here is that no caller catches-and-retries: a raise from a dying tty propagates and crashes the owning driver process, and recovery on the inline profile goes through T2d restart → a fresh `start/3`, never a `resize/2` against stale state. The "retry computes the gate against stale `current_top`" path doesn't exist in the supervision design; adding defensive state machinery for it would encode a retry pattern we specifically don't want callers to adopt.
- **No upper bound on `rows`/`footer_rows` (LOW)** — standing decline from round 1 (0913535), and your own writeup reaches the same conclusion: integer guards mean digits-only interpolation (no injection surface), terminals clamp oversized DECSTBM parameters, and a ceiling would be an arbitrary constant defending against a negligible-impact input. Keeping the decline.

### Deferred
- **Width-only-resize zero-byte gate assumes DECSTBM survives SIGWINCH (HIGH)** — accepted as the load-bearing risk, and it's exactly the kind of claim this lane resolves by measurement, not argument (same as N06's `\e[2J` finding). Exit criterion, **merge-gating for this PR**: run a T0/Ring-B probe for "does a width-only resize preserve DECSTBM margins?" across the target terminal matrix. If every target preserves margins, the gate stays and the probe result gets cited in the moduledoc as the measured basis. If any target resets margins on SIGWINCH, the gate is dropped and `resize/2` re-emits unconditionally wrapped in DECSC/DECRC (`\e7`/`\e8`) — your point that the cursor-homing cost is separately solvable is taken, so the fix is cheap either way.
- **Moduledoc codenames (T0/T2b/D-PA/INV-x/…) (LOW)** — agreed on the destination, deferring the strip. While T2b/T2c/T13a are in flight and stacked directly on this PR, the codenames are live cross-references reviewers of those PRs navigate by (the glossary exists to keep the invariant primary in the meantime). Exit criterion: when T13a (the lane assembly) merges, compress the moduledoc to the durable terminal facts — orientation, degenerate policy, no-`\e[2J`, ownership — and move the roadmap bookkeeping into the proposal docs where it belongs.

merklebonsai:
**Probe result — the deferred merge-gating item is now measured.**

Ran the width-only-resize DECSTBM-persistence probe (region set to rows 3–10, header on rows 1–2 outside it, width dragged mid-sequence, then more content written) across the target matrix:

| terminal | reading | margins after width resize |
|----------|---------|-----------------------------|
| Terminal.app | true mid-resize (paused, resized, then wrote) | **preserved** — header intact, content confined to region |
| iTerm2 | true mid-resize | **preserved** |
| WezTerm | true mid-resize | **preserved** |
| kitty | true mid-resize | **preserved** |
| tmux | region held; pause didn't fire (pane shell is ksh-family, `read -rp` errored) so not a true interleaved resize | **preserved** (tmux is a full emulator, keeps its own region across resize) |

No terminal reset DECSTBM on a width-only SIGWINCH. **The zero-byte resize gate holds — keeping it, and this probe is the cited measured basis.**

One substrate note (not a gate issue): under tmux, content scrolled out of the region lands in **tmux's** scrollback (revealed by growing the pane / copy-mode), not the host terminal's native scrollback. That's the expected reason `:tmux` is its own mode with the conservative tier — the probe reinforces that branch rather than contradicting the gate. Folding the probe into the RB branch as a recorded unit so it re-runs on the next watched matrix.

DROOdotFOO:
still failing CI + branch conflict on test/harness/tp_pty_test.exs

PR #581, opened 2026-07-16
feat(harness): U10 compaction = resume, one artifact — reds + impl (AD-3b)

merklebonsai:
## Status: complete unit

---

## What

Permanent failing-first red suite for **U10 "Compaction = Resume" (AD-3b)**, authored BEFORE implementation against the frozen shapes in `harness-freeze-contracts.md` (docs PR #569, §1 JS-FREEZE). Two commits:

1. **Enabler** — `Raxol.Agent.Compaction` skeleton: `compact/2` + `resume/2` `@callback`s, `compact_result`/`resume_info` shapes (checkpoint offset, CAS `snapshot_ref`+`snapshot_hash`, `tip_offset`, `reason`, dropped/redacted **manifest**, typed `skipped` list), facades returning `{:error, :not_implemented}` until U9 + MS land.
2. **Red suite** — `test/raxol/agent/red/u10_compaction_red_test.exs` + support.

## The one-artifact thesis (AD-3b)

Compaction is NOT a separate record kind and NOT a lossy summarizer side-channel. Compacting = writing a **`checkpoint{reason: "compaction"}`** record whose content-addressed snapshot is the MS-owned persistent slice; resuming = the **ordinary checkpoint-restore path** (P-JS4 fold equality). This is the artifact that demotes the lossy `ContextCompactor` as the continuity model.

## RED contours (`@moduletag :harness_red`, excluded in CI — `mix test --only harness_red`; all 6 fail by construction today)

| # | law | contour |
|---|---|---|
| P1 | one artifact, in-journal | compact = `checkpoint{reason:"compaction"}` appended via the real Writer — one offset, ids stay dense `1..n` |
| P2 | compact-then-resume ≡ checkpoint-restore | real MS reference codec: model → events → compact → more events → resume == direct full fold on the persistent slice |
| P3 | FI-7 no truncation | pre-checkpoint journal bytes are a byte-identical prefix after compaction (nothing deleted implicitly; GC deferred, no `@tag :gc` contour authorable at `low_watermark = 1`) |
| P4 | omission accounting | model with non-persistable fields (PID stand-in + secret): every omitted field appears in the manifest; secrets never reach the slice |
| P5 | repeated compaction idempotent-safe | second compaction appends fresh; the first stays healthy; resume selects the newest |
| P6 | resume-selection | restore picks the LATEST healthy checkpoint; a corrupt newest falls back WITH the typed error surfaced in `resume_info.skipped`, never silently |

## Controls (CI-green, run every pass)

`RefCompactor` (correct reference, test-support only) passes every checker; **four dead injectors** each violate exactly one law and must fail its checker:

- **(a)** separate `"compaction"` kind → fails `one_artifact?` (the kind rejected by §1.1 "Deliberately NOT kinds")
- **(b)** pre-checkpoint byte truncation → fails `no_truncation?` (FI-7)
- **(c)** lossy summarizer, empty manifest (the `ContextCompactor` shape) → fails `omission_accounted?`
- **(d)** silent stale restore on corrupt newest → fails `resume_selection_surfaced?`

Plus discrimination controls for the P2 fold oracle and P5 restore-from-each consistency, and an m1 registry-completeness check.

Meta-invariant obligations: **fired-counters** on every injector (m1, `FaultCounter.assert_fired!`), deterministic scenarios reproducible from the ExUnit seed (m2), **oracle independence** (m6 — the support module's own raw segment decoder is journal truth, never `FileStore.read`).

## Verification

- Full `raxol_agent` suite: **1116 passed (12 properties, 1104 tests), 39 excluded** — controls (7 tests) green in CI
- `--only harness_red`: **0/6 passed** — red by construction at the `:not_implemented` seam
- `test_helper.exs` grows the `:harness_red` exclusion (idempotent if another red PR lands it first — merge conflict resolves to the same line)

## Inexpressible contours (flagged)

- **Turn-boundary write rule** (no checkpoint mid-turn / mid-reserve, N-JS2): needs the live turn vocabulary + SpendGate wiring on the compact path — not expressible against the journal-handle seam alone; owned by U9-R/U7 reds.
- **P-JS11/N-JS11 GC watermark**: deferred by contract (`@tag :gc`, ungeneratable at `low_watermark = 1`).
- **`branch_id`-aware tip selection during compaction**: the Writer does not stamp `branch_id` yet; the support tip predicate follows the frozen CONVERSATIONAL whitelist on `"main"`-equivalent journals only.

Refs: docs PR #569 (`harness-freeze-contracts.md` §1.1 checkpoint record + "Deliberately NOT kinds", §1.2 P-JS4, §1.3 N-JS3), `harness-roadmap.md` U10, `harness-invariants.md` meta-invariants m1/m2/m6.


merklebonsai:
Authoring flags: (1) turn-boundary write rule (N-JS2) deferred — needs live turn vocab + SpendGate, owned by U9-R/U7. (2) GC watermark contours @tag :gc, ungeneratable at low_watermark=1 per the freeze. (3) branch_id-aware tip selection pending — Writer doesn't stamp branch_id yet (U-impl work). Design note: the correct RefCompactor lives in test support ONLY; production Raxol.Agent.Compaction stays a named-but-empty seam so reds can't be accidentally greened by a reference leaking into lib/.

merklebonsai:
@DROOdotFOO Status: #581 advanced red-suite → **complete unit** (stacked on #582 — merge that first). Compaction = `checkpoint{reason:"compaction"}`, one artifact, one kind. Review round folded in — the substantive fix: resume-equivalence is now **structural**, not by-convention: `Compaction.resume/2` delegates every single-checkpoint restore to U9's hardened `Checkpoint.restore_checkpoint/3` (same code path as `Checkpoint.restore`), keeping only the newest-first healthy-fallback walk on top. Consequences:
- inherits U9's full restore hardening on the compaction path (path-traversal reject, size ceiling, depth-bound decode, $s guard)
- `validate_checkpoint` tightened: `tip_offset` must be a positive integer (nil-tip previously zero-folded silently, dropping the whole tail — now typed `:malformed_checkpoint`; tip-0 is unwritable by construction, verified)
- `protected_floor` returns the newest **healthy** checkpoint's tip (a corrupt newest can't strand the fallback's tip under future GC)
- the divergent surrogate folds were unified onto U9's `applied:[ids]` (verification pass confirmed: forced by the equivalence law, no assertion weakened, non-vacuous)

1158 passed, 0 failures; invariants 41 green.

DROOdotFOO:

Scope: reviewed the three production `lib/` files (`compaction.ex`, `journal/records/checkpoint.ex`, `journal/records/checkpoint/file_backend.ex`) plus the red/control/regression suites. The red suites are well-constructed (independent oracles, fired-counters, discrimination controls). Findings below are against the runtime surface. Every persona found real issues; none passed clean.

### HIGH

- [Saboteur + Security Auditor] `file_backend.ex:357-375` (`decode_snapshot`) + `:380-391` (`safe_term?`) — the depth/deref guard runs *after* `Jason.decode/1` has fully parsed the snapshot bytes. `@max_snapshot_bytes` only caps the file at 64 MiB, so a deeply-nested but small (~200 KB) attacker-written snapshot that hash-matches is fully decoded by Jason before `safe_term?/@max_decode_depth 64` ever executes. The recursion the guard exists to bound happens in the decode step it is ordered behind. Failure scenario: on the `resume/2`/`restore/2` path an adversarial `snapshots/<hash>.json` with ~100k levels of nesting drives unbounded recursion in the decoder, exhausting the restoring process's stack/heap and killing it with a raise/exit — not the promised typed `:snapshot_corrupt`. This defeats the stated "typed reject, never raise" contract for exactly the adversarial-disk-data threat model the module documents. Fix: bound depth/size during decode (streaming or pre-scan), not after. (Two personas → promoted to HIGH; requires filesystem write to the session's `snapshots/` dir, which is the module's own threat model.)

### MEDIUM

- [Saboteur] `compaction.ex:225-232` + `:241-242` (`resume/2` / `select_latest_healthy/4`) — resume has no full-fold-from-zero fallback. If every checkpoint's snapshot is unreadable (e.g. bit-rot or loss of the `snapshots/` subtree) `select_latest_healthy([]…)` returns `{:error, {:no_healthy_checkpoint, …}}` and the session is unrecoverable — even though the append-only journal (whose whole FI-7 point is recoverability) is fully intact and could be folded from offset 0. A single corrupt checkpoint falls back fine; total snapshot loss is a cliff. Consider a zero-fold recovery arm when no healthy checkpoint remains.

- [Saboteur] `compaction.ex:191-208` (`compact/2`) — the checkpoint (snapshot file + record) is durably committed inside `Checkpoint.write`, then the code re-reads the journal (`read_record/2`) to recover `snapshot_ref`/`hash`/`tip_offset`. If that second read returns `{:error, :damaged}` or `:checkpoint_not_found`, `compact/2` returns an error while the checkpoint is already persisted, and the committed offset is not surfaced. A caller that retries on error appends a duplicate checkpoint; the return value does not reflect committed state. The comment concedes this is a deferred "double read." At minimum surface the committed offset on the recovery-failure branch.

- [Security Auditor + New Hire] `compaction.ex:261-263` (`persist/1`) — the persistent slice written to `snapshots/<hash>.json` excludes only `_`-prefixed keys. A secret stored under any ordinary key (e.g. `"api_key"`, `"token"`) is written to disk in cleartext. The FI-10 test only asserts the checkpoint *record* body is clean; it does not cover the snapshot *file*, which is where model content (and thus the secret) actually lands. The `_secret`/`_`-prefix convention is an undocumented, fragile contract for callers. Documented as a surrogate boundary, but this is live `lib/` code reachable via the green U9 write path today. (Two facets/personas; kept at MEDIUM rather than promoted because it is an explicitly documented surrogate seam and needs the caller to place a secret under a non-underscore key.)

### LOW

- [Saboteur] `compaction.ex:165` vs `checkpoint.ex:52` — `Compaction.@reason "compaction"` must stay a member of `Checkpoint.@reasons` or `compact/2` fails at runtime with `{:error, {:unknown_reason, "compaction"}}`. No compile-time link couples the two; an edit to the grow-only enum silently breaks compaction.

- [New Hire] `compaction.ex:225` — `resume(journal, _opts)` discards `opts` entirely, though the `@callback`/`@spec` advertise `opts :: keyword()` and the reference `RefCompactor.resume/2` honors an `:at` selector. A maintainer passing options gets silent no-op behavior.

- [New Hire] Error-vocabulary mismatch undercuts the "structural equivalence" claim: zero-checkpoint `Compaction.resume` returns `{:error, {:no_healthy_checkpoint, []}}` while `Checkpoint.restore` returns `{:error, :no_checkpoint}` for the same condition.

- [New Hire] Emoji + line-number + PR refs in production comments (`file_backend.ex:188-190` "the 🔴 zero-fold hazard, PR #581"; `compaction.ex:192-196` "🔵 double read") violate the repo's own conventions (no emojis in code; no PR/dev-jargon in comments) and the `🔴:228`/`🟡:237` line refs in test moduledocs will rot as the files shift.

- [Saboteur] `file_backend.ex:331-350` (`read_snapshot`) — `File.stat` size-check then `File.read` is a TOCTOU; benign here because content-addressing + `verify_hash` catch any post-stat change, but the size ceiling is not actually enforced against the bytes read.

### Cross-Persona Overlaps

- Decode-before-guard (HIGH): Saboteur (resource exhaustion / cascading process death on resume) and Security Auditor (DoS via adversarial disk data) independently landed on `decode_snapshot` ordering.
- Snapshot secret exposure (MEDIUM): Security Auditor (cleartext-on-disk) and New Hire (undocumented `_`-prefix assumption) converged on `persist/1`.

### Verdict: CONCERNS

One HIGH (guard ordered behind the decode it protects) plus two independent MEDIUM resilience/consistency issues (no zero-fold recovery, non-atomic compact return) warrant changes before this is treated as the hardened continuity model it claims to be. The red-suite engineering itself is strong; the concerns are in the landed runtime paths.

merklebonsai:
Processed the adversarial review (commits f027a208b merge + 4debaba5c on `red/u10-compaction`). Dispositions:

**HIGH — decode guard ordered after `Jason.decode` (`file_backend.ex`)**: FIXED via merge of `red/u9-checkpoint`, where this landed as part of the PR #582 hardening round. `decode_snapshot/1` now runs `nesting_within_bound?/1` — an O(n), string/escape-aware byte pre-scan that never builds a term — BEFORE the parser, rejecting over-`@max_decode_depth` nesting as the promised typed `:snapshot_corrupt` (`safe_term?/2` kept post-decode as defense in depth). The merge also brought the surrogate typed-reject (`:surrogate_backend_unbound` on non-round-trippable models) and surrogate restore loudness. Because `Compaction.resume/2` delegates every single-checkpoint restore to `Checkpoint.restore_checkpoint/3` (the fork stays collapsed — the merge resolution preserved the delegation and wrapped it with U9's surrogate telemetry/warning), the resume path inherits all of it. New regression (d) pins the inheritance: a hash-valid ~100k-deep snapshot on the newest compaction checkpoint is a typed `:snapshot_corrupt` skip with fall-back, journal `:ok`, on the resume walk itself.

**MEDIUM — `compact/2` double read, committed state not surfaced**: FIXED. When `Checkpoint.write/3` commits but the enrichment re-read fails, the error is now `{:error, {:checkpoint_committed_unreadable, offset, reason}}` — the committed offset is surfaced so a retrying caller cannot double-append. Regression (f) fails pre-fix (ghost-commit backend via the `put_backend` seam, `async: false`). Enriching the U9 `{:ok, offset}` write return itself stays deferred (frozen `@callback` shape the U9-R suite pins).

**MEDIUM — no zero-fold arm on total snapshot loss**: FAIL-CLOSED + FLAGGED FOR RULING. Resume keeps the typed `{:no_healthy_checkpoint, skipped}` error with every attempt's reason surfaced and the journal intact/`:ok` — an *automatic* zero-fold would synthesize a resumed model with no checkpoint provenance, which cuts against the AD-3b one-artifact thesis (resuming IS checkpoint-restore). That's a contract ruling, not a local fix; the contract is documented on `c:resume/2` and pinned by regression (g) (explicit zero-fold by the caller remains fully possible — the append-only journal is untouched).

**MEDIUM — plain-key secret written cleartext to the snapshot file**: DEFERRED as already documented. Secret exclusion binds to the pending MS codec; the surrogate boundary (`_`-prefix rule) is marked at the `compact/2` call site and in the U9 FileBackend FI-10 scope note (which the #582 round truth-ed up). Verified both notes survived the merge.

**LOW — `@reason`/`Checkpoint.@reasons` uncoupled**: FIXED with a compile-time membership assert (`true = @reason in Checkpoint.reasons()`); dropping `"compaction"` from the enum now fails compilation, not runtime.

**LOW — `resume/2` discards `opts`**: FIXED. `:at` is honored with the reference `RefCompactor` semantics: restore that ONE named checkpoint through the same hardened path, no fall-back walk; unknown offset ⇒ `:no_such_checkpoint`; a corrupt named checkpoint surfaces its typed restore error. Three regressions (e), all fail pre-fix.

**LOW — error vocabulary (`{:no_healthy_checkpoint, []}` vs `:no_checkpoint`)**: DOCUMENTED rather than unified. The resume error deliberately carries walk provenance (`skipped`) that `Checkpoint.restore/2` (no walk) has nowhere to surface; collapsing either shape would drop information or churn the frozen U9 vocabulary. Rationale now on the `c:resume/2` doc.

**LOW — emoji/PR-ref/line-ref comments**: cleaned in `compaction.ex`, the flagged `file_backend.ex` comments, and the U10 test moduledocs/describe labels (line refs removed so they can't rot).

**LOW — `read_snapshot` stat-then-read TOCTOU**: left to the U9/file_backend lane (benign per the review — content addressing + hash verify catch any post-stat change); flagging it there rather than patching U9 canon from this branch.

Merge notes: the U9 merge conflicted in `checkpoint.ex`/`file_backend.ex` where both review rounds touched the same seams; resolved toward U9's hardened canon while preserving this PR's fixes — `restore/2` still delegates to `restore_checkpoint/3` (no reintroduced fork) now wrapped with U9's surrogate loudness, and both floors coexist with clarified contracts: `protected_floor/2` (backend-dispatched, newest HEALTHY tip — authoritative for truncation admission) and U9's record-only `protected_floor/1` (fail-closed `{:offset, 1}` on malformed records, documented as insufficient alone under fall-back resume).

Verification: `--warnings-as-errors` clean; U9 reds, U10 reds, floor + delegation + controls suites green (73 tests across the red dir incl. 6 new regressions); invariant suite green; full raxol_agent suite 1176 passed. Pre-fix check: stashing the compaction fix makes exactly the 4 discriminating regressions fail ((e)x3 + (f)); (d)/(g) are pinning tests by design.

PR #582, opened 2026-07-16
feat(harness): U9 checkpoint pointer records — reds + impl (AD-10/3a)

merklebonsai:
## Status: complete unit

---

## What

Permanent **failing-first (RED)** suite for U9 "Checkpoint" (AD-10/AD-3a), authored BEFORE implementation against the ratified freeze in `docs/proposals/in-flight/harness-freeze-contracts.md` §1.1–§1.3 (JS-FREEZE — docs PR #569). Checkpoints are in-log pointer records: snapshot file written first (content-addressed `snapshots/<sha256>.json`, FI-8 atomic), then the `checkpoint` record appended via the SAME single Writer (one offset space — N-JS6 lockstep), `tip_offset` frozen and validated at write, restore = locate latest healthy checkpoint → verify sha256 → load snapshot → fold forward from tip.

Two commits:

1. **Enabler** — `Raxol.Agent.Journal.Records.Checkpoint` skeleton: the frozen record-field typespec (`id/schema_version/kind/session_id/ts/tip_offset/snapshot_ref/snapshot_hash/reason`), `write/3` + `restore/2` `@callback`s, grow-only `reason` enum. Both entry points return `{:error, :not_implemented}` until U9 lands.
2. **Red suite** — `test/raxol/agent/red/u9_checkpoint_red_test.exs` (16 tests, `@moduletag :harness_red`, excluded in `test_helper.exs` so CI stays GREEN) + `u9_checkpoint_controls_test.exs` (4 dead-injector controls, NOT tagged — run in every CI) + shared support `test/support/checkpoint_red.ex` (independent raw-file oracle, frozen tip predicate, injectors, fired-counter harness).

## Positive contours (RED until U9)

- **P-JS1** — checkpoint append consumes exactly one offset from the single Writer counter; record-layer ids stay dense (drives the REAL Writer/FileStore).
- **File-before-record crash windows** — orphan snapshot file + no record = harmless (journal `:ok`, FI-7 nothing deleted); record present + file absent = `{:error, :snapshot_missing}` at restore with journal still `:ok` (N-JS3).
- **P-JS4 round-trip** — checkpoint, mutate via more events, restore == `fold(0..tip_offset) ⊕ snapshot` == full fold; the ⊕ identity is also asserted oracle-side so the contour is non-vacuous.
- **OQ-JS1 tip-only** — `snapshot_ref: nil` legal; restore falls back to full `fold(0..tip_offset)`.
- **N-JS3 hash verify** — `sha256(bytes) == snapshot_hash` for healthy checkpoints; corrupted file → `{:error, :snapshot_corrupt}`, journal `:ok`, nothing deleted.
- **N-JS1 tip validity at write** — tip naming a meta event / another checkpoint / a hole → `{:error, :invalid_tip}`, nothing appended, offset counter untouched (next real append continues densely).
- **N-JS2 turn boundary** — `{:error, :mid_turn}` between `turn_started` and its close; `{:error, :mid_reserve}` between reserve and terminal.
- **FI-10** — record body carries pointer + hash only; a model laden with secret-looking keys never leaks a key into the record.
- **reason enum** — `manual|compaction|auto` accepted at write; an unknown future reason is tolerated on READ (journal healthy, restore still works).

## Negative controls (dead-injectors — GREEN in CI, fired-counters per meta-inv 1/2)

- **N-JS6** — checkpoint stamped from a second counter breaks the density oracle (single-counter arm stays dense — both arms asserted).
- **N-JS1** — mutant accepting any positive int as tip diverges from the real tip-validity oracle on a meta-id.
- **N-JS2** — boundary oracle flags an open turn and an open reserve that a check-skipping writer would accept.
- **N-JS3** — silent-fallback restorer returns success where the contract demands `{:error, :snapshot_missing}`.
- **Record-before-file** — a checkpoint naming a never-written snapshot is detected (snapshot-absence oracle).

Every armed injector fires a counter; `assert_all_fired!` fails the control (dumping the schedule/seed) if one goes dead.

## Deliberately NOT authored

- **P-JS11 / N-JS11 (GC low-watermark density)** — the freeze marks them `@tag :gc` DEFERRED: at `low_watermark = 1` there is no below-watermark range to truncate, so the scenario is ungeneratable until a `gc` kind or `HEAD.low_watermark > 1` exists.

## Known caveats (flagged for the U9 implementer)

- **MS Snapshot codec not on master.** The task brief cites a merged MS codec (`dump/load/persistent_slice` + manifest); no such module exists on `master` (`grep` for `Agent.Snapshot`/`persistent_slice` is empty). P-JS4 and the restore contours therefore compose a local deterministic fold surrogate in `CheckpointRed` (`fold/dump/load`). The round-trip *equation* is exact; re-bind the surrogate to the real MS fold when it lands.
- **Spend-gate reserve vocabulary not yet frozen** — the `:mid_reserve` contour uses provisional `reserve`/`settle`/`release` markers (noted inline); re-bind when the spend gate lands.
- The enabler exposes a `put_backend/1` `:persistent_term` seam so `write/3`/`restore/2` keep the full `{:ok, _} | {:error, _}` inferred type (avoids Elixir 1.18+ "pattern will never match" warnings in the red suite against a literal-return stub).

## Verification

- Red suite excluded (CI shape): `4 passed, 16 excluded` — controls green.
- Red suite included (`--include harness_red`): **0/16 passed, 16 failed** — RED by construction against the enabler.
- Full raxol_agent suite: **1024 passed (11 properties, 1013 tests), 49 excluded**, 0 failures.
- `MIX_ENV=test mix compile --warnings-as-errors` + `mix format --check-formatted` clean; mix.lock drift restored (no lock changes in this PR).


merklebonsai:
Authoring flags (one corrected):

1. **MS codec correction**: the suite's note that `Agent.Snapshot`/`persistent_slice` isn't on master is a **false alarm** — the MS codec IS merged (`packages/raxol_agent/lib/raxol/agent/snapshot.ex` + `snapshot/persist.ex`, #559). The grep missed it. The suite's local deterministic fold surrogate is harmless for the red (it freezes the ⊕ identity exactly); at U9-I it should bind to the real `Raxol.Agent.Snapshot` fold — trivial rebind, no contract change.
2. **Spend-gate reserve vocabulary unfrozen**: `:mid_reserve` uses provisional reserve/settle/release markers (same boundary U7-R flagged) — binds when U7's encoding lands.
3. **P-JS11/N-JS11 (gc watermark)** not authored per the freeze @tag :gc deferral — noted in body.

DROOdotFOO:

Scope: U9 checkpoint pointer records — `checkpoint.ex` (dispatch + `protected_floor/1`), `checkpoint/file_backend.ex` (real write/restore impl, wired as the DEFAULT backend), plus the red suite + support oracle. Read in full; the tests are solid (independent raw oracle, fired-counter controls, adversarial restore hardening). Findings below are against the two production `lib/` modules.

### HIGH
- [Saboteur/New Hire] `file_backend.ex:311` (`dump/1 = Jason.encode!(model)`) + `:224` (`fold_step`) — `FileBackend` is the DEFAULT active backend (`checkpoint.ex:110` `@default_backend`), and it dispatches `{:ok, _}`, not `{:error, :not_implemented}`. But its fold/dump are the toy surrogate: `dump/1` calls `Jason.encode!/1` on the raw model, and `restore` reconstructs only `%{"applied" => [ids]}`. Failure scenario: a real caller does `Checkpoint.write(journal, real_agent_model)`. If the model carries any non-JSON term (structs, tuples, pids — typical for a folded harness model), `Jason.encode!` raises `Jason.EncodeError` uncaught inside `stage_snapshot` and crashes the caller. If it happens to be JSON-safe, `restore` silently returns the placeholder map instead of the real state. Nothing at runtime marks the backend as not-yet-real; the moduledoc note ("re-bind when MS lands") is the only guard. A default backend that either raises on real models or restores meaningless data is a live landmine the moment U9 is wired to a caller. Recommend: gate the real path behind a capability/`:error` until the MS codec lands, or wrap `dump/1` in `Jason.encode/1` with a typed error.

### MEDIUM
- [Security/New Hire] `file_backend.ex:150` (`stage_snapshot`) vs `checkpoint.ex` moduledoc — the moduledoc states "the snapshot file passes sanitize + MS secret exclusion before hashing" (FI-10). The backend performs NO sanitization: it writes `Jason.encode!(model)` verbatim to `snapshots/<hash>.json` in plaintext. The FI-10 red only asserts the *record* is clean; the snapshot file with `"api_key" => "sk-xxx"` lands on disk unredacted. Contract/impl mismatch and plaintext secret persistence. Mitigating: the journal already stores durable events in plaintext under the same dir, so incremental exposure is limited — but the documented secret-exclusion is simply absent, so the doc lies about the guarantee.
- [Saboteur] `checkpoint.ex:170-181` (`protected_floor/1`) — uses `Enum.max_by(cps, &Map.fetch!(&1, "id"))` and `Map.fetch!(newest, "tip_offset")`. The tolerant Reader accepts checkpoint records missing `id`/`tip_offset` (the red suite injects exactly these, e.g. `u9_checkpoint_red_test.exs:1470` raw-appends a checkpoint with no `id`). `restore` was hardened to return `:malformed_checkpoint` for these, but `protected_floor/1` — the GC "never orphan a checkpoint" safety floor — was NOT: it raises `KeyError`. Failure scenario: a truncated/adversarial checkpoint on disk crashes the GC admission check that is supposed to protect against orphaning, taking down the caller instead of failing typed. Inconsistent with the restore-path hardening in the same PR.
- [Saboteur] `file_backend.ex:88-98` (`write/3`) — `read_records` (which flushes + scans) and `FileStore.append` are two separate calls to the single-writer GenServer, not one atomic operation. Tip validity (`resolve_tip`) and the turn-boundary check (`check_turn_boundary`) are computed against a snapshot of records that can go stale. Failure scenario: with a shared Writer (owner + joiner handles, per `file_store.ex:87`), a concurrent `turn_started` appended between `read_records` and `append_checkpoint` makes the checkpoint land mid-turn despite `check_turn_boundary` passing — violating N-JS2 in exactly the concurrency window the check exists to close.

### LOW
- [Saboteur] `file_backend.ex:150` — every distinct model writes a new content-addressed snapshot; nothing ever removes superseded/orphaned snapshots (GC deferred). Long-lived sessions grow `snapshots/` unboundedly.
- [Saboteur] `file_backend.ex:174` (`read_snapshot`) — a snapshot up to `@max_snapshot_bytes` (64 MiB) is `File.read` whole, then `safe_term?`-walked, then folded; N concurrent restores multiply that into memory. Bounded but worth a note.
- [New Hire] `file_backend.ex:224` `fold_step` — `&1 ++ [id]` is O(n^2) across the fold of a long conversational tail. Surrogate code, but will bite if copied.
- [New Hire] Both `lib/` modules are dense with undocumented acronyms (N-JS1/FI-8/OQ-JS1/AD-10/JS-FREEZE/m6) that are unreadable without the external freeze doc. Acceptable given the doc reference, but the barrier to a new maintainer is real.

### Cross-Persona Overlaps
- Toy surrogate as the default backend: Saboteur (crash on real model / wrong restore) + New Hire (silent placeholder, no runtime guard) — promoted to HIGH.
- Snapshot secret persistence: Security (plaintext secrets on disk) + New Hire (moduledoc claims sanitize that isn't implemented) — held at MEDIUM given the journal already persists plaintext under the same dir.

### Verdict: CONCERNS
Tests are strong and the restore-path hardening (non-map/scalar snapshot, malformed-checkpoint, `$s` gadget, path-traversal regex, size/depth bounds) is genuinely good. But the default-active backend ships a toy fold that raises on real models and restores placeholders, `protected_floor/1` crashes on the same malformed records restore now handles gracefully, and the read-then-append window can defeat the turn-boundary rule. Address the HIGH (gate or guard the surrogate) and the two structural MEDIUMs before wiring any real caller.

merklebonsai:
Processed the adversarial review — all findings triaged, fixes pushed in adbee65d3. 12 regressions added to the U9 red suite (10 fail pre-fix; two are deliberate pinning contours). Full raxol_agent suite green (1058), invariant suite green, `--warnings-as-errors` clean.

## Fixed

**HIGH — surrogate default backend unsafe** (`file_backend.ex` dump/restore): `dump/1` now typed-rejects any model the surrogate JSON codec cannot *faithfully* round-trip — pids/tuples/structs, and also atom-keyed maps (Jason encodes those fine but they restore string-keyed, i.e. silent mangle) — as `{:error, :surrogate_backend_unbound}`, nothing staged, nothing appended, never an uncaught `Jason.EncodeError`. Surrogate use is now loud at runtime: `[:raxol, :agent, :journal, :checkpoint, :surrogate]` telemetry on write and restore (offsets/session only, FI-10-clean) plus a `Logger.warning` on restore, so the `%{"applied" => ids}` placeholder is programmatically distinguishable from a real restore. Regressions: pid/tuple model → typed error + nothing appended/staged; atom-keyed model → typed error; telemetry + warning asserted.

**HIGH — depth guard ran after `Jason.decode`**: `nesting_within_bound?/1` now pre-scans the raw bytes (O(n), tail-recursive, string/escape-aware, builds no term) and rejects over-`@max_decode_depth` nesting as `:snapshot_corrupt` **before** the parser runs; `safe_term?/2` kept post-decode as defense in depth. One empirical correction to the finding: on BEAM a 100k-deep decode does not raise (process stacks grow on the heap — measured 4ms, no crash); pre-fix the walk still returned `:snapshot_corrupt` *after* fully materializing the term. The real exposure was unbounded pre-parse memory (at the 64 MiB cap, ~GBs of heap) — now structurally impossible. Regressions: ~100k-deep hash-valid snapshot → `:snapshot_corrupt`, journal `:ok`; scanner unit contours incl. brackets-inside-strings and escaped quotes (no false rejects).

**MEDIUM — `protected_floor/1` raises `KeyError` on malformed checkpoints** (`checkpoint.ex`): now fail-closed — any checkpoint record with missing or non-positive-integer `id`/`tip_offset` degrades the floor to `{:offset, 1}` (entire journal protected, every truncation proposal rejected) instead of crashing the never-orphan admission check. Protecting too much is safe; guessing is not. Regressions: missing id / missing tip_offset / junk-typed fields → `{:offset, 1}`; healthy path pinned.

**MEDIUM — read-then-append window defeats N-JS2**: added `FileStore.append_checked/3` → `Writer.append_checked/3`: the turn-boundary check re-runs INSIDE the single Writer against the freshest on-disk records, atomically with the append — a concurrent `turn_started` through a shared joiner handle can no longer slip a checkpoint mid-turn. The pre-flight check in `write/3` remains only as a cheap early reject before staging a snapshot. A raising check fails closed (`{:error, {:check_raised, ...}}`) without killing the Writer. Regressions: stale-reject/fresh-accept unit contours; raising check; and a 40-round concurrent joiner-handle stress asserting the raw-journal invariant that every checkpoint sits at turn balance 0.

**MEDIUM — FI-10 doc/impl mismatch (snapshot sanitization)**: partially a doc fix, honestly scoped — the freeze binds sanitize + MS secret exclusion to the MS codec, which is not landed. Both moduledocs now state the surrogate performs NO redaction and writes JSON-safe models verbatim (same trust domain and plaintext posture as the journal segments, as you noted). Combined with the surrogate write-gate + loud markers above, nothing claims a guarantee that isn't implemented. Redaction re-binds together with `dump/1` when MS lands. Flagging for a freeze ruling if snapshot-side redaction must precede MS.

**LOW — `fold_step` O(n²)**: `fold_forward/3` is now one O(n) pass (single concat, no per-record `++ [id]`).

**LOW — acronym density**: added an acronym legend to the FileBackend moduledoc mapping N-JS*/FI-*/P-JS*/OQ-JS1/AD-* to the freeze doc.

## Deferred (verified, with rationale)

- **LOW — snapshots/ grows unboundedly**: deferred by the freeze itself — FI-7 forbids implicit deletion and GC is the explicitly deferred unit (`@tag :gc`, P-JS11/N-JS11). The GC unit owns superseded-snapshot reaping; `protected_floor/1` is already its admission seam.
- **LOW — 64 MiB whole-file read × N concurrent restores**: bounded as reviewed; left as-is. Streaming hash/decode belongs to the MS codec re-bind, not the surrogate.

## Cross-PR note

The decode-guard fix (pre-decode nesting bound) lives in `checkpoint/file_backend.ex`, which **PR #581 (U10) delegates to** — the same finding raised there is fixed by this commit; nothing further needed on #581 for it.

DROOdotFOO:
This branch has conflicts that must be resolved packages/raxol_agent/lib/raxol/agent/journal/records/checkpoint.ex 
packages/raxol_agent/lib/raxol/agent/journal/records/checkpoint/file_backend.ex

merklebonsai:
Conflict resolved (checkpoint.ex + file_backend.ex). Root: U10 (#581) carried the **final** post-review checkpoint impl to master and added `restore_checkpoint/3` + refined `protected_floor/2`, so master became the strict superset — it already contains all three U9 hardening fixes (`surrogate_backend_unbound`, `nesting_within_bound?`, `append_checked`, verified at equal counts). Adopted master's versions for both files; #582's production diff is now redundant-with-#581, so this PR contributes its **red test suite**. Verified: `mix compile` clean + full U9 red suite 42/0 green against master's impl. Re-requesting review.

PR #583, opened 2026-07-16
feat(harness): U12 probe runner + provenance — reds + impl (D2)

merklebonsai:
## Status: complete unit

Red suite (the spec, below) **+ implementation + two adversarial review rounds + verification pass** — reds run **green in CI**. Original red-suite description preserved below as the authored spec.

---

Permanent failing-first RED suite for **U12 "Probe runner"**, authored before the implementation exists — the red-first fan-out against the ratified freeze contracts (docs PR #569, `harness-freeze-contracts.md` §3). Two commits: the frozen-behaviour enabler, then the red suite.

## Commit 1 — enabler: the two frozen behaviours

- `Raxol.Agent.Probe` — `spec/0` (incl. the F5 bounded-parking knobs `max_parked` / `park_timeout_ms`), `build/1`, `interpret/2`, with the §3.1 `context`/`request` typespecs verbatim. Probes are **pure interpreters**; the Runner owns journal/bus/provider/provenance (isolation by construction).
- `Raxol.Agent.Probe.Runner` — `submit/3` / `kill/1` / `status/1` skeleton with the frozen status enum and `charge` split shape; bodies return `{:error, :not_implemented}` so the reds compile and RUN (red) against a real symbol.

### Positive contours (18 tests, `@moduletag :harness_red` — excluded in `test_helper.exs`, CI stays GREEN; run with `--only harness_red`)

| contour | what green will guarantee |
|---|---|
| submit totality | never blocks (timed), never inline results; `{:ok, run_id}` under saturation AND exhaustion; only unknown probe → `{:error, :unknown_probe}` |
| P-U12.1 / N-U12.8 | fold over `probe_run` per run_id: exactly one `:started`-or-`:parked`, exactly one terminal; `kill` → `:killed` once, nothing after it (N-U12.7) |
| P-U12.2 / N-U12.2 | `reserve → call → settle` per provider call (order fold via stub counter); fail-closed |
| P-U12.6 / N-U12.9 | mid-run reserve refusal → `:exhausted` with **atomic** output (zero drafted events) |
| N-U12.4 | `max_calls` Runner-owned — no (n+1)th provider call |
| bounded parking (F5, N-U12.10) | park past `park_timeout_ms` → `:exhausted`; submit past `max_parked` → excess `:exhausted` **never** `:parked` (parking precedence), parked set bounded, one terminal each — additive to N-U12.3 |
| P-U12.3 / N-U12.5 | prefix byte-identity, **provider-free**: captured built requests, N probes at one tip byte-equal to each other and the primary's |
| P-U12.4 / N-U12.1 | kill/crash subset → primary loop trace identical to no-probes run (P-U11.5 fold machinery); `family: :loop` draft rejected whole with `:family_violation` |
| P-U12.5 / N-U12.6 | `source == :probe_<spec.id>`, `trust == context.taint ⊓ refs` — a tainted context produces NO trusted event, even when the probe drafts `:trusted` |
| fingerprint | REQUIRED on every `probe_run` terminal (frozen §2.1 struct keys checked) |
| charge | frozen split shape, cache dividend visible (`cached_prompt_tokens ≤ prompt_tokens`) |

### Negative controls + dead injectors (16 tests, **no tag → run in CI, green**)

All ten §3.3 injectors verbatim — N-U12.1 family-check-removed, N-U12.2 settle-only, N-U12.3 silent-drop, N-U12.4 probe-controlled leash, N-U12.5 re-serializing prefix builder (**the 1-byte divergence is named with its byte offset**), N-U12.6 honors-probe-trust, N-U12.7 post-kill leak, N-U12.8 double **and** omitted terminal, N-U12.9 streaming drafts, N-U12.10 no-cap/no-TTL parking — plus isolation-perturb and missing-fingerprint injectors, well-formed-trace passes in **both** trust directions (checkers not vacuous), and the m1 fired-counter self-test (schedule dumped on a dead site, m2). Schedules seed-reproducible via `U12_SEED` (default 424242).

### NOT authored (flagged, per contract)

- **`:standalone` mode reds** — `@tag`-pending until U17 (**OQ-U12.2**, ruled): the interface is frozen (`spec().mode` includes `:standalone`) but C6 cross-family is its only consumer and lands in Wave 4; a standalone red today would have no implementation surface to validate against.
- **`park_timeout_ms` TTL contour is wall-clock bounded** (the red waits `park_timeout_ms + 2s`); when U12 lands, the implementation may expose a virtual-clock seam if CI time becomes a concern — the observable (`:exhausted` terminal, exactly once) is what's frozen, not the wait mechanics.
- **Two-level budget (OQ-U12.1)** session-then-run reserve *ordering* is not separately contoured — `budget_scope` is opaque in the frozen context and the ordering has no observable seam distinct from P-U12.2 until the Runner exists; the exhaustion contours cover both levels' observable behavior (park / `:exhausted`).

## Verification

- Full `raxol_agent` suite: **1036 passed, 51 excluded** (18 of those the new reds)
- `--only harness_red`: **0/18 passed** — properly red
- Controls in default run: **16/16 green**
- No `mix.lock` drift committed


merklebonsai:
Authoring flags:
1. **:standalone reds** @tag-pending until U17 (OQ-U12.2) — interface frozen, no contour authored.
2. **park_timeout_ms TTL** uses wall-clock wait (observable frozen, wait mechanics not) — virtual-clock seam deferrable to U12-I.
3. **OQ-U12.1 reserve ordering** (session-then-run) has no observable seam distinct from P-U12.2 while budget_scope is opaque — exhaustion contours cover both levels' observables; the ordering itself is the one genuinely inexpressible pre-impl contour.

Follow-up (landed in freeze AFTER this suite was authored, in the Drew-review pass `41bfbd66d`): the two-level budget now specifies **session reservation RELEASED on run-level refusal** + a budget-conservation red (session fold returns to pre-submit value). Add that conservation contour + its leak-injector in a follow-up commit before U12-I, or I'll dispatch it — noting so it's not lost.

DROOdotFOO:

Scope note: this PR has evolved past its "red suite only" description. The branch now ships a full working `Raxol.Agent.Probe.Runner.Pool` (794 LOC) that turns the reds GREEN (moduledoc: "GRADUATED 2026-07-16"). So this judges the Pool as production code, not a `{:error, :not_implemented}` skeleton. Grounded against `red/u12-probe-runner`.

### HIGH

- [Saboteur] `lib/raxol/agent/probe/runner/pool.ex:489-495` (+ `427-436`, `271`) — Reserve is never settled back to the budget on the happy path. `run.reserved` is pinned at `@reserve_estimate` (100) regardless of `max_calls`, but `more_calls/3` reserves another 100 against the budget per subsequent call (`:433`). The `apply_result` `:completed` branch releases nothing; `release_reserve/1` only ever returns the fixed 100. Scenario: a completed 2-call run reserves 200 and releases 0 — the budget's `reserved` grows monotonically and never returns to 0; a killed/timed-out multi-call run strands `100*(calls-1)`. In a long-lived session the run-level Ledger saturates and every subsequent probe parks/exhausts. The moduledoc claims "reserve BEFORE each provider call; settle actuals after ... the authoritative charge stops diverging from the reserved counter" — the code emits `:reserve/:call/:settle` events but never reconciles the budget. No test catches it: every rig uses a throwaway budget and no test asserts `reserved == 0` after a `:completed` run (the budget-conservation checker only fires on the refusal/rollback path).

- [New Hire + Security] `pool.ex:729-743, 751-762, 766-775` — The module presents as the production Runner ("production binds the journal / EmitBridge and the real provider / Ledger to the same seam") but `provider_invoke/3`, `fingerprint/0`, and `charge/1` return hardcoded fabricated data: canned `%{content: "probe-response", usage: %{output_tokens: 12}}`, a literal `"anthropic"/"claude"/seed 7` fingerprint, and a fixed `100/90/12` charge split. The red checkers only assert SHAPE (`charge` has 4 keys and `c <= p`; fingerprint has `provider/name/params_hash`; `cached_prompt_tokens <= prompt_tokens`), never that any value reflects real usage, a real cache hit, or the real model. Scenario: the whole suite passes against a stub that invents its own charges and fingerprints — the "cache-riding dividend visible" and "fingerprint = replay identity" contracts are not actually pinned, and a real provider wiring that reports wrong/zero cached tokens would still be GREEN. A maintainer cannot tell which parts of this "production" module are stub.

### MEDIUM

- [Saboteur] `pool.ex:74-85, 530-536` — The pool is an unsupervised singleton started via `GenServer.start` (not `start_link`, not under a supervisor), yet the moduledoc and `safe_emit` comment repeatedly call it "the supervised pool" and claim a coordinator-context emit failure "fails the supervised pool loudly rather than losing the record." `emit_opening` runs inside `handle_call({:submit,...})` / `park_or_shed` (coordinator context). Scenario: one genuine (non-`:noproc`) journal raise on a single submit's opening re-raises through `safe_emit`, crashes the singleton mid-`handle_call`, and every OTHER in-flight run in the BEAM is orphaned — Tasks hit `:noproc` on `GenServer.call(parent, ...)` and die with no terminal (lifecycle-incomplete), with no supervisor to restart. The "emit durability" test (`u12_test.exs:268`) only raises on a `:call` event in the Task context (isolated `:error` terminal); the cascading coordinator-context crash is untested.

- [Saboteur] `pool.ex:793` (+ `495/499/503/507`) — `state.runs` is append-only; terminal states are never pruned. In production this singleton lives for the node's lifetime, so `runs` grows unbounded across every probe ever submitted — a steady memory leak for a busy agent. No TTL/GC, no cap.

- [Saboteur] `test/raxol/agent/red/u12_probe_runner_red_test.exs:601-673` — Timing-window test asserts a single pressure-shed evicts exactly one parked run while survivors stay "still parked", relying on `@pressure_shed_ms 40 < Process.sleep(120) < park_timeout_ms 300`. Both test modules are `async: true` against the shared singleton pool; under CI load the 120ms sleep can overrun 300ms (or timers fire late), the survivors' park TTLs fire, and the "survivors still parked" refute flakes. This is the repo's documented top flake class (wall-clock asserts).

- [Security] `pool.ex:607-633` — Trust-lattice laundering surface. `derive_trust/3` under a coarse-`:trusted` context defers to `refs_tainted?/2`, which treats a dangling ref as non-tainting (documented). A probe controls its draft `refs`; a draft that references only trusted/dangling offsets while its content derives from tainted data it read would be stamped `:trusted`, contradicting the moduledoc's "no laundering in v1." The only guard is the coarse `context.taint` floor computed upstream (outside this PR). Warrants an explicit fail-closed (dangling-ref-is-tainted) decision or a test pinning the laundering direction.

### LOW

- [New Hire] `pool.ex:37, 94, 384` — Production timing magic numbers tuned to test internals: `@kill_grace_ms 25` adds a `Process.sleep(25)` to EVERY probe run's first side effect, `@pressure_shed_ms 40` debounces shedding, with comments justifying the values by "stay well inside the suite's terminal awaits." Production latency sized to test timing, not a domain requirement.

- [New Hire] `pool.ex:351-353, 347` — `oldest_held` orders parked runs by parsing the trailing integer out of the `run_id` string (`String.split("-") |> List.last() |> String.to_integer()`). Age-ordering is coupled to the `"probe-run-<int>"` format; any change to `gen_run_id` makes `String.to_integer` raise inside the coordinator, crashing the singleton. Carry a real monotonic `seq` on the run instead.

### Cross-Persona Overlaps
- Fabricated stub values (`charge`/`fingerprint`/`provider`) + shape-only checkers is the central weakness: independently reached by New Hire (unmaintainable stub-as-production), Security (trust/replay identity not actually pinned), and Saboteur (the reserve-settle leak is invisible precisely because no test asserts real budget/charge reconciliation). Promoted to HIGH.
- Reserve-settle leak (HIGH) and budget conservation being tested only on the rollback path converge: the "reserve->call->settle" contract is pinned as event ORDER only, never as an actual budget debit/credit.
- The "supervised pool" claim (MEDIUM cascading crash) reinforces the stub-as-production framing: the moduledoc describes guarantees (supervision, durability, real Ledger) the code does not implement.

### Verdict: CONCERNS

Two HIGH plus four MEDIUM. Not a BLOCK (no CRITICAL, fewer than 3 HIGH), but the suite as-shipped GREEN-lights a stub whose budget accounting leaks and whose charge/fingerprint values are fabricated — the contract the reds claim to freeze (real reserve->settle reconciliation, real replay-identity fingerprint, cache-riding dividend) is not actually asserted. Recommend: (1) settle/release the per-call reserve on `:completed` and release `100*calls` on kill/timeout, with a test asserting `reserved == 0` after a completed multi-call run; (2) add a checker pinning charge/fingerprint to observed provider usage rather than shape; (3) supervise the pool (or drop the "supervised" claim) and prune terminated runs; (4) de-flake `u12_test.exs:601` off wall-clock windows.

merklebonsai:
@DROOdotFOO Status: #583 advanced red-suite → **complete unit** and back to ready-for-review (stacked on #584 — merge that first). Probe Runner Pool (D2/U12-CONTRACT): non-blocking submit, genuine held bounded parking (peak-bounded, single-oldest pressure shed, TTL shed), wall-clock timeout leash, two-level session→run budget with rollback (now actually test-covered incl. session-leak injector), provenance through the U11 Meta seam (refs-taint folded, probe cannot self-stamp), fingerprint on every terminal. Two adversarial rounds + a verification pass folded in.

Two **ratified red-suite amendments** (documented in-test, N-U12.3 style): (1) N-U12.3 awaits the terminal instead of reading synchronously — the old form encoded pre-F5 park semantics, contradiction ruled by V; (2) the ~458 refute scoped to `:max_parked`-refused runs per §3.1/§3.3 (pressure/TTL sheds legitimately were parked). Verification confirmed both keep catching their original lies.

Lifecycle-completeness proven under hang/kill(running+parked)/shed — exactly one terminal per path, reserves released. One follow-up commit landing shortly on this PR: post-grace-kill spend leak (a killed multi-call probe could keep reserving+calling the provider until natural completion — kill becomes effective at the call boundary). 1110 passed; default U12 run ~3.6s (10s TTL contours tagged :slow with fast equivalents).

merklebonsai:
Thanks for the review — triaged all of it. Pushed 4 commits to `red/u12-probe-runner`.

## Fixed (with regressions where behavioral)

- **HIGH — reserve/settle leak (`pool.ex:489-495` + `427-436`, `271`).** You're right: `run.reserved` was pinned at 100 while `more_calls/3` reserved another 100 per call, and no terminal settled the difference. Reworked so every per-call reserve routes through the coordinator (`{:reserve_next,…}`), accruing the cumulative held amount onto the run; **every** terminal now settles it (`apply_result` releases on `:completed` too, not just kill/timeout), so `reserved` returns to baseline. Two regressions (seeds 0/424242): a completed multi-call run leaves budget `reserved` back at baseline across a session (no monotonic growth); a post-grace kill mid multi-call stops spend at the call boundary with no leaked reserve. Kill now also terminates the run Task (mirroring the `:run_timeout` leash) so no `:call`/`:settle` can race past `:killed` (that was the deterministic way to hold N-U12.7 for the in-flight call). `2b192ffa5`.

- **MEDIUM — "supervised pool" wording (`pool.ex:74-85, 530-536`, + `runner.ex`).** Corrected. New moduledoc "Process model" section states the real thing: unsupervised lazy singleton (`GenServer.start`), lazy restart on next submit, journal-as-durability-authority (in-memory state is a cache), in-flight runs orphaned by a coordinator crash get no in-band terminal — closure comes from journal replay at recovery. `safe_emit`/`ensure_started` comments aligned. `2085faa61`.

- **HIGH — stub-as-production legibility (`pool.ex:729-743, 751-762, 766-775`).** Marked `provider_invoke/3`, `charge/1`, `fingerprint/0` as STUBs with a moduledoc "Stub vs. production values" section so a maintainer can tell stub from real. `2085faa61`. (Value-level pinning deferred — see below.)

- **LOW — `oldest_held` parses run_id (`pool.ex:351-353`).** Now carries a monotonic `seq` on each run and orders by the stored value; a `gen_run_id` format change can no longer raise `String.to_integer` inside the coordinator. `a575a872c`.

- **Stale pressure-shed comments** (`:timeout` → `:exhausted`/`:pressure`, shed-oldest-one). `19e7fdcf4`.

## Deferred (with rationale)

- **HIGH — pin charge/fingerprint to observed provider usage.** Requires binding the real provider / Ledger / `Fingerprint`; that's production wiring, not the D2 seam this PR ships. Landed the interim honesty fix (STUB markers); value-level checkers come with the real binding. Documented as follow-up.
- **MEDIUM — supervise the pool.** One-way-door production-wiring decision; explicitly out of scope this round (a supervisor changes the crash/restart contract). Documented as the moduledoc follow-up rather than added here.
- **MEDIUM — `state.runs` unbounded growth.** Real, but pruning immediately breaks `status/1`/`kill/1` answerability for recently-terminated run_ids (tests at `u12_test.exs:322,351` rely on it). Needs a TTL/GC design so terminals stay queryable for a window; noted as a follow-up next to supervision.
- **MEDIUM — `u12_test.exs:601` wall-clock flake.** Pre-existing, the repo's documented flake class; de-flaking wants a deterministic-clock seam, separate from this fix. Not touched.
- **Security — trust-lattice laundering / dangling-ref lenient (`pool.ex:607-633`).** Already a ratified v1 decision (see the `refs_tainted?/2` comment): U12 deliberately does not layer U8's strict dangling-ref damage; the coarse `context.taint` floor is the fail-closed guard. Left as-is per §3.1/§2.1.
- **LOW — timing magic numbers.** `@kill_grace_ms`/`@pressure_shed_ms` sized to test timing is a production-tuning follow-up; noted, not re-tuned here.

Verification: `compile --warnings-as-errors` clean, `mix format` clean, U12 suite 45 default + 2 slow green across seeds 0/424242, full `raxol_agent` suite (1112) green both seeds, invariants green.

PR #584, opened 2026-07-16
feat(harness): U11 meta family + provenance/taint — reds + impl (FI-5)

merklebonsai:
## Status: complete unit

---

## What

Permanent **failing-first** red suite for U11 "meta event family + provenance/taint" (FI-5), authored against the ratified freeze in `docs/proposals/in-flight/harness-freeze-contracts.md` §2 (docs PR #569) — BEFORE any U11-I implementation exists.

Two commits, per the red-suite mergeability discipline:

1. **Enabler** — the frozen shapes as compilable skeleton, zero behavior:
   - `Contract.Event` grows `scope` / `provenance` / `actor`, all optional-with-default (`scope: :session`, `provenance: %{source: :primary, trust: :trusted}`, `actor: nil`) — the grandfather clause; I9 additive-only, existing suites stay green (1032 passed).
   - `Raxol.Agent.Meta.Registry` — the frozen v1 meta type table **as data**: 12 types (`gate_decision` / `extract` / `residual` / `calibrate` / `verdict` / `research` / `promote` / `probe_run` / `attach` / `speculation` / `approval_decided` / `policy_amended`) with required payload keys, `refs` uniform on every type, `promote` the sole `:global`; plus the grow-only provenance source registry.
   - `Raxol.Agent.Fingerprint` — frozen exclusion list (`request_id` / `idempotency_key` / `trace_id` / `timestamp` / `ts`; **seed INCLUDED**), `params_inline` cap; `canonical_json/1` + `params_hash/1` declared `:not_implemented` (the ONE normative serializer is U11-I).
   - `Raxol.Agent.Meta` — seam/fold skeleton with frozen signatures, all `:not_implemented`.

2. **Red suite** (`packages/raxol_agent/test/raxol/agent/red/`):
   - `u11_meta_family_red_test.exs` — **19 reds**, `@moduletag :harness_red`, excluded in `test_helper` so CI stays GREEN. Verified `0/19 passed` under `--include harness_red` — failing-first by construction; every assertion demands a positive concrete shape, so `:not_implemented` can never satisfy one vacuously.
   - `u11_meta_controls_test.exs` — **12 control/corpus tests that RUN IN CI**: all ten §2.3 dead injectors (N-U11.1…N-U11.10) fire and are caught against an independent oracle (m1 fired-counters, m2 seed dumps, m6 oracle independence), generator non-vacuity (m5: taint chains ≥3 deep, ≥2 speculation parents, absent-actor generation, X≠Y≠Z fingerprints), registry-freeze pins, and the P-U11.2 grandfather corpus checks (green today, forever).
   - `support/meta_oracle.ex` + `support/meta_journal_gen.ex` — the m6 reference implementation (correct + one broken variant per dead injector) and seeded Reader-shaped journal generators.

## Contour coverage (freeze §2.2/§2.3)

| contour | red | control |
|---|---|---|
| producer-strict seam (unknown type, missing key, promote refs:[], scope) | ✅ | N-U11.1/2/4 |
| reader-tolerant seam (future type: skip-in-typed-folds, preserve raw, never error) | ✅ | N-U11.6 |
| P-U11.3 taint monotonicity, chains ≥3 (tool_result→extract→promote-draft) | ✅ | N-U11.3 |
| no-laundering (:tainted→:trusted forbidden, any path) | ✅ | N-U11.5 |
| OQ-U11.3 replay-taint-violation (real FileStore: `{:ok,_}` + marker, never `{:damaged}`) | ✅ | — (ruled behavior) |
| P-U11.4 scope discipline | ✅ | N-U11.4 |
| P-U11.5 fold independence (interleaved == meta-stripped, dual-oracle) | ✅ | N-U11.7 |
| P-U11.6 actor producer-seam (absent = system by fold rule) | ✅ | N-U11.8 (reader-infer + module-local halves) |
| P-U11.7 fingerprint precedence (item Z > turn Y > head X) + canonical_json byte-stability, seed included | ✅ | N-U11.9 |
| P-U11.8 speculation plural in-session parents | ✅ | N-U11.10 |
| P-U11.1 codec round-trip per registry type | ✅ | — (property class) |
| P-U11.2 grandfather corpus | corpus (CI-green) | — |

## Deliberately NOT here (U11-I work — flagged per the freeze's record-lifecycle diagram)

- EmitBridge/Writer/Reader **carry-through** of the new envelope fields: `durable_record/1` AND `map_event/3` (both paths — a field stamped on one path but stranded on the other is the N-U11.8-class named failure mode), plus the Reader decode.
- The real `Raxol.Agent.Meta` seams/folds and `Fingerprint.canonical_json/1`.

## Inexpressible-as-red (relabeled per F9, encoded as CI-green tests instead)

- P-U11.1 codec round-trip and P-U11.7 hash byte-stability are property/oracle-agreement classes (no single dead injector) — pinned as reds against the real codec, with the canonical-serializer oracle in support.
- P-U11.2 grandfather is a pure-decode **corpus** check — green today against the enabler defaults, in the controls file.
- "Nothing journaled, nothing on the bus" after a producer-seam reject is asserted at the seam's return contract; the bus/journal side-effect half becomes assertable only when U11-I wires `Meta.validate` into EmitBridge — noted in the red's docs.

## Verification

- Full `raxol_agent` suite: **1032 passed, 0 failures** (52 excluded: harness_red + pending_unit + mutation)
- Red suite under `--include harness_red`: **0/19 passed** (all red, as designed)
- `mix format --check-formatted` clean; mix.lock drift restored

Contract: `docs/proposals/in-flight/harness-freeze-contracts.md` §2 (docs PR #569).


merklebonsai:
Authoring flags: (1) the side-effect half of producer-seam rejects ('nothing journaled, nothing on the bus') is only assertable once U11-I wires Meta.validate into EmitBridge — the red pins the seam's RETURN contract now, the emission-suppression half composes at impl. (2) P-U11.1 codec round-trip + P-U11.7 hash byte-stability are property/oracle-agreement classes (no single dead injector, per the freeze's F9 relabeling); P-U11.2 grandfather runs as pure-decode corpus. (3) Envelope carry-through (durable_record/1 AND map_event/3 both paths + Reader decode) explicitly flagged as U11-I work — the enabler grows the struct, wiring is impl. Note: OQ-U11.3 red drives the REAL FileStore (violated journal stays {:ok}, violation surfaces as :taint_violation fold marker — never :damaged), which is exactly Drew's decision-time-vs-replay-seam distinction in test form.

DROOdotFOO:

Scope: `Raxol.Agent.Meta` (taint/actor/scope/precedence folds), `Raxol.Agent.EmitBridge`
(envelope carry-through + producer-seam stamping), `Raxol.Agent.Fingerprint`,
`Raxol.Agent.Meta.Registry`, `Contract.Event` growth. All three personas read the full
diff. Note: an earlier hardening round already closed several taint gaps (fail-closed trust,
branch round-trip, atom-table DoS, nested canonical_json — see `u11_hardening_test.exs`);
the findings below are residual issues NOT covered by that suite.

### HIGH
- [Saboteur] `packages/raxol_agent/lib/raxol/agent/meta.ex:655-680` (`derived_trust/3`) —
  the taint fold has no memoization. `seen` only prevents cycles on the *current* DFS path;
  sibling ref-branches share the parent's `seen`, so a node reachable through two refs is
  recomputed from scratch each time. On a DAG where meta event `n` refs `[n-1, n-2]` the
  recursion is Fibonacci-shaped — `T(n) = T(n-1) + T(n-2)` — i.e. exponential. And
  `derive_taint/1` runs this per meta event. Multi-parent refs are not exotic: the freeze
  *requires* `speculation` with >=2 parents (N-U11.10) and P-U11.8 asserts "plural in-session
  parents". A journal of a few dozen chained multi-parent speculation/extract events makes
  `derive_taint/1` (and `taint_violations/1`, which calls it) hang / OOM. Failure scenario:
  replay or a taint audit over a legitimate long speculative session never returns. Fix is
  cheap: memoize `offset => trust` (accumulate across siblings, not just cycle-guard).

### MEDIUM
- [Security Auditor] `packages/raxol_agent/lib/raxol/agent/emit_bridge.ex:339-343`
  (`stamp_generation/2`) + `contract.ex:22` default `provenance: %{source: :primary, trust:
  :trusted}` — taint *sources* are marked at no enforced seam. The producer seam force-stamps
  `:actor` and `:branch_id` (the moduledoc's whole "modules never invent it" guarantee), but
  it does NOT stamp `:provenance`/trust. Trust is entirely module-supplied and defaults
  `:trusted`. The read side is meticulously fail-closed, yet the *write* side is fail-open:
  a `tool_result` carrying attacker-controlled content that is emitted without an explicit
  `trust: :tainted` (a plain omission, not a forbidden `:tainted -> :trusted` transition)
  becomes a trusted taint leaf, and any `promote` derived from it derives `:trusted`. This is
  laundering-by-omission of exactly the source-marking the U11 taint model exists to prevent.
  The most security-sensitive field is the one left to individual modules. (Wiring
  `Meta.validate` in is flagged U11-I, but the fail-open default provenance ships in this PR.)

- [Security Auditor + New Hire] `packages/raxol_agent/lib/raxol/agent/meta.ex:658-660` — the
  cycle guard returns `:trusted` (fail-OPEN) inside a module that fails closed to `:tainted`
  everywhere else (`stored_trust`, `trust_token`, `decode_provenance` all documented "fail
  CLOSED"). It is currently safe only because DFS tree-edges still reach every tainted leaf,
  but (a) that soundness argument is undocumented — the comment only says "cycle-guarded",
  not "why trusted is safe on a back-edge"; and (b) it becomes a real laundering bug the
  moment someone adds the memoization the HIGH finding calls for keyed on node result, since
  a cached `:trusted` from a truncated cycle would then be reused on a path that should taint.
  The consistent-and-safe choice is `:tainted` on cycle. Promoted to MEDIUM (2 personas).

### LOW
- [New Hire] `meta.ex:723-728` (`check_scope/1`) returns `:ok | [map()]` — a mixed
  atom-or-list return forces every caller into a type-switch and is easy to misuse (`if
  check_scope(x) == :ok` vs pattern match). Prefer `{:ok, []}`/`{:error, violations}` or an
  always-list. Contrast `taint_violations/1` which returns a plain list.
- [New Hire] `meta.ex:812-816` (`head_default/1`) keys on `type(r) == :head_config`, a type
  that is neither in `Meta.Registry` nor produced by `contract_type/2` anywhere in this PR.
  The precedence fold's head-default branch is silently `nil` unless some other unit emits
  `:head_config`; that cross-unit dependency is undocumented, so a maintainer can't tell if
  the branch is dead code or a real seam.
- [Security Auditor] `meta.ex:896-901` (`decode_actor_field`) + `fold_actor/1` — an unknown
  actor `kind` string flows through `decode_token/2` and is preserved as a raw binary, so
  `actor.kind` may be a binary rather than one of `:human|:agent|:system`. Downstream code
  that pattern-matches the kind atom (trust/attribution decisions) silently falls through to
  a default instead of rejecting. Low because the known set is small and interned, but the
  actor-attribution path is the one the moduledoc calls "spoofable".
- [New Hire] `test/.../support/meta_oracle.ex:1778-1784` — the "independent oracle" uses
  `String.to_atom/1` on record `family`/`type`. Fine for the controlled control-corpus, but
  the module is one copy-paste away from being fed generator tokens, reintroducing the exact
  atom-table DoS the impl (`decode_token`) was hardened against.

### Cross-Persona Overlaps
- Fail-open-in-a-fail-closed-system is a recurring theme: the write-side provenance default
  (Security) and the cycle-guard `:trusted` (Security + New Hire) both undercut the taint
  model's fail-closed posture. The exponential fold (Saboteur) and the memoization fix
  intersect the cycle-guard finding — fixing one naively worsens the other, so they must be
  addressed together.

### Verdict: CONCERNS
One HIGH (exponential taint fold on spec-supported multi-parent journals) plus multiple
MEDIUM (fail-open taint-source default; fail-open cycle guard). No CRITICAL and fewer than
three HIGH, so not a BLOCK — but the taint substrate should not be wired into replay/folds
(U11-I) until the exponential fold is memoized and the write-side trust default is enforced.

merklebonsai:
## Round-2 adversarial findings — dispositions (a26e4a7f6)

All five findings triaged; four fixed with regressions, one deferred with rationale.

### FIXED

**HIGH — exponential taint fold (Saboteur)** — confirmed real. Reproduced: the un-memoized fold on a Fibonacci-shaped multi-parent DAG measures ~4.4s at depth 32, growing ~×2.6 per +2 depth (minutes at depth 40, unbounded beyond). One precision on the failure mode: the **tainted** case short-circuits at the first tainted ref (`Enum.any?`), so the true worst case is the **all-trusted** multi-parent journal, where every branch must be exhausted. Fixed with a whole-fold `offset => trust` memo threaded through `derive_taint/1` — each node computes once, linear; measured ~20µs at depth 40 post-fix. Regressions: 40-deep all-trusted Fibonacci DAG folds inside a 30s test timeout (times out pre-fix), tainted variant pins absorption through every chain, `taint_violations/1` pinned on the same shape, and the memoized result is asserted **equal to the un-memoized oracle** (which is a byte-level copy of the pre-fix algorithm) on small DAGs for both leaf trusts.

**MEDIUM — cycle-guard `:trusted` (Security + New Hire)** — addressed together with the HIGH, exactly as the cross-persona note demanded. The memo is **cycle-safe by construction**: a `:tainted` result is always cached (path-independent — a concrete tainted path was found), but a `:trusted` result is cached only when its computation touched no pruned back-edge. A back-edge-contaminated `:trusted` is recomputed fresh at its own top-level fold, so the laundering you predicted a naive node-keyed memo would introduce cannot happen — pinned by a regression on the `A<->X` cycle where X reaches taint only through A (a naive memo caches X `:trusted`; the fix derives `:tainted`). The `:trusted` back-edge itself is **kept, with the soundness argument now documented in-code**: a record is tainted iff a tainted leaf is reachable via refs, any reachable tainted leaf is reached via a simple (acyclic) path which the DFS always explores, so pruning a back-edge can never hide a tainted path — it only stops infinite descent. Cycles only exist in corrupt journals (refs name already-existing offsets in a legal one); an all-trusted-cycle test pins the behavior so any future change to that arm is deliberate, not drift.

**MEDIUM — write-side trust never producer-stamped (Security)** — real gap; fixed to the extent the freeze permits. EmitBridge now takes a write-generation `:provenance` context applied at `stamp_generation/2` — but deliberately **not** as a blind `Map.put` like the actor stamp: an unconditional overwrite would itself launder a module's honest `:tainted` under a trusted generation default. Instead it is the **taint-absorbing meet** (§2.1 taint law pt.3): a tainted generation force-taints module omissions (laundering-by-omission closed — regression pins the omitted-provenance neutral map landing `:tainted` on disk), a module-supplied `:tainted` is never overwritten, and a wrong-shaped module provenance fails closed mirroring `decode_provenance`. Two boundaries stay put by frozen contract: the grandfather default `%{source: :primary, trust: :trusted}` is the frozen §2.1 shape (defaults are load-bearing; I2 byte-identity pinned by a regression), and *which* sources are untrusted is explicitly U8 policy per the entry rule ("frozen is only that the classification happens at the tool_result producer") — the classification wiring lands at U11-I, which now has an enforced seam to plug into instead of an honor system.

**LOW — oracle `String.to_atom` (New Hire)** — fixed. `MetaOracle` now uses a bounded existing-atom conversion (its own independent helper, not the impl's `decode_token` — oracle independence preserved); unknown tokens stay raw binaries.

**LOW — `head_default` keys on unproduced `:head_config` (New Hire)** — documented, not changed. It is a cross-unit seam, not dead code: `:head_config` is the session head-tag record (§2.1 "head config is defaults only" / the replay-fidelity law), produced by the session-head unit landing with U12/U11-I fingerprint wiring. Until then the branch folds `nil` for headless journals, which is the correct answer (no head config ⇒ no default fingerprint, never a guess). In-code comment now says exactly this, and the precedence reds exercise the branch via generator-built head records.

**LOW — unknown actor `kind` preserved as binary (Security)** — documented + pinned, behavior kept. Preserve-raw is the frozen reader-seam rule (§0.2), and rewriting a present-but-unknown kind to `:system` would *misattribute* the event (absence⇒system is the rule for absence only). The real protection is §2.1 pt.6: admission decisions must fold, never read stamps — the new in-code comment requires any consumer making trust/attribution decisions on `actor.kind` to treat a non-atom kind fail-closed. Regression pins that an unregistered kind survives decode+fold verbatim (no error, no fresh atom, no `:system` rewrite).

### DEFERRED

**LOW — `check_scope/1` mixed `:ok | [map()]` return (New Hire)** — agreed it's an awkward shape, deferred. The `:ok` arm is pinned by the landed red suite (`u11_meta_family_red_test.exs` asserts `check_scope(records) == :ok`), all in-tree callers pattern-match it, and normalizing to an always-list is API churn on a red-pinned surface with no security content. Candidate for U11-I alongside the validate wiring if we want it.

### Verification
`mix compile --warnings-as-errors` clean · U11 reds + controls + hardening: 60 green · invariant suite: 41 green · full raxol_agent suite: 1077 green · `mix.lock` untouched.

PR #585, opened 2026-07-16
Harness UI: automated Ring B terminal-matrix driver (device control) [RB]

merklebonsai:
Unit **RB** — the **automated Ring B terminal-matrix driver**. Converts T0's manual real-terminal QA runbook into a device-control autotest: drives real macOS terminals, captures via their APIs, judges the C-1..C-4/N06/N07 claims programmatically, and runs the D-PA resolver. Retires the "unscheduled human action" that was the biggest strategic risk in the lane — D-PA is now re-derivable on command.

## What
- `T0.RingB.Driver` behaviour + real implementations: **iTerm2** (osascript, `contents` incl. scrollback), **Terminal.app** (osascript `history`), **wezterm** (`wezterm cli get-text`), **kitty** (remote-control), **ghostty** documented-skip (no get-text API — Ring-C residual, confirmed via `sdef`).
- `T0.RingB.Measurements` — programmatic judges: C-2 feed (count LINE rows from **`get_scrollback`**), C-1 footer-pin, C-3 cursor, C-4 resize content-preservation, N06 (`\e[2J` history-survival), N07. Marker-polled (no fixed sleeps).
- `T0.RingB.Guard` — safe teardown: kills the probe process first, verifies actual closure (`still_open?`, not just that `close` returned), bounded keystroke-fallback that activates the right app first. **Capacity gate** (R8 calibration): a separate throwaway session confirms scrollback depth ≥ count before C-2/C-4, so a shallow profile returns `:insufficient_scrollback` skip, never a false `lost`.
- `mix t0.ringb` runs the matrix + prints the resolver's D-PA.

## Measured result (live device control)
C-2 **FED on iTerm2 + wezterm + kitty** (3/3 tier-1, 100/100). **C-4: iTerm2 REFLOWS** sealed history (the (B) signal). N06: `\e[2J` wipes scrollback on wezterm/kitty (validates the inline-mode keyframe-ban). Terminal.app corrected to fed 100/100 (was a false 20/100 reading the visible tail). Resolver: **provisional (A)** — ghostty structurally un-measurable, so a fully-definitive automated verdict is impossible; the D-PA ruling (ship (A), (B) runtime-detected per-terminal) is recorded in roadmap §0.

## Safety + tags
Opus safety review: **APPROVE** — window targeting is handle-scoped (never touches the user's live windows), keystroke-dismiss activates the right app first. `@moduletag :ring_b/:macos_gui/:unix_only`, unconditionally excluded from default `mix test` (needs a GUI session + installed terminals) — headless CI opens no windows; `--only ring_b` runs the matrix.

**Operational note (learned the hard way):** the live GUI matrix is a **watched-only** operation — an unattended run can leave close-confirmation sheets stuck on a background Space. The driver *code* builds/compiles unattended; only the live matrix run needs eyes. The measurements already produced the D-PA answer, so re-running is low-value.

## Residuals (documented, not guessed)
ghostty (no get-text API), C-4 perceptual reflow-quality (content-preservation asserted; visual ghost-column judgment stays screenshot/Ring-C), Terminal.app C-3 cursor-column. RB v2 follow-up: a generic `AXUIElement` text-reader would cover ghostty without per-app recipes.

---
*Terms: **Ring A/B/C** = test rings: CI-headless / real-terminal device-control / human-perceptual. **C-1..C-4, N06/N07** = the probe claims (C-2 = sealed lines actually reach scrollback; C-4 = resize reflow; N06 = `\e[2J` wipes scrollback). **D-PA** = the paint-authority decision those measurements settle: (A) seal-time-only, with (B) soft-owned-history as a runtime-detected upgrade where a terminal measurably reflows (iTerm2).*


DROOdotFOO:

Scope: `scripts/harness/t0/ringb/*.ex` + `lib/mix/tasks/t0.ringb.ex`. This is a macOS-GUI-only, `:ring_b`-tagged harness that is unconditionally excluded from `mix test`/CI (verified in `test/test_helper.exs`) and is developer-run only. That bounds blast radius: no production runtime path. Findings judged against that context. All personas found real issues; no clean passes.

### HIGH
- [Saboteur] `scripts/harness/t0/ringb/measurements.ex:562` (`poll_capture`) + every driver's capture/spawn call — the entire measurement path has **no timeout on the external call itself**. `poll_capture` calls the getter (`driver.get_visible/1` -> `Osa.run` -> `System.cmd("osascript", ...)`, or `wezterm/kitty` `System.cmd`) and only checks `expired?` *after* the getter returns. `System.cmd` has no built-in timeout, and `Guard.with_timeout/2` wraps only `close/1` and `still_open?/1` — not `spawn_session`, `run_command`, `get_visible`, `get_scrollback`, or `resize`. Failure scenario: on first run macOS pops a TCC "osascript wants to control iTerm2" Automation-permission dialog, or the terminal beachballs; `System.cmd` blocks forever, and the 6s poll deadline never fires because it is only evaluated between getter returns. `mix t0.ringb` hangs indefinitely with no bound. This directly contradicts the moduledoc/PR claim that the task "will not leave a modal dialog waiting on a human" / is safe to run — the safety net only covers teardown, not the capture path that actually holds the window open.

### MEDIUM
- [Saboteur + New Hire] `runner.ex:86` — teardown reconstructs `marker = "ringb-#{driver.name()}-#{claim}"` (no unique suffix), but the probe was launched with a *different* marker generated in `measurements.ex` via `Guard.marker/2` = `ringb-<drv>-<claim>-<System.unique_integer>` (line 68 etc.). Teardown's `kill_marker` -> `pkill -f "ringb-iterm2-C1"` only kills the probe by unintended prefix-substring match; the `row.marker` the measurement returns is dead code. Fragile: if a claim id ever becomes a prefix of another (e.g. `C1`/`C10`), `pkill -f` would kill the wrong probe. Separately, the capacity-gate teardown (`measurements.ex:518`) passes `ringb-<drv>-CAP-<int>` to `kill_marker`, but `calibrate/3`'s command (`for i in $(seq ...)`) never embeds that marker — so `pkill -f` there matches nothing and is a silent no-op (harmless only because the calibration command has no HOLD, so no live foreground job to kill).
- [Security] `guard.ex:188` `dismiss_modal_best_effort/1` — activates the *whole app* (`tell application "iTerm2"/"Terminal" to activate`) then sends `keystroke return`, which per the code's own comment lands on the "Terminate" destructive default. This is app-scoped, not window/handle-scoped, contradicting the PR's "window targeting is handle-scoped (never touches the user's live windows)" safety claim. Failure scenario: the user has a second, unrelated Terminal/iTerm2 window with a frontmost sheet (save dialog, another confirm) when the guard fires — the synthetic Return dismisses *that* sheet with its destructive default, on a window the harness never created.

### LOW
- [Security] `measurements.ex:481` `held_cmd/3` builds `"T0_HOLD_SECONDS=8 bash '#{probe}' ..."` and types it into a live shell via `run_command`. The probe path is single-quoted, so a checkout path containing a single quote (worktree paths here already look like `/Users/.../.claude/worktrees/agent-<hash>/...`) breaks the quoting and injects into the driven terminal's bash. Low exploitability (local, self-controlled path), but it is an unescaped-path -> shell surface. Prefer `System.cmd`-style arg lists or shell-quote the path.
- [Saboteur] `guard.ex:150` `with_timeout/2` uses an unsupervised `Task.async` + `Task.shutdown(:brutal_kill)`. Brutal-killing the task closes its Port but does not guarantee the spawned `osascript`/CLI OS process is reaped — repeated close timeouts can orphan external processes. Idiomatic OTP would use a `Task.Supervisor`.
- [New Hire] `measurements.ex:78` (`available_claims/1`) and `measure_c4` both gate C4 on `@resize_capable` — double-gating the same capability in two places that must be kept in sync. Also `mark_and_capture/4` has an undocumented `Process.sleep(300)` magic wait, alongside many bare-number settle sleeps (`@settle_ms 1400`/`2200`, `Process.sleep(200)`), several documented but some not.
- [Security] `osa.ex:44` writes the AppleScript (window ids, commands) to a predictable-name file in the world-readable `System.tmp_dir!()` then `osascript`s it — minor local info exposure / TOCTOU window between `File.write!` and exec. Negligible given local-only use.

### Cross-Persona Overlaps
- The PR's central safety narrative ("won't hang / will not wait on a human", "handle-scoped, never touches user windows") is undercut from two independent angles: the Saboteur HIGH (capture/spawn path has no timeout, so a TCC prompt or beachball hangs the run) and the Security MEDIUM (`dismiss_modal_best_effort` is app-scoped and sends a destructive keystroke). Both are real gaps between the stated guarantees and the code.
- The marker divergence was flagged by both Saboteur (wrong-process-kill risk) and New Hire (dead `row.marker`, sync-hazard) — promoted to MEDIUM per the 2-persona rule.

### Verdict: CONCERNS
One HIGH (unbounded hang on the capture path, defeating the PR's own safety claim) plus two MEDIUM. Not a BLOCK given this is CI-excluded, macOS-GUI-only, watched-only developer tooling with no production path — but the timeout gap should be closed (wrap `spawn_session`/`run_command`/`get_*`/`resize` in `Guard.with_timeout` or give `System.cmd` a hard deadline) and the teardown marker should reuse the measurement's actual marker before this is relied on for unattended runs.

merklebonsai:
Thanks Drew — round-2 findings adjudicated below; everything landed in `71db725` on `feat/harness-ui-RB`.

### Accepted + fixed
- **HIGH — unbounded capture/spawn path** (`measurements.ex` `poll_capture` + all driver calls): `71db725` — every driver call site (`spawn_session`, `run_command`, `get_visible`, `get_scrollback`, `mark_cursor`, `resize`) is now wrapped in `Guard.with_timeout/2` — `guarded_osa/1` in the iTerm2/Terminal.app drivers, shared `Guard.run_cmd/4` for the wezterm/kitty CLI calls — so a TCC prompt or beachball can no longer hang `mix t0.ringb` past the deadline. The residual is stated honestly in `Osa`'s and the mix task's moduledocs: the guarantee is that *the run stays bounded*, not that a TCC-stuck `osascript` orphan is reaped (the first-run Automation prompt must be clicked by a human, and the task is documented watched-only).
- **MEDIUM — teardown marker divergence** (`runner.ex:86` / `measurements.ex`): `71db725` — `Runner.run_one` now generates one marker per (driver, claim) via `Guard.marker/2` and threads it through `opts[:marker]`, so the marker in the probe's argv is byte-identical to the one `kill_marker/1` pkills afterward; no more prefix-substring matching and no dead `row.marker`. The capacity-gate's silently-no-op `kill_marker` (calibration command never embeds a marker) was replaced with `Guard.close_only/2`, which closes-and-verifies without claiming a process-kill guarantee that path never had.
- **MEDIUM (claims half) — "handle-scoped" safety narrative vs app-scoped dismissal**: `71db725` — the PR's implication is retracted; `Guard.dismiss_modal_best_effort/1` and the `mix t0.ringb` moduledoc now state plainly that the last-resort dismissal is APP-scoped, sends a Return bound to the destructive default, fires only after kill-marker + close + `still_open?` verification have all failed, and only under a watched run. The wrong-sheet scenario you described is documented as a real, accepted limitation rather than papered over.
- **LOW — `held_cmd` unescaped probe path**: `71db725` — probe path now goes through a `shell_quote/1` POSIX close-quote/escape/reopen (`'\''`) escaper instead of bare `'#{probe}'`.

### Declined
- **LOW — `with_timeout/2` should use `Task.Supervisor`**: these are standalone dev scripts loaded via `Code.require_file` (no OTP application or supervision tree to hang a `Task.Supervisor` off), and supervision wouldn't change the failure mode you identified — neither `Task.shutdown(:brutal_kill)` nor a supervised shutdown can reap an `osascript` blocked on a TCC prompt. That reaping gap is the real issue, and it's now documented as an accepted caveat in `Osa`'s moduledoc rather than hidden behind an idiom swap.
- **LOW — C4 double-gating + magic sleeps**: `@resize_capable` is a single module attribute referenced twice *within the same module* — `available_claims/1` (matrix planning) and `measure_c4` (defense for direct calls, e.g. from the unit tests) share one constant, so there's no second list to drift out of sync. The correctness path is marker-polled; the remaining `@settle_ms`/`Process.sleep` values are pre-poll GUI settle heuristics in a watched-only harness, not judged timing. Fair hit that the `mark_and_capture` 300ms is undocumented — noted for the RB v2 pass, but not worth a churn commit alone.
- **LOW — AppleScript temp file in `System.tmp_dir!()`**: contents are window ids and probe commands (no secrets), the name includes `System.unique_integer`, the file is removed in `after`, and it's a local-only dev tool — declining per your own "negligible" rating.

### Deferred
- **MEDIUM (mechanism half) — sheet-scoped dismissal**: replacing the app-scoped `activate` + `keystroke return` with a dismissal addressed to the exact window/sheet element is deferred, not skipped. Exit criterion: the next **watched live matrix run**, where the sheet-scoped path can be validated against a real confirmation modal before it replaces the current fallback — landing it unvalidated would trade a documented limitation for an untested one. Recorded in `Guard.dismiss_modal_best_effort/1`'s moduledoc and the mix task's operator notes.

PR #586, opened 2026-07-16
test(harness): U4-R red suite — reattach/replay closure + Dormammu (AD-15/FI-12)

merklebonsai:
## What

Permanent **failing-first (RED)** suite for U4 "Reattach/replay" (AD-15/FI-12), authored BEFORE implementation against the ratified freeze in `docs/proposals/in-flight/harness-freeze-contracts.md` §1.1–§1.3 (JS-FREEZE — docs PR #569). Reattach = a client (re)connects to a session and receives history per `historyPolicy` (from offset / tip / none) plus the live durable stream from that point, with NO gap and NO duplicate delivered-as-live (P-JS5: `read(0..o−1) ++ attach_live(o..) ==` full durable record stream, as a **sequence**). It MUST work writerless (dead BEAM, replay-only, tar'd dir) — read-side only, no record written; per §1.1 there is no attach record kind and the attach audit meta event is best-effort only. Tip location = backward scan under the frozen `conversational?` predicate, branch-aware, skipping unknown kinds.

Two commits:

1. **Enabler** — `Raxol.Agent.Journal.Tip`: the FROZEN tip predicate as a pure function (`conversational?/1` + branch-aware `tip/2`, grandfather defaults `kind ⇒ "event"` / `branch_id ⇒ "main"`, CONVERSATIONAL whitelist incl. `item_started` + `approval_requested` per OQ-JS3, `woken`/`idle`/`state_change`/meta/non-event excluded — comment names the freeze section). Shared by U4 and U9 — the false parallel dissolved at the contract layer. Plus `Raxol.Agent.Reattach`: `attach/3` `@callback` surface (`attach{offset, historyPolicy}`) dispatching to a configurable impl, default `Reattach.NotImplemented` → `{:error, :not_implemented}` until U4 lands.
2. **Red suite** — `test/raxol/agent/red/u4_reattach_red_test.exs`: `U4ReattachRedTest` (7 reds, `@moduletag :harness_red`, excluded in `test_helper.exs` so CI stays GREEN) + `U4ReattachGuardsTest` (5 frozen-predicate properties/corpus + 5 dead-injector controls, NOT tagged — run in every CI) + support `test/raxol/agent/red/support/u4_red_support.ex` (independent raw-file tip oracle, closure/publish-ahead detectors, injectors, fired-counter harness).

## Positive contours (RED until U4)

- **P-JS5 replay closure** — ∀ split offset o over generated journals driven through the REAL FileStore: history ++ live == full durable stream as a sequence, live ids never `< o` (no earlier durable delivered as live). Property; seed-reproducible, journal spec dumped on failure.
- **Writerless reattach** — Writer closed (dead BEAM), attach serves full history and journal bytes stay untouched (reattach never writes; nothing may depend on an attach marker).
- **Self-containment** — `tar` + untar the session dir into a fresh base: identical bytes, identical reattach (§0 clause 5: the directory is the unit of portability).
- **Dormammu (P-JS3, FI-12)** — generated journals ENDING in ≥1 checkpoint/meta/idle/woken record (generator shape enforced per meta-inv 5); `attach(:tip)` must resume at the last CONVERSATIONAL loop event, never the tail.
- **historyPolicy slices** — `{:from_offset, n}` delivers exactly ids `n..o−1`; `:tip` exactly the tip record; `:none` empty history + live-only.

## Green guards (run in CI — hold the frozen predicate in place while U4 is built)

- **P-JS2/P-JS8 dual-oracle tip determinism, branch-aware** — raw-file JSONL decoder (own whitelist copy, backward scan — deliberately independent, m6) vs Reader-path `Tip` agree on every branch or both `:no_tip`; closure-rule corollary (a selected tip is conversational and on-branch); `tip/1 == tip(_, "main")`.
- **P-JS3 at the predicate layer** — Dormammu property + all-non-conversational → `:no_tip`.
- **P-JS6 unknown-kind tolerance** — future-kind records replay `{:ok, _}`, keep offsets dense, never move the tip (compared by payload marker vs the journal without them).
- **P-JS7 grandfather corpus** — the checked-in kind-less/branch_id-less `golden-v1` journal reads identically: absent kind ⇒ event, absent branch_id ⇒ main, tip preserved at offset 8 on both oracles. Corpus test, not a red.

## Negative controls (dead injectors — GREEN in CI, fired-counters per meta-inv 1/2)

- **N-JS7 emit-ahead-of-journal** — an EmitBridge-like variant publishing the live id BEFORE append/ack; the probe (a late subscriber's raw read at delivery time) shows the id not yet readable — invariant **I3**'s publish-ahead window, exactly what the P-JS5 closure red catches. The correct append-then-publish order shows no violation (both arms asserted).
- **N-JS5** — tip predicate collapsed to `kind == "event"` (family/type clauses dropped) selects a trailing meta event → fails Dormammu.
- **N-JS8** — branch-blind tip scan selects another branch's higher-offset record as the "main" tip.
- **N-JS4** — a damages-on-unknown-kind reader marks a journal the real Reader replays `{:ok, _}` — fails the tolerance red.
- **Marker-dependent reattach** — a variant requiring a journaled attach marker fails on a writerless session (and "works" once a marker exists — the failure is exactly the forbidden dependence).

Every armed injector fires a counter; `assert_all_fired!` fails the control (dumping the schedule alongside the ExUnit seed) if one goes dead.

## Deliberately NOT authored / flagged

- **Live-delivery transport** — the freeze specifies the closure law but not the live wire shape; this suite FREEZES `{:reattach_live, session_id, record}` messages to the attaching process (documented in the moduledoc) so the implementation is built against the red, not vice versa.
- **Tar'd-dir session resolution** — `attach/3`'s frozen arity has no base-dir parameter; the self-containment red pins the `RAXOL_SESSIONS_DIR` resolution the FileStore already honors (noted inline for the U4 implementer).
- **Broadcast permission bus (AD-15's second half)** — no frozen shape exists for it in JS-FREEZE §1; not authorable without inventing contract surface.
- **P-JS11/N-JS11 (GC low-watermark)** — `@tag :gc` DEFERRED by the freeze; ungeneratable at `low_watermark = 1`.
- The enabler exposes a `config :raxol_agent, :reattach_impl` seam so `attach/3` keeps its full inferred type (avoids Elixir 1.18+ "pattern will never match" warnings in the red suite against a literal-return stub — same device as U9-R's `put_backend/1`).

## Verification

- Red suite excluded (CI shape): `10 passed (2 properties, 8 tests), 7 excluded` — guards + controls green.
- Red suite included (`--include harness_red`): **10/17 passed — all 7 reds FAIL** with `{:error, :not_implemented}` + journal specs + seeds dumped. RED by construction against the enabler.
- Full raxol_agent suite: **1030 passed (13 properties, 1017 tests), 40 excluded**, 0 failures.
- `MIX_ENV=test mix compile --warnings-as-errors` + `mix format --check-formatted` clean; mix.lock drift restored (no lock changes in this PR).


merklebonsai:
Authoring flags — three are genuine contract GAPS the freeze doesn't yet pin (the suite made red-side decisions, documented; worth a freeze addendum before U4-I):

1. **Live-delivery transport unspecified** — JS-FREEZE §1 froze the journal/closure semantics but not the reattach *wire*. Suite freezes `{:reattach_live, session_id, record}` messages as a deliberate red-side placeholder. Needs a real transport freeze at U4-I / with S2 (wire transport, UI lane).
2. **Tar'd-dir resolution** — `attach/3` has no base-dir arg; suite pins `RAXOL_SESSIONS_DIR`. The self-containment law (§0) implies dir-relative resolution but doesn't name the mechanism.
3. **AD-15 broadcast permission bus** — no frozen shape in JS-FREEZE §1; not authorable without inventing contract, so not authored.

Non-gaps: I3 citation applied (per correction) for the emit-ahead/N-JS7 control; P-JS11/N-JS11 @tag :gc deferred. The 3 gaps above are additive freeze work, not red defects — parking them as U4-I-blockers unless you want them pinned now.

DROOdotFOO:

Scope: test-only + two skeleton modules (`Journal.Tip` pure predicate, `Reattach` facade returning `{:error, :not_implemented}`). The RED module (`U4ReattachRedTest`) is `:harness_red`-excluded and does not run in CI today; the guards module (`U4ReattachGuardsTest`) does. Findings below distinguish "runs in CI now" from "activates when U4 lands and the tag is removed" — the latter still matter because the PR explicitly designates the red suite as the future acceptance suite, and its stated purpose is *deterministic* contract enforcement.

### HIGH
- [Saboteur + New Hire] `u4_reattach_red_test.exs:359-370` (`reattach_in_base`, module is `async: true` at :58) — the helper mutates the process/OS-global `RAXOL_SESSIONS_DIR` via `System.put_env`, runs `attach`, then restores in an `after`. In an `async: true` suite this global is shared with every other concurrently-running module; any test that resolves a journal base from `RAXOL_SESSIONS_DIR` during the window reads the wrong session dir, and the restore can clobber a value another test set. Failure scenario: FileStore/base-resolution test runs concurrently with the tar self-containment test → non-deterministic cross-session dir bleed in a suite whose whole reason to exist is replay determinism. New-Hire facet: the tar test's pass/fail silently depends on the *undocumented-in-contract* assumption that FileStore "honors" `RAXOL_SESSIONS_DIR` (comment at :356-358 asserts it, the frozen `attach/3` arity has no base param) — so a green here can mean "env honored" rather than "reattach correct." Same root line, two personas → promoted. Fix: point the read path at an explicit base (test-injected opt) instead of a global env var, or make the helper `async: false`.

### MEDIUM
- [Saboteur] `u4_reattach_red_test.exs:346-353` (`collect_live/3`, `after 1_000`) — a wall-clock receive timeout that on expiry silently returns whatever partial list it has. When U4 lands, a correct-but-slow live delivery yields fewer than `expected_live` messages (callsites :131, :334), and `closure_check` then reports `{:sequence_mismatch, ...}` — a spurious failure disguised as a closure violation rather than an honest "timed out waiting for N live messages." This is exactly the flake class the task flags: a 1s sleep-equivalent gating the core P-JS5 assertion. Prefer `assert_receive` with an explicit count and a clear timeout message, or a synchronization signal, not a partial-return `after`.
- [Security] `reattach.ex:86-87` (and the frozen `@callback` at :69) — `attach/3` validates only `is_binary(session_id) and is_integer(from_offset) and from_offset >= 0`; `session_id` is unsanitized. Reattach is described as a client-facing reconnect surface ("a client (re)connects to a session"), read-side, resolving `session_id` into a filesystem path via FileStore. Freezing the contract shape now without a validation clause bakes in path-traversal exposure (e.g. `session_id = "../../other-tenant/session"`) for the eventual U4 implementation. Current runtime impact is nil (returns `:not_implemented`), but the frozen surface is where a `valid_session_id?/1` guard belongs. Add it before the shape is treated as immutable.

### LOW
- [Saboteur] `u4_reattach_red_test.exs:139` — the P-JS5 property closes the FileStore *after* its assertions, so a failing/shrinking run leaks the Writer process/handle; the guards module (e.g. :467, :525) correctly closes before asserting. Inconsistent cleanup discipline; matters once the red activates under StreamData shrinking.
- [Saboteur] `u4_red_support.ex:308-312` (`emit_ahead_publish!`, runs in CI via the N-JS7 guard) — `predicted = length(FaultJournal.raw_ids!(dir)) + 1` then `{:ok, ^predicted} = FileStore.append(...)` hard-couples a CI-green test to FileStore assigning dense, 1-based, gapless offsets. Any future offset-scheme change (0-based, sparse after damage) breaks this guard for a reason unrelated to the invariant it guards.
- [New Hire] `u4_red_support.ex:298` — the docstring says "predict the next offset from a local counter," but the code reads the file (`length(FaultJournal.raw_ids!(dir)) + 1`), not a counter. Minor doc/code drift. More broadly, the suite's codename density (Dormammu, N-JS7/P-JS5/I3, m1/m2/m6, OQ-JS3) forces constant cross-referencing to the freeze doc; acceptable for a contract-freeze suite but at the edge of unmaintainable without the doc open.

### Cross-Persona Overlaps
- `RAXOL_SESSIONS_DIR` global mutation flagged by both Saboteur (async race) and New Hire (hidden env dependency substituting for the frozen contract) — promoted to HIGH.

### Verdict: CONCERNS

One HIGH (test-isolation via global env in an async determinism suite) plus a determinism MEDIUM (`after 1_000` partial-return) and a forward-looking security MEDIUM (unsanitized frozen `session_id`). None are production-breaking today (skeleton returns `:not_implemented`, red module excluded), but all land squarely in the invariants this PR exists to protect, so they should be resolved before U4 turns the suite on. The `Journal.Tip` predicate, the dual-oracle independence, and the dead-injector fired-counter discipline are well-constructed and genuinely adversarial — the guards that run in CI are sound.

merklebonsai:
## Review response (9e6febe19) — fixed vs. deferred

Triage rule applied: test-determinism/hygiene findings → fixed in this PR; findings about future U4-I behavior → recorded as U4-I obligations (U4-I itself stays out of scope pending the transport freeze addendum).

### HIGH — `RAXOL_SESSIONS_DIR` mutation under `async: true` → **FIXED**
`reattach_in_base/4` and every `System.put_env` are gone. The red module now seeds **all** journals under the run-wide ambient sessions base — pinned once in `test_helper.exs` before any test starts and never written again; per-test isolation comes from unique session ids. This also resolves the New-Hire facet *and* a latent coherence bug you were one step from naming: the frozen `attach/3` has no base param, so sessions created under per-test tmp dirs would have been unresolvable by ANY correct implementation — the non-tar reds silently assumed a resolution surface that didn't exist. The resolution rule (attach resolves ids where `FileStore.open/2` does by default) is now stated in the module doc, not implied by a hidden env write.

The tar red is reformulated as **destroy-then-restore-from-archive**: tar the session dir, `rm_rf` the original, untar back into the ambient base, assert record identity, then attach. Same frozen claim (§0.5 — reattach against the untarred artifact serves identical history), zero global writes, `async: true` kept. Deferred facet: the *moved-to-a-different-base* variant is untestable without a base on the read surface; that's an explicit **U4-I / transport-addendum obligation** (pin session-base resolution, or add an opts-bearing read entry point) — noted below.

### MEDIUM — `collect_live` partial return on `after 1_000` → **FIXED**
The collector now takes exactly `expected` messages or `flunk`s with "timed out (1000ms) waiting for live message i/N … this is a delivery timeout, NOT a P-JS5 closure violation". No partial list can ever reach `closure_check`, so a slow-but-correct U4 can't be misreported as a closure breach. (Per-message timeout, so total patience scales with N.)

### MEDIUM — unsanitized frozen `session_id` → **FIXED**
`Raxol.Agent.Reattach` gains public `valid_session_id?/1` — the same conservative single-segment rule FileStore enforces at write time (`[A-Za-z0-9._-]+`, never `.`/`..`) — and the facade rejects everything else as `{:error, :invalid_session_id}` **before** dispatching to any impl. The frozen shape now carries the guard, so the eventual U4 implementation inherits an already-validated id. Pinned in CI today by two new guards in `U4ReattachGuardsTest` (traversal/separator/NUL/empty rejected; well-formed ids never rejected by the facade).

### LOW — P-JS5 closes FileStore after assertions → **FIXED**
`try/after` around the P-JS5 property body and the `:none`-policy red (close-before-assert is impossible in these two — live records are appended after the attach — so `after` is the correct discipline there). No Writer leak under StreamData shrinking.

### LOW — `emit_ahead_publish!` couples to dense 1-based offsets → **ACCEPTED AS DELIBERATE, doc fixed**
The `raw-count + 1` prediction couples the guard to the *frozen* offset law (§1.1 / P-JS1: dense, 1-based, gapless) — which is ratified, not incidental. If the offset law ever changes, this guard breaking loudly is the desired failure mode (contract change, not flake). The docstring now says exactly that, and the "local counter" drift you flagged is corrected to match the code.

### LOW — codename density → acknowledged, no change
Accepted cost of a contract-freeze suite; every module doc links the freeze doc. Not restructuring.

### Deferred — U4-I obligations (for the transport freeze addendum)
1. Pin session-base resolution on the read path explicitly (ambient base per `FileStore` default), or freeze an opts-bearing read entry point — then add the moved-base portability red.
2. Live-delivery pacing: the 1s per-message timeout is a suite constant; the addendum should state the delivery-latency envelope the closure red may assume.
3. `valid_session_id?/1` charset may be tightened (not loosened — §0 only-grows) when real session-id producers are enumerated.

### Verification
- Guards (CI path): 12 passed (was 10; +2 session-id hygiene)
- Red suite `--include harness_red`: 7 reds fail exactly on `{:error, :not_implemented}` — no errors, no timeouts
- Full `raxol_agent` suite: **1032 passed**, 40 excluded
- `mix.lock` untouched; no `System.put_env`/`delete_env` anywhere in the suite

DROOdotFOO (commented):

Scope: test-only "red suite" (`@moduletag :harness_red`, excluded in `test_helper.exs:64`) plus two enabler modules and CI-green guards/controls. Verified: exclusion tag and `RAXOL_SESSIONS_DIR` pin already exist in `test/test_helper.exs`; `FaultJournal.raw_scan/raw_records!/raw_ids!/segment_paths` exist; golden fixture resolves to `test/invariants/fixtures/golden/v1.0.0/golden-v1`; the CI-green controls are deterministic (messages pre-sent synchronously, raw reads follow synchronous appends). So master CI stays green -- good. Findings below are about the suite's value as the *future acceptance suite* and about opacity.

### HIGH
- **[New Hire + Saboteur] missing contract** `packages/raxol_agent/lib/raxol/agent/journal/tip.ex:8`, `reattach.ex:12`, and both test moduledocs -- Every citation the suite is "authored against" (`docs/proposals/in-flight/harness-freeze-contracts.md` §1.1-§1.3, `harness-invariants.md`, JS-FREEZE, P-JS5/N-JS7, FI-12/AD-15) points to files that **do not exist in the repo**. They live in `#569` (`docs(harness): freeze contracts constitution`), still OPEN. Consequence 1 (New Hire): a newcomer cannot locate the frozen predicate/law these ~1500 lines encode -- the "verbatim from §1.1" block in `tip.ex` is the only ground truth, and it is unreviewable against a source. Consequence 2 (Saboteur): the whole justification ("frozen shape, implementation built against the red") rests on an unmerged sibling PR; if #569's §1.x drifts during review, `@conversational`, the grandfather defaults, and the `{:reattach_live, ...}` wire shape silently diverge from the ratified doc with nothing to catch it. Fix: gate this PR on #569 merging (or land the doc here), and pin the exact doc revision/section hash in the moduledoc.

### MEDIUM
- **[Saboteur] vacuous-pass hole in the CI-green dual-oracle property** `packages/raxol_agent/test/raxol/agent/red/u4_reattach_red_test.exs:1191` -- P-JS2 asserts `a == b`, which is trivially satisfied when both oracles return `:no_tip`. Nothing in the generator (`member_of(@shapes)`, `min_length: 1`) guarantees any run produces a real `{:tip, _}`; a corpus of only `:meta/:idle/:woken/:checkpoint/:unknown` passes green while exercising zero tip logic. Contrast the Dormammu guard at :1242 which has teeth (`refute raw_conversational?(List.last...)`). Fix: assert at least one branch yields `{:tip, _}` across the run, or prepend a guaranteed conversational record so agreement is non-trivial.
- **[Security Auditor] tar-slip pattern encoded in the self-containment red** `u4_reattach_red_test.exs:886-898` -- `:erl_tar.extract/2` into `Path.dirname(dir)` with no member-path validation and no `:keep_old_files`. Harmless for this self-created archive, but §0.5 self-containment explicitly frames the session directory as portable (tar'd/rsync'd -- potentially cross-host/cross-tenant). The frozen facade validates `session_id` (good, `valid_session_id?/1`) but the self-containment contour models blind extract-to-cwd -- exactly the pattern the U4 implementer is invited to copy for *received* archives, where an archive member `../../other-tenant/...` is a path traversal. Fix: add an inline note that real reattach must validate/normalize archive member paths (mirroring the `session_id` rule), so the frozen contour does not bake in the footgun.

### LOW
- **[Saboteur] thin property coverage for a future acceptance suite** `u4_reattach_red_test.exs:784` and `:925` (`max_runs: 8`) over tiny input spaces (pre 2..8, post 1..4). The green guards use 15. When U4 lands and these become the acceptance reds, 8 runs is weak. Bump on graduation.
- **[Saboteur] wall-clock delivery ceiling** `u4_reattach_red_test.exs:1061-1081` (`@live_timeout_ms 1_000`, `receive ... after`). Inert now (reds excluded; `attach/3` fails fast before delivery), but becomes a load-sensitive flake ceiling the day these run as the acceptance suite -- the repo's stated main flake class. Prefer a completion-keyed receive loop over a fixed 1s.
- **[Saboteur] dual-oracle independence is only structural** `u4_red_support.ex:493-521` vs `tip.ex:36-45` -- both `@conversational` whitelists are byte-identical; P-JS2 can catch a scan-structure regression but not a wrong whitelist value (both oracles would be wrong identically). Acknowledged by the author, but worth stating as a known limit of what the property proves.
- **[New Hire] undefined codename** `u4_red_support.ex:459` (`dormammu_tail_pool`) and the "Dormammu" describe blocks -- the moduledoc explains the *shape* (journal ending in checkpoint/meta/idle/woken) but never why the codename maps to it. A one-line gloss ("resume must escape the non-conversational tail loop") would orient a newcomer.

### Cross-persona overlaps
- Missing freeze contract (#569 unmerged) is both a New Hire opacity issue and a Saboteur drift/merge-order risk -> promoted to HIGH.

### Verdict: CONCERNS
Test-only, correctly excluded, CI-green controls verified deterministic -- not a BLOCK. But resolve the freeze-doc availability (HIGH) before/with merge, and close the P-JS2 vacuous-pass hole plus the tar-slip note (MEDIUM) so the frozen contours actually have teeth when U4 graduates them.

merklebonsai:
Per V (2026-07-17): **#586 is explicitly stacked behind #569.** It stays open and does not merge until the constitution is *frozen*, not merely proposed — specifically until PA-6 / PA-7 / PA-8 (the security HIGHs whose ratification would move the shapes these reds assert against) are ratified into binding law. Once #569 freezes, #586 rebases onto the ratified shapes and the reds re-green against the final contract text. This closes the "reds frozen against movable shapes" coupling Drew flagged.

Marking this blocked-by #569.

DROOdotFOO:
**Merge readiness: DRAFT -- not for merge.**

Intentional red suite; keep it draft until its implementation lands. It is also authored against #569's frozen contracts and depends on the #613/#633 confinement fix, so sequence it after those.

PR #587, opened 2026-07-16
feat(harness): U8 BlastRadiusGate + approvals — reds + impl (AD-6b/14)

merklebonsai:
## Status: complete unit

---

## What U8 is

Write/destructive tools are **LOCKED by default**. Unlocking requires an approval flow: the BlastRadiusGate evaluates each call (`effect_class` + `egress` + taint), escalating calls emit `approval_requested` (**family `:loop`** — frozen F3, so a resumed conversation lands on the pending approval via the tip predicate), a decision arrives as an `approval_decided` meta event (`%{request_ref, decision, refs}`, actor from the **envelope only**), and only then does the call proceed. Deny is durable (ReviewDecision semantics, AD-14): after a deny, the same request is never retried into success without a NEW approval cycle. Scopes `:once`/`:session`/`:root` mirror `Authorization.Policy`.

This PR is the **permanent failing-first red suite** for that unit, authored BEFORE implementation against the ratified freeze (`harness-freeze-contracts.md` §1.1/§2.1/§5.2, `harness-yolo-safe-research.md` §2/§5/§7). Enabler commit adds the `Raxol.Agent.Authorization.BlastRadiusGate` behaviour skeleton (`:not_implemented`); red commit adds the suite. When U8 lands, the reds flip green **in place** — no assertion changes.

## Contours

| # | Contour |
|---|---|
| C1 | Locked by default: write tool without approval → typed reject; stub counter proves no execution |
| C2 | `approval_requested` emitted `family :loop` + `request_ref`; independent `conversational?`/tip oracle selects it as tip over a trailing checkpoint + meta record (Dormammu-lite, FI-12) |
| C3 | `approval_decided :approved` with refs to the request unlocks exactly that request (scope `:once`) |
| C4 | Deny is durable: post-deny bare retry never succeeds |
| C5 | `:session` scope persists across calls within the session |
| C6 | Fold-rebuild: approval state folded over `approval_decided` events authorizes probes identically to live state (§2.1 replay/resume law) — behavioral equality, 5 fixed seeds |
| C7 | Forged/dangling `approval_decided` (refs match no live request) rejected |
| C8 | `@tag :action_surface` — §5.2 predicate: escalate iff `:irreversible_external` OR `egress` |
| C9 | `@tag :action_surface` — `:reversible_local` + no egress + trusted auto-proceeds; self-reported `destructiveHint` never gates (N-Y5) |
| C10 | **Taint is a FOLD over refs (HIGH-1)**: a `:trusted`-STAMPED arg whose ref chain reaches a tainted record escalates; direct taint escalates despite a benign class (FI-5); clean chain auto-proceeds |

## Dead injectors (negative controls — run in CI)

Each fails its contour with an exact violation signature (m3 branch probes), with a fired-counter pass over all sites (m1) and seed-reported schedules (m2):

- **(a)** `ExecutesWhilePending` — runs the side effect while the approval is pending → fails C1
- **(b)** `ForgetsDeny` — a denied decision grants like an approve → fails C4
- **(c)** `AcceptsForgedDecision` — enacts a decision with no matching live request → fails C7
- **(d)** `InMemoryOnly` — approval state not rebuildable by fold → fails C6 on every seed
- **(e)** `ReadsDestructiveHint` — gates on the self-reported hint instead of `effect_class` → fails C8 + C9
- `ReadsStampedTrust` (**HIGH-1**) — reads the stamped trust field of the direct arg: passes the honest direct-taint leg, fails ONLY the fold-only mis-stamped leg of C10 (the signature difference is the field-read/fold discrimination)
- `IgnoresTaint` — no taint check at all → fails C10 at the direct-taint leg
- `EmitsMetaFamily` — emits `approval_requested` as `family :meta` → fails C2 (F3 is a frozen decision)

`ReferenceGate` (test support) is the executable spec: discrimination tests prove every contour is satisfiable exactly as written.

## Mergeability discipline

- `@moduletag :harness_red` on the RED module; `:harness_red` added to `test_helper.exs` exclusions — CI stays GREEN
- Negative controls + discrimination live in an untagged companion module in the same file — they run every push
- `@tag :action_surface` marks C8/C9 (depend on the not-yet-landed F2 `Raxol.Action` draft)
- Verified: `MIX_ENV=test mix compile --warnings-as-errors` clean; package suite 1039 passed / 0 failed (reds excluded); with `--include harness_red` exactly the 10 reds fail on `:not_implemented`

### DIFF
--- a/packages/raxol_agent/lib/raxol/agent/authorization/blast_radius_gate.ex
+++ b/packages/raxol_agent/lib/raxol/agent/authorization/blast_radius_gate.ex
@@ -33,6 +33,44 @@ defmodule Raxol.Agent.Authorization.BlastRadiusGate do
   `approval_decided` events (`rebuild/1`) so a resumed/replayed session
   reconstructs identical enforcement state — the journal is the authority.
 
+  ## The approver is authenticated by the gate (fail-closed)
+
+  The gate's reason to exist is "a destructive call requires a **human**
+  decision", so the gate itself enforces it: `apply_decision/3` and the
+  `rebuild/1` fold ENACT a decision **only** when the envelope actor is
+  `%{kind: :human}`. A decision authored by an `:agent`, `:system`, a `nil`/
+  absent actor (absent = system emission, §2.1 frozen fold rule), or a
+  malformed actor is NOT a grant — live it is `{:error, {:unauthorized_actor,
+  kind}}` (the request stays pending, awaiting the human), and on rebuild it
+  folds as a no-op. This closes the self-approval hole: an agent that can emit
+  `approval_decided` events naming its own live `request_ref` still cannot
+  unlock its own escalated write. The trust boundary is thus IN the gate, not
+  deferred to journal-producer discipline.
+
+  ## One request identity everywhere
+
+  A single injective key — the `request_ref`, encoding
+  `{tool, call_id, lineage_digest}` — is the identity for pending requests,
+  `:once` grants, AND durable denies (no bare-`call_id` side keys). So:
+
+    * a `:once` grant unlocks exactly the approved request — a different tool
+      sharing the `call_id`, or the same `(tool, call_id)` re-issued with
+      DIFFERENT argument lineage (the digest differs), does not match and falls
+      through to the taint fold / escalation (AD-14 "unlocks exactly that
+      request");
+    * a deny is durable **per request** (AD-14: "the *same request* may not be
+      retried into success without a NEW approval cycle") — it never
+      blanket-blocks an unrelated tool that happens to share a `call_id`.
+
+  ## Deployment status (declared deferrals)
+
+    * **Not yet wired**: no tool-dispatch path routes through `authorize/3` in
+      this unit — "locked by default" describes the gate's contract, not yet a
+      live control. Wiring the dispatch seam is a follow-up unit.
+    * **State growth**: `pending`/`once`/`session`/`denied` are bounded only by
+      session lifetime (undecided escalations and denies accumulate). A
+      TTL/cap policy is a follow-up GC concern; sizes are O(decisions).
+
   ## The auto-approve predicate (§5.2, inlined normative)
 
   A call **escalates** (requires human approval) iff
@@ -184,9 +222,11 @@ defmodule Raxol.Agent.Authorization.BlastRadiusGate do
 
   The `decision.request_ref` MUST name a live pending request; a forged/dangling
   decision (no matching request) is rejected `{:error, reason}` — a decision with
-  no request is not a fact the gate may enact. `actor` comes from the envelope. A
-  `:denied` decision is **durable**: after it, the same request may not be retried
-  into success without a NEW approval cycle.
+  no request is not a fact the gate may enact. `actor` comes from the envelope
+  and is AUTHENTICATED: only a `%{kind: :human}` actor's decision is enacted —
+  any other actor (agent/system/nil/malformed) is rejected fail-closed, the
+  request left pending. A `:denied` decision is **durable**: after it, the same
+  request may not be retried into success without a NEW approval cycle.
   """
   @callback apply_decision(state(), decision(), actor()) ::
               {:ok, state()} | {:error, reason :: term()}
@@ -201,6 +241,12 @@ defmodule Raxol.Agent.Authorization.BlastRadiusGate do
   folding one reproduces the post-consumption state, so a spent one-shot grant is
   NOT resurrected on resume (a re-issued call must not auto-admit without a new
   approval cycle).
+
+  The fold is reader-tolerant and fail-closed (never a crash, never a grant):
+  a non-`:human` actor, an unknown `decision`/`scope` value, an unparseable
+  `request_ref`, or a malformed record folds as a NO-OP — a version-skewed or
+  partially-corrupt journal degrades to "nothing extra granted", it does not
+  prevent the session from reconstructing enforcement state.
   """
   @callback rebuild([{decision() | consumption(), actor()}]) :: state()
 
@@ -232,14 +278,21 @@ defmodule Raxol.Agent.Authorization.BlastRadiusGate do
   @impl true
   @spec escalate?(call()) :: boolean()
   def escalate?(call) do
+    # Map.get, never dot access: a call map MISSING a "required" structural
+    # field must fail CLOSED to escalation on the authorization path, not raise
+    # a KeyError out of evaluate/2. `:__missing__` is unrecognized, so both
+    # cond arms below treat absence exactly like an unknown value.
+    effect_class = Map.get(call, :effect_class, :__missing__)
+    egress = Map.get(call, :egress, :__missing__)
+
     cond do
-      call.effect_class == :irreversible_external -> true
-      call.egress == true -> true
+      effect_class == :irreversible_external -> true
+      egress == true -> true
       # Auto-proceed ONLY for a recognized benign class with egress EXACTLY
-      # false. An unknown/unrecognized effect_class, or a non-boolean egress,
-      # fails CLOSED to escalation (§2.1 unknown-trust doctrine) — never a silent
-      # proceed on a field we don't understand.
-      known_effect_class?(call.effect_class) and call.egress == false -> false
+      # false. An unknown/unrecognized/absent effect_class, or a non-boolean
+      # egress, fails CLOSED to escalation (§2.1 unknown-trust doctrine) —
+      # never a silent proceed on a field we don't understand.
+      known_effect_class?(effect_class) and egress == false -> false
       true -> true
     end
   end
@@ -341,10 +394,10 @@ defmodule Raxol.Agent.Authorization.BlastRadiusGate do
 
   Order is load-bearing (the YOLO-safe soundness ladder):
 
-    1. an `:once` grant for exactly this `call_id` → `{:proceed, _}` (a completed
+    1. an `:once` grant for exactly this request → `{:proceed, _}` (a completed
        approval cycle covers precisely this request; the grant is consumed);
-    2. a durable deny for this `call_id` → `{:reject, :denied, _}` (deny survives
-       a bare retry — AD-14);
+    2. a durable deny for exactly this request → `{:reject, :denied, _}` (deny
+       survives a bare retry — AD-14);
     3. a tainted lineage (decision-time fold) → `{:escalate, _, _}` REGARDLESS of
        class/egress or any standing session grant (FI-5: the untrusted leg is a
        distinct confirmation);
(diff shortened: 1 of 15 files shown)


merklebonsai:
Authoring flags: (1) :root scope treated session-wide by the reference — route-keyed :root needs U8's real route plumbing (C5 covers :session only). (2) C6 rebuild takes {decision, actor} pairs (no journal-backed approval_decided records yet); asserts behavioral equality so it survives any internal rep — real U8 derefs refs into the journal. (3) tip-when-trailing asserted against an in-suite oracle of the frozen conversational? predicate (Dormammu-lite; U4's live tip locator not landed). (4) envelope-actor-single-source is structural (actor is a separate arg, not in payload) — a payload-actor drift red belongs to U11's emit seam. HIGH-1 baked in: C10 three-leg — direct-tainted escalates, :trusted-stamped-over-tainted-ref-chain escalates via FOLD, clean auto-proceeds; ReadsStampedTrust injector passes the honest leg, fails ONLY the fold-only leg.

merklebonsai:
@DROOdotFOO Status: #587 advanced red-suite → **complete unit** (stacked on #584 — merge that first). BlastRadiusGate implements the decision-time-fold law: taint folded over `arg_refs` via `Meta.derive_taint` at decision time, never reads a stamped trust field (C10 mis-stamp leg pinned). Two adversarial review rounds folded in; second round closed real replay-law holes:

- **string-tool rebuild divergence** — canonical string tool key + injective (length-prefixed) request_ref; rebuild == live pinned for string/colon-bearing tools (C11)
- **once-grant scope** — keyed by full request identity (tool + call_id); a grant for (toolA, c1) can't admit (toolB, c1) or bypass its taint fold (C12)
- **once-consumption folds** — consumed grants aren't resurrected on rebuild (C13)
- **escalate? fail-closed** — unknown effect_class / non-boolean egress escalates (C14)
- **ReferenceGate spec tightened to fail-closed** — matches production, pinned (C15)

1106 passed, 0 failures; invariants 41 green.

**Substrate follow-up (deliberate, not in this PR):** `Meta.derive_taint` treats a dangling ref as non-tainted; U8 layers its own comprehensive fail-closed guard. Proposal: add `Meta.derive_taint_strict/1` to the U11 substrate so U12+ consumers inherit fail-closed instead of re-hand-rolling — deferred to avoid churning the stacked branches mid-flight.

DROOdotFOO:

Scope: the landed implementation in this unit — `blast_radius_gate.ex`, `meta.ex`, `meta/registry.ex`, `emit_bridge.ex`, `contract.ex`. Reviewed the full gate + folds; tests read for intent, not audited for style.

### HIGH

- **[Security Auditor] blast_radius_gate.ex:475, :502 — the approver is never authenticated; `actor` is accepted then discarded.** The gate's whole purpose is "a destructive call requires a *human* decision." Yet `apply_decision(state, decision, _actor)` ignores the actor entirely, and `rebuild/1` folds `{decision, _actor}` with the actor discarded too. Nothing in the gate (nor `Meta.validate/1`, which only checks required keys) verifies `actor.kind == :human`. Failure scenario: an agent (or any producer that can emit a `family: :meta, type: :approval_decided` with `decision: :approved` naming a live `request_ref`) self-approves its own escalated write — the confirmation step is defeated. The moduledoc defers actor auth "upstream / on the journal envelope," but there is no upstream wiring in this PR, and a security gate that never inspects the approver is fail-open by construction. At minimum `apply_decision`/`rebuild` should reject a non-`:human` approving actor (or the contract must state the trust boundary that guarantees only humans can journal an `approval_decided`).

- **[Security Auditor + New Hire] blast_radius_gate.ex:375/:531 vs :518 — deny and grant use asymmetric keys (deny-bypass / over-block).** `denied` stores and is checked by the **bare `call_id`** (`MapSet.member?(state.denied, cid)` at :375; `grant(:denied)` puts `cid` at :531). But `:once` grants are keyed by the **full tool+call_id ref** (`ref_encode` at :518), and the `:once` branch is evaluated *before* the deny branch (:372 before :375). Failure scenario, when `call_id` is reused across tools within a session (the gate offers no uniqueness guarantee): approve `(toolA, c1)` `:once` → `once = {ref(A,c1)}`; deny `(toolB, c1)` → `denied = {c1}`, and `grant(:denied)` only deletes `ref(B,c1)` from `once`, leaving `ref(A,c1)`. A re-issued `(toolA, c1)` hits the `:once` branch first and **proceeds despite `denied` containing `c1`** — a deny that the code otherwise treats as blocking all of `c1`. The inverse reading (deny is meant to be per-request) makes :375 over-block unrelated tools that share a `call_id` (availability). Either way the keying is internally inconsistent; pick one identity and use it in all four sites. (Promoted per the 2-persona rule; exploitability is conditional on `call_id` reuse.)

### MEDIUM

- **[Saboteur] blast_radius_gate.ex:514-533, :436 — `apply_decision`/`rebuild` crash on out-of-domain journal input (replay DoS).** `grant/5` has heads only for `:approved`/`:denied` and scopes `:once|:session|:root`; there is no catch-all. A journaled decision with `decision: :corrupt` or `scope: :global` (a legal meta scope elsewhere) raises `FunctionClauseError`, and `parse_ref/1` (:436) raises on any `request_ref` not shaped `"req:<len>:..."`. Failure scenario: a version-skewed or partially-corrupt journal makes `rebuild/1` crash on resume — the session cannot reconstruct enforcement state and fails to start. This directly contradicts the reader-tolerant doctrine the sibling `Meta.decode/1` upholds (unknown tokens preserved raw, never raised). The gate's fold should degrade fail-closed (treat an unparseable/unknown decision as no-grant or a deny marker), not crash.

- **[Security Auditor] blast_radius_gate.ex — gate has zero call sites; "locked by default" is not enforced at runtime.** `grep` across `packages/` and `lib/` finds no invocation of `BlastRadiusGate.authorize/3` or `evaluate/2` outside the module and its tests (the only other `blast_radius` hit is an unrelated UI preview component). The module is a correct library with a passing suite, but no tool-dispatch path routes through it, so no write/destructive tool is actually gated by this PR. Acceptable if wiring is a declared follow-up unit — but the PR body's "write/destructive tools are LOCKED by default" reads as a live control and it is not one yet. State the deferral explicitly so a reviewer doesn't merge believing protection is active.

- **[Saboteur] blast_radius_gate.ex:411, :517-531 — unbounded state growth.** `pending` grows on every escalation and is pruned only by `apply_decision`; `denied` grows on every deny and is never pruned; `once`/`session` likewise. A long-lived session (or an agent that spams escalating write calls that are never decided) leaks these MapSets without any TTL or cap. Bounded only by session lifetime, but there is no back-pressure.

### LOW

- **[Saboteur] blast_radius_gate.ex:236-242 — `escalate?/1` uses dot access on `call.effect_class`/`call.egress`.** A call map missing a "required" field raises `KeyError` inside `evaluate/2` rather than failing closed to escalation. `tainted_lineage?/1` already uses `Map.get` with defaults; `escalate?` should be equally defensive given it sits on the authorization path.

- **[New Hire] blast_radius_gate.ex:346-354, :372 — the `:once`-before-taint ordering is a fragile invariant.** A consumed one-shot grant proceeds without re-running the taint fold (:372 precedes :378). The ordering comment claims intent, but "a grant for exactly this request was already vetted" silently assumes `call_id` uniquely and stably identifies the same arguments/lineage — an assumption the type contract does not enforce. Reuse of a `call_id` with newly-tainted lineage would proceed on the standing `:once` grant.

- **[New Hire] blast_radius_gate.ex:433-441 — the `"req:<len>:<tool><cid>"` encoding is a hand-rolled, undocumented wire format parsed by binary-size matching.** It is injective and clever, but it is the load-bearing link between a journaled decision and the grant key, with no round-trip test guard visible in the impl module and a hard crash (:439 binary match) if a `request_ref` ever originates from outside the gate.

### Cross-Persona Overlaps

- **Actor-not-authenticated** reached independently by Security (missing `:human` check) and Saboteur (forged/agent-authored `approval_decided` folded into a grant on rebuild) — kept HIGH.
- **Deny/grant keying asymmetry** flagged by Security (deny-bypass) and New Hire (maintainer will assume symmetric keying) — promoted MEDIUM → HIGH.

### Verdict: CONCERNS

Two HIGH findings (approver never authenticated; asymmetric deny/grant keying) plus several MEDIUM. Not a BLOCK: the gate is an unwired library unit with a green suite, so nothing is live-breaking today. But the actor-authentication gap is a design-level fail-open in a control whose entire reason to exist is "a human must approve," and it should be closed (or the trust boundary contractually pinned) before this gate is wired into any tool-dispatch path.

merklebonsai:
## Adversarial review processed (dced1cfa5)

All findings triaged; fixes each carry a regression contour (C16–C20) verified **failing against the pre-fix gate** before landing. U8 suite 49/49, invariant suite 41/41, full raxol_agent suite 1116 green, `compile --warnings-as-errors` clean.

### Fixed

- **HIGH — approver never authenticated** → the gate now owns the trust boundary: `apply_decision/3` and the `rebuild/1` fold enact a decision **only** for a `%{kind: :human}` envelope actor. Agent/system/`nil`-absent (absent = system per §2.1 frozen fold rule)/malformed actors are rejected fail-closed — live it's `{:error, {:unauthorized_actor, kind}}` with the request left pending for its human; on rebuild the entry folds as a no-op, so a forged agent-authored `approval_decided` in the journal grants nothing on resume. **Regression C16**: every non-human actor shape fails to unlock live *and* via rebuild, side effect never runs, and a `:human` decision on the same still-pending request still works.
- **HIGH — deny/grant keying asymmetry** → one identity, all sites: the `request_ref` now keys pending requests, `:once` grants, **and** durable denies (the bare-`call_id` deny set is gone). Deny is request-scoped, matching AD-14's "the *same request* may not be retried into success": no deny-bypass through the once branch, and no over-blocking of an unrelated tool sharing a `call_id`. **Regression C17**: a denied `(tool_a, cid)` stays durably rejected while `(tool_b, cid)` gets its own approval cycle (pre-fix: blanket reject).
- **MEDIUM — folds crash on out-of-domain journal input** → both folds degrade fail-closed, never raise: unknown `decision`/`scope` values are `{:error, _}` live and no-op on rebuild; `parse_ref` returns `:error` (never a binary-match crash) on any foreign/corrupt ref; a malformed record folds as a no-op. A version-skewed journal still rebuilds, and valid grants in the same journal still apply. **Regression C18.** This also resolves the LOW hand-rolled-ref concern: the wire format is now documented at the encoder and parsing is total.
- **LOW — `escalate?/1` dot access** → structural fields read via `Map.get`; a call missing `effect_class`/`egress` escalates (fail-closed, same doctrine as the unknown-value arms) instead of raising `KeyError` inside `evaluate/2`. **Regression C19.**
- **LOW — `:once`-before-taint fragile invariant** → the assumption is now enforced rather than assumed: `request_ref` binds `tool + call_id + lineage digest`, so a standing grant covers exactly the request the human vetted. Re-issuing the same `(tool, call_id)` over newly-tainted lineage produces a different ref, misses the grant, and falls through to the decision-time taint fold. Digest skew fails closed (grant just doesn't match → re-escalate). **Regression C20.**

ReferenceGate (the executable spec) mirrors every fix; the `ForgetsDeny` injector was adjusted to the ref-keyed `grant` signature. All pre-existing contours (C1–C15) unchanged and green.

### Deferred (verified + now declared in the moduledoc)

- **MEDIUM — zero call sites**: confirmed — the only other `blast_radius` hits in the tree are the unrelated UI preview component. This unit is the gate library + its contract suite; wiring the tool-dispatch seam through `authorize/3` is a follow-up unit. The moduledoc now states explicitly that "locked by default" is the gate's contract, not yet a live control.
- **MEDIUM — unbounded state growth**: real but session-bounded; documented in the moduledoc as a declared deferral. A TTL/cap policy is a GC concern for the wiring unit — capping `pending`/`denied` inside the gate would silently drop enforcement facts, which is worse than growth.

### Ruling flagged (fail-closed default chosen)

Non-human **denies** are also rejected/no-op'd, not just approvals — the whole `approval_decided` surface is treated as human-only, so an unauthorized deny leaves the request pending rather than durably denying it. If the contract later wants system-authored denies (e.g. timeout auto-deny), that's a new ruling; today's fail-closed default is "only a human decides, either way."

PR #589, opened 2026-07-16
Harness UI: printed-history append path (seal-once, immutable-prefix keystone) [T2b]

merklebonsai:
Unit **T2b** — the **printed-history append path**, the immutable-prefix keystone. This is the substrate that makes "sealed history is never repainted" true in real code. Builds on T2a's DECSTBM region manager (#574).

## What
`Raxol.UI.Rendering.PaintAuthority.InlineAuthority` — the real `PaintAuthority` impl:
- **`seal/2`** = `with_cursor(:history, &append_sealed/2)` — save → position → emit → restore. `with_cursor/3` is the **sole owner** of the `\e7`/`\e8` (DECSC/DECRC) pair, now `try/after`-guarded so the restore always runs even if the producing fun raises (airtight "sole owner").
- **`append_sealed/2`** fills the history region **top-down** via a `next_row` cursor, positioning at `min(next_row, region_top)`, then clamps to the bottom row and relies on the terminal's natural index-at-boundary scroll once full. (A first cut always-CUP'd to the bottom row — that front-loads content-free filler and silently defeats the high-water/immutable-prefix check; **fill-down-then-scroll is load-bearing** and documented as such.)
- **`resize/3`** delegates to `ScrollRegionManager.resize/2` — a single DECSTBM re-set, **zero re-emitted history bytes** (INV-5-A).
- **`reflow_capable?/1`** — thin (B)-seam predicate, `true` only for `{"iTerm2", _}` (RB's C-4 reflow finding); fires `[:raxol,:ui,:paint_authority,:reflow_capable_resize]` telemetry on a geometry-changing resize but implements **no (B) re-emission** (per the ruling: ship (A), (B) runtime-detected only).

## Keystone invariant proven
The immutable-prefix invariant — `history(E_k)` is a byte-identical **prefix** of `history(E_final)` — holds across a 1000-block deterministic stream and a 25-run property, checked at every checkpoint by `SealOracle.immutable_prefix?/2`.

## Tests bind to reality (Drew's 3 questions)
- **Real producer:** positive/adversarial suites drive `seal`/`append_sealed`/`with_cursor` end-to-end into a real `StringIO` device — no logic reimplemented in the test.
- **Goes RED if broken:** a genuine fail-first — three real seals, then a row-1 repaint fed through the **real** `SealOracle.replay` → `Emulator` → `immutable_prefix?` → `{:violation, 0, ...}`; the paired GREEN test proves the oracle isn't stuck returning violation (oracle-validity guard, not a rubber stamp).
- **Real bytes:** INV-5-A asserts byte-identical prior-history prefix + `new_bytes == Dialect.region_set(1,18)`, not a count/proxy. No path emits `\e[2J`/`\e[3J` (grep-verified; the ban that RB's N06 validated).

## Review
Opus analytic review: **SHIP** — cursor boundary math traced correct at `region_top-1/region_top/region_top+1`, `with_cursor` confirmed sole save/restore owner, fail-first genuine, `reflow_capable?` nil-safe and iTerm2-only, no full-clear anywhere. Four yellows raised; #1 (moduledoc overclaimed the fail-first) and #2 (`with_cursor` try/after) folded into this commit; #3/#4 cosmetic, deferred.

Gates: `compile --warnings-as-errors` clean; 16 property/adversarial tests green + T2a's 19 + the 154-test harness oracle suite still green; `mix format` + `mix credo --strict` clean; no mix.lock drift.

**Base note:** this targets `feat/harness-ui-T2a` (#574), not master — T2a isn't merged yet. Rebase to master once #574 lands.

---
*Terms: **seal** = a block written into the terminal's own scrollback, never repainted after. **Immutable-prefix invariant** = bytes emitted up to any checkpoint are a byte-identical prefix of the final stream (checked by replaying real bytes through the terminal emulator — the `SealOracle`). **DECSC/DECRC** = `\e7`/`\e8` cursor save/restore. **D-PA (A)** = the lane's paint-authority ruling: after seal, nothing in history may be rewritten. **INV-5-A / R8** = lane invariant/methodology ids (R8 = every bug shown red before the fix).*

### DIFF
--- a/lib/raxol/ui/rendering/paint_authority.ex
+++ b/lib/raxol/ui/rendering/paint_authority.ex
@@ -96,6 +96,19 @@ defmodule Raxol.UI.Rendering.PaintAuthority do
                bottom >= top do
       "\e[#{top};#{bottom}r"
     end
+
+    @doc """
+    Cursor Position (CUP), 1-based row, column pinned to `1`: `CSI row;1H`.
+    The single byte-builder for "position at the start of history row
+    `row`" -- callers that used to hand-roll `"\\e[\#{row};1H"` (T2b's
+    `InlineAuthority.append_sealed/2`) route through this instead, so the
+    wire format is defined once and pinned by test rather than duplicated
+    per call site.
+    """
+    @spec cursor_position(pos_integer()) :: binary()
+    def cursor_position(row) when is_integer(row) and row >= 1 do
+      "\e[#{row};1H"
+    end
   end
 
   defmodule IOAuthority do
--- a/lib/raxol/ui/rendering/paint_authority.ex
+++ b/lib/raxol/ui/rendering/paint_authority.ex
@@ -100,10 +100,10 @@ defmodule Raxol.UI.Rendering.PaintAuthority do
     @doc """
     Cursor Position (CUP), 1-based row, column pinned to `1`: `CSI row;1H`.
     The single byte-builder for "position at the start of history row
-    `row`" -- callers that used to hand-roll `"\\e[\#{row};1H"` (T2b's
-    `InlineAuthority.append_sealed/2`) route through this instead, so the
-    wire format is defined once and pinned by test rather than duplicated
-    per call site.
+    `row`" -- callers that used to hand-roll `"\\e[\#{row};1H"` (the append
+    path's `InlineAuthority.append_sealed/2`) route through this instead,
+    so the wire format is defined once and pinned by test rather than
+    duplicated per call site.
     """
     @spec cursor_position(pos_integer()) :: binary()
     def cursor_position(row) when is_integer(row) and row >= 1 do
--- a/lib/raxol/ui/rendering/paint_authority/content_guard.ex
+++ b/lib/raxol/ui/rendering/paint_authority/content_guard.ex
@@ -1,0 +1,165 @@
+defmodule Raxol.UI.Rendering.PaintAuthority.ContentGuard do
+  @moduledoc """
+  Neutralizes control bytes inside agent/LLM-originated content before it
+  reaches `PaintAuthority.InlineAuthority.seal/2` (T2b review fix, HIGH:
+  "our code was wrong" -- `seal/2`/`append_sealed/2` wrote iodata
+  verbatim, so an embedded `\\e[2J`/`\\e[3J`/CUP inside model output could
+  wipe native scrollback or repaint an already-sealed row FROM INSIDE THE
+  CONTENT, defeating every invariant the append path exists to hold, no
+  matter how correct the fill-down/seal-once machinery around it is).
+
+  This module is intentionally SHARED, not private to T2b:
+  `Raxol.UI.Rendering.PaintAuthority.InlineAuthority.seal/2` is its first
+  caller, and T2c's footer/pinned-viewport repaint path (also fed
+  agent-controlled text) is expected to reuse it rather than growing a
+  second, divergent sanitizer.
+
+  ## The allowlist grammar
+
+  Scanning byte-by-byte (UTF-8 multi-byte sequences pass through as
+  opaque bytes -- nothing here decodes or re-encodes text, it only ever
+  recognizes ASCII control structure):
+
+    * **Printable ASCII** (`0x20..0x7E`) and any byte `>= 0x80`
+      (UTF-8 lead/continuation bytes) -- passed through unchanged.
+    * **`\\t`, `\\r`, `\\n`** -- passed through unchanged (the C0 controls a
+      line of legitimate content actually needs).
+    * **SGR (`CSI ... m`)** -- e.g. `\\e[1;31m`, `\\e[0m` -- passed through
+      VERBATIM. This is the renderer's own styling vocabulary; content
+      that legitimately carries color/style resets must keep working.
+    * **Every other C0 control byte** (`0x00..0x1F` except the three
+      above) and **DEL** (`0x7F`) -- stripped silently. These carry no
+      printable residue worth preserving.
+    * **Any other ESC (`0x1B`)-led sequence** -- a non-SGR CSI (cursor
+      moves, erases, ...), OSC (`\\e]...`), DCS (`\\eP...`), or a bare/
+      truncated ESC with no recognized introducer at all -- has ONLY its
+      leading ESC byte stripped. Scanning resumes immediately after that
+      ESC, so whatever printable bytes followed it (the `[2J` in
+      `\\e[2J`, the `]0;title` in an OSC, etc.) are re-scanned as ordinary
+      text and, being printable, survive into the output.
+
+  ## "Visible-honest" neutralization, not silent deletion
+
+  A `\\e[2J` with its ESC stripped becomes the literal, visible text
+  `[2J` -- four printable characters a terminal just prints, doing
+  nothing else. That is the deliberate choice this module makes over
+  silently deleting the whole sequence: a visible `[2J` fragment in the
+  sealed history is an honest record that content tried to inject a
+  control sequence and got stopped, whereas an invisible drop would look,
+  to anyone watching the terminal, exactly like the model simply never
+  said anything unusual. Any C0 control byte embedded in what would have
+  been that sequence's parameters (e.g. a stray BEL terminating an OSC)
+  IS silently dropped, per the point above -- there is no printable
+  residue for a non-printable control byte to leave behind.
+
+  ## Scope note: OSC 8 hyperlinks
+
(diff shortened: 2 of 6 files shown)


DROOdotFOO:

Scope: `lib/raxol/ui/rendering/paint_authority/inline_authority.ex` (+2 property suites). Three adversarial personas, grounded in code. Base modules cross-checked against `origin/feat/harness-ui-T2b` (`ScrollRegionManager`, `PaintAuthority` behaviour, `Dialect`).

### HIGH
- [Security Auditor] `inline_authority.ex` `seal/2` and `append_sealed/2` (~L199-226) write `iodata` to the device verbatim — no escape stripping, no validation. Every invariant this module advertises ("never emits `\e[2J`", "immutable prefix", "no row is ever re-addressed") is true only for the module's *own* bytes. A producer whose block text contains ANSI (`\e[2J`/`\e[3J` in agent stdout, an LLM transcript, or remote/tool output routed into printed history) injects those bytes straight into the terminal: `\e[2J`/`\e[3J` wipes native scrollback — the exact N06 disaster the moduledoc claims to prevent — and `\e[<r>;1H...` inside a "sealed" block repaints an already-sealed row. Both defeat seal-once from *inside* content that the oracle treats as trusted. The property/adversarial suites only ever feed `string(:alphanumeric)` and hand-authored clean blocks, so this surface is never exercised. Either the append path must reject/escape control bytes in `iodata`, or the "sealed content is trusted" precondition must be enforced at the seam, not assumed.

### MEDIUM
- [Saboteur] `append_sealed/2` + `count_lines/1` (~L210-226, L305-310): `next_row` advances by the count of `"\n"` bytes. A seal whose `iodata` is not newline-terminated (e.g. `"partial"`) yields `count_lines == 0`, so `next_row` does not advance; the *next* `seal/2` CUPs to the same `target_row` and overwrites the partially written row — a direct seal-once / immutable-prefix violation. The "whole `\r\n`-terminated lines" contract is stated only in the `seal/2` docstring (~L198-202) and enforced nowhere in code (`seal/2` does no validation). `block_gen` always appends `\r\n`, so no test can catch a caller that forgets. This is the same root cause as the HIGH: `iodata` is trusted without validation.
- [Saboteur + New Hire] `with_cursor/3` (~L246-263) is documented as the "SOLE owner" of the `\e7`/`\e8` register, but nothing enforces non-nesting. The struct carries no depth field; if any caller invokes `seal/2` (or another `with_cursor/3`) *inside* a `with_cursor` fun — e.g. T2c's footer path the moduledoc explicitly anticipates as a caller — the inner `\e7` clobbers the single hardware DECSC register and the outer restore lands on garbage, corrupting cursor position for every subsequent seal. `SealOracle.save_restore_balance/1` checks `decsc_max_depth <= 1` in tests, but production has no guard. "Sole owner" is a prose guarantee with a runtime hole. (Promoted: Saboteur + New Hire.)
- [New Hire] Moduledoc (~L8-111) is ~100 lines of lane-id shorthand — `T2a/T2b/T2c/T2d`, `D-PA (A)`, `INV-5-A`, `R8`, `N06`, "RB's C-4 probe", "RULING §0" — undecodable without the external roadmap. This directly contradicts the repo's own recorded doc preference (strip `Phase N`/meta-narration and lane ids from docstrings). A new maintainer cannot learn what the module *does* without three other documents; the load-bearing 3-line "fill-down then scroll" rule is buried under provenance narration.

### LOW
- [New Hire] `bottom = ScrollRegionManager.region_top(region)` (~L211): a function named `region_top` returns the value the code binds as `bottom` (the history region's last/split row). Inherited upstream naming, but at this call site it reads as a contradiction and invites off-by-one edits.
- [New Hire] `cup/1` (~L303) hand-rolls raw `"\e[#{row};1H"` while save/restore route through `Dialect.cursor_save/0`/`cursor_restore/0`. The module calls `Dialect` "the shared wire vocabulary" yet positioning bypasses it — the one place a future terminal-dialect quirk in CUP would need to live is off the seam.
- [Saboteur] `resize/3` telemetry metadata (~L282-283) reads `old_region.region_top`/`new_region.region_top` as raw struct fields while the rest of the module uses the `ScrollRegionManager.region_top/1` accessor — couples to struct internals inconsistently.
- [Saboteur] `new/5` (~L151) guards only `width > 0`; `rows`/`footer_rows` are spec-only (`pos_integer`/`non_neg_integer`), unenforced. A `footer_rows >= rows` caller silently produces a degenerate 1-row region rather than an error. No crash (ScrollRegionManager's `region_top/2` clamps `>= 1`), but the failure is silent and only visible as squashed output.

### Cross-Persona Overlaps
- Unvalidated `iodata` is the shared root of the HIGH (Security: escape injection wipes scrollback / repaints sealed rows) and the top MEDIUM (Saboteur: unterminated line overwrites a sealed row). One fix — validate/reject control bytes and require newline-termination in `seal/2` before emit — closes both. The `with_cursor` nesting hole is an independent second class of unenforced-prose-contract.

### Verdict: CONCERNS
One HIGH (terminal escape injection through unvalidated sealed content) plus three MEDIUM. No CRITICAL and fewer than 3 HIGH, so not a BLOCK, but the immutable-prefix keystone this unit exists to guarantee is only as strong as its untrusted `iodata`, and the newline/nesting contracts are enforced only in prose. Recommend validating `iodata` (control-byte rejection + newline-termination) and adding a nesting guard to `with_cursor/3` before this substrate is built on.

merklebonsai:
Thanks Drew — round-2 findings triaged below. Everything except the moduledoc pass landed on the branch (review-response commit `75d92d5`, plus the T2a-side rename `c8debeb`).

### Accepted + fixed
- **[HIGH] Escape injection through unvalidated sealed `iodata`** — `75d92d5`: new shared `PaintAuthority.ContentGuard` (reusable by T2c); `seal/2` runs content through `ContentGuard.sanitize_line/1` before `append_sealed/2` ever sees bytes — allowlists printable text, SGR, and `\t`/`\r`/`\n`; all other C0/CSI/OSC/DCS/bare-ESC get the ESC byte stripped so residue stays *visible* (`[2J`) instead of executing. Red-first suite proves each vector (full clear, sealed-row repaint, OSC title side effect, truncated CSI merging the following line) is oracle-visible pre-guard, then neutralized through the real `seal/2` path.
- **[MED] Newline-termination contract enforced nowhere** — `75d92d5`: `seal/2` now raises `ArgumentError` unless `iodata` is `\r\n`-terminated; the docstring MUST is code now, so a caller that forgets can't silently freeze `next_row` and overwrite a sealed row.
- **[MED] `with_cursor/3` "sole owner" has no nesting guard** — `75d92d5`: new `in_cursor_bracket` flag set on the state passed into `fun`, cleared on exit; a nested `with_cursor/3` — or a `seal/2` reached from inside a bracket (the T2c footer path you flagged) — raises instead of clobbering the single DECSC register.
- **[LOW] `region_top` returning the bottom row** — `c8debeb`: renamed `region_top` → `history_bottom` across `ScrollRegionManager` and all call sites; `75d92d5` completed the follow-through in `InlineAuthority`. The `PaintAuthority` behaviour callback keeps its `region_top` name — that contract is shared and unchanged.
- **[LOW] `cup/1` hand-rolls CUP off the `Dialect` seam** — `75d92d5`: `cup/1` deleted; positioning now goes through a new `Dialect.cursor_position/1`, byte format pinned by test, so the wire builder is defined exactly once.
- **[LOW] `resize/3` telemetry reads raw struct fields** — `75d92d5`: metadata now goes through the `ScrollRegionManager.history_bottom/1` accessor, consistent with the rest of the module.

### Declined
- **[LOW] `new/5` should guard `rows`/`footer_rows` (`footer_rows >= rows` silently degenerates)** — declining the raise: a too-short terminal is a runtime environment condition, not a caller bug, and we don't want a resize race to crash the authority. The *silent* part is already fixed upstream in the T2a response (`0913535`): `ScrollRegionManager` records `degenerate?: true` on the struct and emits the honest full-region release (`CSI r`) instead of a lying `CSI 1;1r`, so callers gate on `degenerate?/1` rather than discovering squashed output.

### Deferred
- **[MED] Moduledoc is lane-id shorthand undecodable without the roadmap** — agreed it violates our own recorded doc preference; deferring the de-jargon pass to the already-planned rebase onto master once #574 lands, so the whole T2 lane gets one consistent rewrite instead of per-PR drift. Exit criterion: before this branch merges to master, the moduledoc is self-contained — lane ids (`T2a`/`D-PA (A)`/`INV-5-A`/`N06`/`R8`) replaced with plain statements, provenance narration cut, and the load-bearing fill-down-then-scroll rule promoted to the top. (`ScrollRegionManager` already gained a codename glossary in `0913535` as an interim aid.)

PR #590, opened 2026-07-16
Harness UI: pinned footer viewport (footer-scoped diff, no \e[2J keyframe) [T2c]

merklebonsai:
Unit **T2c** — the **pinned footer viewport**: a buffer-diff pipeline scoped to the footer rows only (live tail + status strip + composer). Completes the inline-substrate spine: T2d (inline driver) → T2a (DECSTBM split, #574) → T2b (append path, #589) → **T2c**.

## What
Extends `InlineAuthority` (+169/-7 — the diff logic composes directly on T2b's struct/cursor protocol, a separate module would only add indirection):
- **`footer_diff/2`** — pure minimal diff over footer line lists; unchanged rows emit zero bytes; a shrunk footer emits `CUP + \e[K` clears for stale rows (no ghosts).
- **`repaint/2`** — diff-driven footer repaint; every emitted CUP lands in `footer_range` (rows `region_top+1..rows`), never a history row. All writes CUP-anchored, bracketed by T2b's `with_cursor` (sole `\e7`/`\e8` owner).
- **`keyframe/2`** — full footer redraw (Ctrl-L recovery): per-row `CUP + \e[K + line`, footer region ONLY — never `\e[2J`/`\e[3J` (the ban RB's N06 validated: full clears wipe scrollback on wezterm/kitty).
- **`resize/3` contract** — delegates region math to `ScrollRegionManager.resize/2` (now geometry-gated per the #574 review fix) and deliberately does NOT auto-repaint: callers compose `resize(w,h) |> keyframe(lines)`. Pinned from the resize side by a T2b regression test; documented for the T13a assembler.
- **`degenerate?/1`** — thin delegation surfacing T2a's degenerate-geometry signal (rows−footer < 2 ⇒ footer can't be pinned) so the assembler can adapt instead of silently losing the pin.

## Acceptance (roadmap T2c, all byte-stream properties on the REAL producer → StringIO → real oracle replay)
1. Footer repaint touches only footer rows — property over random content changes; O1 `cup_rows` scanner (fail-closed) on real emitted CSI.
2. Resize: no `\e[2J`, no history rewrite — `resize |> keyframe` composition asserts footer-only CUPs in the NEW range + O2 immutable-prefix over the whole raw stream.
3. Ctrl-L keyframe: footer-scoped clears + repaint, zero full-clear, zero history addressing.
4. R-N4 contrast — characterizes `Backends.build_terminal_frame`'s `\e[2J`-on-width-change (the wrong behavior T2c exists to replace) vs this path emitting none.

**Fail-first is genuine, twice over:** adversarial suite drives BuggyAuthority bleed variants (absolute CUP / CUU-relative / VPA), a raw-bypass write, and a row-1 history stomp through the real `SealOracle.replay → Emulator → immutable_prefix?` — each RED; paired GREENs prove the oracle isn't stuck. Plus a content-correctness pin: temporarily stripping `\e[K` makes the shrink test fail with the exact stale-residue class (`"hiAAAA…"` vs `"hi "`), restored green — rendered footer content is asserted via emulator replay, not just CUP confinement.

**Degenerate geometry tested:** rows=2/footer=3 (clamps, in-bounds), rows=1/footer=2 (empty range, zero CUPs), rows=4/footer=2 (control). Footer line width is a documented CALLER contract (pre-truncate via `Raxol.UI.TextMeasure`) — stated for T13a.

## Review
Opus analytic review: **SHIP, no RED.** DECSTBM cursor-home side effect traced SAFE through `resize |> keyframe` (every content write CUP-anchored — no path assumes inherited cursor position); diff minimality verified for shrink/equal/multi-byte-UTF-8; 3 of 4 yellows folded into this commit (footer content-render pin, degenerate tests + signal, width-contract doc); Y4 declined (interleaving property's confinement+append-only+no-2J proxy deemed sufficient).

Gates: `compile --warnings-as-errors` clean; 42 tests green across the 4 renderer suites (8 properties, 34 tests) + T2a/T2b suites re-run green on the rebased chain; `format` + `credo --strict` clean.

**Base note:** stacks on `feat/harness-ui-T2b` (#589), which stacks on #574. Merge order: #574 → #589 → this.

---
*Terms: **keyframe** = full repaint of the footer region (e.g. Ctrl-L recovery) — footer rows only, per-row `\e[K`, never `\e[2J` (which wipes native scrollback on wezterm/kitty — measured, #585). **R-N4** = the characterization test pinning today's full-frame path doing exactly that wrong thing. **Degenerate geometry** = `rows − footer_rows < 2`, where DECSTBM can't pin (see #574's `degenerate?`).*

### DIFF
--- a/lib/raxol/ui/rendering/paint_authority/inline_authority.ex
+++ b/lib/raxol/ui/rendering/paint_authority/inline_authority.ex
@@ -612,7 +612,7 @@ defmodule Raxol.UI.Rendering.PaintAuthority.InlineAuthority do
   or whose width-correct truncation went stale / whose sealed history rewrapped
   (horizontal). This is independent of `reflow_capable?/1`/the telemetry hook
   below: the ghost-content risk applies to every terminal's footer, not just the
-  (B)-detection subset.
+  reflow-capable subset.
   """
   @impl true
   def resize(%__MODULE__{region: region, next_row: next_row} = t, width, height) do
@@ -636,13 +636,14 @@ defmodule Raxol.UI.Rendering.PaintAuthority.InlineAuthority do
     reflow_relevant? = geometry_changed? or width_changed?
 
     if reflow_relevant? and t.reflow_capable? do
-      # The (B)-detection seam firing: this session's terminal is known,
-      # from RB's real-hardware C-4 probe, to reflow sealed history
-      # cleanly on a geometry-changing resize. NOTHING is re-emitted here
-      # — re-emission is a FUTURE unit's job (D-PA RULING: ship (A), (B)
-      # is a runtime-detected ADDITIVE upgrade). This telemetry event is
-      # the thin hook that unit gates on — it carries BOTH axes so the (B)
-      # unit can tell a rewrapping width change from a pure row change.
+      # The reflow-aware detection seam firing: this session's terminal
+      # is known, from the real-hardware terminal-matrix probe, to reflow
+      # sealed history cleanly on a geometry-changing resize. NOTHING is
+      # re-emitted here — re-emission is a FUTURE unit's job (ship
+      # seal-time-only now; reflow-aware re-emission is a runtime-detected
+      # ADDITIVE upgrade). This telemetry event is the thin hook that unit
+      # gates on — it carries BOTH axes so the future unit can tell a
+      # rewrapping width change from a pure row change.
       :telemetry.execute(
         [:raxol, :ui, :paint_authority, :reflow_capable_resize],
         %{},
--- a/lib/raxol/ui/rendering/paint_authority/inline_authority.ex
+++ b/lib/raxol/ui/rendering/paint_authority/inline_authority.ex
@@ -174,6 +174,44 @@ defmodule Raxol.UI.Rendering.PaintAuthority.InlineAuthority do
   simply operate over whatever footer capacity `footer_range/1` actually
   reports for that geometry, which may be smaller than `footer_rows`
   requested, or empty.
+
+  ## T2c review response: footer `ContentGuard`, and the `needs_keyframe` latch
+
+  Two FIX-NOW findings from review, both closed here:
+
+    * **Footer content is not trusted either (HIGH).** T2b's review
+      already established that `seal/2` cannot write agent/LLM-originated
+      iodata verbatim (`ContentGuard.sanitize_line/1`); the footer path
+      carries the SAME kind of content (live-tail/agent text) through the
+      SAME risk (a footer line smuggling `\\e[3;1H`/`\\e[2J`/etc. would
+      execute against the footer confinement invariant exactly like an
+      unguarded history append would against the seal-once invariant).
+      `repaint/2` and `keyframe/2` both run every caller-supplied line
+      through `ContentGuard.sanitize_line/1` at entry — BEFORE padding or
+      diffing — so `footer_diff/2` and `footer_lines` only ever see
+      already-neutralized content. `footer_lines` itself is therefore an
+      invariant: once sanitized in, never re-sanitized out, so `footer_diff/2`
+      comparing old (already-sanitized) against new (freshly-sanitized) is
+      always an apples-to-apples comparison.
+    * **A stale post-resize repaint could leave ghost content (MED).**
+      `resize/3` clamps `next_row` but, by design (see above), never
+      repaints the footer — a geometry-changing resize can relocate the
+      footer's on-screen rows to different absolute row numbers while their
+      CONTENT (and thus `footer_diff/2`'s logical, index-based comparison)
+      is unchanged. A subsequent `repaint/2` call with unchanged lines then
+      computes a no-op diff and writes NOTHING, leaving those rows showing
+      whatever was on screen at that position before the resize (leftover
+      history text, or a stale row `repaint/2` previously left blank via
+      `\\e[K` at the OLD position). The `needs_keyframe` flag closes this:
+      a geometry-changing `resize/3` sets it (a pure state change — the
+      pinned T2b regression test asserting resize's ONLY new bytes are
+      `ScrollRegionManager`'s single DECSTBM re-set is untouched, since
+      setting a struct field emits no bytes); the NEXT `repaint/2` call
+      checks it FIRST and, if set, self-promotes to a full `keyframe/2`
+      (which clears the flag) instead of running its normal diff — so the
+      first repaint after any geometry change always fully re-renders the
+      footer at its current position, regardless of whether the logical
+      content changed.
   """
 
   @behaviour Raxol.UI.Rendering.PaintAuthority
@@ -190,7 +228,8 @@ defmodule Raxol.UI.Rendering.PaintAuthority.InlineAuthority do
     :reflow_capable?,
     :next_row,
     in_cursor_bracket: false,
-    footer_lines: []
+    footer_lines: [],
+    needs_keyframe: false
   ]
 
   @type t :: %__MODULE__{
@@ -199,7 +238,8 @@ defmodule Raxol.UI.Rendering.PaintAuthority.InlineAuthority do
           reflow_capable?: boolean(),
           next_row: pos_integer(),
           in_cursor_bracket: boolean(),
-          footer_lines: [binary()]
+          footer_lines: [binary()],
+          needs_keyframe: boolean()
         }
 
   @doc """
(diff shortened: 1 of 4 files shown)


DROOdotFOO:

Scope: `lib/raxol/ui/rendering/paint_authority/inline_authority.ex` (+207/-7) plus the two new property suites. Read in full against the T2a `ScrollRegionManager` it composes on. The footer geometry math (CUP row = `region_top+1+idx`, bounded by `footer_range/1`) checks out for the normal and degenerate geometries. The findings below are about what the diff logic does NOT defend against.

### HIGH
- [Security Auditor / Saboteur / New Hire] `inline_authority.ex:398` (`footer_row_bytes/2`), reached via `repaint/2:357` and `keyframe/2:385`, emitted at `repaint_footer/2:305` / `keyframe_footer/2:311` — **the footer content `line` is written to the device verbatim, with no control-sequence sanitization.** The entire PR's headline invariant ("every emitted CUP lands in `footer_range`, never a history row; `\e[2J` is never emitted") is enforced only over the bytes THIS module generates (`cup/1` + `\e[K`). It is not enforced over the caller's content. The footer is documented as "live tail + status strip + composer" — the live tail is streamed agent/LLM/tool output, which is untrusted and routinely contains ANSI. Concrete failure: a footer line `"tail: \e[1;1Hpwned"` executes the embedded CUP at emit time, moving the cursor to row 1 (inside the sealed history region) and writing "pwned" there — a direct INV-2 confinement break and an immutable-prefix (seal-once) violation. A line containing `\e[2J` wipes native scrollback on wezterm/kitty — the exact outcome N06 and this whole module exist to prevent — defeated through content rather than through an emitted clear. This needs no malice: ordinary colored CLI/tool output (embedded SGR is fine, but embedded cursor moves / `\e[K` / erase sequences are common) corrupts the layout. The "Caller contract" moduledoc section (`inline_authority.ex:150-166`) addresses only line *width* (wrap) and points callers at `TextMeasure` — display-width truncation does not strip escapes and would not catch this, so a diligent caller who follows the documented contract still ships the bug. Untested: every property/adversarial test drives alphanumeric-only content (`line_gen` = `string(:alphanumeric, ...)`, `renderer_footer_property_test.exs:120`), so the O1 `cup_rows` confinement scanner never sees an escape smuggled through legitimate `repaint/2`. Fix: strip/neutralize C0/C1 and CSI/OSC sequences from each `line` before `footer_row_bytes/2` (or assert the caller contract as a hard, documented, tested precondition), and add an adversarial content generator that injects `\e[...H`/`\e[2J` into footer lines. Promoted (3 personas, same root, defeats the core PR guarantee).

### MEDIUM
- [Saboteur] `resize/3:447-478` updates `region`, `width`, and `next_row` but deliberately does NOT reset `footer_lines`. The on-screen footer content is not moved by resize (only the DECSTBM split is re-emitted), so after a geometry change the previously-painted rows sit at their OLD absolute positions while `footer_range/1` now points at DIFFERENT rows. The only protection is the documented `resize |> keyframe` composition. Failure: a caller that does `resize |> repaint` (documented as needing `keyframe`, but nothing enforces it) diffs the new lines against stale `footer_lines`; unchanged rows emit zero bytes and are never repainted at their new positions, leaving ghost rows / stale history-adjacent residue. This is a one-line safety improvement: setting `footer_lines: []` in `resize/3` turns the next `repaint/2` into an automatic full repaint (all rows diff as changed) without touching the "resize emits only DECSTBM" byte-count regression test (resize still emits no footer bytes). A silent-ghost footgun becomes a self-healing full redraw.

### LOW
- [New Hire] `footer_diff/2:322-324` is a public, `@doc`'d function whose clause guard requires `length(old_lines) == length(new_lines)`. Called with mismatched lengths it raises a cryptic `FunctionClauseError` rather than a clear contract error. Internal callers always pre-pad, but the function is exported as if it were a general-purpose diff; a `raise ArgumentError` with a message, or explicit docs that it is internal-only, would be kinder to the next caller.
- [New Hire] `keyframe/2:388-389` always opens the `with_cursor` save/restore bracket even when `footer_row_count/1` is 0 (degenerate `rows=1/footer=2` geometry), emitting `\e7`/`\e8` with nothing between. Harmless, but contradicts the "footer-scoped, zero bytes when nothing to do" mental model the `repaint/2` no-op path (`:349`) establishes; a maintainer reasoning about byte output will be briefly surprised.
- [Saboteur] `repaint/2:349-350` no-op branch returns `t` without writing back `padded_new` to `footer_lines`. After a resize-induced count change, `footer_lines` can retain the old length; because `pad_rows/2` is deterministic subsequent diffs stay consistent, so this is benign today, but it means `footer_lines` does not always equal "what is on screen at the current count" — an implicit invariant a future reader may over-trust.

### Cross-Persona Overlaps
- The escape-injection gap (`footer_row_bytes:398`) was reached independently by all three personas: Saboteur (benign ANSI tool output corrupts layout / cascading render breakage), Security Auditor (untrusted live-tail content escapes footer confinement and wipes scrollback), New Hire (the confinement guarantee rests on an undocumented "content is escape-free" assumption that the documented `TextMeasure` width contract does not cover). Same root, promoted to HIGH.

### Verdict: CONCERNS
One HIGH (footer confinement is not enforced over caller content — the PR's central invariant is bypassable through the footer's own untrusted live-tail data, and untested) plus a MEDIUM resize/footer_lines ghost footgun. Neither is a correctness bug in the byte-generation math, which is sound; both are missing defenses at the content boundary. Recommend: sanitize control sequences from footer lines (or make it a hard, tested caller precondition) and add an escape-injection content generator to the adversarial suite before merge.

merklebonsai:
Thanks Drew — round-2 findings adjudicated below; all five landed in e037876 (`fix(harness-ui): T2c review response`), with RED/GREEN regression tests in `test/property/renderer_t2c_review_fixes_test.exs`.

### Accepted + fixed
- **HIGH — footer content escape injection (`footer_row_bytes/2`)**: e037876 — every footer line now passes through `ContentGuard.sanitize_line/1` (the same guard T2b's review added on the history append path) at `repaint/2` and `keyframe/2` entry, via `sanitize_and_pad/2`, *before* padding/diffing — so `footer_diff/2` and the `footer_lines` field only ever hold neutralized content. Adversarial RED/GREEN pairs added for a smuggled history-row CUP (`\e[3;1H`), `\e[2J`, and `\e[K`, plus a GREEN pinning that legitimate SGR (`\e[1;31m…\e[0m`) passes through byte-identical.
- **MEDIUM — `resize/3` leaves stale `footer_lines` (ghost rows on `resize |> repaint`)**: e037876 — fixed with a stronger mechanism than the suggested `footer_lines: []` reset: `resize/3` sets a `needs_keyframe` latch (pure state change — the pinned "resize emits only DECSTBM" byte regression is untouched), and `repaint/2` checks it first, self-promoting to a full `keyframe/2` (which clears the flag). Wrong composition is now structurally impossible rather than documented-only; tests cover the latch lifecycle and a stale blank row getting its own `\e[K` at the NEW footer position after resize.
- **LOW — `footer_diff/2` cryptic `FunctionClauseError` on length mismatch**: e037876 — now raises `ArgumentError` with an explicit message; regression test added.
- **LOW — `keyframe/2` emits empty `\e7`/`\e8` bracket at zero footer rows**: e037876 — zero-row geometry now early-returns with no `with_cursor` bracket at all (zero bytes), matching the `repaint/2` no-op mental model; tested.
- **LOW — `repaint/2` no-op branch doesn't write back `padded_new`**: e037876 — the no-op branch now writes back the padded list, so the "`footer_lines` equals what's on screen at the current count" invariant holds unconditionally; tested.

Gates re-run green on the amended branch: `compile --warnings-as-errors`, `format`, `credo --strict`, and the T2a/T2b/T2c suites including the new review-fix suite.

merklebonsai:
Follow-up from the watched RB width-resize probe (5-terminal matrix): `5d6fc5f7a`.

The probe showed reflow is a **horizontal** phenomenon — 4/5 terminals soft-wrap sealed history on a width change (kitty is the no-reflow exception; iTerm2 bottom-anchors). But `resize/3` gated both `needs_keyframe` and the `reflow_capable_resize` telemetry on `geometry_changed?/2`, which watches the **vertical** axis (`history_bottom`) and is blind to width. So a width-only resize — the exact thing that triggers reflow — set neither the latch nor the seam the future (B) reflow unit gates on.

One predicate was doing two jobs on orthogonal axes. Region re-emission stays correctly vertical-only (a width resize needs zero DECSTBM bytes — the pinned T2b regression is untouched). But keyframe + reflow are horizontal: width-shrink can wrap an untruncated footer line past the pin (and with no content change, no repaint fires to fix it), so `needs_keyframe` now sets on `geometry_changed? or width_changed?`, and the reflow telemetry fires on the width axis carrying `old_width`/`new_width` so (B) can tell a rewrapping width change from a pure row change.

Red-first: inverted the prior test that asserted "width-only resize does NOT set needs_keyframe" (that was the bug); added a telemetry-on-width-shrink test. T2c suite 20/20, T2b/footer/seal 66/66 unchanged.

PR #591, opened 2026-07-16
Harness UI: status strip (stage+elapsed, needs-input, missing-data honesty) [T10]

merklebonsai:
Unit **T10 — status strip**. Pure projection `state → [footer line]` for the pinned strip: turn stage + elapsed ticker (never a bare spinner — pain point P4), needs-input, context %, session cost. Consumed by T13a as pre-truncated footer lines for T2c's `repaint/keyframe`.

- **Working-vs-hung**: stage + elapsed always ("plan 4s" → "plan 22s ⏳" at 15s → "HUNG plan 1m05s" at 60s; thresholds overridable). Time is injected ints (R11 — zero clock reads; a property proves time-shift invariance).
- **Missing-data honesty**: every absent field renders `—` in its labeled slot; the roadmap's "never a stale % from the prior turn" is enforced structurally (`context_pct` gated on `turn_completed === true` — fail-safe, never a wrong number). The live-vs-snapshot semantics of `context_pct` are explicitly surfaced in the moduledoc as **T13a's open decision** (if the producer is live, the gate relaxes to a `context_fresh` liveness flag — the "never stale" rule stays).
- **Width discipline**: display-width via `Raxol.UI.TextMeasure` everywhere (the strip emits its own wide glyphs — ⏳ — so naive `String.length` would overflow the pinned region); priority drop at narrow widths Cost→Ctx→Stage→Input, with a property asserting survivors are always a strict prefix of the priority order.
- Red-first: truncation disabled → width test + property red (17 cols into width 5), restored green.

Opus review: **SHIP, no RED.** Real-producer suite, genuine red-first, display-width confirmed correct on the multi-byte path, HUNG display honors "never a bare spinner", R11 clean. Strong yellow (context_pct producer-model ossification) resolved by making the moduledoc surface it as T13a's decision instead of asserting it settled.

Gates: compile --warnings-as-errors clean; 26/26 green (+739 harness/status-bar suite green); format + credo --strict clean.


DROOdotFOO:

Scope: `lib/raxol/harness/status_strip.ex` (+ two test files). Pure `state -> [footer_line]` projection. Reviewed against idiomatic Elixir/OTP and the repo's own pinned-region / TextMeasure contracts.

### HIGH

- **[Saboteur] status_strip.ex:277 (`render_stage_elapsed` warn branch) + :298 (`fit_to_width`)** — the warn-state glyph `⏳` (U+23F3, HOURGLASS WITH FLOWING SAND) is measured as **1 column** by `TextMeasure.display_width`, but renders as **2 columns** in every modern terminal (iTerm2/kitty/wezterm/gnome-terminal; U+23F3 has `Emoji_Presentation=Yes` -> UAX#11 Wide). `Raxol.Terminal.CharacterHandling.wide_char?/1` (character_handling.ex:17-49) has no range covering 0x2300–0x23FF — its widest low range is `0xFF01–0xFF60` / `0x1F300–0x1FAFF`, so `get_char_width(0x23F3)` falls through to `1`. `fit_to_width` measures every candidate with this undercount, so during the 15s–60s window the emitted footer is **1 column wider than the requested width**, overflowing the pinned region T2c writes byte-level (wrap/corruption of the pinned strip). This directly defeats the module's headline guarantee and contradicts the moduledoc's own claim (status_strip.ex:~168) that it uses TextMeasure *because* "naive `String.length` would overflow" on ⏳ — TextMeasure returns exactly the same wrong `1` that `String.length` would. Failure scenario: turn at 20s elapsed, terminal width == full-strip width -> strip renders `width+1` columns -> pinned region overflows by one cell every render tick.

- **[Security] status_strip.ex:259 (`stage_label/1` -> `to_string(stage)`)** — `turn_stage` (`@type turn_stage :: atom() | String.t()`) is interpolated verbatim into the footer line with zero sanitization, then handed to T2c's byte-level pinned-region writer. A stage string containing an ANSI/control sequence (`"\e[2J\e[H"`, `"x\e[31m"`) or a newline (`"a\nb"`) is written straight into the terminal: escape sequences corrupt/clear the screen (classic terminal-escape injection, CWE-150) and a `\n`/`\r` breaks the single-line pinned-region contract. `TextMeasure.display_width` counts ESC and each following byte as width 1 (character_handling.ex `get_char_width` returns 1 for all control chars), so the width bound *passes* while the real byte stream both overflows and injects. In an agent harness, `turn_stage` plausibly derives from tool names or model-emitted event labels, i.e. it can carry adversarial bytes. (Promoted to HIGH: shares the "display_width does not guarantee safe/real column output" root with the ⏳ finding above.)

### MEDIUM

- **[New Hire] t10_status_strip_test.exs:~205 (`without_cost_width = full_width - 12`)** — the drop-Cost-first test hardcodes the magic constant `12` tied to the exact rendered length of `" | Cost: $1.23"`. Any change to the cost format, separator, or the `@wide_state` cost value silently shifts the boundary and the test no longer exercises "just barely too narrow for Cost," yielding false-green coverage of the degradation edge. Derive the width from the actual rendered segments instead.

- **[New Hire] t10_status_strip_property_test.exs:~63 (`maybe_add_turn_completed/1`)** — the generator force-injects `turn_completed: true` whenever `context_pct` is present, so the "each present field renders its slot" property *never* generates the real producer state (`context_pct` present, `turn_completed` false/absent). The whole "never a stale %" gate — the unit's most safety-relevant branch — is therefore covered only by hand-picked example tests, and the property gives false confidence that "context_pct present => renders a number." Property space and the structural gate are silently decoupled.

### LOW

- **[Saboteur] status_strip.ex:249 (`cost_value`)** — `:erlang.float_to_binary(cost / 1, decimals: 2)` has no domain guard: a negative `cost` renders `"$-1.23"` and a non-finite/huge value renders oddly; `cost` is trusted numeric today so impact is cosmetic, but the "honest slot" intent argues for a guard.
- **[Saboteur] status_strip.ex:192 (`render/2` guard)** — `when is_map(state) and is_integer(width)`; a non-integer `width` (e.g. a float from an upstream layout calc) matches no clause -> `FunctionClauseError`, despite the moduledoc promising "never raises." Minor given callers pass integer columns.
- **[New Hire] status_strip.ex:1-186** — ~130 lines of moduledoc prose (open-question essays, cross-unit rationale) before any code. Thorough, but the ⏳ width claim in it is now factually wrong (see HIGH #1); at minimum that paragraph must be corrected.

### Cross-Persona Overlaps

- Saboteur (⏳ undercounted) and Security (control bytes undercounted + injected) are the same root defect: `TextMeasure.display_width` / `CharacterHandling.get_char_width` is an incomplete, non-sanitizing width oracle, and this module treats its output as a hard safety bound for a byte-level pinned region. The module's entire value proposition (width discipline, no raw ANSI in the pinned region — a CLAUDE.md hard rule) is not actually enforced against real terminal bytes. This overlap is why the Security injection finding is promoted to HIGH.

### Verdict: CONCERNS

Two HIGH findings, both undermining the module's central width/safety contract, plus two MEDIUM test-integrity gaps. Not a BLOCK (no CRITICAL, fewer than 3 HIGH), but the ⏳ overflow and unsanitized `turn_stage` should be fixed before merge: (1) add U+2300–U+23FF emoji-wide handling (or special-case the emitted glyph set) and (2) strip control/ANSI bytes from `stage_label/1` before interpolation.

merklebonsai:
Thanks Drew — round-2 response below. All fixes landed in `adbf6ec` (`fix(harness-ui): T10 review response`); 4 of 7 findings accepted and fixed, 2 declined, 1 root-cause item deferred upstream.

### Accepted + fixed
- **HIGH — ⏳ (U+23F3) measured 1 col, renders 2 (width overflow in the 15s–60s window)**: `adbf6ec` — replaced the hourglass with the ASCII word `SLOW` (width-honest by construction); added a `glyphs/0` accessor enumerating every non-ASCII glyph the module can emit (`—`, `…`) plus a width-honesty regression test asserting each measures exactly 1 column per `TextMeasure`, so a reintroduced ambiguous glyph fails loudly.
- **HIGH — `turn_stage` interpolated unsanitized into the byte-level pinned writer (CWE-150)**: `adbf6ec` — added `sanitize_stage/1` stripping C0 controls + DEL (`[\x00-\x1F\x7F]`, covers ESC/CR/LF) before interpolation; red-first tests cover an ESC-laden stage (`"plan\e[2J"` → `"plan[2J"`), newline collapse, and sanitize-plus-elapsed composition.
- **MEDIUM — drop-Cost test hardcodes magic constant `12`**: `adbf6ec` — boundary now derived from the actual rendered `Cost` segment + separator via `TextMeasure.display_width`, so format changes shift the boundary with the test instead of rotting it.
- **MEDIUM — property generator force-injects `turn_completed: true`, never exercising the staleness gate**: `adbf6ec` — new fourth property generates `turn_completed` independently (`:absent`/`false`/`true` crossed with `context_pct` absent/present) and asserts `context_pct` renders numerically **iff** `turn_completed === true`, `—` otherwise. The gate is now property-covered, not just example-covered.
- **LOW — moduledoc's ⏳ width claim factually wrong**: `adbf6ec` — the incorrect paragraph is corrected and replaced by a "Glyph width honesty" section documenting the exact `CharacterHandling` gap and the upstream fix path.

### Declined
- **LOW — `render/2` `is_integer(width)` guard raises on float width, "despite moduledoc promising never raises"**: the moduledoc promises "never raises **on missing keys**" — a state-honesty guarantee, not a type-laundering one. A float width is a contract violation from an upstream layout calc; silently rounding it would paper over the real bug, and a `FunctionClauseError` at the boundary is the loud failure we want. Guard stays as-is.
- **LOW — `cost_value` domain guard for negative/non-finite cost**: declined for the same honest-slot reason the finding cites. `cost` is producer-trusted numeric from the harness's own accounting; if it goes negative, `"$-1.23"` is the *honest* rendering of a real accounting bug — masking it to `—` would hide exactly the signal an operator needs.

### Deferred
- **Cross-persona root cause — `TextMeasure`/`CharacterHandling` is an incomplete width oracle (no Emoji_Presentation coverage below 0x1F300)**: the module-level mitigation is complete (ASCII-safe glyph set + control-byte sanitization + the `glyphs/0` regression gate), but the true fix belongs upstream in `Raxol.Terminal.CharacterHandling`'s width table, which is shared infrastructure beyond this PR's scope. **Exit criterion**: once `wide_char?/1` correctly flags Emoji_Presentation glyphs as 2 columns, the width-honesty test unblocks richer glyphs here — the path is documented in the moduledoc and the regression test.

Gates re-run post-fix: compile `--warnings-as-errors`, format, credo `--strict`, full T10 suite (examples + 4 properties) green.

DROOdotFOO:
still failing CI

PR #592, opened 2026-07-16
Harness UI: keybind layer (canonical events → typed commands, composer-safe) [T12]

merklebonsai:
Unit **T12 — keybind layer**. Pure keymap: T27 canonical input events (+ mode context) → typed commands. The data-table design (`binds/0`) is the invocation-parity seam T15's palette enumerates — no hidden case clauses (asserted by test: the emitted-command union equals the table).

**v1 binds:** ESC → `:interrupt` (`:always` — AD-1: interrupt must never be swallowed by focus) · Tab → `:steer` (`:always`; "Tab=queue / Enter=now" per the protocol spec — Enter stays the composer's) · `z` → `:fold_toggle` · `j`/`k` → `:jump_next`/`:jump_prev` (all three `:not_composing`-guarded — the direct fix for the prototype's "keymap stole j/k/s/z from typing" bug).

**Shape-match is proven, not assumed:** the suite drives the REAL T27 emitters (`EventTranslator.translate/1`, `InputParser.parse/1`, `Event.key_event/3` — the exact driver dispatch path) through the keymap across all three wire shapes. If the driver emitted `:esc` instead of `:escape`, the suite goes red.

**Command shape:** plain `%{type: atom, payload: map}` — field-identical to `%Raxol.Agent.Command{}` (main raxol cannot depend on raxol_agent; map-across-boundary per repo convention; conversion is literally `struct(Command, cmd)` for the agent-bound types).

Red-first: guard mutated to always-pass → composing-guard property red, StreamData shrank to exactly `"j"`; restored green (23/23).

Opus review: **SHIP, no RED.** Four yellows, all T13a/T13b seam preconditions (recorded in the STATE doc as T13a acceptance inputs):
1. T13a MUST dispatch keymap-first for `:always` binds — T11's composer catch-all delegates every unhandled key into MultiLineInput, so component-first wiring would kill both ESC-interrupt and Tab-steer.
2. T13b conversion bifurcation: `:interrupt`/`:steer` cross to the agent lane as `%Command{}`; `:fold_toggle`/`:jump_*` stay UI-local.
3. ESC-always vs ESC-to-close overlays (T14) — resolve in T13a's focus model.
4. `streaming?` context field forward-declared, currently unread.

Gates: compile --warnings-as-errors clean; 23/23 + T27 (66) + T11 composer (46) green together; format + credo --strict clean.

### DIFF
--- a/lib/raxol/ui/harness/keymap.ex
+++ b/lib/raxol/ui/harness/keymap.ex
@@ -22,7 +22,22 @@ defmodule Raxol.UI.Harness.Keymap do
 
     * **tui-steal rule** (chords replace keys later without restructuring
       the dispatch logic -- a bind's `match` grows a `mods:` requirement,
-      the walking loop does not change).
+      the walking loop does not change). This is not aspirational: a
+      `key:`-kind bind with no declared `mods:` matches only a BARE
+      keypress (no ctrl/alt/meta held) -- Ctrl+Tab, Alt+Tab, and Meta+Tab
+      all fall through to `:passthrough` rather than firing `:steer`,
+      because they are shortcuts wired to something else, not "Tab". An
+      entry that DOES declare `mods:` (a full `InputEvent.mods()` map)
+      requires an EXACT match against the normalized event's `mods`
+      instead of the bare-keypress default -- that is the chord's match
+      spec growing a field, per the promise above, with `resolve/2`'s
+      walking loop untouched. Shift alone is not restricted by the
+      bare-keypress default (mirrors `InputEvent.text?/1`: shift-only is
+      not a shortcut), so Shift+Tab still resolves through the plain
+      `:tab` bind. `char:`-kind binds need no separate mods check --
+      `InputEvent.printable_char/1` already returns `nil` whenever
+      ctrl/alt/meta is held, so a modifier-qualified letter never reaches
+      the char-kind match branch in the first place.
     * **Invocation parity with T15's palette**: the palette can enumerate
       `binds/0` and invoke `command_for/2` on any entry directly (a command
       palette is just another way to select a bind, not a second source of
@@ -49,6 +64,38 @@ defmodule Raxol.UI.Harness.Keymap do
       fix: they only resolve to a command when focus is NOT the composer
       (browsing the transcript), and fall through to `:passthrough`
       (ordinary character insertion) while composing.
+
+  ### Fail-safe default: missing `composing?` means composing
+
+  A caller that omits `composing?` from `context()` entirely (forgets it,
+  or is a code path -- palette invocation, a future non-composer
+  interaction -- that never renders a composer at all) gets the
+  fail-SAFE reading: **treated as composing**, so the guarded binds
+  (`fold-toggle`, `jump_next`, `jump_prev`) resolve to `:passthrough`,
+  never firing. This is deliberately asymmetric with what would be
+  simplest to implement (`Map.get(context, :composing?, false)`, "missing
+  means not composing"), because the two failure directions are not
+  equally bad:
+
+    * **Fail-safe (this module's choice)**: a caller that never sets
+      `composing?` gets dead `j`/`k`/`z` navigation. That is immediately
+      discoverable -- the keys visibly do nothing -- and ESC/Tab are
+      unaffected (`:always` binds never consult `composing?`), so nothing
+      load-bearing is silently broken.
+    * **Fail-open (the rejected alternative)**: a caller that never sets
+      `composing?` gets the guarded binds firing unconditionally --
+      exactly the named prototype bug this module exists to fix (see
+      below), except now triggered by an *absent* flag instead of a
+      flat keymap with no guard at all. Typed `j`/`k`/`z` would silently
+      turn into fold-toggle/jump commands instead of inserting characters,
+      with no crash and no visible error -- the worst kind of bug, because
+      it looks like normal typing until content goes missing.
+
+  Only an explicit `composing?: false` opts a caller INTO guarded-bind
+  resolution. `context()` also normalizes a bare `nil` (e.g. "no block
+  focused" callers that pass `nil` instead of `%{}`) to `%{}` before any
+  guard consults it, so the crash surface is not bind-dependent -- see
+  `resolve/2`'s doc.
 
   `context()` is deliberately small: `composing?` (is the composer
   focused/receiving text), `streaming?` (is a turn currently running), and
@@ -123,7 +170,8 @@ defmodule Raxol.UI.Harness.Keymap do
           required(:command_type) => command_type(),
           optional(:key) => atom(),
           optional(:char) => String.t(),
-          optional(:guard) => guard()
+          optional(:guard) => guard(),
+          optional(:mods) => InputEvent.mods()
         }
 
   # Plain keys only (v1) -- see moduledoc's "tui-steal rule": a future chord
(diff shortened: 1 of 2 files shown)


DROOdotFOO:

Scope: `lib/raxol/ui/harness/keymap.ex` (+199) and its test. Pure keymap, no process state, well-tested against the real T27 emitters. The findings below are real defects/traps, not a rejection of the design.

### HIGH
- [Saboteur + Security Auditor] `keymap.ex:167` (`resolve(norm, context \\ %{})`) + `keymap.ex:186-187` (`Map.get(context, :composing?, false)`) — **the default fails open toward the exact bug this unit exists to fix.** A missing/omitted `composing?` is treated as *not composing*, so `z`/`j`/`k` resolve to `:fold_toggle`/`:jump_next`/`:jump_prev`. The whole raison d'etre (moduledoc + `harness-ui-STATE.md`: "the demo's flat keymap steals j/k/s/z from typing") is defeated the moment a caller forgets to thread composer-focus state — which is precisely the fragile part deferred to T13a's wiring. Failure scenario: T13a dispatches keymap-first (its own stated precondition #1) but a code path building `context` drops `composing?` (e.g. `%{focused_block_id: id}` with no focus flag) → the user typing `j` mid-message gets a jump instead of a character. The safe default for a key-stealing guard is "assume composing (passthrough)", not "assume free to steal". The test at `keymap_test.exs` ("composing?: unset defaults to the guarded (non-composing) behavior") codifies the fail-open default rather than flagging it. Base MEDIUM, promoted (2 personas, same root cause).

### MEDIUM
- [Saboteur / New Hire] `keymap.ex:176-177` (`matches?(%{key: key}...)` → `InputEvent.key(norm) == key`) — **key binds are modifier-blind; the "tui-steal rule" claim is aspirational.** `InputEvent.key/1` returns the bare atom (`:tab`, `:escape`) regardless of `mods`, so `Ctrl+Tab`, `Alt+Tab`, and `Shift+Tab` (via `Event.key_event(:tab, :pressed, [:ctrl])`, a shape T27 supports) all resolve to `:steer`. Meanwhile the `char` binds *are* modifier-safe, but only as an accidental side effect of `printable_char/1`'s internal `text?` check (`ctrl/alt/meta` held → `nil`). This asymmetry is undocumented and load-bearing. The moduledoc promises a future chord "grows a `mods:` requirement, the walking loop does not change" — but `matches?/3` has **no** code path that inspects `mods`, so adding a chord *does* require restructuring `matches?`. A maintainer trusting the doc will ship `Ctrl+Tab` = steer.

### LOW
- [New Hire] `keymap.ex:155` (`command_types, do: Enum.map(@binds, & &1.command_type)`) — spec'd and documented as "the exact **set** … can ever emit", but returns a plain list that would contain duplicates if two binds ever shared a `command_type`. The parity test itself calls `Enum.uniq` on it, revealing the author already knows it is not a set. `Enum.map |> Enum.uniq` (or `MapSet`) here would make the name honest for T15's palette consumer.
- [Saboteur] `keymap.ex:186-187` — `resolve/2` promises totality ("never raises on a well-formed `InputEvent.t()`") but `resolve(norm, nil)` raises `BadMapError` inside `guard_passes?(%{guard: :not_composing}, nil)` (`Map.get(nil, ...)`), while an `:always` bind with the same `nil` context returns fine. Inconsistent crash surface; an `is_map(context)` guard or a normalizing clause would restore the totality guarantee. Easy to hit since `context` is optional and a caller may compute it as `nil` when no block is focused.
- [New Hire / Security Auditor] `keymap.ex:189-194` (`build_command(:fold_toggle)`) — `focused_block_id` is passed straight from context into `payload.block_id` as opaque `term()` with no shape check. Benign at this pure layer, but it is an unvalidated trust boundary: downstream fold/jump handlers (T13a/T15) must not assume it references a live block. Worth a `@type` narrowing or a note that consumers validate.
- [New Hire] moduledoc "Command shape" — `command()` is hand-kept field-identical to `Raxol.Agent.Command` with the real conversion (`struct(Command, cmd)`) deferred to T13b, and nothing (test or type) enforces parity across the package boundary. Silent drift risk: if `Raxol.Agent.Command` gains a required field, `struct/2` fills a default and the mismatch surfaces only at the live wiring seam, far from this file.

### Cross-Persona Overlaps
- Fail-open default context (HIGH) surfaced independently as a production-correctness regression (Saboteur) and a safe-default/least-surprise violation where a lost UI-focus signal escalates typed keys into privileged transcript commands (Security Auditor). Promoted per the 2-persona rule.
- Modifier handling recurs: Saboteur sees `Ctrl/Shift+Tab` mis-firing `:steer`; New Hire sees the undocumented mechanism asymmetry between the two bind classes. Same underlying gap in `matches?/3`.

### Verdict: CONCERNS
One HIGH (fail-open default re-opens the named key-stealing regression) plus a MEDIUM (modifier-blind key binds contradicting the stated extensibility contract). No CRITICAL and not 3+ HIGH, so not a BLOCK — but the HIGH should be addressed (invert the `composing?` default to fail safe, or require the flag) before T13a wires this in, since T13a is exactly where the missing-context failure mode lives.

merklebonsai:
Round-2 responses below — the review response landed as e550ba9 (`fix(harness-ui): T12 review response`); each finding is dispositioned individually.

### Accepted + fixed
- **HIGH — fail-open `composing?` default** (`keymap.ex:186-187`): e550ba9 — default inverted to fail-safe: `Map.get(context, :composing?, true)`, so a missing/omitted flag is treated as *composing* and the guarded `j`/`k`/`z` binds resolve to `:passthrough` (dead nav, discoverable) instead of stealing typed keys; only an explicit `composing?: false` opts a caller into guarded-bind resolution. The test that codified the fail-open default was rewritten to assert the fail-safe direction, plus a new explicit opt-in test, and the moduledoc now documents why the two failure directions aren't symmetric.
- **MEDIUM — modifier-blind `key:` binds** (`keymap.ex:176-177`): e550ba9 — new `mods_match?/2` in `matches?/3`: a `key:`-kind bind with no declared `mods:` now matches only a bare keypress (ctrl/alt/meta held → fall through to `:passthrough`; shift deliberately exempt, mirroring `InputEvent.text?/1`), and a bind that declares `mods:` requires an exact match against the normalized event's mods — the chord-growth promise in the moduledoc is now real, not aspirational. Covered by tests for Ctrl+Tab/Alt+Tab in both the translator and `Event.key_event/3` wire shapes, plus an exact-match fixture-chord test; the `bind()` type gained `optional(:mods)` and the char/key asymmetry is now documented.
- **LOW — `command_types/0` returns a list, not the documented set** (`keymap.ex:155`): e550ba9 — uniqs internally (`Enum.map |> Enum.uniq`); the parity test dropped its compensating `Enum.uniq/1`.
- **LOW — `resolve(norm, nil)` raises `BadMapError`** (`keymap.ex:186-187`): e550ba9 — a `resolve(norm, nil)` head normalizes to `%{}` before any guard runs, restoring the totality guarantee with a consistent crash surface across `:always` and `:not_composing` binds; both nil-context paths are tested.

### Declined
- **LOW — opaque `block_id` passed unvalidated into `payload.block_id`** (`keymap.ex:189-194`): declined. The keymap is a pure, stateless layer with no block registry to validate against — checking liveness here would either duplicate downstream state or make the module impure. The trust boundary is real but belongs to the consumers: T13a/T15 fold/jump handlers must treat `block_id` as a hint and validate against live blocks at dispatch time (already noted as a seam precondition in the STATE doc).

### Deferred
- **LOW — `command()` shape hand-kept field-identical to `Raxol.Agent.Command` with no parity enforcement**: deferred to **T13b**, which owns the actual `struct(Command, cmd)` conversion seam (main raxol cannot depend on raxol_agent, so the assertion can't live in this package). Exit criterion: T13b lands a cross-boundary parity test asserting `command()`'s fields are exactly `%Raxol.Agent.Command{}`'s struct keys, so drift fails in CI rather than surfacing at the live wiring seam.

DROOdotFOO:
still failing CI

PR #593, opened 2026-07-16
Harness UI: degradation ladder (mode select + flat authority) [T3]

merklebonsai:
Unit **T3 — degradation ladder**: the startup mode pick + the `:flat` substrate. Answers AD-U2 (block-haters), CI/pipes, and screen readers with one honest rule: when the fancy substrate can't keep its promises, drop to the tier that can.

## What
- **`ModeSelect.select/3`** — pure `(caps, env, opts) → :inline_log | :tmux_conservative | :flat`. First match wins: `RAXOL_HARNESS_MODE` override → headless (`TERM=dumb` / non-tty / CI-without-tty) → **degenerate geometry → :flat** → tmux/screen → `:inline_log`. No `System.get_env` inside — env is an argument.
- **`FlatAuthority`** — third `PaintAuthority` impl: append-only, **zero escape bytes of any kind** (not just movement — documented v1 decision; SGR-when-tty is a later opt-in flag, additive). Footer/keyframe no-ops, resize writes nothing.
- **`:tmux_conservative` is not a new module** — it's `InlineAuthority` under T1's already-clamped caps; T3 only names the tier.

## The review catch (why degenerate outranks tmux)
The first cut checked tmux before degenerate geometry, with a "the byte layer degrades gracefully" rationale. Opus review **byte-traced it and proved data loss**: at degenerate geometry `region_top` pins at 1, so every seal CUPs to row 1 and **overwrites the prior block before it can scroll into scrollback** (`\e[1;1HL1…\e[1;1HM1…` — L1 destroyed). Flat prints the same transcript linearly, loses nothing, and can't violate any tmux constraint (zero escapes). Order swapped; the corner is regression-pinned through both tmux signal paths (`TMUX` env and `caps.multiplexer`); the rationale now documents the trace instead of contradicting it.

## Tests bind to reality
- The mechanical flat assert is **fail-closed, not vacuous**: the scanner routes every `0x1B` to a non-text token (an unknown/unterminated escape can never be classified as text), so "every token is text" ≡ "no ESC byte anywhere". R8 red-first: a buggy flat writer emitting one CUP fails the same scanner on the same fixture; the real `FlatAuthority` passes.
- Same-fixture parity: real `InlineAuthority` and real `FlatAuthority` both driven; line-content sequences equal (order-sensitive) — "same transcript, different substrate".
- Mode-pick matrix: property + table over caps × env × geometry, incl. garbage override ignored (falls through, no crash), `CI=true`+tty stays inline.

Opus review: RED on the ordering (fixed as above, red-first shown), everything else clean — scanner fail-closure, purity, PaintAuthority conformance, zero-escape all verified. Side-find logged to backlog: the shared `SequenceScanner` hangs on a lone trailing `\e` at EOF (hangs, not false-passes — fail-closure holds).

Gates: `compile --warnings-as-errors` clean; 15/15 T3+renderer green, 183/183 full harness+property; format + credo --strict clean.

**Base note:** stacks on `feat/harness-ui-T2b` (#589) — needs `InlineAuthority` + `degenerate?/2`. Merge after #589.

---
*Terms: **AD-U2** = the cohort-research finding that a segment of operators reject block chrome entirely — flat mode is their first-class answer, not a fallback. **degenerate geometry** = terminal too short to pin a footer (#574). **R8** = demonstrated-red methodology rule.*

### DIFF
--- a/lib/raxol/ui/rendering/paint_authority/flat_authority.ex
+++ b/lib/raxol/ui/rendering/paint_authority/flat_authority.ex
@@ -1,0 +1,122 @@
+defmodule Raxol.UI.Rendering.PaintAuthority.FlatAuthority do
+  @moduledoc """
+  T3's `:flat` tier: the append-only, zero-region, zero-cursor-jump
+  `PaintAuthority` implementation
+  (`docs/proposals/in-flight/harness-ui-roadmap.md`, unit T3 —
+  "degradation ladder"). Picked by
+  `Raxol.UI.Rendering.PaintAuthority.ModeSelect.select/3` for `TERM=dumb`,
+  non-tty, CI-without-tty, and degenerate-geometry sessions — the
+  screen-reader answer, the CI/pipe answer, the block-hater answer (AD-U2).
+
+  ## The SGR/escape decision: zero escape bytes, full stop
+
+  Unlike `InlineAuthority` (which never emits a full clear but DOES emit
+  DECSTBM/CUP/cursor-save bytes) this module never writes ANY byte in the
+  `0x1B` (ESC) family — not just cursor movement, but also SGR styling.
+  This is a deliberate v1 simplification, not an oversight:
+
+    * A conditionally-styled flat mode (SGR when the destination happens
+      to be a real tty, plain when it's a genuine pipe) would need its
+      own tty-detection seam threaded into the AUTHORITY itself, when
+      `ModeSelect` already computed that exact fact once, at startup, to
+      pick `:flat` in the first place. Re-deriving it a second time inside
+      the authority (or plumbing the `:tty?` flag through as authority
+      state) is more machinery than a v1 flat writer needs.
+    * Zero escape bytes is also the easiest claim to make PROVABLY true:
+      the mechanical acceptance check ("flat output contains no
+      cursor-move/CUP/scroll sequences") reduces to "every scanned token
+      is `{:text, _}`" with no exceptions to special-case for allowed SGR
+      runs.
+    * This is additive, not a rewrite, to upgrade later: a future unit
+      that wants styled flat output for a genuine tty destination can
+      thread `ModeSelect`'s already-computed `:tty?` fact into `new/3` as
+      an opt-in flag and add SGR emission ONLY on that path, without
+      touching the append-only/no-region/no-cursor contract this module
+      ships today.
+
+  ## Line-terminator convention: plain `\\n`
+
+  `InlineAuthority.seal/2` requires `\\r\\n`-terminated content because a
+  real terminal in raw mode needs the explicit carriage return (output
+  translation is off). `:flat`'s destination is a pipe, a log file, a
+  screen reader, or a CI capture buffer — none of which need or want the
+  extra `\\r`. Callers building content for `FlatAuthority` should
+  terminate each line with plain `\\n`. This module does not enforce or
+  rewrite terminators itself (it writes `iodata` verbatim, exactly like
+  `PaintAuthority.IOAuthority`'s minimal-stub convention) — the `\\r\\n`
+  vs. `\\n` choice lives entirely in what the CALLER passes in, so a
+  caller that already has `\\r\\n`-terminated content (e.g. replaying the
+  same fixture through both tiers for a parity check) gets it written
+  through unchanged rather than silently rewritten.
+
+  ## What every callback does
+
+    * `append_sealed/2` — writes `iodata` verbatim. No CUP, ever.
+    * `repaint_footer/2` / `keyframe_footer/2` — NO-OP (return state
+      unchanged, write nothing). Flat has no footer to repaint or
+      keyframe; a caller that calls these on a flat authority gets
+      silence, not stray bytes.
+    * `with_cursor/3` — runs `fun` directly against the state. No save,
+      no restore: there is no cursor position worth protecting when
+      nothing ever moves it.
+    * `resize/3` — updates the tracked `width`/`height` but writes zero
+      bytes. There is no region to re-pin.
+    * `region_top/1` — returns `height`: with no footer carved out, the
+      entire screen is (conceptually) history/content.
+  """
+
+  @behaviour Raxol.UI.Rendering.PaintAuthority
+
+  @enforce_keys [:device, :width, :height]
+  defstruct [:device, :width, :height]
+
+  @type t :: %__MODULE__{
+          device: IO.device(),
+          width: pos_integer(),
+          height: pos_integer()
+        }
+
+  @doc "Builds a new flat authority state. No device writes happen at construction."
+  @spec new(IO.device(), pos_integer(), pos_integer()) :: t()
+  def new(device, width, height)
+      when is_integer(width) and width > 0 and is_integer(height) and
+             height > 0 do
+    %__MODULE__{device: device, width: width, height: height}
+  end
+
+  @doc """
+  Sugar mirroring `InlineAuthority.seal/2`'s call shape so a caller can
+  swap authorities without changing its call site: `seal/2` is exactly
+  `append_sealed/2` here (no cursor bracket to wrap it in).
+  """
+  @spec seal(t(), iodata()) :: t()
+  def seal(%__MODULE__{} = t, iodata), do: append_sealed(t, iodata)
+
+  @impl true
+  def append_sealed(%__MODULE__{device: device} = t, iodata) do
+    IO.write(device, iodata)
+    t
+  end
+
+  @impl true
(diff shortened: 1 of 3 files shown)


DROOdotFOO:

Reviewed against `feat/harness-ui-T2b` (the stacked base, which supplies `InlineAuthority`, `ScrollRegionManager.degenerate?/2`, and the `PaintAuthority` behaviour). Three files, all additive. Findings grounded in the actual seam contracts.

### HIGH

- [Saboteur / New Hire / Security] `flat_authority.ex:96-98` (`append_sealed/2`) — **the module's central "zero escape bytes, full stop" guarantee is unenforced; it is a property of the caller, not this module.** `append_sealed` does `IO.write(device, iodata)` verbatim. The moduledoc promises flat output "never writes ANY byte in the `0x1B` (ESC) family... the screen-reader answer, the CI/pipe answer" and calls this "PROVABLY true." It is only true for escape-free input. Every other `PaintAuthority` (Inline) routes SGR-styled content through the identical `(t, iodata)` seam; nothing in this PR guarantees the sealed blocks handed to `FlatAuthority` are plain text (the moduledoc itself defers that to "T7's journal-fold projection... T13a's job to bridge" — future work). Failure scenario: colored assistant/tool output (or crafted escape sequences from untrusted remote-agent/tool content) is sealed and routed to flat at a `TERM=dumb` / pipe / screen-reader / CI sink — `FlatAuthority` faithfully forwards the raw ANSI, corrupting screen-reader output and passing terminal-injection sequences straight through to whatever consumes the flat stream. The mechanical test (`t3_degradation_ladder_test.exs`, `@fixture_blocks`) only proves purity for a hand-picked plain-text fixture, so the acceptance criterion "flat output contains no cursor-move/CUP/scroll sequences" is validated only for clean input. The module makes an absolute safety claim it cannot itself hold. (3-persona overlap; base MEDIUM promoted.)

- [Saboteur / Security Auditor] `mode_select.ex:112-127` (`select/3` -> `override/1` before `select_without_override`) — **`RAXOL_HARNESS_MODE=inline` defeats the degenerate-geometry safety floor and silently reintroduces the exact data-loss bug this PR exists to fix.** Rule 1 (override) runs before rule 3 (degenerate geometry). The test even pins this: `override=inline wins over ... degenerate geometry -> :inline_log`. But per this PR's own byte-trace, `InlineAuthority.append_sealed` positions at `min(next_row, region_top)`, and at degenerate geometry `region_top` pins at ≤1, so every seal CUPs to row 1 and clobbers the prior block before it scrolls (verified: `InlineAuthority` moduledoc confirms it always CUPs to `min(next_row, region_top)`). Failure scenario: an operator who prefers inline and exports `RAXOL_HARNESS_MODE=inline` in their shell profile (a reasonable, documented-as-honored choice) then runs the harness in a 2-row split pane — transcript blocks are silently overwritten, never reaching scrollback. If that transcript is an agent/tool audit trail, this is silent integrity/observability loss with no error and no warning. There is no safety floor (override could still clamp to `:flat` when `degenerate?` is true) and no log. (2-persona overlap; base MEDIUM promoted.)

### LOW

- [Saboteur] `mode_select.ex:185-192` (`degenerate_geometry?/1`) — `:rows` is guarded (`is_integer and > 0`, else skipped), but `:footer_rows` from `Keyword.get(opts, :footer_rows, 0)` is passed straight into `ScrollRegionManager.degenerate?/2`, whose guard requires `footer_rows >= 0` and integer. `select(caps, env, rows: 5, footer_rows: -1)` (or a non-integer) raises `FunctionClauseError` instead of the documented fail-open-to-`:inline_log`. Caller-controlled, so low, but the moduledoc claims "purely a function of its three arguments" with a uniform fail-open default that this path violates.

- [New Hire] `mode_select.ex:130-137` (`override/1`) — a misspelled explicit override (`RAXOL_HARNESS_MODE=Flat`, `flatt`, uppercase) matches the `_other -> nil` catch-all and silently falls through to auto-detection. The moduledoc bills the override as the signal that "wins over every other signal," so an operator who typo'd it gets the opposite of what they set with zero feedback (no log/warn). The PR treats "garbage override ignored, no crash" as a feature; for an explicit operator directive, silent no-op is a footgun.

- [New Hire] `flat_authority.ex` / `mode_select.ex` (moduledocs) — ~100 lines of rationale prose per ~40 lines of code, and the "why degenerate outranks tmux" essay is duplicated near-verbatim across the mode_select moduledoc, the PR description, and two test comments (`t3_degradation_ladder_test.exs` table row + regression test). Three-plus copies of the same byte-trace argument to keep in sync as the ordering evolves.

### Cross-Persona Overlaps

- **Verbatim passthrough** (HIGH #1) independently surfaced by all three personas: Saboteur (data corruption at the safe sink), New Hire (moduledoc asserts a guarantee the code doesn't keep), Security (unsanitized ANSI/terminal-injection passthrough at the tier advertised as the safe one). Strongest finding.
- **Override safety floor** (HIGH #2) surfaced by Saboteur (reintroduced clobber/data loss) and Security (env-controlled integrity loss of transcripts/audit trail).

### Verdict: CONCERNS

Two HIGH findings (both promoted via multi-persona overlap), no CRITICAL and fewer than three HIGH, so not a BLOCK. Both are addressable without a redesign: (1) either sanitize/assert escape-free content in `FlatAuthority.append_sealed` or soften the moduledoc's absolute guarantee to state the caller owns escape-freedom and add a mechanical test with styled input; (2) add a degenerate-geometry safety floor (or at minimum a one-line warning) so `RAXOL_HARNESS_MODE=inline` cannot silently drop transcript blocks on a terminal too short to pin a footer.

merklebonsai:
Both HIGH findings and the guard LOW landed in `10c3fe2`; per-finding disposition against your round-2 review below.

### Accepted + fixed
- **HIGH — `FlatAuthority.append_sealed/2` verbatim passthrough / unenforced zero-escape guarantee** (`10c3fe2`): the guarantee is now module-enforced, not a caller contract — `append_sealed/2` scrubs every C0 byte (`0x00`–`0x1F`, ESC included) except `\t`/`\r`/`\n` before writing (byte-wise scrub is UTF-8-safe: C0 never overlaps lead/continuation bytes). Mechanical tests now drive hostile styled input (CSI clear + SGR color + bare BEL) through the real authority, plus `\t`/`\r`/`\n` survival and multi-byte UTF-8 passthrough. Moduledoc rewritten to state the property the code actually keeps. Design note: stripping the ESC lead and leaving a visible `[31m` fragment is the deliberate honest-failure mode — a visible fragment beats an invisible injection.
- **HIGH — `RAXOL_HARNESS_MODE=inline` defeats the degenerate-geometry floor** (`10c3fe2`): took exactly the shape you suggested. Mode-pick is now two passes: resolve a candidate (override, else auto-detect), then apply the degenerate-geometry FLOOR last. Any non-`:flat` candidate — override included — clamps to `:flat` at degenerate geometry; a `:flat` override is a no-op clamp, and adequate geometry leaves overrides untouched. Red-first regression proof (pre-fix `select/3` honored the override unconditionally), both `=inline` and `=tmux` override paths pinned, and the property test updated to "override wins UNLESS the floor clamps".
- **LOW — `footer_rows: -1` raises `FunctionClauseError` instead of failing open** (`10c3fe2`): `degenerate_geometry?/1` now guards `footer_rows` (non-negative integer) the same way `:rows` was already guarded, failing open to non-degenerate; table row added (`rows: 2, footer_rows: -1` → no raise, `:inline_log`).
- **LOW — misspelled override silently falls through** (`10c3fe2`, mechanism; warning emission deferred below): override matching is now trim + downcase, so `Flat` / ` flat ` / `FLAT` are honored as typed. A genuinely unrecognized non-empty value still falls through to auto-detect, but is no longer indistinguishable from "no override": the new `select_with_reason/3` seam surfaces it as `:override_unrecognized`.

### Declined
- **LOW — moduledoc prose density, byte-trace essay duplicated 3+ places**: keeping the prose. The floor-vs-override ordering is a one-way-door decision and the moduledoc is its canonical home; `10c3fe2` already cut the copy count — test comments now cite the moduledoc section by title ("Why the degenerate floor applies after override resolution") instead of restating the trace. The PR-description copy is a reviewer snapshot, not a maintained artifact, so it doesn't add sync burden.

### Deferred
- **Operator-visible warning on unrecognized override** (tail of the LOW above, and the "no log" tail of HIGH #2): `ModeSelect` stays pure — no IO or logging inside — so the warning itself can't live there. Exit criterion: T13a's assembler consumes `select_with_reason/3` and prints a startup notice on `:override_unrecognized` (typo'd override) and `:degenerate_clamp` (why the session ended up in `:flat`). The seam and both reason atoms are already in place and tested.

PR #594, opened 2026-07-16
Harness UI: block bodies (BodyProvider contract + fold-aware component mount) [T5]

merklebonsai:
Unit **T5 — block bodies**: mounts the merged block components (#535–541) as fold-aware bodies behind an explicit per-kind content contract. The last M1 leaf before T13a assembly.

## What
- **`BodyProvider`** (new) — the contract-first seam (per the lane's binding advisory: schema BEFORE mount). Per-kind required keys: `:message`/`:reasoning` `[:text]` · `:tool_call` `[:name,:args]` (+optional result/taint) · `:diff` `[:path,:old,:new]` · `:approval` `[:action,:blast_radius,:options]`. Presence-validated with plain errors; kind→component mapping: MessageBlock / ReasoningBlock / ToolCallBlock (+ToolResultBlock composed when a result landed) / DiffViewer / ApprovalPrompt (embeds BlastRadiusPreview).
- **`BlockBody`** (new) — fold-aware entry: folded = verbatim delegation to T4's `Block.render/2` (header + outcome row; prominence/markdown-fade flow through untouched — no D-PA re-render logic); expanded = `BodyProvider.mount/3`.
- **`Block.extract_content/2`** gained a `:diff` clause — verified truly additive (no producer emits `:diff` today, the touched clauses are disjoint from T26/T8's block.ex regions, zero behavior change for existing kinds).

## The review catch (total-safety made true, not just claimed)
Opus review found the expanded mount path could let a component **raise escape to the render loop** — reachable from a REAL producer: an `approval_requested` event without `blast_radius` passed presence-validation, then `BadMapError`'d inside BlastRadiusPreview. On the highest-stakes surface (approval prompts), while the moduledoc claimed total-safety. Fixed two layers deep:
1. **Rescue at `BlockBody`** (the layer that makes the promise): any component raise converts to the existing `{:error, "component raised …"}` → same visible fallback block + `Logger.warning` + `[:raxol,:harness,:block_body,:recovered]` telemetry (name verified distinct from BOTH existing `:recovered` events — the W1 collision lesson). Never silent.
2. **`blast_radius || %{}` default** in extract_approval_content — a partial approval now renders the REAL prompt ("No tracked effects.") instead of falling back at all.
Both vectors red-proven before the fix (`BadMapError`/`FunctionClauseError` escaping uncaught), then asserted green with fallback content + telemetry.

## Tests bind to reality
Real merged components driven through real `Block.from_events` producers (no stubs); assertions on rendered buffer content. Four re-runnable red-first proofs: diff-extraction revert, wrong-kind-refusal neuter, tool-result composition force-off, folded-delegation perturbation — plus the two raise-escape repros above.

Gates: `compile --warnings-as-errors` clean; 463 harness-component tests green; full suite 5503/5507 (4 pre-existing documented flakes, reproduced identically without this change); format + credo --strict clean.

Base: master. Independent of the substrate stack — mergeable any time.

---
*Terms: **fold** = collapsing a block to its header+outcome row. **D-PA (A)** = post-seal content is frozen; fold state only affects pre-seal/live rendering. **taint** = tool-result provenance badge. The **`:recovered` telemetry** name is verified distinct from the two existing recovered events (a real collision bit W1).*

### DIFF
--- a/lib/raxol/ui/components/harness/blast_radius_preview.ex
+++ b/lib/raxol/ui/components/harness/blast_radius_preview.ex
@@ -17,6 +17,15 @@ defmodule Raxol.UI.Components.Harness.BlastRadiusPreview do
       optional; missing lists default to `[]`, missing `:reversible`
       defaults to `true` (a missing/absent flag is not itself a danger
       signal -- but deletes are still shown loud regardless, see below).
+      A missing PROP defaults to `%{}` -- a caller that never mentions a
+      blast radius at all is presumed to have nothing to declare, and
+      renders the calm "No tracked effects." line. An explicit `nil`
+      VALUE is a different claim entirely: it means a producer somewhere
+      upstream had an approval to show but no blast radius payload for
+      it (see `Raxol.UI.Components.Harness.Block.extract_approval_content/1`)
+      -- that renders a standalone warning instead (see below), never the
+      "No tracked effects." line, because an undeclared radius is not
+      known to be safe.
 
   ## Visual hierarchy
 
@@ -57,7 +66,7 @@ defmodule Raxol.UI.Components.Harness.BlastRadiusPreview do
 
   @type t :: %{
           id: String.t() | atom(),
-          blast_radius: blast_radius(),
+          blast_radius: blast_radius() | nil,
           style: map(),
           theme: map()
         }
@@ -96,29 +105,48 @@ defmodule Raxol.UI.Components.Harness.BlastRadiusPreview do
     base_style =
       StyleHelper.merge_component_styles(state, context, :blast_radius_preview)
 
-    br = state.blast_radius
-    reversible = Map.get(br, :reversible, true)
-
-    sections =
-      [
-        marker_section(state.id, reversible)
-        | group_sections(state.id, br, reversible)
-      ]
-      |> Enum.reject(&is_nil/1)
-
-    children =
-      case sections do
-        [] ->
-          [Components.text(content: "No tracked effects.", style: %{dim: true})]
-
-        list ->
-          list |> Enum.intersperse(blank_line()) |> List.flatten()
-      end
+    children = body_children(state.id, state.blast_radius)
 
     # gap: 0 is load-bearing: the layout engine defaults an unset gap to 1,
     # which would double every row; section spacing is the explicit
     # interspersed blank lines, counted by estimate_rows/1.
     %{type: :column, style: base_style, gap: 0, children: children}
+  end
+
+  # `nil` means no producer ever supplied a blast_radius payload for this
+  # approval -- distinct from a producer-declared-empty `%{}` (handled
+  # below, still the calm "No tracked effects." line). An undeclared
+  # radius is not a known-safe one, so this renders a standalone warning
+  # instead, reversibility assumed false, so a destructive action with no
+  # declared radius never reads as harmless.
+  defp body_children(id, nil) do
+    [
+      Components.text(
+        id: "#{id}-not-declared",
+        content: "⚠ blast radius not declared — treat as unsafe",
+        fg: :red,
+        style: %{bold: true}
+      )
+    ]
+  end
+
+  defp body_children(id, br) do
+    reversible = Map.get(br, :reversible, true)
+
+    sections =
+      [
+        marker_section(id, reversible)
+        | group_sections(id, br, reversible)
+      ]
+      |> Enum.reject(&is_nil/1)
+
+    case sections do
+      [] ->
+        [Components.text(content: "No tracked effects.", style: %{dim: true})]
+
+      list ->
+        list |> Enum.intersperse(blank_line()) |> List.flatten()
+    end
   end
 
   @doc """
@@ -128,7 +156,11 @@ defmodule Raxol.UI.Components.Harness.BlastRadiusPreview do
   `Raxol.UI.Components.Harness.ApprovalPrompt.estimate_height/1`). Generous
   by design, mirroring `Raxol.UI.Components.Modal.Rendering.estimate_height/1`.
   """
-  @spec estimate_rows(blast_radius()) :: pos_integer()
+  @spec estimate_rows(blast_radius() | nil) :: pos_integer()
+  # `nil` (not declared) renders the single-line warning from
+  # `body_children/2` above -- one row, no groups to size.
+  def estimate_rows(nil), do: 1
+
   def estimate_rows(blast_radius) do
     reversible = Map.get(blast_radius, :reversible, true)
     marker? = not reversible
--- a/lib/raxol/ui/components/harness/block.ex
+++ b/lib/raxol/ui/components/harness/block.ex
@@ -490,6 +490,13 @@ defmodule Raxol.UI.Components.Harness.Block do
     first_line(action)
   end
 
+  defp summary(%__MODULE__{kind: :diff, content: %{path: path}}) do
+    case path do
+      "" -> "(no path)"
+      p -> p
+    end
+  end
+
   defp summary(%__MODULE__{
          kind: :opaque,
          raw_kind: raw_kind,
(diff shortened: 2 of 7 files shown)


DROOdotFOO:

Scope: `block.ex` (`:diff`/`extract_diff_content`, `blast_radius || %{}`), new `BlockBody`, new `BodyProvider`, plus tests. Read in full; BlastRadiusPreview/ApprovalPrompt (unchanged) read for context.

### HIGH
- [Security + Saboteur] `block.ex` `extract_approval_content` (diff L43, `blast_radius = find_in_events(...) || %{}`) — This "fix" is a fail-open regression on the highest-stakes surface. An `approval_requested` event with no `blast_radius` (the PR itself confirms this is reachable from a **real** producer) now defaults to `%{}`. `BlastRadiusPreview.render/2` then reads `reversible = Map.get(br, :reversible, true)` (blast_radius_preview.ex:100) → no `IRREVERSIBLE` marker (blast_radius_preview.ex:155), all groups empty → it renders the authoritative-looking line **"No tracked effects."** (blast_radius_preview.ex:112). Failure scenario: a buggy/adversarial agent emits an approval for `rm -rf /` with `blast_radius` omitted; the operator is shown a clean approval prompt asserting "No tracked effects.", actively contradicting the danger, instead of the pre-change fail-loud crash → plain fallback summary. The PR's own RED test (`block_body_test.exs:743-766`) encodes exactly this: `action: "rm -rf /"`, no blast_radius, asserting "No tracked effects." renders. Trading a fail-safe crash for a reassuring-but-false safety claim on an approval prompt is the wrong direction. Distinguish "no effects declared" from "no effects" (e.g. render an explicit "blast radius unknown — treat as unsafe" state, or keep the fallback for approvals specifically).

### MEDIUM
- [Saboteur] `block_body.ex:129-171` — `render/2` is on the per-frame hot path and `emit_recovered/2` fires a `Logger.warning` + `:telemetry.execute` on **every** failing render with no dedup / rate-limit / already-warned guard. Failure scenario: one persistently-raising block (latent component bug on otherwise-valid input) floods logs and telemetry at frame rate (e.g. 60/s) indefinitely, and the user silently sees a degraded folded summary forever with no crash to force a fix. The logging avoids "silent swallow," but the unbounded volume + permanent silent degradation is its own production hazard.
- [Saboteur + New Hire] `body_provider.ex:344-356, 407-410` — `mount/3`'s `@spec` promises `{:ok, map()} | {:error, String.t()}`, but `mount_one/3` does `{:ok, state} = component.init(props)` (hard match → `MatchError` if any component ever returns non-`{:ok, _}`) and calls `component.render/2`, which can raise. Only `BlockBody.mount_body/2`'s rescue (block_body.ex:156-160) makes this safe. The moduledoc bills `BodyProvider` as a "pure, independently test-facing schema seam" and explicitly the direct-call surface — yet a direct caller (a test, or any future non-`BlockBody` caller) gets no total-safety and a spec that lies about raising. Two personas flag → promoted. Either the spec should admit it can raise, or the match should be softened to an `{:error, _}` path.
- [New Hire] `block_body.ex:77-104`, `body_provider.ex:180-234`, `block.ex` comment (diff L49-56) — Moduledocs/comments are saturated with undefined internal shorthand: `T4`/`T5`/`T7`/`T13a`, `M1 leaf`, `Y1 default`, `W1 collision lesson`, `D-PA (A)`, `R9`. None are defined in-repo; a new hire cannot decode them. This also directly violates the project's own documented voice rule (strip roadmap phase codes / review-iteration labels / dev meta-narration from comments). Replace with plain descriptions of behavior.

### LOW
- [New Hire] `body_provider.ex:466-476` `tool_status/2` — the catch-all clause is named `_present` but also matches `{present_result, nil_exit_code}` and `{present_result, 0}`; the name reads like it means "result present" while it really means "present-and-not-failed." Minor readability trap given the three-way `:pending/:failed/:done` semantics live entirely in the guard ordering.
- [New Hire] `block.ex` `extract_diff_content` (diff L57-64) plus `summary/2` `:diff` clause (diff L9-14) — a `:diff` block whose events carry no `:old`/`:new` yields `old: ""`/`new: ""` (via `to_display_text/1`), so `DiffViewer` renders an empty proposed change with no signal that data was absent, mirroring the approval blind spot at lower stakes.

### Cross-Persona Overlaps
- blast_radius default: Security (misleading safety claim) + Saboteur (fail-open on safety-critical path) → promoted to HIGH.
- `BodyProvider.mount/3` raising vs its `{:ok}|{:error}` spec: Saboteur (uncaught raise for direct callers) + New Hire (spec/contract mismatch, implicit cross-module rescue dependency) → promoted to MEDIUM.

### Verdict: CONCERNS
One HIGH (approval fail-open) plus multiple MEDIUM. The total-safety rescue in `BlockBody` is well-executed and the tests genuinely drive real producers/components; the blocking concern is that the same PR relaxes the approval-content path to fail *open* on the one surface where fail-*safe* matters most.

merklebonsai:
Thanks Drew — round 2 adjudicated. Everything in the FIX-NOW batch landed in `ddff55a`; the two remaining items have explicit exit criteria below.

### Accepted + fixed
- **HIGH — `blast_radius || %{}` fail-open** (`ddff55a`): the `|| %{}` default is gone. `Block.extract_approval_content/1` now preserves an undeclared radius as `nil`, and `BlastRadiusPreview` renders `nil` as a bold red standalone warning — `⚠ blast radius not declared — treat as unsafe` (reversibility assumed false, `estimate_rows(nil) == 1`) — exactly the "distinguish undeclared from declared-empty" state you asked for. A producer-declared `%{}` still renders the calm "No tracked effects." line. The former RED test asserting "No tracked effects." for the `rm -rf /` repro was inverted to assert the warning; a new `block_test.exs` case pins that extraction keeps `nil` so absence can never silently re-collapse into `%{}`.
- **MEDIUM — `BodyProvider.mount/3` spec lies about raising** (`ddff55a`): took the first of your two options — the contract now states plainly that `{:ok, _} | {:error, _}` covers only the seam's own guard failures, that a mounted component's `init/1`/`render/2` raise passes through uncaught for direct/test callers, and that production code reaches `mount/3` only via `BlockBody.render/2`'s rescue. The seam deliberately keeps no try/rescue of its own so it stays a pure schema layer; the docs no longer imply otherwise.
- **MEDIUM — undefined internal shorthand in comments** (`ddff55a`): review-iteration codes (`R6`/`R9`) and `D-PA`'s bare-letter option tags replaced with plain behavioral descriptions. One deliberate scope note: `T4`/`T5`/`T7`-style unit references were kept — they're the lane's stable roadmap vocabulary (defined in `harness-ui-roadmap.md`), not review-iteration noise.
- **LOW — `_present` misnomer in `tool_status/2`** (`ddff55a`): renamed to `_result_present_not_failed`, its actual meaning.

### Deferred
- **MEDIUM — `emit_recovered/2` fires log + telemetry per failing frame, unbounded**: real, but unreachable today — nothing drives `BlockBody.render/2` at frame rate yet; T5 is a leaf unit exercised only by tests until T13a assembly wires it into the live render loop. Exit criterion: before T13a mounts `BlockBody` in the per-frame path, add a warn-once / rate-limit guard keyed on `{block id, reason}` (telemetry can stay per-occurrence for counters, but `Logger.warning` must dedup). Tracked as a T13a precondition. The "permanent silent degradation with no crash to force a fix" half is by design — the total-safety contract is the point — but the recovered-telemetry event exists precisely so a persistent fallback is alertable rather than invisible.
- **LOW — empty `:diff` renders blank with no absent-data signal**: same shape as the HIGH at lower stakes, agreed. Also unreachable today — no producer emits `:diff` (`BlockBuilder` never resolves it; the clause is forward-looking). Exit criterion: the first real `:diff` producer lands together with the same absence-vs-empty treatment the approval path now has (extraction preserves `nil` for absent `:old`/`:new`, viewer renders an explicit "no diff payload" state instead of an empty proposed change).

PR #595, opened 2026-07-16
Harness UI: S1 fixture assembly — the assembled harness (M1) [T13a]

merklebonsai:
Unit **T13a — S1 fixture assembly**: the **M1 milestone**. `Raxol.Harness.Surface` composes the whole reviewed stack end-to-end against replayed TF golden sessions — the first assembled, demoable harness. Fish, not htop: blocks seal into native scrollback exactly once, the live tail + status strip + composer stay pinned in the DECSTBM footer, folds work pre-seal, and the same fixture degrades to a clean linear transcript in flat mode.

## Composition (all previously reviewed units)
T2a scroll regions (#574) · T2b append path (#589) · T2c footer viewport (#590) · T3 mode select/flat (#593) · T5 block bodies (#594) · T10 status strip (#591) · T12 keymap (#592) — plus merged T4/T7/T11/T26/T27/TF. Fusion merges were all clean (disjoint files).

## What the new commit adds
- **`Raxol.Harness.Surface`** — the TEA core: fixture events → T7 Projection → Blocks; sealed blocks flow to `InlineAuthority.seal` exactly once (a strictly-monotonic `painted_count` high-water — double-seal is structurally impossible and test-asserted); footer = StatusStrip + pending-block preview + Composer via T2c `repaint`. All **7 assembly preconditions** from the lane STATE are honored with their own moduledoc sections — notably **keymap-first dispatch**: every input goes `InputEvent.normalize → Keymap.resolve → (only :passthrough) → Composer`, so ESC/Tab are never swallowed by focus.
- **Foldable-before-seal**: T7 only hands over completed(=sealed) blocks, so the assembler holds the newest completed block back one step — **visible meanwhile in the footer preview** (not hidden) — then seals. The one-per-advance limit of this window and the real fix (a `completed_but_unsealed` phase in T7's Block lifecycle) are documented honestly; follow-up tracked.
- **Operator honesty**: fold on an already-sealed block emits a footer notice ("block N sealed — fold unavailable") instead of a silent no-op; interrupt/steer render honest fixture-mode stubs. History bytes stay byte-identical through all of it (asserted).
- **`ViewText`** — view→iodata bridge, display-width truncation via TextMeasure.
- **Demo**: `mix run --no-start examples/harness_fixture_demo.exs [fixture] [--speed ms]` — real tty run with newline push-up startup (never `\e[2J`), `\e[K` per sealed line (dirty-screen safe), auto-degrades to flat when piped.

## Acceptance → tests (10 tests, all real-oracle)
1. Full chrome renders from fixture. 2. Sealed blocks in native scrollback (emulator replay). 3. **End-to-end invariants at every checkpoint** of the assembled run: immutable-prefix, footer confinement, zero `\e[2J`, resize (real 60×20→80×24 geometry change) with byte-identical prior history. 4. Fold/jump on replayed content; post-seal fold = notice + unchanged history. 5. Flat parity (fail-closed scanner: every token is text; order-sensitive line equality). 6. R8 red (oracle catches a duplicated-seal stream) + the real-Surface GREEN exactly-once guard. 7. **Unicode end-to-end**: CJK/Arabic/Hebrew/emoji through the whole assembly — display-width footer bounds, grapheme survival through emulator replay (NFC-normalized comparison; wide-placeholder-aware row reconstruction).

## Review
Opus (M1-stakes review): **SHIP, no RED.** Held-back block confirmed visible (no UX lie); `painted_count` monotonicity traced (fold of a sealed block can't even record an override); keymap-first verified on every path; checkpoint-level invariant coverage confirmed non-golden-walk. Yellows folded: sealed-fold notice, unicode suite, real-geometry resize, hold-back-limit + teardown-ownership docs. Declined-documented: the synthetic RED half of test 6 (the GREEN real-Surface guard is the load-bearing assertion). Follow-up logged: T7 `completed_but_unsealed` phase.

Gates: `compile --warnings-as-errors` clean; 1040 tests green across harness/property/component/keymap suites (multi-seed verified); format + credo --strict clean.

**Base note:** stacks on the T2c chain (#590) and fusion-merges #593/#594/#591/#592. Merge order: the substrate stack + leaves first, then this.

---
*Terms: **S1/M1** = the lane's first assembled scenario / first milestone. **TF golden sessions** = the merged fixture toolchain's recorded event streams (#551). **seal / immutable-prefix / D-PA (A)** = see #589. **fish, not htop** = the north-star shape: an inline, scrollback-native instrument (like fish's prompt), not a fullscreen dashboard.*

### DIFF
--- a/examples/harness_fixture_demo.exs
+++ b/examples/harness_fixture_demo.exs
@@ -1,0 +1,189 @@
+# Harness fixture demo (the assembled harness — golden-fixture assembly).
+#
+# Runs the assembled `Raxol.Harness.Surface` against a golden fixture
+# session, replayed on a real tty: blocks seal into native scrollback one
+# event at a time, the status strip + composer stay pinned in the footer,
+# and `z`/`j`/`k` (fold/jump) + Tab (steer stub) + ESC (interrupt stub)
+# work via the real `Raxol.Terminal.InlineDriver` raw-mode input
+# path -- the same canonical event route `Raxol.UI.Harness.Keymap`
+# resolves in production.
+#
+# Usage (run from the repo root, or this worktree's root):
+#
+#   mix run --no-start examples/harness_fixture_demo.exs
+#   mix run --no-start examples/harness_fixture_demo.exs multi-tool-turn
+#   mix run --no-start examples/harness_fixture_demo.exs simple-chat --speed 250
+#
+# `--no-start` avoids interleaving `:raxol` application boot logs into the
+# byte stream -- InlineAuthority and Surface are plain modules, no
+# supervision tree required.
+#
+# Keys while running: `z` fold/unfold the focused block, `j`/`k` move
+# focus (once off the composer -- see `Surface.focus_transcript/1`), `Tab`
+# queues a steer stub, `Esc` shows an interrupt stub (no agent lane in
+# fixture mode -- see `Raxol.Harness.Surface`'s moduledoc, precondition
+# #6), `q` quits cleanly WHEN THE COMPOSER BUFFER IS EMPTY (Ctrl-C does
+# not: raw mode disables SIGINT delivery, per `Raxol.Terminal.InlineDriver`'s
+# own moduledoc) -- a non-empty buffer means `q` is just a character the
+# composer is focused to receive, same as any other letter; quitting mid-
+# typed-message on that same keystroke silently ate every "q" that would
+# ever end up in a prompt (see `loop/3`/`linger/3` below, both call sites).
+
+alias Raxol.Harness.{Fixture, Surface}
+alias Raxol.Terminal.InlineDriver
+alias Raxol.UI.Components.Harness.Composer
+alias Raxol.UI.Harness.InputEvent
+
+defmodule Raxol.Examples.HarnessFixtureDemo do
+  @moduledoc false
+
+  @default_fixture "simple-chat"
+  @default_speed_ms 500
+  @footer_rows 6
+  @post_done_linger_ms 15_000
+
+  def run(argv) do
+    {fixture_name, speed_ms} = parse_args(argv)
+    path = fixture_path(fixture_name)
+
+    session =
+      case Fixture.load(path) do
+        {:ok, session} ->
+          session
+
+        {:error, reason} ->
+          IO.puts(:stderr, "failed to load fixture #{path}: #{inspect(reason)}")
+          System.halt(1)
+      end
+
+    {width, rows} = geometry()
+    tty? = Raxol.Terminal.TerminalUtils.has_terminal_device?()
+
+    # The driver's raw-mode + real key-event input path -- until the
+    # assembled harness wires the real Dispatcher, this seam is the whole
+    # contract (InlineDriver's own moduledoc). `subscriber: self()` routes every
+    # parsed keypress here as `{:inline_input, %Event{}}`. Same `tty?`
+    # fact both InlineDriver AND Surface's own `ModeSelect.select/3` pick
+    # (a non-tty run -- piped/CI -- degrades InlineDriver's stty/reader
+    # AND Surface's own render mode consistently to :flat, never one
+    # without the other).
+    {:ok, driver_pid} =
+      InlineDriver.start_link(
+        subscriber: self(),
+        device: :stdio,
+        rows: rows,
+        probe?: false,
+        tty?: tty?
+      )
+
+    # Startup discipline: push whatever is currently
+    # on screen up into scrollback via plain newlines -- never `\e[2J`.
+    Surface.startup_push_up(:stdio, rows)
+
+    model =
+      Surface.new(session,
+        device: :stdio,
+        width: width,
+        rows: rows,
+        footer_rows: @footer_rows,
+        tty?: tty?
+      )
+
+    try do
+      loop(model, driver_pid, speed_ms)
+    after
+      GenServer.stop(driver_pid, :normal)
+    end
+  end
+
+  defp loop(model, driver_pid, speed_ms) do
+    receive do
+      {:inline_input, event} ->
+        norm = InputEvent.normalize(event)
+
+        if quit_key?(norm, model) do
+          :ok
+        else
+          model |> Surface.handle_input(event) |> loop(driver_pid, speed_ms)
+        end
+    after
+      speed_ms ->
+        case Surface.advance(model, System.monotonic_time(:millisecond)) do
+          {model, :ok} -> loop(model, driver_pid, speed_ms)
+          {model, :done} -> linger(model, driver_pid)
+        end
+    end
+  end
+
+  # Fixture fully replayed: keep the elapsed ticker honest and keep
+  # listening for fold/jump/quit for a while before exiting on its own,
+  # rather than snapping the terminal away the instant the last block
+  # seals.
+  defp linger(model, driver_pid, remaining_ms \\ @post_done_linger_ms)
+
+  defp linger(_model, _driver_pid, remaining_ms) when remaining_ms <= 0, do: :ok
+
+  defp linger(model, driver_pid, remaining_ms) do
+    tick_ms = 1_000
+
+    receive do
+      {:inline_input, event} ->
+        norm = InputEvent.normalize(event)
+
+        if quit_key?(norm, model) do
+          :ok
+        else
+          model
+          |> Surface.handle_input(event)
+          |> linger(driver_pid, remaining_ms)
+        end
+    after
+      tick_ms ->
+        model
+        |> Surface.tick(System.monotonic_time(:millisecond))
+        |> linger(driver_pid, remaining_ms - tick_ms)
+    end
+  end
+
+  # `q` quits ONLY while the composer buffer is empty -- otherwise it is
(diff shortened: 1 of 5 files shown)


DROOdotFOO:

Reviewed all changed source: `surface.ex`, `view_text.ex`, `status_strip.ex`, `keymap.ex`, `mode_select.ex`, `flat_authority.ex`, `block_body.ex`, `body_provider.ex`, the `block.ex` diff-kind extension, `agent_stream.ex`, and the demo. Three independent personas (Saboteur, New Hire, Security Auditor).

### HIGH
- [Saboteur + Security, promoted MEDIUM->HIGH] `lib/raxol/harness/surface/view_text.ex:1362-1414` (and its callers `surface.ex` `seal_block/2`, `pending_preview_lines/1`, `notice_line/2`) — `ViewText.lines/3` performs display-width truncation but does **no control-byte or newline sanitization** on arbitrary content before that content is written to the terminal / sealed into scrollback. The moduledoc only *assumes* "every harness Component already builds its multi-line bodies as one text child per line," but nothing enforces it. Two concrete failures: (1) any `%{type: :text, content: ...}` whose content contains an embedded `\n` (tool-call results, diff `old`/`new`, agent messages, or composer-submitted text echoed in the stub notice) is emitted as a single iodata row that actually spans multiple physical lines, desyncing `InlineAuthority`'s one-binary-per-row footer/scroll-region accounting -> corrupted pinned region and clobbered history; (2) embedded ESC (`\x1B`) bytes in that same untrusted content pass straight through to the terminal as an escape-injection vector (also miscounts `TextMeasure` width). `Composer.value/1` returns the raw MultiLineInput buffer with no stripping seam, so submitted/steered text flows in unsanitized too. Fixture-only mode limits *today's* blast radius to curated golden data, but this is the assembled production harness path and the invariant is load-bearing and unchecked.

### MEDIUM
- [Security] `packages/raxol_payments/lib/raxol/payments/xochi/agent_stream.ex:2820-2822` — `validate_base_url!/1` is a prefix match: `defp validate_base_url!("http://localhost" <> _)` and `"http://127.0.0.1" <> _`. `http://localhost.evil.com/...` and `http://127.0.0.1.evil.com/...` (or `http://127.0.0.1@evil.com`) all satisfy the guard, defeating the HTTPS-only control and sending the signed announce over plaintext to an attacker-controlled host. The signed payload deliberately omits the human wallet, so exposure is bounded, but a security guard that is trivially bypassable by a crafted hostname is a real defect. Anchor on exact host equality (`localhost`, `localhost:`, `127.0.0.1`, `127.0.0.1:`) instead of a bare prefix.
- [Saboteur] `agent_stream.ex:2573-2586` + `2732-2738` — `announce/2` runs `announce_sync -> build_req -> validate_base_url!` inside an unlinked `Task.start`. The moduledoc promises "Any failure -- config, signing, network, or rate limit -- is logged and dropped; it never propagates," and `announce_sync/2`'s `@spec` is `{:ok, 202} | {:error, term()}`. But a bad base_url makes `validate_base_url!` **raise** after `build_body` has already succeeded, so the `case announce_sync(...)` never matches — the Task dies with an unhandled crash report (noisy SASL), not the clean logged drop the contract states, and the spec is violated. Either return `{:error, :bad_base_url}` from the sync path, or wrap the Task body so config failures degrade to a `Logger` line.
- [New Hire] `surface.ex`, `keymap.ex`, `mode_select.ex`, `status_strip.ex` moduledocs/comments — saturated with private-roadmap jargon (`T13a`, `T2b/T2c`, `T7`, `D-PA (A)`, `AD-1`, `AD-U2`, `R10`, `R11`, `N06`, `SESSION LESSONS #4`, `harness-ui-STATE.md`) and AI meta-narration ("this is not a defect... worth naming honestly", "OPEN QUESTION for T13a"). None of it resolves without the unpublished in-flight docs, and it directly violates this repo's own recorded convention against dev jargon / meta-narration in comments. A new maintainer cannot follow the fold/seal invariants without the private lane state.

### LOW
- [Saboteur] `surface.ex:896-914` — `advance/2` calls `Projection.project(events_so_far, ...)` over a growing prefix on every step, plus repeated `length/1` over `events`/`projection.blocks` lists per call, giving O(n^2) replay cost. Fine for the small golden fixtures; degrades on a long session.
- [New Hire / Saboteur] `examples/harness_fixture_demo.exs:105,134` — the loop hard-intercepts `"q"` before `handle_input`, so the letter `q` can never be typed into the composer during the demo. Confusing dead key for anyone trying to compose.
- [Security] `agent_stream.ex:2577` — jitter uses non-crypto `:rand.uniform/1` and defaults to only 250 ms. The moduledoc sells jitter as a timing-correlation privacy defense; a 0-250 ms predictable-PRNG delay is negligible against network round-trips and is weak as a privacy control (acceptable as defense-in-depth, but do not oversell it).
- [New Hire] `status_strip.ex:499-508` and `view_text.ex:1394-1403` carry a near-identical `truncate`/`truncate_to_width` helper (ellipsis at `width-1`). Duplicated width logic is a drift risk; extract one.

### Cross-Persona Overlaps
- ViewText unsanitized-content-to-terminal was flagged independently by the Saboteur (row desync / pinned-region corruption from embedded newlines) and the Security Auditor (ESC-byte escape injection from untrusted tool/agent content). Promoted MEDIUM -> HIGH.

### Verdict: CONCERNS
One HIGH (promoted) plus three MEDIUM. Not a BLOCK — the HIGH is bounded to curated fixture data in the current fixture-only mode — but the ViewText sanitization gap should be closed before the live agent lane (T13b) feeds real tool output through this path, and the `validate_base_url!` prefix bypass + raise-in-Task contract break are quick, worthwhile fixes.

merklebonsai:
Reconciled your `b2c31128c` (thanks for the direct fix) with the finisher chain that landed in parallel — merged as `3f22b57fe`, your commit preserved as a parent.

Kept the **split-newline** variant over `sanitize_cell`'s strip-newline, because they diverge on multi-line leaf content:

- `sanitize_cell` strips `\n` as a C0 byte → `"a\nb"` collapses to `"ab"` on one row.
- `add_lines/3` splits on `\n` → one collected line per row (correct one-row-per-line accounting; not lossy).

Folded your **DEL-strip** into `add_lines`'s scrub — `0x7F` is `>= 0x20`, so the prior allowlist wrongly passed it; good catch. Removed the `lines/3` seam as redundant (`add_lines/3` is the sole builder of the `{content, style}` entries and scrubs each before `lines/3` ever sees it).

Your injection test stayed (renumbered describe `"9."` — `"8."` is the memory-residency block) and passes against the split impl: ESC stripped, printable residue (`before`/`after`/`home`/`injected`) survives. Suite green, 18 passed.

merklebonsai:
Thanks Drew — dispositions for the round-2 findings below. The ViewText sanitize reconciliation (your `b2c31128c` `sanitize_cell` vs the finisher's split-newline `add_lines/3`, merged as `3f22b57fe`) is already covered in the comment above, so it's only referenced here, not re-explained.

### Accepted + fixed
- **HIGH — ViewText unsanitized content (embedded `\n` row desync + ESC injection)**: `e91f5832` makes `ViewText` the enforced trust boundary — embedded newlines split into real collected rows (restoring `InlineAuthority`'s one-binary-per-row accounting) and ESC/C0 bytes (except `\t`) are stripped from every line, covering tool output, streamed content, and composer echo. Reconciled with your `sanitize_cell` in `3f22b57f` — see the earlier comment for the merge rationale.
- **LOW — demo `q` dead key**: `e91f5832` — `"q"` now quits only when the composer buffer is empty, at both call sites (`loop/3`, `linger/3`), so the letter q is typeable in a prompt again.

### Deferred
- **MEDIUM — `validate_base_url!` prefix bypass**, **MEDIUM — raise-in-Task contract break**, and **LOW — jitter oversell** (`agent_stream.ex`): all three are valid, but this file shows in the diff only because the stacked base drifted — it landed on master via #576, and its hardening went through a separate payments PR that did not cover these three. Routing them to a raxol_payments follow-up rather than this UI lane. Exit criteria: exact-host anchoring for the plaintext allowlist, `{:error, :bad_base_url}` returned so the Task degrades to the documented logged drop (spec restored), and the jitter moduledoc reworded to defense-in-depth — all before the announce path is enabled against non-localhost hosts.
- **MEDIUM — roadmap jargon in moduledocs** (`T13a`/`AD-*`/`R*`/`N06`/STATE references): fair from a fresh-eyes standpoint, but these are load-bearing cross-references into the lane's in-flight decision docs while the stack is under construction. Exit criterion: a plain-language scrub pass over all harness moduledocs when the lane graduates from stacked branches to master (at which point the referenced docs publish alongside).
- **LOW — O(n²) replay in `advance/2`**: deferred to T13b as previously agreed. Golden fixtures keep the prefix small today; exit criterion: incremental projection (project only the appended suffix, drop the per-step `length/1` walks) before the live agent lane feeds long sessions through this path.

### Declined
- **LOW — duplicated truncate helper** (`status_strip.ex` / `view_text.ex`): keeping both local. They're two ~4-line helpers at different layers — StatusStrip's component-local cell ellipsis vs ViewText's post-sanitize iodata truncation — and a shared module would couple the footer component to the view bridge for less code than the extraction costs. Will extract on the third copy.

PR #597, opened 2026-07-16
payments: gate agent-stream announce host + topic format

DROOdotFOO:
Follow-up to #596. Closes the one open adversarial-review finding #596 did not cover: the announce host was context-controlled and only HTTPS-gated, so a poisoned Action context could redirect every signed announce (event + `agentWallet` + signature) to an attacker-controlled endpoint. `topic_id` was also accepted unvalidated.

## Changes
- **Announce-host allowlist** (`resolve_url/2`): an explicit `agent_stream.xochi_api_url` is honored only when its host matches the swap's own Xochi host (`xochi_config.base_url`) or a new `config :raxol_payments, :agent_stream_hosts` allowlist. Otherwise it resolves to `:error`, which #596's diagnostics already surface as `announce_skipped` reason `:no_announce_host`. Default posture: same-host only.
- **Topic-id format validation** (`config/1`): `topic_id` must match Xochi's `TOPIC_ID_RE` (`^[A-Za-z0-9_-]{16,64}$`, byte-matched from `agentStream.ts`) before anything is announced; a malformed topic skips with new reason `:bad_topic_id`.
- Moduledoc updated for both fields.

## Tests
4 new `config/1` tests: malformed topic skipped, non-allowlisted host skipped, allowlisted host honored (plus the existing suite). `mix test test/raxol/payments/xochi test/raxol/payments/actions` -> 296 passed, 0 failures.

## Manual testing
- [ ] Configure an agent with `agent_stream.xochi_api_url` pointing at a non-Xochi host; confirm no announce is sent and `announce_skipped{reason: :no_announce_host}` fires.
- [ ] Add that host to `config :raxol_payments, :agent_stream_hosts`; confirm the announce is sent.
- [ ] Configure a malformed `topic_id`; confirm `announce_skipped{reason: :bad_topic_id}`.

## Notes
Not addressed here (needs mandate infrastructure): full cryptographic binding of `topic_id` to the authorizing Mandate. This PR is format validation + host allowlist only.

### DIFF
--- a/packages/raxol_payments/lib/raxol/payments/xochi/swap_announcer.ex
+++ b/packages/raxol_payments/lib/raxol/payments/xochi/swap_announcer.ex
@@ -22,9 +22,13 @@ defmodule Raxol.Payments.Xochi.SwapAnnouncer do
 
     * `:topic_id` -- the capability handed to the agent out of band at Mandate
       authorization (see the "Capability handoff" section of `AgentStream`).
-      Required; absent, every entry point here is a silent no-op.
+      Required; must match Xochi's topic format (16-64 char base64url). A missing
+      or malformed topic makes every entry point here a silent no-op.
     * `:xochi_api_url` -- announce host. Optional; defaults to
-      `xochi_config.base_url` (the same host the swap ran against).
+      `xochi_config.base_url` (the same host the swap ran against). An explicit
+      value is honored only when its host matches that swap host or the
+      `config :raxol_payments, :agent_stream_hosts` allowlist; otherwise the
+      announce is skipped rather than sent to an unrecognized host.
     * `:mandate_hash` -- telemetry-only, never on the wire. Optional.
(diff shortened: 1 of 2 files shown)


DROOdotFOO:
Closing as superseded by the base branch `feat/wire-agent-stream-announce-578`.

Base already carries a more complete host-gate + topic-format implementation. Diffing this branch against the **current** base is `+43 / -177`: merging would tear out base's implementation and replace it with this earlier, smaller design — dropping the Mandate-binding trust boundary and `announce_stranded`, and renaming the config key (`:agent_stream_hosts` vs base's `:xochi_announce_hosts`) and reason atoms (`:bad_topic_id` vs `:invalid_topic_id`). The host-gate + topic-format goal of this PR is fully met on base.

Nothing is lost functionally. Reopen if the `:agent_stream_hosts` naming/config approach is worth revisiting on its own.

PR #579, opened 2026-07-16
raxol_payments: wire AgentStream.announce into the swap execution path

DROOdotFOO:
Follow-up to #575 / #576. Wires the merged `Raxol.Payments.Xochi.AgentStream`
into the agent's swap execution path so a Mandate-authorized swap auto-emits a
signed activity row to the user's Xochi live feed.

## What's wired

The two agent Actions that already run a Mandate-authorized swap:

- **`ExecuteXochiIntent`** announces the **execute** (non-terminal) event -- full
  route from the quote request + response (`intent_id`, chains, tokens,
  `from_amount`, `to_amount`, `status`, `ts`).
- **`PollXochiStatus`** announces the **terminal** event (`completed` / `failed`
  / `refunded` / `expired`) so the subscriber's row advances on its own.

Both emits are **best-effort and non-blocking** -- `AgentStream.announce/2` is
fire-and-forget; a failed or unconfigured announce never touches the swap.

## Config plumbing (the crux)

New `Raxol.Payments.Xochi.SwapAnnouncer` resolves the per-Mandate config from an
optional `:agent_stream` key on the Action context:

```elixir
context = %{
  wallet: MyAgentWallet,
  xochi_config: %{base_url: "https://api.xochi.fi", auth: {:mandate, w}},
  agent_stream: %{topic_id: "abc...", mandate_hash: "0x..."}   # <- new, optional
}
```

- `topic_id` is the out-of-band capability handed to the agent at Mandate
  authorization -- never derived, never inside the signed Mandate envelope, never
  alongside the delegator wallet. **Absent it, every emit is a silent no-op.**
- `xochi_api_url` defaults to the swap's own `xochi_config.base_url`.
- `wallet` (signing + on-wire `agentWallet`) is the agent's own wallet. The
  human/delegator wallet never touches the wire.
- `mandate_hash` is telemetry-only.

## The terminal-route problem

`IntentStatus` carries no route, and the Xochi browser **merges feed rows by
`intentId`** (`localActivity.ts` `{...existing, ...item}`), so a terminal event
with a blank route would clobber the row. `PollXochiStatus` only sees the
`intent_id`. Rather than thread route data through the poll Action's LLM-tool
context, `ExecuteXochiIntent` stashes the execute-time route in
`Raxol.Payments.Xochi.SwapRouteStore` (a best-effort, self-starting, TTL-pruned,
size-capped ETS singleton) keyed by `intent_id`; the poll takes it back to
[...]

PR #598, opened 2026-07-16
raxol_payments: gate announce host + bind topic to Mandate (#3)

DROOdotFOO:
Stacked on #579. Closes the remaining **HIGH** exposure from the adversarial review (the last open blocker): a context-influenced announce host or an unvalidated `topic_id` would exfiltrate a signed activity row to an attacker host or into another user's feed.

`SwapAnnouncer.config/1` now fails closed on three axes before it will announce, each with a specific `:announce_skipped` telemetry reason:

## 1. Announce-host allowlist

The host must be the swap's **own** Xochi host (already trusted to move the funds), `localhost`/`127.0.0.1`, or a host on the operator allowlist (`config :raxol_payments, :xochi_announce_hosts`, default `["api.xochi.fi", ...]`). An injected `agent_stream.xochi_api_url` pointing anywhere else is rejected (`:host_not_allowed`). Same-host announces (the normal case) are unaffected, so existing config keeps working.

## 2. topic_id format

The topic must match the Xochi capability format `[A-Za-z0-9_-]{16,64}` (mirrors `TOPIC_ID_RE` in the shared contract), so a truncated or malformed topic never ships (`:invalid_topic_id`).

## 3. Topic bound to the authorizing Mandate

When the swap surfaces the authorizing mandate (`context.authorizing_mandate_hash`), the topic's declared `mandate_hash` must equal it — a topic issued for one mandate cannot ride a swap under another (`:mandate_binding_required`). Absent that context, a binding-required deployment (production by default, or `config :raxol_payments, :require_topic_mandate_binding`) still requires the topic to name its mandate. Full cryptographic ownership is a Xochi-side check (the worker verifying the signature against the mandate the topic was issued for) and lands with the authorization handoff — documented in the module's new "Trust boundary" section.

## Tests

New `config/1 trust boundary` describes: host allowlist (own host / default / injected-reject / allowlisted-accept), topic format (valid / too-short / illegal-chars), and Mandate binding (match / mismatch / known-but-undeclared / required-without-mandate / no-op). Plus `:announce_skipped` telemetry assertions for `:invalid_topic_id`, `:host_not_allowed`, `:mandate_binding_required`.

## Manual testing

- [x] `MIX_ENV=test mix test` in `raxol_payments` -- 1112 passed, 0 failures
- [x] `mix format --check-formatted` + `mix credo` clean on changed files
- [x] Existing same-host / valid-topic / non-production usage unaffected

With this, all three of the review's live HIGH+CRITICAL items are resolved (redaction, route-store hardening, and now host/topic binding).

### DIFF
--- a/packages/raxol_payments/lib/raxol/payments/xochi/swap_announcer.ex
+++ b/packages/raxol_payments/lib/raxol/payments/xochi/swap_announcer.ex
@@ -47,6 +47,28 @@ defmodule Raxol.Payments.Xochi.SwapAnnouncer do
   and defeat the very unlinkability those modes exist for. The execute and
   terminal rows share the same pseudo id so the browser still merges them into
   one advancing row.
+
+  ## Trust boundary: where and under whose topic a signed row may go
+
+  The `topic_id` is a capability and the announce host is signature-free, so a
+  context that names a hostile host or a wrong topic would exfiltrate a signed
+  activity row to an attacker or into another user's feed. `config/1` therefore
+  fails closed on three axes before it will announce:
+
+    * The announce host must be the swap's own Xochi host, `localhost`, or a
+      host on the operator's allowlist (`config :raxol_payments,
+      :xochi_announce_hosts`). An injected `agent_stream.xochi_api_url` pointing
+      elsewhere is rejected.
+    * The `topic_id` must match the Xochi capability format
+      (`[A-Za-z0-9_-]{16,64}`), so a truncated or malformed topic never ships.
+    * The topic is bound to its authorizing Mandate. When the swap surfaces the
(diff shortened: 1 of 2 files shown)


DROOdotFOO:
Closing as superseded by the base branch `feat/wire-agent-stream-announce-578`.

Diffing this branch against the **current** base (not the stale merge-base), the only net change here is `-29 / +0`: this branch is *missing* `announce_stranded/2`. The entire host-gate + topic-format + Mandate-binding implementation this PR proposes is already on base, byte-identical — base is effectively `this PR's lineage + #599's stranded-row`. Merging would regress base by deleting the `announce_stranded` feature (#599) and add nothing.

Nothing is lost: base is a strict superset of this branch. Reopen if I'm missing something.

PR #601, opened 2026-07-16
test(harness): raise timeouts on two CPU-heavy substrate property tests (Windows CI)

merklebonsai:
The Windows matrix job on #595 failed with two 60s `ExUnit.TimeoutError`s in merged substrate suites — not a T13a regression:

- `test/property/renderer_seal_once_property_test.exs:66` — the 1000-block immutable-prefix keystone (full emulator replay at each of 9 checkpoints)
- `test/property/renderer_footer_property_test.exs:498` — the footer cursor round-trip property (many generated repaint/keyframe rounds, each byte-checked)

Both are legitimately CPU-heavy and pass comfortably on Unix runners and locally (22 tests, seconds). The Windows runner is CPU-starved enough to blow the default 60s — a runner-speed artifact, not a hang.

Fix: `@tag timeout: 180_000` on exactly these two tests, with a comment explaining why. Coverage is preserved on every platform (no `skip_on_windows`), and a genuine infinite loop still fails — just at the raised bound.

After this merges, re-running the failed Windows job on #595 should clear it.

### DIFF
--- a/.github/workflows/ci-unified.yml
+++ b/.github/workflows/ci-unified.yml
@@ -537,7 +537,12 @@ jobs:
 
       - name: Run tests (Windows)
         if: runner.os == 'Windows'
-        run: mix test --exclude slow --exclude docker --exclude integration --exclude skip_on_ci --exclude skip_on_windows
+        # --timeout 180000: the Windows runner is CPU-starved enough that
+        # emulator-replay-heavy property tests (substrate keystones, SSH
+        # rendering round-trips) blow the 60s default as a runner-speed
+        # artifact — a different one each run. Raising the job-wide ceiling
+        # keeps full coverage; genuine hangs still fail at the raised bound.
+        run: mix test --timeout 180000 --exclude slow --exclude docker --exclude integration --exclude skip_on_ci --exclude skip_on_windows
         env:
           MIX_ENV: test
           SKIP_TERMBOX2_TESTS: "true"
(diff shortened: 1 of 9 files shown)


merklebonsai:
The re-run failed on a *different* heavy property (`SSHRenderingTest` CRLF round-trip, same 60s `ExUnit.TimeoutError` via `AnsiReplayer`) while the two tagged tests passed — the slow runner flakes a different emulator-replay-heavy test each run, so per-test tagging is whack-a-mole. Added a commit raising the Windows job's test timeout to 180s job-wide (`mix test --timeout 180000`, Windows step only; Unix keeps the 60s default). The two per-test tags stay as self-documentation on the heaviest substrate tests.

merklebonsai:
Root cause found (V's bisect was right about the trigger commit, and it's load, not breakage):

**Bisect:** master Windows CI green through T2a (`b49e0c613`), first red exactly at the T2b squash (`0923528ea`). But T2b's diff is disjoint from every failing test — and on that first red run, T2b's own tests PASSED while `SSHRenderingTest` (untouched since long before) timed out.

**Mechanism:** all the emulator-replay property suites are `async: true` and CPU-bound. Measured in isolation: SSH suite ~10s CPU, T2b keystone ~28s CPU on a fast machine — 4-6x that on the 2-core Windows runner. T2b added enough concurrent load that ExUnit's async pool starves whichever heavy suite loses the scheduling draw past the 60s default. Victim rotates per run — which is why the per-test tags kept chasing the wrong test.

**Fix (this PR now):** `async: false` on the seven emulator-replay suites. CPU-bound work on 2 cores gains nothing from concurrency — serializing removes the contention class entirely at ~10s local wall cost (305 property tests green in 35s). The job-wide 180s ceiling stays as defense-in-depth; the two per-test tags stay as documentation.

PR #604, opened 2026-07-16
Harness UI: polarity-aware 16-color salience mapping (semantic roles survive degradation)

merklebonsai:
On 16-color terminals, nearest-RGB quantization erases exactly the signal the harness exists to show — measured on our own solver output: 8 of 11 semantic fields collapse to gray on a dark ground (6 of 11 on light), and the solved error color quantizes to ANSI 3 — yellow. Error arriving as yellow is the instrument lying.

## What this adds

`Raxol.UI.Theming.Ansi16Salience` — a pure role -> ANSI-slot table for the 12 semantic roles the harness paints (the 8 salience seeds + block chrome + diff add/del + running). Callers gate on terminal capability and ask by ROLE, never by RGB. Additive only: truecolor and 256-color paths untouched, byte-exact theme fixtures untouched.

Design rules:
- Category preserved: error stays red-family, success green, warning yellow, accent blue, running magenta — on both polarities (bright 9-15 loud on dark canvas, normal 1-7 on light; tiers fold pairwise at the 0.8 threshold).
- Emphasis deliberately trades its warm hue for the max-contrast neutral slot so it can never be confused with the warning signal (yellow is reserved for state).
- Muted/border are the documented recede floor — the only roles exempt from the loud/soft tier distinction.

## Regression tripwires (all red-first)

- `@chromatic_gray_budget 0`: no chromatic role may land on a gray slot, both polarities, both tiers.
- `@cross_category_collision_budget 0`: no slot hosts two hue families.
- Measured-collapse pins on the naive path (>=6/11 gray dark, >=5/11 light, error-quantizes-to-yellow) — proving the audit is testing a real failure, not a vacuous claim.

## Verification

Full test file written before implementation: 10/12 red. After: 12/12 green; theming+harness sweep 199 passed, zero regressions; full suite 5814/5816 (both failures pre-existing/environmental, confirmed present with these changes removed). `--warnings-as-errors` + format clean.

### DIFF
--- a/lib/raxol/ui/theming/ansi16_salience.ex
+++ b/lib/raxol/ui/theming/ansi16_salience.ex
@@ -18,7 +18,32 @@ defmodule Raxol.UI.Theming.Ansi16Salience do
   measuring distance in RGB space, each semantic role is PINNED to a
   hue-preserving ANSI slot chosen so the role's category (red/green/yellow/
   blue/magenta/neutral) is never lost and never confused with another
-  role's category.
+  role's category. (`:running` is reserved for activity state -- no RGB
+  seed exists in the harness yet, which is why measured naive-collapse
+  counts say "11 fields" while `roles/0` returns 12.)
+
+  ## Legibility floor
+
+  Every slot assignment meets a WCAG-style 3:1 contrast ratio against its
+  polarity's canonical ground, measured on the module's reference palette
+  (`Raxol.UI.Theming.Colors.ansi_to_rgb/1`), with one small, named
+  exemption set on light polarity only:
+
+    * Green/yellow state roles on light ground: no green or yellow slot in
+      the reference palette is legible on light ground (best available:
+      normal green 1.98:1, normal yellow 1.56:1). The best-effort normal
+      slots (2 and 3) are pinned anyway, because swapping hue family would
+      be a category lie worse than low contrast for a state signal, and
+      gray would erase the signal entirely. Real light-mode terminal
+      themes darken these slots; the pin keeps the category-true handle
+      for them to interpret.
+    * Muted/border recede chrome on light ground: silver is a subtle
+      light-UI border by design; receding below the floor is these roles'
+      function.
+
+  Dark polarity has NO exemptions: every dark-canvas assignment meets the
+  floor outright. The ranked priority throughout the table is explicit:
+  legibility > category preservation > tier separation.
 
   ## Polarity-awareness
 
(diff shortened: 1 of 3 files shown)


DROOdotFOO (commented):

Reviewed `ansi16_salience.ex` (new, 220 LOC), the doc pointer in `colors.ex`, and the
test. Verified against `salience.ex` (`resolve_polarity/2` threshold), `salience_theme.ex`
(8 seeds), and the ANSI palette in `colors.ex:318`. The mapping design is genuinely sound
where it is tested (no cross-category collisions at any single prominence, polarity
threshold matches the solver at `ground < 0.5`). The findings below are what the tripwires
do NOT cover.

### HIGH
- **[Saboteur + New Hire] legibility is never asserted anywhere** `ansi16_salience.ex:216-219`, test whole file -- The module's entire purpose is "survive degradation *and stay legible*", yet no test asserts any contrast floor against the ground; the audit only checks gray-collapse and category. The polarity flip `chromatic_slot(:light, :soft, base) -> base + 8` (line 219) sends every receded chromatic role to a **bright** slot on a light canvas: soft `warning` -> slot 11 (bright yellow `#FFFF00`) and soft `success`/`diff_add` -> slot 10 (bright green `#00FF00`). On a near-white ground (`0.97`) bright-yellow-on-white is ~1.05:1 contrast -- effectively invisible, not merely "receded". `warning`'s own seed tier is `:differentiate` (`salience_theme.ex:29`), i.e. it naturally resolves to the soft tier, so on a light terminal a warning can render unreadable. Fix: add a contrast-ratio assertion (WCAG floor, e.g. >= 3:1) per role against the polarity's canonical ground, and clamp/re-pin the light-soft yellow/green slots that fail it.

### MEDIUM
- **[Saboteur] the module has zero callers -- the behavioral claim is not delivered** `ansi16_salience.ex:1` -- Nothing in `lib/` or `packages/` references `Ansi16Salience`; `find_closest_basic_color/1` is still the only live 16-color path (`colors.ex:425`, `convert_to_basic`). The PR title asserts "semantic roles survive degradation" but no render/capability-gate path consumes the table, so today they still don't. Additive scaffolding is fine as a seam, but the title/body over-claim relative to what is wired. Confirm a follow-up wires it, or soften the claim.
- **[Saboteur] undocumented neutral distinction-loss on dark @ receded** `ansi16_salience.ex:122-138` -- At soft tier on dark, `foreground` (8), `chrome` (8), `muted` (8) and `border` (8) all collapse to slot 8. The moduledoc only documents the `:muted`/`:border` fold as the "recede floor"; it never says `foreground` and `chrome` also merge into that floor when receded. The `@cross_category_collision_budget` test can't catch this (all `:neutral`), and the tier-separation test exempts only `:muted`/`:border`. So four distinct semantic roles become indistinguishable, silently. Either document this merge explicitly or move `foreground`/`chrome` soft off slot 8.

### LOW
- **[Security Auditor] the degradation path itself crashes on its own fallback input** `ansi16_salience.ex:154` -- `polarity/1` guards `is_number`, so a `nil` ground (the exact case the moduledoc names: "no OSC 11 detection") raises `FunctionClauseError` instead of defaulting to `reference_ground()`. Likewise `slot/3`/`category/1` raise on any role atom outside the closed set. Fail-loud on unknown roles is defensible (and correctly avoids `String.to_atom`), but a 16-color path is a *fallback*; crashing the harness when its background probe is unavailable is fragile. Add a `polarity(nil)` clause (or a documented caller contract enforced by a guard) that lands on `reference_ground()`.
- **[New Hire] "11 fields" vs 12 roles is unexplained** `ansi16_salience.ex:11` -- Docs and tests repeatedly say "8 of 11" / "11 harness-painted fields", but `roles/0` returns 12. `:running` is a role with no RGB seed, so it is excluded from the naive-collapse count -- correct, but never stated. A maintainer updating the table can't reconcile 11 vs 12 without reverse-engineering it. One sentence naming `:running` as the count-exempt reserved role fixes it.

### Cross-persona overlaps
- Saboteur and New Hire independently landed on legibility: illegible slots can ship (Saboteur) *because* no contrast/legibility floor is asserted (New Hire). Promoted to HIGH.

### Verdict: CONCERNS

The design is careful and the tripwires are real, but two gaps keep it short of CLEAN: the
mapping can emit invisible chromatic slots on light grounds with no test guarding contrast,
and the whole module is currently unwired so the headline behavior is not yet delivered.
Neither is a hard production break (nothing calls it yet), so not BLOCK.

merklebonsai:
Review response landed as `185c7be6f`. All findings verified by measurement against `Colors.ansi_to_rgb/1` on the canonical grounds before touching anything. Point-by-point:

**HIGH — light-soft flip emits illegible bright slots; legibility never asserted.** *Confirmed — and understated.* The reported failures verified exactly (soft warning -> slot 11 = 1.02:1, soft success -> 10 = 1.26:1), and measurement found MORE: light-LOUD warning (slot 3, 1.56:1) and loud success (slot 2, 1.98:1) also fail — the reference palette has NO green/yellow slot >= 3:1 on a light ground — plus a dark-side failure the review missed: receded accent (slot 4, #0000EE) = 1.93:1. Red-first: the 3:1 floor test (role x tier x polarity), a fade-direction invariant (soft may never read louder than loud), and an exemption-pin test written first (7/22 red); then the table re-pinned: dark accent -> cyan 14/6; light warning/success/diff_add/running fold BOTH tiers onto their single legible slot; light foreground/chrome soft -> 8. Exemptions are an exact named set, each pinned to its documented slot so the list can't drift; "dark needs no exemptions" is itself asserted. Ranked priority documented: legibility > category > tier.

**MED — zero callers; PR over-claims delivery.** *Confirmed.* Moduledoc now states plainly the render path does not consume the table yet and `find_closest_basic_color/1` remains the live lossy fallback until the capability gate is wired. PR title updated accordingly ("mapping table + tripwires; wiring is a follow-up"); the wiring follow-up touches paint authorities/capability gates and is deliberately out of this unit's scope.

**MED — undocumented neutral merge on dark receded.** *Confirmed; documented + pinned rather than re-pinned:* slot 8 is the only legible sub-body neutral on dark (4.52:1) — moving fg/chrome soft off it would collide with the loud body slot. Now in the accepted-losses section with a dedicated pin test; the light-side mid-ramp compression documented the same way.

**LOW — `polarity/1` crashes on nil.** *Confirmed, fixed:* nil falls back to the reference ground's polarity; other non-numbers still raise (fail-loud on garbage, graceful on the documented unknown). Red-first test included.

**LOW — "11 fields" vs 12 roles.** *Confirmed, fixed:* `:running` is reserved with no RGB seed in the harness, hence excluded from measured naive-collapse counts — now stated in both moduledoc and the enumeration test.

Gates: target file 22/22 (red-first both rounds: 10/12 red round 1, 7/22 red round 2); full suite 5816/5822 with all 4 failures proven environmental/pre-existing; warnings + format clean.

PR #605, opened 2026-07-16
Harness UI: shared seal-frontier classifier (one committability decision, every consumer)

merklebonsai:
The foundation unit for hardening the seal pipeline before live traffic. One pure classifier decides which finished blocks may be sealed into terminal scrollback; every consumer (the seal pass today; the resize gate and viewport sizing when they land) consults it, so they can never disagree about where the frontier stops — the disagreement failure mode is a block's height flipping between the live region and native scrollback, jumping the prompt on every seal.

## The module

`Raxol.Harness.SealFrontier` (pure, 433 lines with full design-note moduledoc):
- `classify/3` -> `:commit | :skip | :stop` with a fixed decision order: pending-input check FIRST and unconditional on turn state (a block awaiting the human must never seal — print-once would freeze the "waiting" form in scrollback), idle relaxation past stale running flags, and two committable-while-running exceptions (a non-last message is provably complete; a background-task lifecycle block's running flag is animation-only).
- `frontier_scan/1` — read-only projection (`tail_start`, `will_commit`) that must mirror the mutating walk exactly; the agreement is pinned by a property test over arbitrary generated states.
- `commit_walk/5` — the ONE mutating walk: emit-then-mark, halting on write failure with the cursor held strictly before the failed entry (a print-once mode can never re-emit a block marked committed that never reached the terminal — it would silently vanish).
- Per-kind display-mode policy frozen at seal (reasoning collapsed, tool calls truncated, diffs expanded).

## The wiring (behavior-identical by construction)

- Projection-agnostic entries: plain maps produced by `Surface.frontier_entries/1`; committed-set = the painted high-water prefix; the live streaming tail never enters the list.
- The surface's one-advance foldable window maps to `pending_input?: true` on the newest block — a fold toggle is exactly the still-mutating-rendered-form hazard the pending gate exists for. The scan lands where the old inline arithmetic did; the existing acceptance suite proves it.
- Cursor is caller-held and a lower-bound hint only; `cursor_after_removal/2` / `cursor_after_truncate/2` cover list mutations.

## Follow-up seams (documented, deliberately not invented)

Resize-path `will_commit` gate + post-commit viewport sizing (the frame-order unit); two-phase write-confirmed sealing building on the `{:error, :write_failed, _}` branch; resize-during-commit recorded as a named red-first obligation in the test file.

## Verification

Red-first: corpus written and failing before the module existed. New suite 17/17 (16 corpus cases + scan/walk agreement property). Full harness+property sweep: 612 passed, 0 failures. `--warnings-as-errors` + format clean.

### DIFF
--- a/lib/raxol/harness/seal_frontier.ex
+++ b/lib/raxol/harness/seal_frontier.ex
@@ -1,0 +1,433 @@
+defmodule Raxol.Harness.SealFrontier do
+  @moduledoc """
+  The shared classifier for "which blocks may seal (commit) this frame" --
+  the one piece of logic every consumer that needs to answer that question
+  goes through, instead of each restating its own walk.
+
+  ## The committed frontier
+
+  The committed frontier is the leading contiguous run of already-committed
+  entries plus whatever newly-committable entries follow them. Everything
+  past that point -- `tail_start` in `scan/0` and `commit_walk/5`'s return
+  -- stays in the repaintable live region (in our substrate, the
+  `InlineAuthority` footer viewport) until it, in turn, becomes
+  committable. Sealed history is print-once: once a block is written
+  through `InlineAuthority.seal/2`, it is never repainted. Getting the
+  frontier boundary wrong in either direction is user-visible: too eager
+  and a still-mutating block freezes mid-change in permanent scrollback;
+  too conservative and a done block never leaves the footer.
+
+  ## Why one shared classifier
+
+  Three concerns all need to agree on exactly where the frontier stops, or
+  a block's rendered height flips between the live region and native
+  scrollback and the prompt visibly jumps at commit time:
+
+    * the seal pass -- which blocks get physically written into terminal
+      scrollback this frame;
+    * the footer/tail composition -- which blocks still render in the
+      pinned live region (the trailing "pending" preview);
+    * (future, documented seam, not yet built) the resize path's
+      `will_commit` gate and post-commit viewport sizing -- the frame-order
+      unit that must decide sizing at the width a possibly-just-committed
+      block was laid out at.
+
+  Consumers never inline their own walk over entries. The consumer table
+  today:
+
+    * `Raxol.Harness.Surface.paint_pending_blocks/1` -- the one mutating
+      walk, via `commit_walk/5`.
+    * `Raxol.Harness.Surface`'s footer pending-preview (`pending_block/1`,
+      via `frontier_scan/1`, itself `scan_frontier/3`) -- read-only.
+
+  Documented seams (not yet built, but must call `scan_frontier/3` before
+  the commit pass in the frame and mirror its stop condition exactly when
+  they land): the resize-path `will_commit` gate, and post-commit viewport
+  sizing.
+
+  ## The entry contract
+
+  Entries are projection-agnostic plain maps (`t:entry/0`) so this
+  classifier stays decoupled from any one producer's data model. Today
+  `Raxol.Harness.Surface.frontier_entries/1` builds them from projection
+  blocks plus that module's own painted high-water mark. The live tail
+  (still-streaming, not-yet-a-block items) never enters the entry list at
+  all: a still-streaming item has no committable form until it completes
+  into a block, so it is definitionally past the frontier already -- there
+  is nothing for this classifier to say about it.
+
+  ## Decision order and rationale (`committable?/3`)
+
+  The order below is load-bearing -- each step assumes the ones before it
+  already ran:
+
+    1. `pending_input?` true -> not committable, UNCONDITIONALLY, regardless
+       of turn state. An entry awaiting a user answer (e.g. a permission
+       prompt) still has a rendered form that changes when the prompt
+       resolves; committing it now (print-once) would freeze the "waiting"
+       form forever. This check runs BEFORE the idle relaxation below on
+       purpose: idle relaxation exists to forgive a *stale running flag*,
+       never a live pending-input mark, which is exactly why pending input
+       is checked first and unconditionally.
+    2. `not turn_running?` -> committable (the idle relaxation). A producer
+       can leave a `running?` flag set on an entry after the turn has
+       already ended (a finalize event missed at a transition boundary);
+       if idleness didn't override a stale flag, that entry would
+       permanently wedge the frontier for the rest of the session. This
+       relaxation only ever reaches entries that already passed the
+       pending-input check, so it can never release something still
+       awaiting input.
+    3. `not running?` -> committable (the ordinary, no-exceptions case: a
+       finalized entry with nothing outstanding always commits).
+    4. Running, mid-turn: not committable UNLESS one of exactly two
+       exceptions applies:
+         * `kind == :background_task` -- always committable, even as the
+           last entry of a still-running turn. A background-task entry is
+           a lifecycle marker; its `running?` flag drives an animation
+           only, and its *content* never subsequently changes --
+           completion arrives later as a wholly SEPARATE entry. Gating it
+           on the flag would wedge the frontier for the rest of the turn
+           for no reason: nothing about this entry will ever change.
+         * `kind == :message and not is_last?` -- committable. A message
+           with a LATER entry already in the list is provably complete:
+           the producer has moved on, which could only happen once this
+           message stopped changing, regardless of what its own `running?`
+           flag says. Tools get NO such relaxation: a running tool call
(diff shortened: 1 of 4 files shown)


DROOdotFOO (commented):

Scope: `Raxol.Harness.SealFrontier` (new shared classifier) + its two live consumers in
`Raxol.Harness.Surface` (`paint_pending_blocks/1` via `commit_walk/5`, `pending_block/1` via
`frontier_scan/1`). Behavior-identity with the old inline arithmetic checks out: the detach
target and the committed range both map to the same indices the old `target`/`slice`
arithmetic produced, so nothing regresses functionally today. The issues below are a hot-path
scalability regression that contradicts the module's own stated design, plus two documented
"single source of truth" seams that are not actually wired and whose docs overstate coverage.

### HIGH
- **[Saboteur] HIGH** `lib/raxol/harness/seal_frontier.ex:292` and `:352` -- the walks are
  O(n^2), directly contradicting the moduledoc's own guarantee. Lines 129-134 claim the walks
  are "implemented over the suffix list carrying the absolute index along (to avoid `length/1`
  and `Enum.at/2` costing O(n) per step, which would make a full walk O(n^2))". The actual
  `scan_walk`/`commit_step` do the opposite: they keep the FULL `entries` list and re-index it
  with `Enum.at(entries, index)` every step -- O(index) on a linked list, so O(n^2) over a
  walk. This is on the hot path: `frontier_scan/1` runs from cursor 0 on every footer render
  (`surface.ex:1011`) and `paint_pending_blocks/1` runs a cursor-0 scan on every advance
  (`surface.ex:644`), while `model.projection.blocks` accumulates every completed block for the
  whole session. A long agent run degrades quadratically per frame. This is exactly the cliff
  the PR says it is avoiding, and the PR's own framing is "before live traffic." Fix: walk the
  suffix by pattern-matching `[entry | rest]` and carrying the absolute index as a counter (as
  the docstring already describes), never `Enum.at` the full list; `classify/3` similarly should
  not `length/1` + `Enum.at/2` the whole list per call.

- **[Saboteur + Security] HIGH (promoted)** `lib/raxol/harness/surface.ex:610` vs
  `seal_frontier.ex:64-77` -- the pending-input gate's only real feed does not implement the
  semantics the gate documents. The moduledoc motivates step 1 with "an entry awaiting a user
  answer (e.g. a permission prompt) ... committing it now (print-once) would freeze the
  'waiting' form forever" and makes the check unconditional precisely for that case. But the
  sole producer sets `pending_input?: not reveal_finished? and index == total - 1` -- purely the
  newest-block-during-reveal window, with no relationship to any block actually awaiting input.
  A real `:approval`/permission block that is not the newest entry, or any pending block once
  the reveal finishes, gets `pending_input? = false` and seals into print-once scrollback while
  still live -- the exact failure the gate exists to prevent (frozen "waiting" form / a
  human-facing prompt committed mid-question). Dormant today (no approval/permission producer
  emits these yet), but the classifier ships the gate with a feed that cannot honor its
  contract. Fix: derive `pending_input?` from the block's actual awaiting-input lifecycle, or
  explicitly scope approval/permission blocks out of frontier eligibility until that feed
  exists, and stop using the reveal window as a proxy for "awaiting the human."

### MEDIUM
- **[New Hire + Security] MEDIUM** `lib/raxol/harness/seal_frontier.ex:435-438` and `:250` --
  two APIs documented as the canonical single place are unwired dead code, with docs that
  overstate their status. (1) `seal_display_mode/1` is called only by tests; no seal-path code
  (`seal_block/2`, `render_block_lines/3`, `BlockBody`) consults it, yet the moduledoc says the
  per-kind fidelity policy "lives here, in one place." Its `:tool_call -> :truncated` rule is
  the only documented bound on tool output reaching permanent scrollback, and it is never
  applied -- unbounded, tool-controlled content (which can carry raw ANSI/control sequences) is
  committed verbatim. (2) `classify/3` has ZERO callers and ZERO test coverage, yet the
  moduledoc calls it "the public single-step primitive" and claims (line 134) its equivalence to
  the walks "is covered by the scan/walk-agreement property test" -- the property test
  (`seal_frontier_test.exs:992`) never calls `classify/3`; it only ties `scan_frontier`<->
  `commit_walk`. So `classify/3` is a third, independent copy of the committability step order
  whose asserted test coverage does not exist, free to drift from the walks undetected. A new
  hire trusting these docs would wire a consumer to the "canonical, tested" primitive and get an
  untested path. Fix: either wire them for real (have `seal_block` consult `seal_display_mode`;
  route the walks through `classify/3`, or add a direct `classify/3` equivalence property) or
  delete them and correct the moduledoc's "one place" / "covered by the property test" claims.

### LOW
- **[New Hire] LOW** `lib/raxol/harness/seal_frontier.ex:232` -- `committable?(entry,
  turn_running?, is_last?)` takes two adjacent positional booleans with no compile-time
  protection; transposing them silently miscomputes the frontier. The wrapper naming also
  invites confusion: `Surface.frontier_scan/1` vs `SealFrontier.scan_frontier/3` are
  near-homographs for wrapper-vs-impl. Consider a keyword/struct for the flags and a distinct
  wrapper name.

### Cross-persona overlaps
- The `pending_input?` feed gap is raised by both Saboteur (mis-seal / frontier correctness) and
  Security (a block awaiting the human sealed into print-once history); same root cause ->
  promoted to HIGH above.
- The unwired-policy finding is shared by New Hire (dead code + false single-source docs) and
  Security (the only tool-output truncation bound is never applied); consolidated into one
  MEDIUM.

### Verdict: CONCERNS

Nothing here breaks correctness for today's traffic (running flags always false, no
approval/background-task producer, behavior-identical to the prior arithmetic). But the module's
central design claim -- avoiding an O(n^2) walk -- is factually violated on a per-frame hot path
that grows unboundedly over a session, and two of the classifier's documented "one place"
policies (`classify/3`, `seal_display_mode/1`) are unwired with docs that overstate their
coverage/authority. Given the PR's explicit "before live traffic" goal, fix the O(n^2) walk and
reconcile the docs before merge.

merklebonsai:
Review response landed as `7987da408` (pending-gate feed) + `38a82260b` (O(n) walks + doc reconciliation). Point-by-point:

**HIGH-1 — O(n^2) walks contradicting the moduledoc's suffix-list claim.** *Confirmed.* Resolved code-to-doc (not doc-downgrade): the unit's whole framing is "before live traffic" and the entry list grows per session, so the doc's promise is the right contract. `scan_walk`/`commit_step` now pattern-match `[entry | rest]` carrying the absolute index (`rest == []` is the is-last check); the commit walk rebuilds via reversed walked-prefix (O(1) mark, one `Enum.reverse/2`); `classify/3` uses a single `Enum.drop`. **Measured linear**: 1k/10k/100k entries -> scan 12/120/912us, walk 57/527/7194us.

**HIGH-2 — pending-input gate's only feed implements the reveal window, not the documented awaiting-the-user semantics.** *Confirmed (dormant today — no producer emits live blocks yet — but the feed couldn't honor the contract).* Reconciled both directions: the feed now derives genuine awaiting-input (`Block.live?/1 and kind == :approval` — Block's own contract) OR the foldable window, and both moduledocs restate the gate's actual invariant with both instances named. Red-first: new `surface_frontier_feed_test.exs`, 2/4 cases red against the window-only mapping (non-newest position, idle-turn-post-reveal), green after.

**MED-a — `seal_display_mode/1` dead code with a false "lives here in one place" claim.** *Confirmed (zero non-test callers).* Doc made honest, table kept: it's the spec-mandated policy the commit-cap follow-up consumes; wiring truncated tool rendering now would be renderer-level feature work the banked spec explicitly defers, and would change golden history output. Doc now states plainly: "NOT YET WIRED — sealed tool output is NOT bounded until the follow-up lands."

**MED-b — `classify/3` zero coverage, false coverage claim.** *Confirmed.* Added 3 direct unit tests + a classify-agreement property (a reference walk of repeated `classify/3` calls must agree with both suffix walks for arbitrary states) — which then guarded the HIGH-1 rewrite. Coverage claim corrected.

**LOW — positional booleans; near-homograph naming.** *Partially accepted*: @specs now name every argument and `committable?/3`'s doc states the transposition hazard. Renames refuted: `classify`/`scan_frontier` signatures are pinned by the ported reference spec; `Surface.frontier_scan/1` is deliberately the Surface-side consultation wrapper (doc states the relationship).

Gates: new suites 25/25 (2 properties), full harness+property sweep 620 passed / 0 failed, warnings-as-errors + format clean.

PR #606, opened 2026-07-16
Harness UI: streaming render cadence layer (input-first backpressure for live sessions)

merklebonsai:
The backpressure layer between an LLM token stream and the harness surface: live tokens arrive in network-rate bursts; painting per-token floods the terminal and starves input. Design: unbounded ingest, cadence-throttled egress, bounded drain, input always ahead of paint.

## `Raxol.Harness.CadencePolicy` (pure — no processes, no clocks)

`decide(now_ms, last_flush_ms, pending_count, input_pending?)` -> `:flush_now | {:defer, ms} | :yield_to_input`, decision order:
1. Input pending -> yield unconditionally — input beats even the first paint.
2. First delta -> flush immediately (no artificial 16ms latency on turn start).
3. Elapsed >= interval -> flush.
4. Else defer to the window's FIXED deadline — a burst coalesces into one timer instead of resetting per delta.

Constants with rationale: 16ms flush interval (one 60fps frame), 32-delta drain bound (input waits at most one bounded batch, never the flood), 1ms input-yield recheck.

## `Raxol.Harness.StreamCadence` (thin GenServer)

Flush delivery is a MESSAGE (`{:render_batch, batch}` to the owner), never a call — the future live-session loop prioritizes input via selective receive. The injected `input_check` additionally holds flushes at the source so batches don't queue behind input at all. Pinned guarantees: ingest-order batches, concat == ingest sequence (no loss/dup/reorder), batch <= 32, `flush_now/1` force-drains at stream end so the tail never sits 16ms stale.

Documented divergence from the reference design: the flush message IS the paint trigger here (no separate application step), so the drain bound doubles as a throughput ceiling — 32 per 16ms = 2,000 deltas/sec, comfortably above real token rates.

## Verification

Red-first: 0/10 before the modules existed -> 40/40 green. Zero sleep-based synchronization: policy tests are table-driven over virtual timestamps (exact defer arithmetic asserted); GenServer tests use zero-interval injection + Agent-backed input_check + receive assertions <= 50ms. Full harness dir 333 passed; full suite 5837/5840 (3 pre-existing real-PTY flakes under parallel load, pass 12/12 in isolation). Format + warnings clean.

### DIFF
--- a/lib/raxol/harness/cadence_policy.ex
+++ b/lib/raxol/harness/cadence_policy.ex
@@ -12,10 +12,13 @@ defmodule Raxol.Harness.CadencePolicy do
   which is what makes this testable with plain integers instead of
   `Process.sleep/1`.
 
-  1. **Input always wins.** `decide/4,5` checks `input_pending?` first,
-     before even asking whether the cadence window is open. A live
-     agent session must never let a token flood delay keystroke
-     handling -- not even by one scheduling decision.
+  1. **Input wins, within a budget.** `decide/4,5` checks
+     `input_pending?` first, before even asking whether the cadence
+     window is open -- a live agent session must never let a token
+     flood delay keystroke handling. But the win is bounded: after
+     `max_consecutive_yields/0` consecutive yields (~one frame at the
+     1ms retry) the decision falls through to the cadence rules, so
+     continuous input can never starve rendering indefinitely either.
 
   2. **First paint is unconditional.** A stream's first delta
      (`last_flush_ms == nil`) flushes immediately. Perceived latency
@@ -43,8 +46,24 @@ defmodule Raxol.Harness.CadencePolicy do
 
   # Recheck delay after yielding to input. Input dispatch is cheap, so
   # retry nearly immediately rather than busy-looping or waiting a full
-  # frame interval.
+  # frame interval. This is a ~1kHz wakeup loop while input stays
+  # pending and deltas are buffered; it is acceptable only because
+  # @max_consecutive_yields bounds it to at most 16 consecutive wakeups
+  # (~one frame) before a flush or cadence defer takes over.
   @input_yield_retry_ms 1
+
+  # Bounds the pending backlog. 10_000 items is 5 seconds of
+  # maximum-rate egress (10_000 / 2_000 per sec at the shipped
+  # defaults); pending older than that is history, not a live tail, and
+  # shedding it bounds the heap no matter how fast the producer floods.
+  @default_max_pending 10_000
+
+  # Bounds consecutive :yield_to_input verdicts between flushes.
+  # 16 yields x the 1ms retry = one full frame interval -- continuous
+  # input may hold a token flush for at most ~one extra frame before
+  # the cadence forces forward progress. This also terminates the 1ms
+  # repoll loop (bounded wakeups per flush window).
+  @max_consecutive_yields 16
 
   @type verdict :: :flush_now | {:defer, pos_integer()} | :yield_to_input
 
@@ -60,6 +79,14 @@ defmodule Raxol.Harness.CadencePolicy do
   @spec input_yield_retry_ms() :: pos_integer()
   def input_yield_retry_ms, do: @input_yield_retry_ms
 
+  @doc "The default pending-queue watermark above which the oldest deltas are shed."
+  @spec max_pending() :: pos_integer()
+  def max_pending, do: @default_max_pending
+
+  @doc "The maximum consecutive `:yield_to_input` verdicts between flushes."
+  @spec max_consecutive_yields() :: pos_integer()
+  def max_consecutive_yields, do: @max_consecutive_yields
+
   @doc """
   Decides what a caller holding `pending_count` buffered deltas should
   do right now.
@@ -73,9 +100,16 @@ defmodule Raxol.Harness.CadencePolicy do
   Decision order (the order IS the design, not an implementation
   detail):
 
-    1. `input_pending?` true -> `:yield_to_input`, unconditionally --
-       input is always scheduled ahead of token flushes, even when the
-       cadence gate is fully open or this is the very first delta.
+    1. `input_pending?` true AND `yields_since_flush <
+       max_consecutive_yields` -> `:yield_to_input` -- input is
+       scheduled ahead of token flushes, even when the cadence gate is
+       fully open or this is the very first delta. Once the yield
+       budget is exhausted, the decision FALLS THROUGH to the cadence
+       rules below (so continuous input holds a flush for at most
+       `max_consecutive_yields x input_yield_retry_ms` ~= one frame
+       before forward progress is forced; a defer landing with the
+       budget still exhausted falls through again and flushes at the
+       window edge).
     2. `last_flush_ms` is `nil` -> `:flush_now` -- the first delta of a
        stream paints immediately.
     3. `now_ms - last_flush_ms >= interval` -> `:flush_now`.
@@ -89,6 +123,14 @@ defmodule Raxol.Harness.CadencePolicy do
       interval. This is both the deterministic-test seam (set it to
       `0` to force `:flush_now` on every decision) and the per-instance
       config seam.
+    * `:yields_since_flush` (default `0`) -- how many consecutive
+      `:yield_to_input` verdicts the caller has already acted on since
+      its last flush. This is one piece of CALLER STATE deliberately
+      carried in opts to preserve the positional-arg shape of this
+      function; the policy itself stays stateless. Callers reset their
+      counter to `0` on every flushed batch.
+    * `:max_consecutive_yields` -- override the module's default yield
+      budget.
   """
   @spec decide(
           now_ms :: integer(),
@@ -99,8 +141,13 @@ defmodule Raxol.Harness.CadencePolicy do
         ) :: verdict()
   def decide(now_ms, last_flush_ms, pending_count, input_pending?, opts \\ [])
       when is_integer(now_ms) and pending_count > 0 do
+    yields_since_flush = Keyword.get(opts, :yields_since_flush, 0)
+
+    max_yields =
+      Keyword.get(opts, :max_consecutive_yields, @max_consecutive_yields)
+
     cond do
-      input_pending? ->
+      input_pending? and yields_since_flush < max_yields ->
         :yield_to_input
 
       is_nil(last_flush_ms) ->
@@ -132,5 +179,25 @@ defmodule Raxol.Harness.CadencePolicy do
   def drain_count(pending_count, opts \\ []) do
     max_drain = Keyword.get(opts, :max_drain_per_flush, @max_drain_per_flush)
     min(pending_count, max_drain)
+  end
+
+  @doc """
+  The number of pending items a caller must shed (from the OLDEST end)
+  to get back under the `:max_pending` watermark: `max(pending_count -
+  max_pending, 0)`.
+
+  Zero at or below the watermark. Callers drop from the queue front --
+  the newest deltas are the live tail's value; the oldest are already
+  history.
+
+  ## Options
+
+    * `:max_pending` -- override the module's default watermark.
+  """
+  @spec drop_count(pending_count :: non_neg_integer(), opts :: keyword()) ::
+          non_neg_integer()
+  def drop_count(pending_count, opts \\ []) do
+    max_pending = Keyword.get(opts, :max_pending, @default_max_pending)
+    max(pending_count - max_pending, 0)
   end
 end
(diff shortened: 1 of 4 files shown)


DROOdotFOO (commented):

Reviewed `lib/raxol/harness/cadence_policy.ex` (pure policy) and `lib/raxol/harness/stream_cadence.ex` (GenServer). The pure policy is clean, well-tested, and deterministic. The GenServer is where the risk lives: for a layer sold as "backpressure," it applies none toward the producer and caps nothing.

### HIGH

- **[Saboteur + Security Auditor] HIGH** `stream_cadence.ex:84` (`ingest/2` cast) + `handle_manager_cast({:ingest, delta}, ...)` + moduledoc section 4 -- **Unbounded pending queue / mailbox, no backpressure despite the name.** `ingest/2` is a fire-and-forget `GenServer.cast` that by design "never blocks the producer," and the internal `:queue` (`pending` / `pending_count`) has no cap anywhere in the diff. Egress is ceiling-bounded at ~2,000 deltas/sec (32 per 16ms); ingress is unbounded. The moduledoc's own section 4 poses "what happens when the consumer stalls" and section 2 admits "this server has no timeout protection against a slow sink" -- but neither path bounds `pending`. If ingest rate exceeds the egress ceiling, or the owner/sink stalls (slow paint, blocked owner mailbox), the queue and the process heap grow without bound -> OOM. This is the security DoS surface too: a hostile or compromised token source (the LLM endpoint the harness streams from) floods `ingest` casts and exhausts memory. "Input-first backpressure" is a misnomer -- there is no backpressure. Fix: add a `:max_pending` bound with an explicit overflow policy (block via `call`-based ingest above the watermark, or drop-oldest with a telemetry event + a documented loss guarantee that then contradicts the "no delta ever dropped" claim in section 3 -- pick one and state it). At minimum emit telemetry and cap heap.

### MEDIUM

- **[Saboteur] MEDIUM** `stream_cadence.ex` `decide_and_act/1` `:yield_to_input -> schedule_flush(state, state.input_yield_retry_ms)` (1ms) -- **Render starvation + 1ms busy-repoll under persistent input-pending.** When `input_check.()` stays true (continuous typing, or a buggy/hostile owner whose input gate never clears), every decision returns `:yield_to_input` and reschedules a 1ms `:flush_due`. There is no maximum-yield bound and no "flush at least once per N ms regardless of input." Consequences: (a) visible output freezes indefinitely while `pending` grows (compounds the HIGH above), and (b) a ~1000 wakeups/sec poll loop burns CPU for as long as input is pending. The policy guarantees "input waits at most one bounded batch"; the inverse -- render waiting on input -- is unbounded. Fix: bound consecutive yields (force a flush after K yields or T ms) so rendering makes forward progress and the poll loop terminates.

- **[New Hire] MEDIUM** `stream_cadence.ex` moduledoc section 5 ("Not wired yet") + `init_manager` default `input_check = fn -> false end` -- **The headline "input-first" guarantee is aspirational and unenforced.** Section 2 states input priority is "the owner's responsibility" via selective receive, and section 5 admits the owning live-session loop "does not exist yet." The only in-module mechanism (`:input_check`) defaults to always-false, so out of the box the "belt-and-suspenders" source-side hold is OFF and the module provides zero input prioritization. A maintainer wiring this up has no compile-time or runtime guardrail that the contract is honored -- the PR title's central promise ("input always ahead of paint") rests entirely on a consumer that isn't in this diff. Fix: either land a minimal owner/reference consumer that exercises the contract, or make the doc state plainly that this module provides no input priority on its own and the `input_check` seam is mandatory (not optional) for the advertised behavior.

### LOW

- **[New Hire] LOW** `stream_cadence.ex` `flush_one_batch/2` vs `full_drain_loop/2` -- Near-identical drain bodies (`drain_count` -> `:queue.split` -> `to_list` -> `sink` -> update `pending`/`pending_count`/`last_flush_ms`) duplicated across the cadence path and the forced-drain path. Two copies of the money logic drift apart under maintenance (e.g. a future bound/telemetry added to one and not the other). Fix: extract a single `drain_one_batch/2`.

- **[New Hire] LOW** `cadence_policy.ex:@input_yield_retry_ms 1` / moduledoc -- The 2,000 deltas/sec ceiling is derived in prose, but the 1ms recheck constant ships without noting its cost (a ~1kHz timer loop while yielding, per the MEDIUM above). Document the CPU implication next to the constant.

### Cross-persona overlaps

- The unbounded-queue / no-backpressure finding was reached independently by the Saboteur (production OOM under stall or ingress > egress) and the Security Auditor (DoS via token-source flood, no resource limit). Same root cause -> promoted MEDIUM to **HIGH**.
- The Saboteur's render-starvation finding and the Security Auditor's input-flood angle share the `:yield_to_input` mechanism; they compound the HIGH (starvation keeps the unbounded queue growing) but are reported once as MEDIUM.

### Verdict: CONCERNS

The pure `CadencePolicy` is solid and the ordering/no-loss tests are genuine. But a module whose stated purpose is "backpressure" ships with an unbounded internal queue, an unbounded ingest cast, no consumer-stall protection, and an unbounded render-yield loop -- the one property it is named for is the one it does not provide. Not a merge-blocker for standalone/unwired code, but the queue bound and the yield bound must land before this is wired to a live session.

merklebonsai:
Review response landed as `53bf559b2` (+404/-54). Every finding confirmed against the code before fixing; 15 red tests captured before implementation. Point-by-point:

**HIGH — unbounded queue, no backpressure despite the framing, DoS surface.** *Accepted, fixed with a stated design decision.* `:max_pending` watermark (default 10,000 = 5 seconds of max-rate egress backlog) with **drop-oldest** shedding: a supervision instrument must stay alive and current — the newest deltas are the live tail's value, and call-based blocking ingest would stall the SSE reader into transport timeouts. Loss is loud twice: `[:raxol, :harness, :stream_cadence, :overflow]` telemetry at drop time, plus an in-band `{:cadence_dropped, n}` marker at the exact position of loss in the next batch (reserved element type, not counted against the drain bound). Section 3 rewritten exactly as demanded: "lossless below the watermark; explicit, in-band loss above it." Contract addition: a sink that treats every batch element as a delta fails loudly at the first loss — deliberate; silently rendering a gapless stream over shed data would be worse than a crash. Invariant verified: `dropped_since_flush > 0` implies `pending_count > 0`, so a marker can never strand.

**MEDIUM — render starvation + 1kHz busy-poll under persistent input.** *Accepted, fixed.* `max_consecutive_yields` (default 16 x 1ms = one frame interval); policy law renamed "Input wins, within a budget"; exhausted budget falls through to the cadence rules, so continuous input holds a flush at most ~one extra frame and the repoll loop always terminates. Red-first: forced-progress under permanently-true input_check, config seam, exact fall-through arithmetic.

**MEDIUM — "input-first" aspirational, default input_check always false.** *Accepted, fixed.* Section 2 now opens: "Out of the box this module enforces NO input priority" — the seam is mandatory for the advertised behavior, as is owner-side selective receive. Section 1 states this layer never applies backpressure toward the producer: cadence + load-shedding, excess dropped not pushed back.

**LOW — duplicated drain bodies / undocumented 1ms cost.** *Both fixed*: single `drain_one_batch/2`; the retry constant documents the ~1kHz wakeup implication and the 16-wakeup bound.

**"Must land before wiring to a live session":** done in this PR — both bounds landed before any live-session wiring exists.

Gates: targeted 58/58, test/harness 351, full suite 5856/5858 (both failures reproduce byte-identically at clean HEAD with changes stashed), warnings + format clean.

DROOdotFOO (commented):

Since last review: cb59fb82..53bf559b

Single new commit `53bf559b` ("Bound the stream cadence layer: overflow shedding, yield budget, honest contract"). Re-read `cadence_policy.ex`, `stream_cadence.ex`, and both test files at the new head.

- **HIGH (unbounded pending queue -> OOM/DoS) -- RESOLVED for retained state.** `shed_overflow/1` now runs on every ingest (`stream_cadence.ex:200`) and drops from the OLDEST end via `CadencePolicy.drop_count/2` (`cadence_policy.ex:200-201`, `max(pending_count - max_pending, 0)`). After each cast the in-state `:queue` is pinned at exactly `:max_pending` (default 10_000). Loss is observable twice: a `[:raxol, :harness, :stream_cadence, :overflow]` telemetry event at drop time and an in-band `{:cadence_dropped, n}` marker on the next batch. Off-by-one check: at `pending == max_pending` `drop_count` returns 0, at `max_pending + 1` returns 1 -> watermark inclusive, no fencepost error. Tested (`stream_cadence_test.exs` overflow describe block). The moduledoc no longer claims backpressure it does not provide -- it explicitly says "load-shedding, not backpressure" (`stream_cadence.ex:18,129`). The dishonest "no timeout protection" framing is replaced by a stated shed-on-overflow contract (section 3).
- **MEDIUM (1ms yield repoll starves renders indefinitely) -- RESOLVED.** `@max_consecutive_yields 16` (`cadence_policy.ex:66`) caps consecutive `:yield_to_input` verdicts; once `yields_since_flush >= max_yields` the `cond` falls through to the cadence rules (`decide/5`, `cadence_policy.ex:150-152`). Confirmed forward progress: `drain_one_batch` resets `yields_since_flush` to 0 per flush, so permanent input yields ~16x1ms then forces one bounded flush, repeating -- throttled, never starved. Tested ("forced progress under permanent input").

### MEDIUM

- **Security Auditor -- the DoS is narrowed, not closed: the bound is on retained queue state, the mailbox ingress is still unbounded.** `ingest/2` remains a fire-and-forget `GenServer.cast` (`stream_cadence.ex:132-134`), and shedding happens only *after* a `{:ingest, delta}` cast is dequeued and processed (`handle_manager_cast`, `stream_cadence.ex:195-200`). A producer that casts faster than the server drains its mailbox grows the process mailbox without bound -> heap OOM, exactly the failure `:max_pending` is advertised to prevent. The moduledoc asserts "The pending queue is therefore always bounded" (`stream_cadence.ex:89`) -- true, but the *process* is not: the unbounded buffer moved from the `:queue` to the mailbox. For the intended SSE producer this is network-rate-limited and low-risk, but the "always bounded" claim overstates what is actually guaranteed.
- **Security Auditor -- `:max_pending` bounds item COUNT, not bytes.** `delta` is `term()` (`stream_cadence.ex:132`) and the watermark counts elements, not payload size. 10_000 retained deltas x an adversarially large per-delta binary (a hostile/buggy endpoint streaming a multi-MB "token") is unbounded heap under a count-based cap. Same root as above (the bound guards the wrong quantity); two findings, same root -> the residual DoS surface is real even though narrowed. Not exploitable via the current real producer, so MEDIUM not HIGH.

### LOW

- **Saboteur -- drop-oldest is correct for a snapshot sink but corrupts an append-only token stream.** The shed justification is "the queue head is history, the queue tail is the live view" (`cadence_policy.ex:185`, `stream_cadence.ex:263`). That holds for a footer/state repaint where only the latest snapshot matters. But section 1 describes the consumer as "apply deltas to the live tail" -- LLM text deltas are *additive*, not snapshots. Dropping the oldest un-rendered deltas leaves a permanent gap in the middle of the rendered transcript; there is no "latest full state" to catch up to. The `{:cadence_dropped, n}` marker makes the loss observable (good, and the crash-on-naive-join is deliberate), but the transcript is still torn, not merely "trimmed to the live tail". The moduledoc conflates the snapshot model (where drop-oldest is lossless-of-meaning) with the append model (where it is content-corrupting-with-a-flag). Worth stating the semantic explicitly so the future owner picks the sink type knowingly.
- **New Hire -- the input-first guarantee is now honestly documented but still unwired end-to-end.** Section 2 concedes "Out of the box this module enforces NO input priority" (`stream_cadence.ex:34-37`, default `input_check` returns `false`) and section 5 concedes the owning live-session loop "does not exist yet". The prior "aspirational guarantee" finding is thus addressed by *honesty*, not by *enforcement*: the yield mechanism exists and is tested, but no real caller exercises the source-side hold or the mandatory selective-receive half. Credit for the candor; the guarantee remains unproven against a real owner. Magic numbers (16/32/1/10_000/16) are all now documented with rationale (`cadence_policy.ex:40-66`) -- that half of "honest contract" is genuinely delivered.
- **New Hire -- batch-size contract has a sharp edge.** Section 3 first states "Each batch carries at most `max_drain_per_flush` deltas" then later "at most `max_drain_per_flush + 1` elements" when a marker rides along. A consumer that reads the first sentence and asserts `length(batch) <= 32` breaks on the first loss (the `marker prepends without displacing a delta` test proves a 33-element batch). Documented, but the two statements are 20 lines apart; lead with the +1.

### Verdict: CONCERNS

The promoted HIGH (unbounded pending queue) is genuinely resolved for retained state, both prior MEDIUMs are addressed (yield starvation fully; input-first honestly documented), and the contract doc no longer claims backpressure it does not provide. Remaining items are residual and non-blocking for a not-yet-wired standalone module: the ingress DoS is narrowed but not eliminated (unbounded mailbox + count-not-bytes cap), and drop-oldest on an append-only token stream is a semantic the moduledoc should name outright. Fix the "always bounded" overstatement and the drop-oldest/append semantics before the live-session owner wires this in.

PR #607, opened 2026-07-16
Harness UI: wire message bodies through the streaming-safe markdown path

merklebonsai:
The streaming-safe markdown machinery (provisional-close for live blocks, sanitization, never-raise) landed earlier as #563 — but the BodyProvider-mounted message path ignored all of it: live message blocks did a raw full parse of partial streams (unclosed constructs flashing raw markers) with no sanitization. This wires the mounted path end-to-end and holds scope to the genuine gaps.

## What changed

- `BlockBody` -> `BodyProvider.mount(seal: block.seal)` -> `MessageBlock` -> `MarkdownBody`: live blocks stream through provisional close, so unclosed fences/emphasis never flash raw markers mid-stream; content is control-char/ANSI-sanitized (no ESC bytes reach the element tree); garbage content falls back to plain text, never raises. Defaults are `:sealed` everywhere — existing callers see zero change.
- Fence info-string language now renders as a dim label line above the code body (first word only). Syntax highlighting stays a documented seam.
- Both parser paths (builtin + opportunistic Earmark AST in dev) now share one fence-assembly helper so they cannot drift.
- `:reasoning` bodies stay deliberately plain through this path (documented).

## Latent bug found and fixed

The rework surfaced a builtin-path bug: multi-line fenced code rendered bottom-to-top (reversed-accumulator misuse). Fixed, pinned with a source-order regression test.

## Parser decision

No new dependency: `earmark_parser` exists only transitively via ex_doc (`only: :dev`), so the hand-rolled builtin parser is the de-facto test/prod path already.

## Verification

Red-first: 12 failing tests before implementation (mode/sanitization, seal-threading, end-to-end leak, language labels) + a golden-fixture replay through the full BlockBody path at every delta prefix and at seal. Harness component suite 526 passed, test/harness/ 292, playground (the other MarkdownRenderer consumer) 324 — zero regressions. Warnings-as-errors + format clean.

### DIFF
--- a/lib/raxol/ui/components/harness/block.ex
+++ b/lib/raxol/ui/components/harness/block.ex
@@ -607,8 +607,11 @@ defmodule Raxol.UI.Components.Harness.Block do
   defp markdown_source(%__MODULE__{content: %{text: text}}),
     do: to_display_text(text)
 
-  defp markdown_mode(%__MODULE__{seal: :sealed}), do: :sealed
-  defp markdown_mode(%__MODULE__{seal: :live}), do: :streaming
+  # The seal->mode mapping lives in exactly one place --
+  # `MarkdownBody.mode_for_seal/1` -- so this call site and
+  # `BodyProvider`'s `:message` props can never drift apart.
+  defp markdown_mode(%__MODULE__{seal: seal}),
+    do: MarkdownBody.mode_for_seal(seal)
 
   # Recursively applies the resolved prominence colour to every text node
   # in a rendered view tree (the MarkdownBody result), so the Markdown
--- a/lib/raxol/ui/components/harness/block_body.ex
+++ b/lib/raxol/ui/components/harness/block_body.ex
@@ -77,7 +77,11 @@ defmodule Raxol.UI.Components.Harness.BlockBody do
   defp mount_body(block, context) do
     BodyProvider.mount(block.kind, block.content,
       context: context,
-      outcome: block.outcome
+      outcome: block.outcome,
+      # threads the block's own seal so a live :message body streams with
+      # provisional close instead of a plain full parse -- see
+      # `BodyProvider.mount/3`'s `:seal` option doc.
+      seal: block.seal
     )
   rescue
     e ->
--- a/lib/raxol/ui/components/harness/body_provider.ex
+++ b/lib/raxol/ui/components/harness/body_provider.ex
@@ -73,6 +73,7 @@ defmodule Raxol.UI.Components.Harness.BodyProvider do
     ApprovalPrompt,
     Block,
     DiffViewer,
+    MarkdownBody,
     MessageBlock,
     ReasoningBlock,
     ToolCallBlock,
@@ -276,7 +277,7 @@ defmodule Raxol.UI.Components.Harness.BodyProvider do
       role: Map.get(body, :role, :assistant),
       content: Map.fetch!(body, :text),
       width: width_from(context),
-      mode: if(seal == :live, do: :streaming, else: :sealed)
+      mode: MarkdownBody.mode_for_seal(seal)
     ]
   end
 
--- a/lib/raxol/ui/components/harness/body_provider.ex
+++ b/lib/raxol/ui/components/harness/body_provider.ex
@@ -59,6 +59,14 @@ defmodule Raxol.UI.Components.Harness.BodyProvider do
   rendering (one-line summary + outcome row) is `Block.render/2`'s own
   proven code path -- `BlockBody` reuses it directly rather than
   reimplementing a second summary renderer here.
+
+  ## `:reasoning` stays plain through this path, by design
+
+  Unlike `:message`, a `:reasoning` body is never threaded through
+  `Harness.MarkdownBody` here -- `ReasoningBlock` renders dim plain text
+  on purpose (reasoning is meant to read as quieter, secondary content).
+  Styled Markdown reasoning is an explicit opt-in that lives entirely in
+  `Block.render/2`'s own `context[:markdown]` path, not in this seam.
   """
 
   alias Raxol.UI.Components.Harness.{
@@ -160,6 +168,14 @@ defmodule Raxol.UI.Components.Harness.BodyProvider do
     * `:context` -- render context passed to every mounted component's
       `render/2` (default `%{}`); `:width` inside it sizes `:message`,
       `:reasoning`, and `:diff` bodies.
+    * `:seal` -- `:live | :sealed` (default `:sealed`), the block's own
+      seal state. Threaded only into the `:message` body, as
+      `MessageBlock`'s render `:mode` (`:live` -> `:streaming`, `:sealed`
+      -> `:sealed`) -- a live message renders with the provisional-close
+      streaming treatment, a sealed one as a plain full parse. Defaulting
+      to `:sealed` means existing direct callers of `mount/3` see zero
+      behavior change; `BlockBody` is the one caller that passes the
+      block's real `seal`. Every other kind accepts and ignores it.
     * `:outcome` -- a `Block.outcome()` map (default `%{}`), consulted
       only for `:tool_call` to derive the mounted status glyph
       (`:pending` with no result yet, `:failed` on a non-zero
@@ -191,12 +207,13 @@ defmodule Raxol.UI.Components.Harness.BodyProvider do
   def mount(kind, body, opts \\ []) do
     context = Keyword.get(opts, :context, %{})
     outcome = Keyword.get(opts, :outcome, %{})
+    seal = Keyword.get(opts, :seal, :sealed)
 
     with {:ok, expected_component} <- component_for(kind),
          component = Keyword.get(opts, :component, expected_component),
          :ok <- ensure_component(kind, component, expected_component),
          :ok <- validate(kind, body) do
-      {:ok, build_view(kind, component, body, outcome, context)}
+      {:ok, build_view(kind, component, body, outcome, context, seal)}
     end
   end
 
@@ -210,23 +227,23 @@ defmodule Raxol.UI.Components.Harness.BodyProvider do
 
   # -- per-kind view construction ---------------------------------------
 
-  defp build_view(:message, component, body, _outcome, context) do
-    mount_one(component, message_props(body, context), context)
+  defp build_view(:message, component, body, _outcome, context, seal) do
+    mount_one(component, message_props(body, context, seal), context)
   end
 
-  defp build_view(:reasoning, component, body, _outcome, context) do
+  defp build_view(:reasoning, component, body, _outcome, context, _seal) do
     mount_one(component, reasoning_props(body, context), context)
   end
 
-  defp build_view(:diff, component, body, _outcome, context) do
+  defp build_view(:diff, component, body, _outcome, context, _seal) do
     mount_one(component, diff_props(body, context), context)
   end
 
-  defp build_view(:approval, component, body, _outcome, context) do
+  defp build_view(:approval, component, body, _outcome, context, _seal) do
     mount_one(component, approval_props(body), context)
   end
 
-  defp build_view(:tool_call, component, body, outcome, context) do
+  defp build_view(:tool_call, component, body, outcome, context, _seal) do
     call_view = mount_one(component, tool_call_props(body, outcome), context)
 
     case Map.get(body, :result) do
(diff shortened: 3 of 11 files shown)


DROOdotFOO (commented):

Scope: the diff wires `BlockBody -> BodyProvider.mount(seal:) -> MessageBlock -> MarkdownBody`, and
adds a fenced-code language label to the shared `MarkdownRenderer` (plus a reversed-accumulator fix).
The streaming-safe machinery in `MarkdownBody` itself (provisional close, control/ESC sanitization,
never-raise, UTF-8 recovery) is pre-existing (#563) and is correctly reused here.

### HIGH
- **[Security + Saboteur] HIGH** `lib/raxol/ui/components/markdown_renderer.ex:299-306` (builtin), `:238` (AST `pre`), helper `:276-283` -- the fence info string, which this PR previously *discarded* (`"```" <> _`), is now rendered as a `text()` label via `lang_word/1`. `MarkdownRenderer` does not sanitize its input: it relies on callers. The message path is safe because `MarkdownBody.to_text/1` strips C0/C1/DEL/ESC *before* the parse -- but `MarkdownRenderer` is a shared component with direct, un-sanitized callers today: `lib/raxol/playground/demo_helpers.ex:76-83` calls `MarkdownRenderer.init/render` with raw content. `lang_word/1` splits on `~r/\s+/`, and ESC (`0x1B`) is not whitespace, so a fence like ` ```elixir<ESC>[2J ` yields the label `elixir<ESC>[2J`, smuggling a raw escape sequence into `text()` for any non-`MarkdownBody` caller -- exactly the "never embed raw ANSI in text()" boundary this codebase enforces. Saboteur angle on the same root: `lang_word/1` takes the entire first non-whitespace run with no length bound, and `MarkdownRenderer` has no byte cap of its own (only `MarkdownBody` caps at 256KB), so a fence with a megabyte-long no-whitespace info string renders one unbounded dim label line for direct callers. Two personas, one root -> promoted. Fix: sanitize control/ESC bytes and clamp length inside `lang_word/1` (or `code_label_elements/1`) so the new label surface is safe regardless of caller, since `MarkdownRenderer`'s contract is "callers may pass untrusted text."

### MEDIUM
- **[New Hire] MEDIUM** `lib/raxol/ui/components/harness/body_provider.ex:108`, `message_block.ex:167`, `block.ex:610-611` -- the same binary distinction carries two names: block-level `seal` is `:live | :sealed`, but `MessageBlock`'s render `mode` is `:streaming | :sealed`, glued by `if(seal == :live, do: :streaming, else: :sealed)`. `:sealed` is shared across both vocabularies while the "not sealed" state is `:live` in one place and `:streaming` in another, and the mapping is duplicated (BodyProvider and Block both re-derive it). A maintainer must hold "`:live` == `:streaming`" in their head. It is documented, but the split is a standing trip-hazard. Suggest a single named helper (e.g. `mode_for_seal/1`) reused by both call sites, or aligning the vocabulary.

### LOW
- **[Saboteur] LOW** `lib/raxol/ui/components/harness/markdown_body.ex:121-129` -- the stable-prefix optimization is deferred, so every render frame re-parses the *entire* accumulated body for every live block (the 256KB cap is the only ceiling). `MessageBlock` already re-parsed per frame, so this is not new cost, but routing live blocks additionally through `provisional_close/1` per frame widens it. With many concurrent live blocks in a transcript this is O(blocks x buffer) per frame. Fine for demos; worth the follow-up (or a concurrent-live-block ceiling) before sustained high-throughput streams.
- **[New Hire] LOW** `lib/raxol/ui/components/markdown_renderer.ex:259-268` -- `code_language_from_attrs/1` is `@doc false` yet `def` (public) solely so the test can call it; and `code_label_elements("")` at `:352` is dead, since `lang_word/1`'s `String.split(..., trim: true)` can never return `""` (it returns `nil` for absent tags). Harmless but adds noise to a "streaming-safe contract" a new hire is trying to learn.
- **[Security] LOW** PR body claims "Defaults are `:sealed` everywhere -- existing callers see zero change." Not strictly true: `MessageBlock.render` moved from a raw `MarkdownRenderer.init/render` to `MarkdownBody.render(mode: :sealed)`, which now runs `to_text/1` (control/ESC strip + UTF-8 recovery). For control-char-bearing content the sealed output is no longer byte-identical to pre-PR. This is a net security improvement, but the "zero change" claim is inaccurate and could mask a real behavioral diff for a downstream snapshot test.

### Cross-persona overlaps
- The fenced-code **info-string label surface** (`markdown_renderer.ex:299-306`, `lang_word/1`) is flagged independently by Security (ESC smuggling for direct callers) and Saboteur (unbounded label length, no cap in `MarkdownRenderer`). Promoted to HIGH.

### Verdict: CONCERNS

The reversed-accumulator fix is correct (verified: `fenced_code_elements/2` returns reading order; the builtin path reverses it, the AST path uses it directly -- consistent) and well-pinned. The message streaming wire-up is sound and the primary untrusted path (`MessageBlock -> MarkdownBody`) is sanitized. Not a BLOCK because the HIGH is a defense-in-depth gap on a shared component rather than an active break on the path this PR ships. But the new label surface should sanitize/clamp at the `MarkdownRenderer` boundary before merge, since that module's own contract is "callers may pass untrusted text," and it already has one un-sanitized direct consumer.

merklebonsai:
Review response landed as `139a5b5bb`. Point-by-point:

**HIGH — fence-label ESC smuggling + unbounded label.** *Confirmed, both legs verified empirically*: `String.split(~r/\s+/)` does not treat ESC as whitespace (`lang_word("elixir\e[2J")` returned the raw escape intact), and `playground/demo_helpers.ex:76-83` is a live un-sanitized direct caller. **Fixed inside the renderer boundary**: `lang_word/1` strips C0/DEL/C1 before word extraction and clamps to a 32-cell budget via `TextLayout.truncate(..., :ellipsis)`; sanitized-to-empty yields no label; the Earmark `class`-attr path routes through the same helper. Lesson now encoded in the `MarkdownRenderer` moduledoc: a new output surface on a shared component inherits the component's own trust contract, not the calling path's. Red-first: 4 failing tests (ESC injection, control-only info string, 100KB info string, untrusted Earmark attrs) — green after.

**MEDIUM — dual `:live`/`:streaming` vocabulary, duplicated mapping.** *Confirmed.* Extracted the single bridge `MarkdownBody.mode_for_seal/1` (explicit clauses, garbage fails loud), used by both call sites; doc mandates use over re-derivation. Red-first unit test on the bridge.

**LOW — per-frame full re-parse of live bodies.** *Acknowledged, no code change*: not a new cost on this path, and the stable-prefix optimization is already the documented follow-up seam (256KB cap as ceiling). Concurrent-live-block ceiling noted for the streaming-optimization unit.

**LOW — `@doc false` public helper + dead `""` clause.** *Partially accepted*: dead clause removed (with the nil-or-nonempty guarantee commented); `code_language_from_attrs/1` kept deliberately — follows the module's existing precedent and is the only dep-free way to exercise the Earmark attr path in test (earmark_parser is dev-only); it now carries the sanitize/clamp tests.

**LOW — inaccurate "zero change" claim.** *Accepted*: overreached; sealed renders of control-char-bearing content are intentionally no longer byte-identical. Made explicit in `MessageBlock`'s "Behavioral note".

Gates: 5 red before / 0 after; component suite 531, harness+playground 616, warnings-as-errors + format clean.

DROOdotFOO (commented):

Since last review: afd87f2c..139a5b5b

- **[HIGH] Fence-language label ESC smuggling -- RESOLVED.** The label is now sanitized at the renderer's OWN boundary, so every caller (including the un-sanitized playground path `DemoHelpers.markdown/2`, `demo_helpers.ex:76-83`) is covered, not just `MarkdownBody`. `lang_word/1` (`markdown_renderer.ex:341-350`) strips controls BEFORE word extraction: `String.replace(~r/[\x00-\x1F\x7F\x{0080}-\x{009F}]/u, "")` (`:332,344`). Verified end-to-end: `"elixir\e[31m"` -> `"elixir[31m"` (ESC gone); the full C0 range, DEL 0x7F, backspace, CR, tab, and the proper-UTF-8 C1 CSI (U+009B = `<<0xC2,0x9B>>`) all strip. Both entry points are covered -- the builtin `parse_blocks` clauses (`:377,383`) and the Earmark AST attr path `code_language_from_attrs/1` (`:315`). The label is length-bounded via `TextLayout.truncate(lang, 32, :ellipsis)` (`:354`), which clips on grapheme/display-width boundaries (`split_at_display_width`), so the Saboteur "cap splits a multibyte char" and "no length bound" concerns are both closed.
- **[MEDIUM] `:live`/`:streaming`/`:sealed` vocabulary drift -- RESOLVED.** Single bridge `MarkdownBody.mode_for_seal/1` (`markdown_body.ex`), now the only mapping used by both `BodyProvider` and `Block` (`block.ex:614`).

### Medium

- **[Saboteur] The new sanitizer itself CRASHES on invalid UTF-8, on the exact untrusted caller the fix was written for.** `lang_word/1` calls `String.replace/3` with a `/u` regex (`markdown_renderer.ex:344`). A `/u` (PCRE-unicode) match over a binary containing an invalid byte raises `ArgumentError`. Reproduced: a fence info string `<<0x9B, "31m">>` (a lone 8-bit C1 CSI byte, invalid UTF-8) raises `ArgumentError` from `lang_word`. There is NO rescue in `MarkdownRenderer.render/2` (`:81-95`) nor in `DemoHelpers.markdown/2` (`demo_helpers.ex:76-83`), so on the playground path this is an unhandled crash. This is a NEW crash path introduced by this commit: before the PR the fence info string was discarded (`"```" <> _`) and never processed, so malformed bytes there were inert. The module's own contract is "callers may pass untrusted text" -- untrusted text can contain invalid UTF-8. The harness path is safe only because `MarkdownBody.render_to_text` runs `ensure_valid_utf8/1` (`markdown_body.ex:196`) upstream; the direct caller the trust note cites has no such recovery. Fix options: run `lang_word`'s replace on a validity-checked/scrubbed binary (reuse `ensure_valid_utf8`), or use a byte-class regex without `/u`, or wrap the label derivation in a rescue.

- **[Security Auditor] Same boundary, sibling surface left raw: the code BODY is not sanitized.** The fix sanitizes the label line but the code-body lines produced one function over -- `fenced_code_elements/2` maps each line to `Components.text(content: "  " <> line, style: @code_style)` (`markdown_renderer.ex:647-652`) with NO control/ESC stripping. So an ESC byte inside a fenced code body still reaches `text()` on the identical un-sanitized `DemoHelpers.markdown/2` path the label fix targets. This is pre-existing behavior, but the PR's new moduledoc trust note ("a new output surface ... inherits the component's OWN trust contract ... so the label is control/ESC-sanitized here, at the boundary that produces it, regardless of which caller supplied the input") now reads as applying to the whole fenced block while in fact it protects only the label. Same root as the Saboteur finding -- boundary sanitization is applied narrowly (label only, valid-UTF-8 only) rather than to the surface the trust note describes. Two findings, one root -> promoted to Medium.

### Low

- **[New Hire] Two independently-maintained control-char regexes with divergent ranges and no shared source of truth.** `markdown_renderer.ex:332` uses `~r/[\x00-\x1F\x7F\x{0080}-\x{009F}]/u` (strips ALL C0 incl. tab/LF/CR); `markdown_body.ex:264` uses `~r/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F\x{0080}-\x{009F}]/u` (deliberately PRESERVES tab 0x09 / CR 0x0D for body prose). The divergence is defensible per-surface, but the boundary contract lives in moduledoc prose while the enforcement is duplicated in two modules. There is no shared `text()`/ViewText sanitizer helper the two derive from, so a future "newly-dangerous codepoint" fix must be remembered in both places. Consider extracting one canonical `sanitize_controls/1` (with a `preserve_whitespace?` flag) so the two surfaces can't drift.

### Verdict: CONCERNS

The specific HIGH (ESC in fence label reaching `text()` for the un-sanitized caller) is genuinely resolved at the correct boundary, with a bounded, grapheme-safe cap and red-first tests -- clean content renders identically. But the fix's own sanitizer introduces an unhandled `ArgumentError` on invalid-UTF-8 info strings on that same un-sanitized caller, and the sibling code-body surface at the same boundary is still raw, so the "sanitized at the boundary regardless of caller" claim is not yet fully true. Neither is a leak of a live escape sequence for well-formed input, so this is CONCERNS, not BLOCK.

PR #608, opened 2026-07-16
Harness UI: byte-golden snapshots per degradation tier

merklebonsai:
Pins each tier of the degradation ladder with a reviewable byte-golden, so a rendering change can never silently degrade a lower tier nobody manually runs.

## Layout + bless flow

Follows the existing fixture-bless convention (`mix raxol.harness.fixtures.bless` / FATE `--gen` precedent): `mix raxol.harness.goldens.bless [--check]` with the same `:written | :current | :drift` statuses and raise-on-drift CI half. Goldens at `test/fixtures/harness/goldens/<fixture>.<mode>.golden` — {simple-chat, multi-tool-turn} x {inline_log, tmux_conservative, flat} at the acceptance suite's fixed 60x20/6 geometry. Drift output goes through a red-first diff formatter: first divergent byte offset, both sizes, +/-32-byte escaped context — never a raw binary dump.

## Determinism audit (documented in the moduledoc)

Time (already clock-free by contract — elapsed derives solely from fixture `ts`), environment (`env: %{}` + explicit `mode:` bypassing ModeSelect), capabilities (pinned `nil`, keeping the persistent_term probe cache out of the byte stream; `nil` is also the conservative tmux clamp), geometry (constants), map ordering (render-twice byte-equality per pair + the cross-machine comparison itself).

## The catch that mattered

Git text normalization would have rewritten the goldens' `\r\n` sealed-line protocol bytes to LF on fresh checkout — corrupting every inline/tmux golden for anyone cloning. Added `goldens/*.golden binary` to `.gitattributes`; staged blobs verified byte-identical to disk.

## Semantic guards (survive re-bless)

Flat goldens: zero ESC bytes + all-tokens-text via the real ANSI tokenizer. Inline/tmux: no `\e[2J`/`\e[3J`. Guard falsifiability pinned against synthetic corrupted streams. tmux goldens are byte-identical to inline today (same authority; reflow seam is detection-only) — each tier still pinned independently so a future divergence is visible.

## Verification

31 new tests (24 golden + 7 red-first diff-formatter), test/harness/ 316 passed, `--check` verified `current` across two separate VM runs. Warnings-as-errors + format clean.


DROOdotFOO (commented):

Reviewed the full diff (789 additions: `Golden`, `GoldenDiff`, the bless mix task, 6 golden blobs, 2 test files, `.gitattributes`) against the assembled `Raxol.Harness.Surface` (`new/2`, `advance/2`) it drives. The determinism argument (explicit `mode:`, `env: %{}`, `capabilities: nil`, fixed geometry, `now` never passed) holds up against the source. Findings below are the gaps.

### MEDIUM
- **[Saboteur] `lib/raxol/harness/surface/golden.ex:287`** -- `drive_to_completion/1` is unbounded recursion with no iteration cap and no timeout. Termination depends entirely on `Surface.advance/2` converging (`revealed >= len(events) and painted_count >= len(blocks)`). That holds for today's fixed fixtures, but this is a golden *generator*: a future `Surface`/`Projection` change or a malformed fixture where `painted_count` never catches `blocks` turns every `render/2` call into a silent hang -- CI and `mix ...bless` block with no diagnostic. Add a cap derived from event count (e.g. `2 * length(events) + footer_rows`) and raise loudly on overflow instead of spinning.
- **[Security + New Hire, promoted] `.gitattributes:12` + `lib/raxol/harness/surface/golden.ex:369`** -- The PR's thesis is "a *reviewable* byte-golden," but `test/fixtures/harness/goldens/*.golden binary` makes every golden show as "Binary files differ" in `git diff`/PR review, and the non-check bless path (`compare_or_write(_, _, false)`) prints the same `bless <name> -> <path>` line whether it *created* a new golden or *silently clobbered a drifted one*. Together the two mechanisms that are supposed to catch an unintended rendering regression -- reviewer eyeballs on the diff, and a signal from the bless tool -- are both defeated. A dev who accidentally blesses a regression gets no louder signal than a fresh checkout, and the reviewer can't see the byte delta. At minimum: have bless distinguish `create` vs `overwrite (was N bytes, now M)`, and/or keep a textual, `inspect`-escaped sidecar so the escape-stream delta is reviewable.

### LOW
- **[Saboteur] `lib/raxol/harness/surface/golden.ex:190-201`** -- Determinism moduledoc claims the single-VM "render twice" tripwire "catches iteration-order ... nondeterminism within a single VM run." It cannot: identical maps iterate identically within one VM, and OTP map iteration order is a deterministic function of the keys (sorted for <=32 entries, fixed-hash HAMT above) and is platform-stable -- so neither the single-run pin nor the cross-machine golden actually exercises the map-ordering class the doc credits them with. The tripwire *does* catch unseeded randomness / process-dictionary / clock leakage; just fix the map-ordering claim so a maintainer doesn't over-trust a guard that doesn't cover that case.
- **[New Hire] `lib/raxol/harness/surface/golden.ex:369-377`** -- The `_missing_or_stale ->` catch-all also swallows genuine `File.read` failures (EACCES, EISDIR) and treats them as "absent," then unconditionally `File.write!`. A real IO error is masked as a normal bless write. Match `{:error, :enoent}` (and `{:ok, ^rendered}` / `{:ok, _stale}`) explicitly and let other errors surface.
- **[Security] `test/harness/golden_snapshot_test.exs:798-812`** -- The inline/tmux semantic guards only forbid `\e[2J`/`\e[3J`; the goldens are otherwise unconstrained control-byte blobs marked non-diffable. A tampered/poisoned fixture emitting OSC 52 (clipboard write) or OSC 8 (hyperlink) would pass every guard and be invisible in review. Consider asserting the golden's sequence set against an allowlist of the CSI/SGR types this surface is actually expected to emit (via `SequenceScanner`) rather than only blocklisting full-clear.
- **[Saboteur] `lib/raxol/harness/surface/golden.ex:269`** -- `StringIO.open("")` device is never `StringIO.close/1`d, leaking a StringIO process per render for the caller's lifetime. Harmless in ExUnit (per-test process exit reaps it) and it matches the existing `t3_degradation_ladder_test.exs` precedent, but the bless task drives the whole matrix + per-drift re-renders in one long-lived process; close the device after `StringIO.contents/1`.

### Cross-persona overlaps
- `*.golden binary` (Security: defeats review surface) and the silent bless overwrite (New Hire: no create-vs-clobber signal) are the same root weakness -- the regression-catching net the PR is built to provide has two independent holes that compound. Promoted LOW -> MEDIUM.

### Verdict: CONCERNS

merklebonsai:
Review response landed as `e78e10379` (the six findings) + `a41e12b84` (a real rendering bug the new tooling caught). Point-by-point:

**M1 — unbounded drive_to_completion recursion.** *Confirmed, fixed.* Step-budgeted loop (`max_steps` = 2x event count + 32, derivation documented), raises naming the budget. Red-first via a synthetic never-done advance fun.

**M2 — `binary` gitattribute defeats PR review; bless silently clobbers.** *Confirmed, fixed — with one carve-out.* The `binary` attribute STAYS: it is load-bearing (the goldens' `\r\n` are sealed-line protocol bytes; eol normalization corrupts them on checkout, verified with `git ls-files --eol`). Both compensations landed: bless now splits `:written` into `:created`/`:overwritten` with old/new sizes plus the full diff report at bless time; and every golden gains an inspect-escaped line-oriented `.golden.txt` sidecar (round-trip tested, LF-pinned), staleness-checked by `--check`. Sidecar diffs are PR-reviewable — proven immediately, see below.

**L1 — moduledoc overcredits the double-render tripwire.** *Confirmed, fixed.* Reworded: map-order determinism is OTP's, not the tripwire's; the tripwire catches unseeded randomness / process-state / clock leakage; cross-machine compare is the environment backstop.

**L2 — catch-all masks EACCES/EISDIR as a fresh write.** *Confirmed, fixed.* Explicit matches; other errors raise with `:file.format_error/1`; red-first testable via a directory occupying a golden's path.

**L3 — only 2J/3J blocklisted; OSC 52/8 would pass.** *Confirmed, fixed with an allowlist:* zero OSC/DCS tokens; CSI finals within `H K m r`, ESC chars within `7 8` — the substrate's MEASURED vocabulary (enumerated from the real goldens first; matched exactly). Falsifiability test proves injected OSC 52 scans as `{:osc,_}`.

**L4 — StringIO leak.** *Confirmed, fixed.*

**New finding of our own (`a41e12b84`, touches `surface.ex` — one clause):** reading the first M2 sidecar revealed every sealed history line began with a literal `[K`: `seal_block` embedded `\e[K` in CONTENT, and ContentGuard's visible-honest neutralization stripped the ESC and kept the printable residue — so the EL never functioned and real terminals were painting `[K` garbage at the start of every sealed row, invisible to every substring-based assertion. Fix: drop the embedded EL (erasing is the authority's protocol business; DECSTBM scroll already blanks target rows). Red-first: a new permanent semantic guard rejecting ESC-less control-sequence residue in golden text tokens (grammar-scoped so `[assistant]` labels never false-positive, falsifiability-tested) failed against the old goldens, green after fix + re-bless — and the re-bless dogfooded M2: `bless (overwrote) ... (was 866 bytes, now 862)` with the exact `[K` removals visible in the sidecar diffs.

Gates: harness suite 329 passed (incl. the full end-to-end invariant suite over the changed seal path); `--check` current across separate VM runs; full suite 5833/5837 with the repeatable failure proven pre-existing on clean HEAD.

DROOdotFOO (commented):

Since last review: 9694d013..a41e12b8

Scope: `e78e1037` (harden bless flow) + `a41e12b8` (strip ContentGuard `[K` residue). Re-verified all four prior findings and reviewed the new residue fix through the Saboteur / New Hire / Security Auditor personas.

- **MEDIUM (binary gitattribute defeats review) -- RESOLVED.** The `.golden` files stay `binary` by design (justified: `\r\n` are sealed-line protocol bytes, eol-normalizing them corrupts the golden), but every golden now gets a reviewable, `inspect/1`-escaped `.golden.txt` sidecar pinned `text eol=lf`. Verified with `git check-attr -a`: `simple-chat.inline_log.golden` -> `binary: set / diff: unset`; `.golden.txt` -> `text: set / eol: lf`. The `*.golden` pattern provably does not also capture `*.golden.txt`. The PR-diff review surface is restored (`.gitattributes`; `golden.ex` `escape_lines/1` + `write_sidecar!/2`; sidecar staleness is `:drift` in `--check` at `golden.ex` `check_sidecar/2`).
- **MEDIUM (bless create-vs-clobber emits identical output) -- RESOLVED.** `:written` is split into `:created` and `:overwritten` (`golden.ex` `bless_pair/2`); the task prints `bless (new)` vs `bless (overwrote) ... (was N bytes, now M bytes)` followed by the `GoldenDiff` report (`raxol.harness.goldens.bless.ex` `report/1`). A silent regression-bless is no longer possible. Demonstrated live by `a41e12b8` itself (reported offset 664, 866 -> 862 bytes).
- **MEDIUM (unbounded `drive_to_completion`) -- RESOLVED.** Now step-budgeted: `max_steps/1 = 2 * event_count + 32`, and `drive_step/4` raises a named `RuntimeError` on non-convergence instead of hanging (`golden.ex`). Covered by `test/harness/golden_snapshot_test.exs` "bounded drive_to_completion" (raise path, convergence-within-budget, and `max_steps/1` arithmetic all asserted).
- **LOW (moduledoc mis-credits the double-render tripwire) -- RESOLVED.** Determinism-audit point 5 (`golden.ex`) now attributes map-iteration determinism to the runtime's own key-stable layout and credits the double-render test only with what it structurally can catch: unseeded randomness, process-dictionary / `persistent_term` leakage, and clock leakage within one VM.

### MEDIUM (new -- test-only, non-blocking; three personas converge)

**`escless_residue/1` is a two-sided heuristic that both under- and over-matches the residue class it advertises** (`test/harness/golden_snapshot_test.exs`, regex `@escless_residue ~r/\[(?:\d{1,3}(?:;\d{1,3})*[A-Za-z]|[A-HJKST])/`). The `[K` fix itself is correct -- dropping the embedded EL from `Surface.seal_block/2` is right, and Security Auditor confirms it drops nothing that mattered (the EL's ESC was already neutralized by `ContentGuard.sanitize_line/1`, `content_guard.ex:88-96,104-106`, so it never functioned as an erase on a real terminal; `\r\n` per-line boundaries are preserved, so no sealed lines merge). The problem is the *guard* written to pin it:

- **Saboteur -- false negatives on the exact siblings of `[K`.** The empty-param branch `[A-HJKST]` catches paramless `[K`/`[J`/`[H` but MISSES the paramless line/char editors most likely to be wrongly embedded next to an EL. Verified against the live regex: `[P` (DCH), `[L` (IL), `[M` (DL), `[I`, `[N`, `[O`, `[Q`, `[X`, `[@` all -> MISSED. A future seal path that embeds `\e[P`/`\e[L`/`\e[M` produces `[P`/`[L`/`[M` residue that this "catch the ContentGuard-stripped `[K` class" guard sails straight past.
- **Security Auditor / New Hire -- false positives on ordinary uppercase bracket labels.** The comment claims the branch is "deliberately uppercase-only so the fixtures' legitimate lowercase bracket labels never false-positive" -- but that protects only *lowercase*. Verified flagged as residue: `[ERROR]`, `[DONE]`, `[FAIL]`, `[KEY]`, `[SKIP]`, `[TODO]`, `[BUILD]`, `[HELP]`, `[Kernel panic]`. All are plausible LLM/agent output. The current fixtures contain only `[assistant]` (grep of the sidecars confirms it, so the guard is green today), but the first fixture line carrying an uppercase bracket label will fail the residue guard as a false positive and block a legitimate `bless`.

Root cause (shared): a fixed letter-set regex cannot distinguish "ESC-stripped CSI residue" from "bracket label" in already-tokenized, ESC-free text -- `[K` as EL-residue and `[K` as the start of `[KEY]` are byte-identical post-strip. The honest containment is upstream (`ContentGuard` neutralizing at content ingress, which is intact), so this is defense-in-depth on a tripwire, not a production hole -- hence non-blocking. Suggest either (a) narrow to residue that is unambiguous (require a params run: `[` + digits + final, dropping the empty-param branch entirely, accepting that a paramless stripped final is indistinguishable from a label and is better caught at the `ContentGuard`/seal seam), or (b) at minimum document that the empty-param branch trades false negatives on `[L]`/`[M]`/`[P]` for false positives on uppercase labels, so a future re-blesser understands why a `[ERROR]` line trips it.

### Verdict: CONCERNS

All four prior findings are resolved with evidence, and the core `[K` fix is sound. The single new concern is a fragile, two-sided test guard (false negatives on `[L]`/`[M]`/`[P]` residue, false positives on common uppercase labels) -- test-only and non-blocking, but it undercuts the very tripwire the fix commit was written red-first to establish.

merklebonsai:
Round-2 response landed as `be42e7bd2` (test-only, one file). Ledger:

**New MEDIUM — two-sided residue guard.** *Confirmed on both sides against the live regex before touching anything*: `[ERROR]`/`[DONE]`/`[KEY]`/`[Kernel panic]` all false-positive under the old empty-param branch; `[P`/`[L`/`[M`/`[X` all sail past it. Your root cause is right — post-strip, paramless residue and label text are byte-identical; no letter set separates them.

**Fix (option a):** the guard is now `\[\d{1,3}(?:;\d{1,3})*[A-Za-z]` — bracket + numeric parameter run + final, a shape prose never produces. The empty-param branch is gone. The comment documents the trade explicitly: the paramless class (`[K`, `[P`, `[L`, `[M`) is indistinguishable from labels post-strip and is owned upstream by the ContentGuard/seal seam — content handed to `seal/2` must never carry escapes (the `a41e12b84` fix, intact, is the honest containment).

**Red-first:** the falsifiability test was rewritten before the regex changed — `[ERROR] build failed` pass-clean failed red against the old regex, green after narrowing. It now pins: `[12;24H`/`[1;14r`/`[2J` residue trips; `[assistant]`/`[reasoning]`/`[ERROR]`/`[DONE]`/`[KEY]`/`[Kernel panic]` pass clean; and the ACCEPTED paramless false negative (`[K`) is asserted and documented in-test, not left implicit.

Gates: golden suites 37/37, test/harness 329, `--check` current, warnings + format clean. The PR's CI failures are the pre-existing pair (#611 ListScorer bound, #612 OSC parser totality) — a re-run after those merge should clear this PR. One more pre-existing flake identified in passing for the backlog: `PlatformGraphicsTest` mutates global `TERM` env mid-test and flaps across identical solo runs.

DROOdotFOO (commented):

Since last review: a41e12b8..be42e7bd

One new commit (`be42e7bd`, test-only, `test/harness/golden_snapshot_test.exs`) narrows the red-first ESC-less residue guard from the two-sided heuristic `~r/\[(?:\d{1,3}(?:;\d{1,3})*[A-Za-z]|[A-HJKST])/` to the parameterized-only form `~r/\[\d{1,3}(?:;\d{1,3})*[A-Za-z]/`, dropping the empty-param branch and documenting the paramless class as owned upstream by the ContentGuard/seal seam.

- **False-positive on common uppercase labels (MEDIUM, one of two sides) -- RESOLVED (reproduced).** I re-ran the exact new regex against the labels my prior finding named. All pass clean now: `[ERROR]`, `[DONE]`, `[FAIL]`, `[KEY]`, `[WARN]`, `[SKIP]`, plus `[Kernel panic]` and the pre-existing `[assistant]`/`[reasoning]`. The empty-param `[A-HJKST]` branch that would have tripped `[ERROR]`/`[KEY]` is gone. The falsifiability test grew the corresponding uppercase pass-clean cases and a `[12;24H` trip case; the three real-residue inputs (`[1;14r`, `[2J`, `[12;24H`) all still trip. The parameterized regression net works.

### Low / Nit (test-only tooling, modest ceiling)

- **The paramless false-negative I originally flagged persists -- now accepted, not fixed (reproduced).** `[K`, `[L`, `[M`, `[P` (the exact ESC-stripped erase/insert/delete-line siblings my finding named) all pass clean under the new regex. This is the honest half of the trade: the fix bought the false-positive relief by keeping the false-negative. The commit documents it and defers ownership to the ContentGuard/seal seam ("content handed to `InlineAuthority.seal/2` must never carry escapes"). Fair reasoning, but note the owner is a *source* fix (`seal_block/2` dropped its one `\e[K`), not an enforced invariant -- there is no assertion at the seam that seal content is ESC-free. The original bug this whole tripwire was born to catch was a paramless `\e[K`; if `seal_block` (or any future seal caller) re-embeds `\e[K`/`\e[L`, ContentGuard strips it to a text token with no ESC, so neither the sequence-vocabulary allowlist (keys on ESC presence) nor this narrowed guard flags it. The compensating catch is the byte-golden diff + escaped sidecar eyeball (which is in fact how the original bug was found) -- but the automated semantic net for that exact class is now blind. Belt intact, suspenders removed.

- **"Unambiguous / provably not text" is overstated -- a narrower false-positive class remains (reproduced).** Digit-leading bracket labels are byte-identical in shape to real residue after ESC-strip, so the regex cannot distinguish them either. Confirmed trips on plausible agent output: `[4K] video`, `[8K]`, `[2FA] enabled`, `[3D] printer`, `[1st] place`, `[2nd] run`, `[100x] speedup`. (`[200 OK]`, `[1080p]` pass -- the digit run must be immediately followed by a letter and capped at 3 digits.) So the class of false-positives shrank from "any uppercase-led label" to "1-3 digits then a letter," but it is not empty, and the moduledoc/commit claim that this shape is one "prose never produces" / "provably not text" is not quite true. Undocumented residual. Practically harmless -- the current chat fixtures carry lowercase labels -- but a future fixture line rendering e.g. `[4K]` would wrongly block a legitimate bless, which is the same failure mode the narrowing set out to remove, just rarer.

### Verdict: CONCERNS

The flagged MEDIUM is materially improved: the common-label false-positive that would have blocked a legitimate bless is genuinely fixed, verified against every label case I named. What remains is low-severity for test-only tooling: the paramless false-negative is now honestly documented and deferred upstream (though the upstream owner is a source fix, not an enforced ESC-free-content invariant, leaving the original `\e[K` bug class without an automated semantic net on recurrence), and a residual digit-leading-label false-positive (`[4K]`/`[2FA]`/`[8K]`) contradicts the "provably not text" framing. No blocker.

merklebonsai:
Round-3 response landed as `c1d3f790f` (test-only). Ledger:

**LOW 1 — paramless residue class has no enforced owner.** *Confirmed — your structural point was right: the upstream owner was a source fix with no assertion at the seam.* The seam contract is now enforced by two complementary tests (describe "seal-seam ingress invariant"):

1. **Fixed-point at ingress** — content handed to `InlineAuthority.seal/2` must satisfy `ContentGuard.sanitize_line(content) == content`. One deliberate correction to the review's framing: the invariant is NOT "zero ESC at ingress" — sealed styled content legitimately carries allowlisted SGR. The enforceable invariant is that sanitization is the IDENTITY: if nothing would be stripped, no ESC-less residue can ever be produced.
2. **Anti-drift replay pin** — the reconstruction could silently diverge from what `seal_block` really emits, so a second test replays the actual emitted bytes through the seal oracle and requires sealed-history rows to match the reconstructed ingress line-for-line.

**Red proof, exactly as the finding deserved:** temporarily re-embedded `"\e[K"` in `seal_block/2` — the anti-drift guard tripped with its named divergence message while the narrowed residue regex stayed blind, demonstrating the new guard covers precisely the class the regex cannot. Mutation reverted, `git diff` on surface.ex confirmed empty; no production changes ship in this commit.

**LOW 2 — "provably not text" overstated.** *Confirmed — your reproductions verified* (`[4K]`/`[2FA]`/`[8K]`/`[1st]` trip; `[200 OK]`/`[1080p]` pass). The comment now states the residual class honestly (1-3 digits immediately followed by a letter), notes current fixtures carry none, and directs that a future `[4K]`-bearing fixture requires revisiting the guard rather than silent widening. The falsifiability corpus pins BOTH sides: pass-clean cases asserted, and `[4K]`/`[2FA]` asserted as documented deliberate false positives — regex drift in either direction fails a named test.

Gates: golden suites 39/39, test/harness 331, `bless --check` current (goldens byte-unchanged — purely test-side round), warnings + format clean.

DROOdotFOO (commented):

Since last review: be42e7bd..c1d3f790

Scope: one new commit, `c1d3f790` ("Enforce the seal-seam ingress invariant;
document the residual label class"). Diff is test-only (+195/-14 in
`test/harness/golden_snapshot_test.exs`); no production change in this delta.
The `\e[K`-removal from `Surface.seal_block/2` was a PRIOR commit on this branch
(verified: `git show c1d3f790:lib/raxol/harness/surface.ex` inline path is now
`&[&1, "\r\n"]`, with an explanatory comment; the erase is the authority's job,
not content's).

**The paramless-residue gap is now genuinely ENFORCED, not merely documented.**
Round-2's CONCERN was that the paramless class (`[K]/[L]/[M]/[P]` after ESC
strip) was deferred "to the ContentGuard/seal seam" as prose, with nothing
automated to catch a re-embed. This commit turns that seam into a two-part
test-side invariant, and both parts have teeth:

- Fixed point at ingress: for every `fixture x {inline_log, tmux_conservative}`
  it reconstructs the per-block content handed to `InlineAuthority.seal/2`
  (`BlockBody.render |> ViewText.lines(:styled)`) and asserts
  `ContentGuard.sanitize_line(bin) == bin`. Verified the enforcement mechanism
  directly: `sanitize_line("\e[Khello") -> "[Khello"` (identity? = false), while
  `sanitize_line("\e[1;31mred\e[0m")` is the identity. So a non-SGR escape
  arriving via CONTENT makes the assertion trip; legit SGR passes. This is a
  real invariant, not "goldens happen to be clean today."
- Anti-drift replay pin (inline_log): reconstructed-ingress plain text vs.
  emulator-replayed sealed history, row-for-row. This is the guard that catches
  a `seal_block` re-embed (the exact regression class the residue regex is blind
  to); commit red-proved it by re-embedding `\e[K` and reverting. Confirmed a
  leading `[K` cannot hide: `plain_lines`/`row_text` trim trailing only, so
  leading residue diverges the comparison.

Net: the residue class is owned by an enforced invariant at ingress, materially
stronger than the prior "documentation + same regex." Geometry constants (60/20/6)
match `Golden.render/2`, so the reconstruction mirrors the real golden path; all
referenced helpers (`Golden.fixtures/drive_to_completion/max_steps`,
`SealOracle.replay/history`, `ViewText.lines`, `BlockBody.render`, `Fixture.load`,
`Surface.new`) exist on the branch.

**Digit-leading-label LOW: addressed honestly as documented+pinned (not fixed).**
The regex claims are exact (reproduced standalone): `[200 OK]` and `[1080p]`
pass clean; `[4K]`, `[2FA]`, `[8K]`, `[3D]`, `[1st]`, `[100x]` false-positive;
`[2J`/`[1;14r`/`[12;24H` caught; `[K]/[L]/[M]/[P]` NOT caught. The falsifiability
corpus now pins `[200 OK]`/`[1080p]` clean and `[4K]`/`[2FA]` as deliberate
false positives, and the comment states the accepted residual class plainly.
The "provably not text" overclaim from Round-2 is withdrawn. Accepting rather
than widening the tripwire is the right call for a residue guard.

### Low / Informational

**ContentGuard does not neutralize 8-bit C1 controls (reproduced).** The
fixed-point invariant equates "sanitize is the identity" with "no escape would
be stripped into residue" -- true only for 7-bit ESC-introduced sequences.
Verified: `sanitize_line` is the IDENTITY on `<<0x9B>>"2J"` (8-bit CSI),
`<<0x9D>>"8;;..."` (8-bit OSC), and `<<0x84>>` (IND), because every byte
`>= 0x80` is treated as opaque UTF-8. So a non-SGR control introduced in C1
form would (a) survive ContentGuard into a golden, (b) satisfy the fixed-point
assertion (identity), and (c) evade the `[...`-shaped residue regex -- all three
nets miss it. This is pre-existing ContentGuard behavior, out of this test-only
commit's scope, and practically low-risk (0x9B is invalid UTF-8 in a UTF-8
terminal), but the harness explicitly models degraded/legacy tiers where 8-bit
C1 can be honored, so it is worth an explicit note behind the "fixed point ==
safe" confidence.

**Enforced-invariant coverage is asymmetric across tiers.** The fixed-point test
alone is blind to a `seal_block` framing re-embed -- its reconstruction is
hardcoded `&[&1, "\r\n"]` and never reads `seal_block`, so only the anti-drift
replay pin re-anchors reconstruction to reality. That pin runs `:inline_log`
ONLY. For `:tmux_conservative`, a re-embedded escape's actually-emitted bytes
are re-anchored by the pre-existing byte-golden, not by this new invariant. Not
blocking (byte-goldens backstop every tier), but the "enforced invariant" for
the framing class is inline_log-only; a one-line loop over both modes in the
replay pin would close it.

**Minor (New Hire):** the replay pin has no non-empty guard analogous to the
fixed-point test's `assert binaries != []`; if a fixture reduced both sides to
`[]` the `history_lines == expected_lines` assertion passes vacuously. Fixtures
are non-empty in practice; low.

### Verdict: CLEAN

Test-only tooling (modest ceiling). The Round-2 CONCERN is resolved: the
paramless-residue class is now enforced by a real ingress invariant
(ContentGuard fixed-point + anti-drift replay pin, red-proven), not merely
documented, and the digit-leading LOW is honestly documented and falsifiability-
pinned. Remaining items are pre-existing (C1 gap) or modest polish (inline_log-
only replay pin, vacuity guard) and do not block a test-only commit.

PR #609, opened 2026-07-16
Harness UI: overlay picker primitive (footer-region, substrate-safe)

merklebonsai:
The shared overlay primitive every picker (palette, session picker, jump, search) will build on. Substrate law shapes the whole design: history above the footer is terminal-owned scrollback that must never be painted over — so the overlay lives INSIDE the DECSTBM footer region, growing it downward-anchored and shrinking back on dismiss. Never a centered modal; never alt-screen.

## Component (`Raxol.UI.Harness.OverlayPicker`, pure)

- `new(items, opts)` with `:label_fn`, `:max_visible` (8), `:filter_fn` (arity-3 — the documented fuzzy seam; default case-insensitive substring).
- Height FIXED at open (`1 + min(max(#items,1), max_visible)`) so filtering never re-pins the scroll region per keystroke.
- `handle_key/2` -> `{:continue, t} | {:picked, item} | :dismissed`; grapheme-safe filtering, window-follow selection.
- `render/1` emits exactly `height/1` text leaves — row accounting correct by construction; width truncation stays in the ViewText trust boundary.

## Footer growth (the substrate-critical part)

New `InlineAuthority.set_footer_rows/2` + `ScrollRegionManager.set_footer_rows/2`: growing over OCCUPIED history rows first scrolls them up via index-at-region-boundary (evicted into native scrollback — never painted over); shrinking clears vacated rows while still footer-owned; both reuse the existing `needs_keyframe` latch. Immutable-prefix property verified across grow-over-occupied-rows.

## Keymap/focus

One head-of-table bind: `escape -> :overlay_dismiss, guard: :overlay`, firing before the always-ESC-interrupt (table order documented as load-bearing for escape only). Fail-safe direction deliberately opposite `composing?`: absent flag = closed = interrupt unchanged. Chars/arrows/Enter stay passthrough, routed to the overlay while open — no parallel dispatch mechanism.

## Degenerate geometry

Open REFUSES (`{:error, :insufficient_footer_capacity}`, zero bytes) unless history keeps >= 2 rows and >= 2 overlay rows fit; taller lists clamp rather than refuse; flat mode -> `{:error, :no_footer}`; resize below capacity force-closes; `set_footer_rows` independently refuses degenerate targets so a temporary overlay can never unpin the live footer.

## Verification

Red-first: all three suites written and failing before implementation. One test-design flaw surfaced during review and fixed: blanket CUP-confinement wrongly flagged DECSTBM's legitimate row-1 homing — replaced with explicit-CUP extraction plus a fail-closed vocabulary sweep and the emulator immutable-prefix nets. Target suites 79/79; full suite 5843/5847 (4 pre-existing/environmental, zero overlap); raxol_terminal 2203/2206 (3 pre-existing, confirmed identical with the package change reverted). Warnings + format clean.

### DIFF
--- a/lib/raxol/harness/surface.ex
+++ b/lib/raxol/harness/surface.ex
@@ -113,6 +113,46 @@ defmodule Raxol.Harness.Surface do
   still calls `repaint/2`/`keyframe/2` (they never crash), just over
   whatever footer capacity `footer_range/1` reports for that geometry.
 
+  While an overlay picker is open (see "The overlay picker" section
+  below), the layout is instead status ++ overlay lines
+  (`ViewText.lines(OverlayPicker.render(picker), width, :styled)`) ++
+  Composer's lines ++ notice -- the pending/live-tail preview lines are
+  SUPPRESSED for exactly as long as the overlay is open (the space they'd
+  occupy is now claimed by the overlay), and return the moment it closes.
+
+  ## The overlay picker (footer-region overlay)
+
+  `open_overlay/3` hosts a `Raxol.UI.Harness.OverlayPicker` by GROWING the
+  DECSTBM footer viewport (`InlineAuthority.set_footer_rows/2`) -- never a
+  centered modal painted over history, never the alternate screen. The
+  overlay's rows live entirely inside the (now larger) pinned footer, the
+  same substrate the composer/status/preview lines already share.
+
+  ESC closes the overlay, not the running turn: `Keymap`'s `:overlay`
+  guard (see that module's moduledoc) captures ESC as `:overlay_dismiss`
+  BEFORE the global `:always` ESC-interrupt bind ever sees it, as long as
+  `handle_input/2`'s context carries `overlay_open?: true` -- which it
+  does whenever `model.overlay` is non-`nil`. Enter, printable characters,
+  and the arrow keys are deliberately NOT added to `Keymap.binds/0` for
+  the overlay; they stay `:passthrough`, and THIS module is what routes a
+  `:passthrough` event to `OverlayPicker.handle_key/2` instead of the
+  Composer while an overlay is open (see `handle_input/2`'s routing,
+  below) -- Enter commits the overlay's current selection instead of
+  submitting the composer's buffer.
+
+  `open_overlay/3` refuses rather than degrading silently whenever the
+  current geometry cannot safely host even a minimal overlay: history
+  must keep at least 2 rows, and the overlay itself needs at least 2 rows
+  (a query row plus one item row) -- `{:error,
+  :insufficient_footer_capacity}`, zero bytes, model untouched. A taller
+  item list than the available capacity is CLAMPED to fit (via
+  `OverlayPicker`'s own `:max_visible` option), not refused -- only a
+  geometry too small for even the 2-row minimum is a hard refusal.
+  `model.footer_rows` always stays the BASE value the caller originally
+  configured; the grown row count lives only in `model.authority` for as
+  long as the overlay is open, and `close_overlay/1` restores the
+  authority back to exactly that base value on dismiss or commit.
+
   ## Precondition #6 -- command bifurcation (fixture mode = honest UI stubs)
 
   `:interrupt`/`:steer` are the two commands that cross to the agent lane
@@ -135,6 +175,15 @@ defmodule Raxol.Harness.Surface do
 
   `:fold_toggle`/`:jump_next`/`:jump_prev` never leave this module -- they
   are pure UI-local state per Keymap's own documented bifurcation.
+
+  While an overlay picker is open (`model.overlay != nil`), `:steer` is a
+  documented no-op instead of queuing the composer's buffer: the composer
+  is frozen mid-pick (its buffer is not what the operator is currently
+  interacting with), so queuing a steer built from THAT hidden state
+  would be dishonest UI -- it would claim to queue "what you were about
+  to send" when what's actually on screen is a filter query, not a
+  prompt. `:interrupt` is unaffected (an overlay is transient UI-local
+  state, not a reason to block the honest interrupt stub).
 
   ## Fold/jump and the seal-time-only gate -- a translation, not a reuse
 
@@ -295,7 +344,7 @@ defmodule Raxol.Harness.Surface do
   alias Raxol.Harness.Surface.ViewText
 
   alias Raxol.UI.Components.Harness.{Block, BlockBody, Composer}
-  alias Raxol.UI.Harness.{InputEvent, Keymap}
+  alias Raxol.UI.Harness.{InputEvent, Keymap, OverlayPicker}
 
   alias Raxol.UI.Rendering.PaintAuthority.{
     FlatAuthority,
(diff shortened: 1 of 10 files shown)


DROOdotFOO (commented):

Reviewed the substrate-critical footer-growth path (`InlineAuthority.set_footer_rows/2`,
`ScrollRegionManager.set_footer_rows/2`), the keymap/routing seam, and the pure
`OverlayPicker` against the actual diff and surrounding source. The core substrate math
(grow-scrolls-into-scrollback via index-at-region-boundary, shrink-clears-vacated-rows,
the `grow_reclaim_count` conservative branch, the `>=2` history / `>=2` overlay thresholds,
`needs_keyframe` reuse, immutable-prefix nets, ViewText + ContentGuard double-sanitization of
untrusted labels) holds up under scrutiny -- I could NOT find a scrollback-overpaint. The
blocking issue is in input routing, not the substrate.

### CRITICAL
- **[Saboteur + Security Auditor] (promoted from HIGH)** `lib/raxol/ui/harness/keymap.ex:185-187` + `:267` -- **The overlay filter silently loses `z`/`j`/`k` and mutates hidden transcript state whenever the overlay is opened from transcript-browse mode.** The design contract (OverlayPicker moduledoc; Surface handle_input doc) is "printable characters stay `:passthrough` and get routed to `OverlayPicker.handle_key/2` while an overlay is open." But `z`/`j`/`k` are `char:` binds (`fold_toggle`/`jump_next`/`jump_prev`) guarded by `:not_composing`, and `guard_passes?(%{guard: :not_composing}, ctx)` reads ONLY `composing?` -- it never consults `overlay_open?`. `open_overlay/3` does not touch `composing?`. So the sequence `focus_transcript/1` (sets `composing?: false`, the documented mode that "enables jump/fold" and the natural mode to open a jump/search/palette picker from) then `open_overlay/3` leaves `composing? == false` AND `overlay != nil`. Pressing `z`/`j`/`k` then resolves to a *command* (guard passes), NEVER reaches `route_passthrough`, so it (a) is dropped from the picker's filter query and (b) toggles a fold / moves the transcript jump cursor behind the open overlay. `j`/`k` are extremely common in search text, so the picker is functionally broken in its primary open-path, and the side effects are invisible. **Masking:** the new keymap test `"only ESC gains an overlay meaning..."` asserts `resolve_from(Event.key("a"), %{overlay_open?: true, composing?: true})` -- it pins `composing?: true` (where `:not_composing` fails anyway) and tests `"a"` (unbound), so it never exercises `z`/`j`/`k` with `composing?: false` and gives false confidence. **Fix:** make `:not_composing` also fail when an overlay is open -- `not (Map.get(ctx, :composing?, true) or Map.get(ctx, :overlay_open?, false))` -- or have `open_overlay/3` assert a routing context that suppresses these binds; then add the `composing?: false` case to the test matrix.

### MEDIUM
- **[New Hire] `lib/raxol/harness/surface.ex` (`open_overlay/3`, `do_open_overlay/4`, `force_close_overlay?/2`)** -- The 2-row history minimum is re-encoded as a bare literal in two places (`model.rows - 2 - model.footer_rows` in the capacity check, `new_rows - 2 - footer_rows` in force-close), duplicating `ScrollRegionManager.degenerate?/2`'s own `history_bottom < 2` definition instead of calling it; and the `max_visible` default `8` in `do_open_overlay`'s `Keyword.get(opts, :max_visible, 8)` duplicates `OverlayPicker`'s `@default_max_visible 8`. If either threshold ever changes, these silently drift and the substrate-capacity guard is exactly where drift is most dangerous. Route both through named helpers / the module attribute. (Also flagged: the moduledocs are very heavy on undefended jargon -- DECSTBM / substrate law / seal frontier / index-at-region-boundary -- with no single glossary anchor for a new maintainer.)

### LOW
- **[New Hire] `lib/raxol/ui/harness/keymap.ex:182` (`@binds` order)** -- "table order is load-bearing for escape only" is defended by exactly one test (`ESC with overlay_open?: true -> :overlay_dismiss`). Nothing structural stops a future editor from inserting another `:escape` entry or reordering; the invariant "no two entries can both match one event" that made order incidental is now false for the escape pair. Consider a structural guard (assert at compile time that the `:overlay` escape precedes the `:always` escape) rather than relying on prose + a single behavioral test.
- **[Security Auditor / Saboteur] `lib/raxol/ui/rendering/paint_authority/inline_authority.ex` (`grow_reclaim_count/3`, `scroll_history_up/3`)** (PLAUSIBLE) -- The grow scroll count is derived from `next_row` as the occupancy high-water mark. `resize/3` (seal-time-only) clamps `next_row` DOWN on a shrink without moving on-screen content, so after a shrink-resize `next_row` can under-report the rows physically occupied on some terminals. An `open_overlay` immediately after such a resize could then under-scroll and let the overlay paint over a still-occupied row. This is the same latent assumption `append_sealed/2` already makes (so not new to this PR), but the overlay grow is the first path that reclaims a *range* rather than appending one line, which widens the blast radius. Worth an explicit test: shrink-resize with a full history region, then `open_overlay`, assert immutable-prefix.

### Cross-persona overlaps
- The CRITICAL routing desync was reached independently by the **Saboteur** (production break: filter drops keystrokes + corrupts fold/jump state) and the **Security Auditor** (input routing desynchronized -- keys leak to the wrong handler, a confused-deputy over `overlay_open?` vs `composing?`). Per the promotion rule, HIGH -> **CRITICAL**.
- No other overlaps; the substrate-overpaint hypotheses (grow off-by-one, shrink clearing wrong rows, degenerate-refusal bypass) were investigated and did NOT survive verification against the code.

### Verdict: BLOCK

merklebonsai:
Review response landed as `52e811059`. Point-by-point:

**CRITICAL — z/j/k resolve as fold/jump commands through an overlay opened from transcript-browse mode; masked by a test pinning the wrong context.** *Confirmed exactly as described.* Reproduced red at three levels before fixing: keymap unit (z/j/k with `%{overlay_open?: true, composing?: false}` resolved as commands), a full guard x context matrix, and an end-to-end Surface test (typed "jkz" never reached the filter; the jump cursor would have moved behind the overlay). Fixed with the review's exact guard expression — `:not_composing` now fails when `composing? or overlay_open?` (fail-safe direction preserved: absent flag = closed). The masking test was rewritten to sweep BOTH composing states, plus a dedicated z/j/k case, plus a full guard x context matrix (every bind x {true, false, absent} x {true, false, absent}) so no pair goes untested again. The Surface test also asserts fold/jump state stays untouched behind the overlay.

**MEDIUM — 2-row minimum literal duplicated.** *Confirmed.* `Surface.max_overlay_rows/2` is the single helper and derives the threshold by asking `ScrollRegionManager.degenerate?/2` directly — no re-encoded literal anywhere; both call sites route through it.

**MEDIUM — max_visible 8 duplicated.** *Confirmed.* `OverlayPicker.default_max_visible/0` exposed; Surface reads it.

**MEDIUM — jargon density.** *Accepted.* Glossary anchor added at the top of the Surface moduledoc (DECSTBM, pin, seal-once, index-at-region-boundary, keyframe vs repaint, degenerate geometry — one line each).

**LOW — escape table order fragile.** *Accepted.* Compile-time structural check below `@binds` raises unless the escape guards are exactly `[:overlay, :always]`.

**LOW — shrink-resize under-scroll ("plausible").** *Investigated fully: the substrate is correct — but the suggested test found a real bug in the test-oracle emulator.* `resize/3`'s clamp lands on the conservative full-delta branch, so no under-scroll exists in the emission. The suggested immutable-prefix test went red anyway; root cause: `CSIHandler.handle_other_csi/3` had NO `"r"` clause — every mid-stream DECSTBM through the real parser was silently dropped (`handle_r/2` existed, unreachable); only the head-of-chunk preprocess regex ever applied one, which is why every prior test passed. Fixed by routing `"r"` -> `handle_r/2`, with three red-first parser regression tests (mid-stream re-set replaces region; bare `CSI r` clears; LF scrolls at the NEW boundary). The shrink-resize-then-open scenario is pinned green.

Gates: overlay/keymap/harness suites 193/193; full suite 5850/5851 (1 known environmental); raxol_terminal 2206/2209 (3 pre-existing, verified failing without these changes); warnings + format clean.

PR #610, opened 2026-07-16
Harness UI: hand the composer draft to $EDITOR with a suspend/resume bracket

merklebonsai:
Long prompts don't belong in a 6-row footer composer. Ctrl-E hands the draft to `$VISUAL || $EDITOR || vi` in a temp file, suspends the harness's terminal claim while the editor owns the tty, then resumes the substrate and reloads the edited draft.

## The terminal handoff (each step grounded in evidence)

**Suspend:** resolve editor + write temp file (nothing to unwind) -> disable the prim_tty stdin reader via a new `ReaderGate` (BYTE-FOR-BYTE OTP's own shell open-editor mechanism — `user_drv:open_editor/2`'s disable/enable alias protocol against the reader process InlineDriver already traces; disable failure ABORTS, never hand the tty to a competing reader) -> emit suspend bytes while still raw (teardown steps 1-3 verbatim + a BARE cursor park: teardown's CRLF would scroll a stale footer row into un-repaintable history; pinned by `teardown_bytes == suspend_bytes <> "\r\n"`) -> `stty sane` last so the editor starts cooked and Ctrl-C works if it wedges.

**Spawn:** `Port.open({:spawn, cmd}, [:nouse_stdio, :exit_status])` — `:nouse_stdio` moves port pipes to fds 3/4 so the child inherits fds 0/1/2 = the REAL tty (`System.cmd` can never do this). `{:spawn, string}` via sh supports arg-carrying editors ("code -w") and gives exit-127-as-not-found.

**Resume:** re-query size while still cooked (terminal may have been resized under the editor) -> raw before reader re-enable (no echo race) -> reader enable -> init bytes -> `resize |> reassert` at the Surface. `ScrollRegionManager.resize/2` is geometry-gated (zero bytes when unchanged), so resume adds an unconditional `reassert/1`; `InlineAuthority.reassert/1` latches `needs_keyframe` so the next repaint self-promotes to a full keyframe — the editor scribbled the screen; a logical diff would lie. Exit 0 reloads the draft (exactly one trailing newline stripped — POSIX editor discipline); nonzero/127/crash/reload-failure keep the original draft + a one-frame notice. Temp file removed on every path. Sealed history untouched by construction.

## Pure/impure split

`Raxol.Harness.EditorSuspend` (pure): a 13-step state machine whose per-failure-point compensation is DERIVED from a resource ledger `{tty, reader, screen, modes, tmp}` in safe resume order — tested exhaustively over every failure point. `Raxol.Harness.EditorSession`: thin interpreter, every seam injectable; exceptions run compensation then propagate.

## Keybind

Ctrl-E -> `:edit_draft` (`:always` class) — the keymap's first char-kind CHORD, using the documented declared-mods exact-match extension; plain `e` still inserts.

## Verification

Red-first machine tests (0/3 before, 20/20 after). 51 new tests total: 20 machine, 8 session-runner (interleaving snapshots via recording fakes), 9 substrate byte-level, 8 surface dispatch, 5 keymap chord cases, and **1 real-pty round trip (PASSED: real reader gate, scripted fake $EDITOR, release->re-pin ordering verified in captured bytes, cooked post-mortem stty)** under `:pty :unix_only :skip_on_ci`. Main suite 5843/5852 and raxol_terminal 2202/2206 — every non-green reproduced identically on pristine master/with changes stashed. Format + warnings clean.

### DIFF
--- a/examples/harness_fixture_demo.exs
+++ b/examples/harness_fixture_demo.exs
@@ -20,7 +20,10 @@
 #
 # Keys while running: `z` fold/unfold the focused block, `j`/`k` move
 # focus (once off the composer -- see `Surface.focus_transcript/1`), `Tab`
-# queues a steer stub, `Esc` shows an interrupt stub (no agent lane in
+# queues a steer stub, `Ctrl-E` opens the composer draft in $VISUAL/$EDITOR
+# (the harness suspends its terminal claim while the editor owns the tty,
+# then resumes and reloads the edited draft -- see
+# `Raxol.Harness.EditorSession`), `Esc` shows an interrupt stub (no agent lane in
 # fixture mode -- see `Raxol.Harness.Surface`'s moduledoc, precondition
 # #6), `q` quits cleanly WHEN THE COMPOSER BUFFER IS EMPTY (Ctrl-C does
 # not: raw mode disables SIGINT delivery, per `Raxol.Terminal.InlineDriver`'s
@@ -86,7 +89,11 @@ defmodule Raxol.Examples.HarnessFixtureDemo do
         width: width,
         rows: rows,
         footer_rows: @footer_rows,
-        tty?: tty?
+        tty?: tty?,
+        # Ctrl-E hands the composer draft to $VISUAL/$EDITOR (real tty
+        # runs only -- a piped/CI run has no terminal to hand over, and
+        # Surface renders its honest stub notice instead).
+        editor_session: if(tty?, do: Raxol.Harness.EditorSession)
       )
 
     try do
--- a/lib/raxol/harness/editor_session.ex
+++ b/lib/raxol/harness/editor_session.ex
@@ -1,0 +1,326 @@
+defmodule Raxol.Harness.EditorSession do
+  @moduledoc """
+  The THIN impure runner for the composer's external-editor handoff:
+  drives `Raxol.Harness.EditorSuspend`'s pure state machine and
+  interprets each step against the real device / stty / reader gate /
+  editor process. All sequencing, ordering, and failure-recovery policy
+  lives in the pure machine; this module only performs effects.
+
+  ## The spawn mechanism (why a Port with `:nouse_stdio`)
+
+  The editor must own the REAL tty. `System.cmd/3` can never provide
+  that -- its child's stdio is a pipe pair back to the BEAM. A
+  `Port.open({:spawn, cmd}, [:nouse_stdio, :exit_status])` moves the
+  port's own pipe protocol to fds 3/4, so the child INHERITS the BEAM's
+  fds 0/1/2 -- the terminal itself. This is byte-for-byte the mechanism
+  OTP's own shell uses for its open-in-editor feature (`user_drv`'s
+  Ctrl+O path: `disable_reader` -> `open_port(..., [nouse_stdio])` ->
+  reload on port exit -> `enable_reader`), so the approach is
+  OTP-sanctioned, not novel. `{:spawn, string}` goes through `/bin/sh`,
+  which buys two things: `$EDITOR` values carrying arguments
+  (`"code -w"`) work unmodified, and a missing editor binary surfaces
+  as the shell's exit status 127 -- mapped here to a keep-the-draft
+  `{:editor_not_found, cmd}` outcome instead of a crash. The temp-file
+  path is single-quote shell-escaped before it joins the command line.
+
+  The `:spawn_fun` seam is `(shell_command :: String.t()) ->
+  non_neg_integer() | :crashed` -- synchronous, returning the editor's
+  exit status (`:crashed` when the port died without ever delivering
+  one). The default implementation is the Port described above.
+
+  ## Outcomes
+
+    * `{:ok, %{text: edited, width: w, rows: h}}` -- editor exited 0 and
+      the temp file read back; `text` is the decoded draft
+      (`EditorSuspend.decode_draft/1`).
+    * `{:kept, reason, %{width: w, rows: h}}` -- the terminal was
+      suspended and RESUMED, but the draft is kept unchanged:
+      `:editor_nonzero` | `{:editor_not_found, cmd}` | `:editor_crashed`
+      | `:reload_failed`.
+    * `{:error, {:reader_disable, reason}}` / `{:error, {step, reason}}`
+      -- a step failed before the handoff completed; the machine's
+      compensation ran, so the terminal is back in a recoverable state.
+      In particular a reader-disable failure aborts BEFORE any byte
+      touches the device: never hand the tty to an editor while the
+      BEAM reader still competes for its keystrokes.
+
+  A step that RAISES (rather than returning an error) still runs the
+  machine's `recovery/1` compensation first, then the exception
+  propagates unchanged -- callers that must not crash (the harness UI
+  loop) wrap their call; the terminal is already restored by the time
+  the exception reaches them either way. The temp file is additionally
+  removed in a `try/after`, so no path -- including the raise path --
+  leaks it.
+
+  ## What this module deliberately does NOT do
+
+  It never writes DECSTBM region bytes. The machine's
+  `:reassert_region` step is interpreted as a no-op HERE because region
+  emission is owned by the paint authority
+  (`Raxol.UI.Rendering.PaintAuthority.InlineAuthority.reassert/1` over
+  its `ScrollRegionManager` state -- the single-DECSTBM-owner rule): the
+  caller that owns the authority (`Raxol.Harness.Surface`'s edit-draft
+  dispatch) composes `resize |> reassert` on EVERY return from this
+  function -- ok, kept, or error -- so the pin is guaranteed
+  belt-and-braces regardless of where a failure landed. Geometry is
+  re-queried here (while still cooked, before re-entering raw mode) and
+  returned so that caller re-pins at the terminal's CURRENT size.
+
+  Every injectable seam has a real default: `:stty` (module),
+  `:reader_gate` (module), `:reader` (pid), `:env`, `:tmp_dir`,
+  `:spawn_fun`, `:size_fun`, `:device`. Tests drive the full sequence
+  with fakes and a StringIO device; the real tty is never touched by
+  the suite.
+  """
+
+  alias Raxol.Harness.EditorSuspend
+  alias Raxol.Terminal.InlineDriver.ReaderGate
+  alias Raxol.Terminal.InlineDriver.Sequences
+
+  @type outcome ::
+          {:ok, %{text: String.t(), width: pos_integer(), rows: pos_integer()}}
+          | {:kept, term(), %{width: pos_integer(), rows: pos_integer()}}
+          | {:error, {atom(), term()}}
+
+  @spec run(String.t(), keyword()) :: outcome()
+  def run(draft, opts) when is_binary(draft) and is_list(opts) do
+    rows = Keyword.fetch!(opts, :rows)
+    stty = Keyword.get(opts, :stty, Raxol.Terminal.Driver.Stty)
+
+    {:ok, editor} =
+      opts
+      |> Keyword.get_lazy(:env, fn -> System.get_env() end)
+      |> EditorSuspend.resolve_editor()
+
+    path =
+      opts
+      |> Keyword.get_lazy(:tmp_dir, &System.tmp_dir!/0)
+      |> Path.join(EditorSuspend.tmp_filename(unique()))
+
+    ctx = %{
+      draft: draft,
+      editor: editor,
+      path: path,
+      cmd: nil,
+      device: Keyword.get(opts, :device, :stdio),
+      rows: rows,
+      width: Keyword.get(opts, :width, 80),
+      stty: stty,
+      gate: Keyword.get(opts, :reader_gate, ReaderGate),
+      reader:
+        Keyword.get_lazy(opts, :reader, fn ->
+          Process.whereis(:user_drv_reader)
+        end),
+      spawn_fun: Keyword.get(opts, :spawn_fun, &default_spawn/1),
+      size_fun: Keyword.get(opts, :size_fun, fn -> stty.size() end),
+      exit_status: nil,
+      outcome: nil
+    }
+
+    try do
+      drive(EditorSuspend.new(), ctx)
+    after
+      # Belt-and-braces: the machine's own :cleanup_tmp (happy path) or
+      # compensation (failure path) already removes the file; this
+      # `after` guarantees it even on the raise path. Removing an
+      # already-removed file is a harmless error tuple.
+      _ = File.rm(path)
+    end
+  end
+
+  # -- the machine loop --------------------------------------------------
+
+  defp drive(machine, ctx) do
+    case EditorSuspend.advance(machine, :ok) do
+      {:done, _machine} ->
+        finish(ctx)
+
+      {:effect, step, machine} ->
+        run_step(step, machine, ctx)
+    end
+  end
+
+  # A raise inside a step still runs the machine's compensation (the
+  # `recovery/1` rescue seam) before propagating unchanged -- the
+  # terminal must be recoverable even when the failure is an exception,
+  # not an error return.
+  defp run_step(step, machine, ctx) do
+    interpret(step, ctx)
+  rescue
+    error ->
+      Enum.each(EditorSuspend.recovery(machine), &compensate(&1, ctx))
+      reraise error, __STACKTRACE__
+  catch
+    kind, reason ->
+      Enum.each(EditorSuspend.recovery(machine), &compensate(&1, ctx))
+      :erlang.raise(kind, reason, __STACKTRACE__)
+  else
+    {:ok, ctx} ->
+      drive(machine, ctx)
+
+    {:error, reason} ->
+      {:abort, compensation, _machine} =
+        EditorSuspend.advance(machine, {:error, reason})
+
+      Enum.each(compensation, &compensate(&1, ctx))
+      {:error, {error_tag(step), reason}}
+  end
+
+  defp finish(%{outcome: nil} = ctx),
+    do: {:ok, %{text: ctx.text, width: ctx.width, rows: ctx.rows}}
+
+  defp finish(%{outcome: {:kept, reason}} = ctx),
+    do: {:kept, reason, %{width: ctx.width, rows: ctx.rows}}
+
+  # The one step whose public error tag differs from its machine name:
+  # callers see the RESOURCE that failed (the reader gate), not the
+  # machine-internal step atom.
+  defp error_tag(:disable_reader), do: :reader_disable
+  defp error_tag(step), do: step
+
+  # -- step interpretation ----------------------------------------------
+
+  defp interpret(:write_tmp, ctx) do
+    case File.write(ctx.path, EditorSuspend.encode_draft(ctx.draft)) do
+      :ok -> {:ok, ctx}
+      {:error, posix} -> {:error, posix}
+    end
+  end
+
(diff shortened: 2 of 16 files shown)


DROOdotFOO (commented):

Reviewed the full diff and the surrounding substrate (`InlineDriver.start_stdin_reader/0`,
`Driver.Stty`, `ScrollRegionManager`, `InlineAuthority`, `Surface`). The pure state
machine, the compensation ledger fold, the suspend/teardown byte kinship
(`teardown_bytes == suspend_bytes <> "\r\n"`), the geometry-gated resize vs unconditional
`reassert`, and the `:user_drv_reader` handle (InlineDriver arms the SAME registered reader
the gate targets, so that coupling is at least internally consistent) all check out against
the code. The findings below are where prose and code diverge.

### CRITICAL
- **[Saboteur + New Hire] (promoted from HIGH)** `lib/raxol/harness/editor_session.ex` `interpret(:enable_reader)` (~L241-245) -- the reader re-enable result is **swallowed**: `_ = ctx.gate.enable(ctx.reader); {:ok, ctx}`. `ReaderGate.enable/2` can return `{:error, {:reader_down, reason}}` (reader died under the editor) or `{:error, :timeout}`. On the `:reader_down` path the prim_tty stdin reader is gone and never rearmed -- **the tty is permanently deaf** -- yet the step reports `:ok`, the resource ledger records `reader: :enabled` (a lie -- `editor_suspend.ex` `apply_step(_, :enable_reader)` models enable as infallible, ~L259), the outcome is `{:ok, text}` / `{:kept, ...}`, and the composer happily shows the edited draft with **no notice and no telemetry**. The moduledoc claims "an enable failure leaves input degraded... the caller may notify rather than crash" -- but the failure never reaches the caller: `Surface.run_editor`'s `{:ok, ...}` branch has no channel for "resumed but input dead." The exhaustive compensation property in `editor_suspend_test.exs` is *vacuous over this case* because the ledger model assumes `:enable_reader` cannot fail. This is exactly the "leaked reader gate / tty permanently deaf" failure the ledger is advertised to prevent. Two personas, one root -> promoted. Fix: propagate the gate result into the outcome (add an `{:ok_degraded, ...}` / notice channel) and telemetry-emit on enable failure; do NOT let the ledger assert `:enabled` on a failed enable.

### HIGH
- **[Security Auditor]** `lib/raxol/harness/editor_session.ex` temp-file creation (path build ~L95-99, `File.write` ~L184) -- the draft is written with plain `File.write/2` (default mode, honors umask -> typically `0644`, world-readable) into `System.tmp_dir!` (the shared, world-writable, sticky `/tmp`). The draft is a composer prompt and can contain secrets / API keys / private text. For the whole time the editor is open any local user can read it. There is no `0600`, no O_EXCL/O_NOFOLLOW, no per-user `0700` subdir. The name `raxol_draft_<monotonic_int>_<microsecond>.md` is only weakly unpredictable, so `File.write` following a **pre-planted symlink** is a live race (truncate/redirect of an arbitrary file the harness user can write, or feeding attacker content back as the "edited" draft). Fix: create in a `0700` per-process subdir, open with `[:exclusive]` (O_EXCL) and chmod `0600`, or use an mkstemp-equivalent.

### MEDIUM
- **[Saboteur]** `lib/raxol/harness/editor_session.ex` `default_spawn/1` (~L303-315) -- the `receive` awaiting `{:exit_status, ...}` / `:DOWN` has **no timeout**. A wedged/daemonizing `$EDITOR`, or one that never exits after the reader-gate race ate its only keystroke, blocks this synchronous call **forever**. Because `run_editor` runs inline in the TEA `handle_input/2` loop, the entire harness UI freezes with the terminal in the editor's mode and no escape hatch from the BEAM side. Fix: bound the receive (configurable), and on timeout kill the port and run compensation -> `{:kept, :editor_timeout, geo}`.
- **[Security Auditor]** `lib/raxol/harness/editor_session.ex` `interpret(:spawn_editor)` (~L218) builds `ctx.editor <> " " <> shell_quote(ctx.path)` and hands it to `Port.open({:spawn, cmd})` = `/bin/sh -c`. Only the path is quoted; `$VISUAL`/`$EDITOR` is interpolated **unquoted and unvalidated** by design (arg-splitting for `"code -w"`). Draft content correctly never reaches the shell (good). But the moduledoc frames this as safe; it is only safe under the assumption that env is fully trusted. In any context where env is attacker-influenceable (SSH `ForceCommand`/`AcceptEnv`, sudo `env_keep`, a service manager's environment) a hostile `$EDITOR` is arbitrary command execution **on the real controlling tty**. There is no allowlist, no basename check, no `[:args]`-form spawn. Fix: at minimum document the trust boundary at the call site; consider rejecting `$EDITOR` values containing shell metacharacters, or spawn via argv where the editor string is a single non-arg token.
- **[New Hire]** `packages/raxol_terminal/lib/raxol/terminal/inline_driver/reader_gate.ex` (`call/2`, ~L80) -- the gate speaks a **reverse-engineered private OTP wire protocol** (`monitor(..., alias: :reply_demonitor)` + `{ref, :disable | :enable}` against a hardcoded `:user_drv_reader`, "verified against OTP kernel prim_tty.erl's reader_loop/2"). The version-coupling risk is admitted in prose but pinned/guarded **nowhere** -- no OTP version assertion, no dep constraint, no runtime capability probe. When OTP changes the reader message shape or registration, `disable` degrades to a silent 2s timeout on every Ctrl-E (best case) or sends a stray message to whatever now holds that name (worse). Fix: add an OTP-version guard or a feature-probe that fails the bind cleanly, and pin the tested OTP range.

### LOW
- **[New Hire]** The suspend park sequence `"\e[#{rows};1H"` is hand-inlined in `sequences.ex suspend_bytes/1` while every sibling uses a named builder (`modes_off/0`, `release_region/0`, `autowrap_cursor/0`). Extract a `park_bottom/1` for parity and to keep the "magic byte" in one place.
- **[Saboteur]** The ledger treats `:release_screen`'s `IO.write(suspend_bytes)` as all-or-nothing; a real device that partially writes before raising would leave modes/region in a state the ledger doesn't model. Narrow window, real device only; noting for completeness alongside the CRITICAL.

### Cross-persona overlaps
- **enable_reader swallow** -- Saboteur (permanent tty-deaf leak) + New Hire (the "exhaustive compensation" safety claim is unsound because the ledger models enable as infallible). Same root -> promoted HIGH -> CRITICAL.
- Temp-file handling is touched by both Security (world-readable/symlink) and, tangentially, Saboteur (leftover-on-crash -- but the `try/after File.rm` genuinely covers the leak side, so only the disclosure/symlink angle stands). No further promotion.

### Verdict: BLOCK

One CRITICAL (silent permanent input death on reader re-enable failure, with the compensation
proof not covering it) plus a HIGH world-readable/symlink-prone temp file. The pure machine,
byte vocabulary, and resume composition are genuinely well-built and well-tested; the blockers
are on the impure edges the ledger model does not actually observe.

merklebonsai:
Review response landed as `fa9b313c9` (10 files, +861/-80). All fixes red-first (10 failing tests captured before any implementation change). Point-by-point:

**CRIT-1 — enable_reader swallowed; ledger models enable as infallible; compensation property vacuous.** *Confirmed, fixed.* `:enable_reader` is now the machine's one *degradable* step: `advance(machine, {:degraded, reason})` continues the resume but records it; the ledger honestly keeps `reader: :disabled` (a later failure's compensation retries the enable — tested); `degradations/1` + a `degraded:` field on both `{:ok,..}` and `{:kept,..}` outcomes; `[:raxol, :harness, :editor, :reader_enable_failed]` telemetry; the Surface renders a visible footer warning on both paths. `{:degraded,_}` on a non-degradable step raises. Bonus catch from the new test: the first warning implementation overflowed 80 columns and ViewText truncation would have eaten it — the warning is now width-budgeted.

**HIGH (security) — world-readable draft in shared tmp; symlink race.** *Confirmed, fixed.* Per-run directory with a `strong_rand_bytes` suffix chmod 0700; draft created with `[:exclusive]` (O_CREAT|O_EXCL — refuses pre-existing paths, never follows a symlink final component) then chmod 0600; whole dir removed on every path. Tests assert 0600/0700, the per-run shape, and `:eexist` refusal for both a pre-existing file and a planted symlink (target proven untouched).

**MED — unbounded editor wait.** *Confirmed, fixed with a documented default choice.* `:editor_timeout_ms` threaded through a new `:editor_opts` seam; on timeout: best-effort SIGTERM + port close + `{:kept, :editor_timeout, geo}` + notice. Default stays `:infinity` deliberately — a human legitimately edits for arbitrary time on their own tty; the bound is for embedders/automation. Tested against a real port.

**MED — $EDITOR sh-interpolated unvalidated.** *Confirmed, addressed by explicit trust-boundary documentation; allowlist rejected with rationale.* Same contract as git/crontab/OTP's own open_editor: whoever set `$EDITOR` could already run commands as this user; draft content never reaches the shell; embedders crossing privilege boundaries must inject a vetted `:env`. A metacharacter allowlist would break real configs (`emacsclient -a ""`) while defending a boundary that doesn't exist here.

**MED — ReaderGate private-protocol coupling unpinned.** *Confirmed, fixed.* Verified range OTP 26-29 pinned; below-floor calls refuse with `{:error, {:unsupported_otp, release}}`; drift is fail-closed (unknown message -> ignored -> bounded timeout -> disable failure aborts the suspend). New 4-test protocol suite with scripted readers.

**LOW — park bytes inline.** *Fixed:* `park_bottom/1` extracted, byte-equality pinned.

**LOW — partial-write ledger.** *Fixed beyond the ask:* suspend-phase failures now fold WORST-CASE into the compensation ledger (partial bytes / created file / reply-lost disable count as landed), so compensation over-repairs idempotently. New exhaustive worst-case property; affected pinned expectations updated red-first.

Gates: editor-lane suites 69/69 incl. the real-pty round trip re-run green after the fixes; full suite 5867/5875 (7/8 load flakes pass on rerun, 1 pre-existing MetricsCollector proven failing with changes stashed); raxol_terminal 2203/2206 (same 3 pre-existing as pristine master); format + warnings clean.

merklebonsai:
CI triage on the two matrix failures: **Windows** was this PR's own new wedged-editor timeout test — its fake editor is `sleep 3` via sh, Unix-only tools, so the spawn fails instantly on Windows and the timeout path never runs; tagged `:unix_only` (same convention as the pty tests) in `c0e709e01`. **macOS** was `ListScorerTest`'s 1M-grapheme perf assert (`millis < 16.0` wall-clock, from merged #561, not this PR's diff) taking 24.9ms on a loaded shared runner — absolute-time perf asserts on shared CI are the same disease class as the earlier Windows timeouts; fixing forward on a separate master-based PR by raising the bound to a blowup-detection threshold.

DROOdotFOO (commented):
Since last review: 5f910ed1..c0e709e0

Re-reviewed the two new commits against the actual source at HEAD `c0e709e0`
(`editor_session.ex`, `editor_suspend.ex`, `reader_gate.ex`, `surface.ex`,
`surface/view_text.ex`), not the commit messages. The two prior BLOCK-level
findings are genuinely addressed; the safety property behind the CRITICAL is
now only PARTIALLY closed at the render layer, and two doc/behavior gaps
surfaced.

- **CRITICAL (reader re-enable swallow) -- resolved at the machine + telemetry
  layer.** `:enable_reader` is now a real DEGRADABLE step, not a swallow.
  `editor_session.ex` `interpret(:enable_reader, ...)` returns `{:degraded,
  reason}` and emits `[:raxol, :harness, :editor, :reader_enable_failed]`
  telemetry before returning; the machine records it (`editor_suspend.ex:260`
  degradable clause, `:340` a degraded completion applies NO ledger
  transition so `reader` stays `:disabled`, `:297` `degradations/1` exposes
  it), and a `{:degraded, _}` on any non-degradable step raises
  (`editor_suspend.ex:265`). Surface surfaces it (`surface.ex:924-940`). The
  old vacuous "infallible enable" is gone. (But see MEDIUM below on the
  footer render.)

- **HIGH (world-readable temp file / symlink race) -- resolved.**
  `write_draft_file/2` opens with `[:write, :exclusive, :binary, :raw]`
  (`O_CREAT|O_EXCL`; POSIX does not follow the final component as a symlink
  under `O_EXCL`) then `chmod 0600`; the per-run parent dir is `File.mkdir`
  (exclusive-create) + `chmod 0o700` in `interpret(:write_tmp, ...)`; cleanup
  is `try/after File.rm_rf(dir)` in `run/1` plus the `:cleanup_tmp` step plus
  the compensation clause -- removed on ok/kept/error/raise. Confidentiality
  window is confined. Symlink-planting into the draft path is refused with
  `:eexist`.

- **LOW (OTP user_drv version coupling) -- resolved.** `reader_gate.ex:104-112`
  pins the verified range (OTP 26-29, `@min_otp 26`) and refuses below the
  floor with `{:error, {:unsupported_otp, release}}` instead of speaking an
  unverified protocol; the drift failure mode is documented fail-closed
  (`reader_gate.ex:59-77`).

- **MEDIUM ($EDITOR word-splitting) -- accepted as documented WONTFIX.** The
  generated path is single-quote escaped (`shell_quote/1`); the editor token
  itself stays sh-word-split by design, with the trust boundary spelled out
  (`editor_session.ex` "Trust boundary" moduledoc: same contract as
  git/crontab/OTP, embedders across a privilege boundary must inject a vetted
  `:env`). Reasonable.

- **MEDIUM (unbounded spawn receive) -- addressed, opt-in.** `default_spawn/2`
  now has an `after timeout` arm that SIGTERMs + closes the port and returns
  `:timeout`. Default is `:infinity` by deliberate policy (a human edits for
  arbitrary time); the bound is threaded through `Surface`'s `:editor_opts`.
  See LOW below for a residual edge.

### MEDIUM

- **The reader-dead warning can still be render-truncated away -- same safety
  property as the original CRITICAL, relocated from machine-swallow to
  footer-truncate.** `apply_degraded_notice/2` (`surface.ex:926-940`) APPENDS
  ` · input reader failed to re-enable` to the END of an existing kept
  notice. `ViewText.lines/3` end-truncates a notice to the width budget with
  an ellipsis (`surface/view_text.ex` `truncate/2`, lines ~230-243:
  `split_at_display_width(width-1)` + ellipsis) -- it does not wrap. So the
  warning, being last, is the first casualty of truncation. On an 80-col
  terminal a long kept notice makes it partially or fully vanish, e.g.
  `$EDITOR="code --wait --user-data-dir=/tmp/x"` yields
  `"» editor not found: code --wait --user-data-dir=/tmp/x — draft kept"`
  (~65 cols) + the 35-col warning = ~100 cols, truncated to 79 -> the entire
  `input reader failed to re-enable` substring is gone. The new test
  (`editor_surface_test.exs:176`) only exercises `:editor_nonzero` (the
  SHORTEST kept notice, ~72 cols combined), so the overflow the commit claims
  to defend ("width-budgeted so truncation cannot eat it -- a real overflow
  the new test caught") is NOT actually pinned; the code comment at
  `surface.ex:928` asserting the warning "must survive ... on an 80-column
  terminal" is contradicted by the append-at-end ordering. Mitigation that
  keeps this MEDIUM not HIGH: `[:raxol, :harness, :editor,
  :reader_enable_failed]` telemetry fires unconditionally regardless of the
  footer, so a wired embedder still sees it -- but an interactive operator
  reading only the footer would not. (Saboteur + New Hire converge on this
  root: behavior loses the message, and the in-code contract comment claims
  it does not.) Fix: prepend the warning, or render it as its own notice line
  exempt from the shared truncation, and add an overflow test with a long
  kept notice.

### LOW

- **Temp-dir name is not the cryptographic random the docs claim.** The
  moduledoc (`editor_session.ex:71`) and both fix commits say "unpredictable
  `:crypto.strong_rand_bytes`-suffixed name" / "strong-random-suffixed
  per-run directory", but `unique/0` (`editor_session.ex:472`) is
  `:erlang.unique_integer([:positive, :monotonic]) <> "_" <>
  System.os_time(:microsecond)` -- a monotonic counter plus a timestamp, i.e.
  predictable, NOT cryptographic. Impact is bounded because `File.mkdir` is
  exclusive-create: a correctly-predicted pre-planted directory causes an
  abort (DoS), not disclosure, and `O_EXCL` + `0700` still hold for the
  secret window. So this is a documentation-integrity / missing
  defense-in-depth finding, not a disclosure hole -- but the confidentiality
  section makes a specific crypto claim the code does not honor, and a future
  reader/auditor would trust it. (Security Auditor + New Hire, same root.)
  Fix: either use `:crypto.strong_rand_bytes(...) |> Base.url_encode64` as the
  suffix, or correct the doc to describe what `unique/0` actually does.

- **`default_spawn/2` timeout can strand a port message and does not bound the
  default wedge.** On the `after timeout` arm, `kill_editor/2` demonitors
  (flushing only the `:DOWN`) and closes the port, but a `{port,
  {:exit_status, N}}` delivered in the race window just before close is not
  drained; `default_spawn` runs in the harness dispatcher / TEA-loop process,
  so a stray message can land in that mailbox. Minor. Separately, with the
  default `:infinity` a genuinely wedged (non-interactive) editor -- the
  original MEDIUM's scenario -- still freezes the loop forever unless an
  embedder opts into `:editor_timeout_ms`; the interactive-human rationale is
  sound but the default leaves the automation case unprotected. Consider
  draining the port mailbox after close, and documenting that non-interactive
  embedders should set a bound.

### Verdict: CONCERNS

Both prior blockers are genuinely fixed (O_EXCL/0600/0700 confinement; the
enable-reader swallow is now a recorded, telemetry-emitting degradation with a
non-vacuous compensation ledger). What remains is one MEDIUM where the
resolved CRITICAL's safety message can still be truncated out of the footer
for long kept notices on narrow terminals (telemetry mitigates; the claimed
test does not cover it), plus two LOW doc/edge gaps. Not a BLOCK -- the tty is
no longer left silently deaf at the machine layer -- but the footer-truncation
and the strong-random doc claim should be corrected before merge.

DROOdotFOO:
have conflicts here

merklebonsai:
Rebased onto master (through #604-#607, #609) + round-2 response, landed as `21862dd88`.

**Rebase notes:** 4 conflicted files, all both-sides-appended; master canonical throughout. One semantic merge decision, documented at the clause: master's "open overlay freezes the composer, `:steer` no-ops" rationale now extends to `:edit_draft` — suspending to edit the hidden mid-pick buffer, then keyframing back at the overlay's wrong row split, would be the same dishonest UI. 727 tests green across the combined suites (master's overlay/frontier/markdown/cadence + this branch's editor lane).

**Round-2 ledger:**

**MEDIUM — reader-dead warning truncated for long kept notices; test only covered the shortest notice; comment overclaims.** *Confirmed — your exact scenario reproduced red* (long `{:editor_not_found, cmd}` + append + end-truncation ate the warning). Fixed: the warning is now its OWN footer notice line — `stub_notice` accepts a list, one width-truncated line per entry, so no notice can truncate another. Overclaiming comment removed. New test pins your exact 80-col scenario asserting BOTH messages present.

**LOW — doc claims strong_rand_bytes; `unique/0` is counter+timestamp.** *Confirmed.* Fixed by making the code honor the doc rather than weakening it: 6 bytes of `strong_rand_bytes` (url-base64) appended. Perms test extended to assert the random segment's shape (fails against numeric-only names).

**LOW — timeout arm can strand a raced exit_status; `:infinity` default leaves automation unprotected.** *Confirmed both halves.* `kill_editor` drains a raced exit_status with a zero-timeout receive after close (the race isn't deterministically forcible, so no dedicated test — noted honestly; the real-port timeout test still covers the path). Moduledoc now states plainly: non-interactive embedders MUST set `:editor_timeout_ms` or a wedged editor blocks the loop forever.

Gates: editor-lane + keymap 115/115; **real-pty round trip green post-rebase and post-fixes**; full suite 6051/6058 with the single repeatable failure the known pre-existing MetricsCollector timing test (proven on pristine master); format + warnings clean.

PR #611, opened 2026-07-16
test: order-of-magnitude bound on the 1M-grapheme scorer guard (CI flake)

merklebonsai:
The macOS matrix on #610 flaked on `ListScorerTest`'s 1M-grapheme adversarial test: `assert millis < 16.0` wall-clock took 24.9ms on a loaded shared runner — with the grapheme clamp working correctly. Same disease class as the earlier Windows property timeouts: absolute-time asserts on shared CI turn runner load into false correctness failures.

The test's purpose is blowup detection (an unclamped O(query x key) DP over a 1M-grapheme label costs seconds), and the file's own sibling test already documents the right convention: "Regression guard only (order-of-magnitude), not the 16ms target" at 200ms. This applies the same convention to the adversarial case, with a comment recording the flake data.

### DIFF
--- a/test/raxol/ui/list_scorer_test.exs
+++ b/test/raxol/ui/list_scorer_test.exs
@@ -281,7 +281,12 @@ defmodule Raxol.UI.ListScorerTest do
 
       millis = micros / 1000
 
-      assert millis < 16.0,
+      # Order-of-magnitude blowup guard, matching the sibling test's
+      # convention above: without the clamp, the O(query * key) DP over a
+      # 1M-grapheme label costs seconds — 200ms catches that class while
+      # absorbing shared-CI-runner load (a 16ms wall-clock target flaked
+      # on loaded runners at ~25ms with the clamp working correctly).
+      assert millis < 200.0,
              "expected the 1M-grapheme adversarial label to stay bounded, took #{millis}ms"
 
       assert [%{item: matched, key: key}] = results


DROOdotFOO (commented):

Scope: test-only change, single commit `b367f976`, raising the 1M-grapheme scorer guard's wall-clock assert from `16.0ms` to `200.0ms` in `test/raxol/ui/list_scorer_test.exs`. Reviewed against `lib/raxol/ui/list_scorer.ex` (the `@max_score_graphemes` clamp under test).

### Medium

- **[Saboteur] medium** test/raxol/ui/list_scorer_test.exs:~289 (new `assert millis < 200.0`) -- The fix keeps the exact mechanism the PR body itself names as the repo's #1 flake class: an absolute wall-clock assert on a shared CI runner. It only swaps one magic number for a larger one. It does not switch to a deterministic proxy, so it still measures runner load, not the scorer. Worse, on the *correct* (clamped) path `rank/4` touches at most `@max_key_bytes` (~4096 bytes) via the O(1) `bounded_key_prefix/1` sub-binary -- real scorer work is sub-millisecond, yet the flaking run measured ~25ms, i.e. the assertion is dominated by scheduling/GC noise, not the code under test. Fix: assert the actual invariant the clamp provides -- input-size independence -- by ranking a 1M-grapheme label and a 2M-grapheme label and asserting their times are within a small ratio (post-cap work is O(1) in label length). A ratio bound is far more robust to shared-runner noise than any absolute millisecond number.

- **[Saboteur] medium** lib/raxol/ui/list_scorer.ex:115 (`@max_score_graphemes 1024`) -- An order-of-magnitude bound only catches *removal* of the clamp (unclamped: `String.graphemes` over 1M plus the `O(query * key)` DP = well over 1s, so 200ms does separate correct from catastrophic here). It does not catch a *weakening* of the clamp: raise `@max_score_graphemes` from 1024 to, say, 50_000 and the DP + grapheme split cost stays comfortably under 200ms while the guard has silently regressed 50x. The test therefore does not pin the value it exists to protect. Fix: additionally assert against the clamp constant directly (e.g. that `positions` never exceed `@max_score_graphemes`, or drive two labels straddling the cap and assert equal work), so a loosened cap fails a test rather than sliding through.

- **[Security Auditor] medium** lib/raxol/ui/list_scorer.ex:153 (`query_graphemes = normalize_string_graphemes(query, ...)`) -- The guard this test defends is a real DoS control on untrusted input (pasted labels / log lines used as keys), and the key axis is clamped. But the *query* axis is not: `query` is grapheme-split with no cap before use. The subsequence gate + key clamp bound the DP itself (a query longer than the 1024-grapheme clamped key fails `subsequence?/2` at list:312 and never reaches `align/3`), so the quadratic blowup is contained -- but `normalize_string_graphemes/2` still does unbounded O(query) `String.downcase` + `String.graphemes` on the raw query on every keystroke. A user pasting a multi-MB string into the filter box pays linear-in-paste work with no ceiling. This PR touches only the key-side adversarial test; the query-side DoS axis has no clamp and no test. Fix: cap the normalized query length (mirror `@max_score_graphemes`) and add a large-query adversarial test.

### Low

- **[New Hire] low** test/raxol/ui/list_scorer_test.exs:~283 (new comment) -- The comment justifies `200.0` as "matching the sibling test's convention above," but the sibling bound (list_scorer_test.exs:263) is 200ms for a *different* scenario -- 10,000 full Smith-Waterman DP runs -- not a single 1M-grapheme label. The two `200`s are numerically coincidental, not a shared derivation, so "same convention" overstates the link. The comment records the correct-path flake (~25ms) but not what a *regressed* run costs, leaving the reader unable to confirm 200ms sits safely between correct and broken. The margin is only ~8x over the observed 25ms flake; a comment that the earlier target flaked at 1.5x over (16 -> 25) without bounding the loaded-runner worst case does not establish 8x is enough. Fix: record the measured regressed-path number (or the observed p99 correct-path time) so the threshold's safety margin is auditable.

### Cross-persona overlaps

- Saboteur ("cannot distinguish clamp-works from clamp-loosened") and Security Auditor ("query axis untested") share one root: the test under-covers the guard's real invariant surface (it pins neither the clamp value nor the second DoS axis). Promotes the "this test proves less than it appears to" concern to the strongest theme of the review.
- Saboteur ("still wall-clock") and New Hire ("unexplained magic 200") share the fragility-of-absolute-thresholds root; both are resolved by the same move to a ratio/invariant assertion.

### Verdict: CONCERNS

The immediate change is a defensible flake mitigation -- 200ms does separate the correct path from a full clamp removal, so it is not vacuous for that one failure mode. But it perpetuates the exact absolute-wall-clock pattern the project has repeatedly flagged as its top flake source instead of asserting the deterministic input-size-independence invariant the clamp actually provides, it cannot detect a loosened (vs removed) clamp, and it leaves the unbounded-query DoS axis untested. Not a BLOCK; the residual issues are substantive enough to warrant follow-up before this is treated as the guard's real protection.

PR #613, opened 2026-07-16
feat(core): centralized boundary-confinement seam [thread-2]

merklebonsai:
## What this is

The **decide-once seam** for PR #569 thread 2 ("same gap, four patches", per V's "centralize it" ruling, 2026-07-17). The same class of bug — *untrusted external bytes crossing a boundary without confinement* — was patched (or unpatched) independently at four sites. This lands the centralized contract in `raxol_core`; consumers migrate as follow-ups.

Design source: `docs/proposals/in-flight/confinement-seam-proposal.md` (option **b**). These are **two** confinements sharing a threat narrative but no types/algorithm/failure-modes, so two functions in one boundary namespace, not one.

## The two functions

```elixir
Raxol.Core.Boundary.Path.confine(root :: String.t(), requested :: String.t(), opts :: keyword())
  :: {:ok, real_path :: String.t()} | {:error, atom()}
# opts: ref_format: Regex.t()

Raxol.Core.Boundary.TermText.sanitize(binary(), opts :: keyword()) :: String.t()
# opts: allow: [0..31]  (default [?\n])
```

- **`Path.confine/3`** — path-traversal confinement, ported from the Agent Client Protocol package's proven `FsSandbox`. Rules, all *before any target syscall*: (1) optional ref-shape gate on the RAW ref → `:malformed_ref`; (2) lexical `Path.expand`-under-root prefix check → `:path_traversal`; (3) hand-rolled realpath of BOTH root and target following symlinks — leaf symlink AND symlinked-ancestor-dir, POSIX relative targets, not-yet-existing write leaf, depth-cap 40 → `:too_many_symlinks`; (4) post-resolution re-check → `:symlink_escape`. Total; never raises (`:invalid_input` on bad types).
- **`TermText.sanitize/2`** — terminal-injection confinement. Total function: strips ESC + all introduced sequences (CSI/OSC/DCS/APC/PM/SOS, two-byte forms, truncated trailing ESC), C0 controls except an allow-list, DEL, raw C1; invalid UTF-8 → U+FFFD. Enforces the repo "never embed raw ANSI in `text()`" rule *at the untrusted boundary*.

## The shared-vector drift guard

`raxol_core` cannot be consumed by the deliberately zero-raxol-dep Agent Client Protocol package (option **c** rejected: breaks its standalone-MIT story). So instead of sharing *code*, the two implementations share a **data-only conformance fixture** — `packages/raxol_core/test/support/boundary_vectors/`:

| File | Vectors | Drives |
| ---- | ------- | ------ |
| `path_reject_vectors.json` | 9 | `Path.confine/3` must-reject |
| `path_accept_vectors.json` | 8 | `Path.confine/3` must-accept |
| `term_text_vectors.json` | 26 | `TermText.sanitize/2` |

A raxol_core runner drives the functions over these vectors. The Agent Client Protocol package (follow-up, in the Agent Client Protocol repo) **copies `path_*_vectors.json` verbatim** and asserts its `FsSandbox` agrees (skipping `ref_format` vectors, a `confine/3`-only gate). A divergence is then a red test, not a silent security fork. Documented in the fixture `README.md`.

## Migration table (from the proposal)

| # | Site | Adopts | Lane / owner | Status |
| - | ---- | ------ | ------------ | ------ |
| 1 | #607 fenced-code info-string render | `TermText.sanitize/2` at render-prep | **UI lane — Drew** | follow-up (cross-lane, needs Drew's sign-off) |
| 2 | #569 / PA-6 CAS `$blob`/`snapshot_ref` deref | `Path.confine/3` + `ref_format:` at deref time | harness — ours | follow-up (PA-6/#586 harness-side) |
| 3 | #586 `:erl_tar.extract` per member | `Path.confine/3` per member vs extraction root | harness — ours | follow-up |
| 4 | Agent Client Protocol `fs_sandbox` | **no code change** — canonical seed, binds to shared vectors + duplicate-marker moduledoc | Agent Client Protocol pkg — ours | follow-up (binds to the shared vectors) |

**Note:** this is the decide-once seam only. Consumers migrate as follow-ups — #607 by the UI-lane agent, PA-6/#586 harness-side, and the Agent Client Protocol `fs_sandbox` binding to the shared vectors.

## Validation

- `MIX_ENV=test mix compile --warnings-as-errors` clean.
- 74 boundary tests pass (vector-driven + FS assertions + properties: accepted paths never escape root; sanitize output has no ESC/no disallowed control byte/valid UTF-8/idempotent for any input).
- Full raxol_core suite stays green: **869 passed**.
- No new deps (raxol_core stays telemetry-only; test JSON via built-in `JSON`).

https://claude.ai/code/session_015Rc719iBNk3hPRNKGrUPgN

### DIFF
--- a/packages/raxol_core/lib/raxol/core/boundary/path.ex
+++ b/packages/raxol_core/lib/raxol/core/boundary/path.ex
@@ -1,0 +1,151 @@
+defmodule Raxol.Core.Boundary.Path do
+  @moduledoc """
+  Path-traversal confinement: resolve an untrusted `requested` reference under
+  a trusted `root` and prove the result stays inside that root, rejecting every
+  escape *before* any syscall touches the target.
+
+  This is the centralized seed of the "same gap, four patches" boundary
+  (PR #569 thread 2). It is one of **two** boundary confinements — the other is
+  `Raxol.Core.Boundary.TermText` (terminal-injection). They share a threat
+  narrative but nothing else, so they are two functions, not one.
+
+  Seeded from the proven `Raxol.AgentClientProtocol.Client.FsSandbox.resolve/2`
+  (leaf-symlink *and* symlinked-ancestor resolution, cycle-guarded, rejecting
+  before any target syscall) plus PA-6's ref-shape gate. The Agent Client Protocol package keeps
+  its own tested copy as a documented intentional duplicate; both bind to the
+  same shared conformance vectors (`test/support/boundary_vectors/`) so drift is
+  a red test, not a silent fork.
+
+  ## Rule order
+
+  `confine/3` enforces these in order. Everything below the ref gate happens
+  before any syscall on the *target* (the only I/O is `:file.read_link` during
+  symlink resolution, which reads links, never opens/creates the target):
+
+  1. **Ref-shape gate** — when `ref_format:` is given, the RAW `requested`
+     string must match the regex, else `{:error, :malformed_ref}`. Refs are
+     validated as opaque tokens before they are ever treated as paths. Skipped
+     when the option is absent.
+  2. **Lexical confinement** — `root = Path.expand(root)`;
+     `lexical = Path.expand(Path.join(root, requested))`. Unless `lexical`
+     equals `root` or starts with `root <> "/"`, `{:error, :path_traversal}`.
+     (An absolute `requested` is jailed under `root` by `Path.join`, not
+     honored.)
+  3. **Symlink resolution** — a hand-rolled realpath of BOTH `root` and
+     `lexical`: follows a leaf symlink, recurses into the parent so a symlinked
+     ancestor directory is caught too, resolves POSIX relative link targets
+     against the link's own directory, and works for a not-yet-existing write
+     leaf by resolving the deepest existing ancestor. Depth-capped at 40;
+     a cycle yields `{:error, :too_many_symlinks}`.
+  4. **Post-resolution re-check** — the real path must still be within the real
+     root, else `{:error, :symlink_escape}`.
+
+  Total function: never raises on bad input (a non-binary `root`/`requested`
+  yields `{:error, :invalid_input}`).
+  """
+
+  @max_symlink_depth 40
+
+  @typedoc "Rejection reason returned by `confine/3`."
+  @type reason ::
+          :malformed_ref
+          | :path_traversal
+          | :too_many_symlinks
+          | :symlink_escape
+          | :invalid_input
+
+  @doc """
+  Confine `requested` under `root`.
+
+  Returns `{:ok, real_absolute_path}` when the fully symlink-resolved target
+  provably stays inside the resolved root, or `{:error, reason}` (see the
+  moduledoc rule order for the exact reason per rejection).
+
+  ## Options
+
+    * `:ref_format` — a `Regex.t()` the RAW `requested` string must match before
+      any resolution (e.g. PA-6's
+      `~r/^(blobs|snapshots)\\/[0-9a-f]{64}(\\.json)?$/`). Absent = no gate.
+  """
+  @spec confine(String.t(), String.t(), keyword()) :: {:ok, String.t()} | {:error, reason()}
+  def confine(root, requested, opts \\ [])
+
+  def confine(root, requested, opts)
+      when is_binary(root) and is_binary(requested) and is_list(opts) do
+    with :ok <- ref_gate(requested, Keyword.get(opts, :ref_format)),
+         expanded_root = Path.expand(root),
+         lexical = Path.expand(Path.join(expanded_root, requested)),
+         :ok <- lexical_gate(expanded_root, lexical),
+         {:ok, root_real} <- real_path(expanded_root),
+         {:ok, real} <- real_path(lexical),
+         :ok <- escape_gate(root_real, real) do
+      {:ok, real}
+    end
+  end
+
+  def confine(_root, _requested, _opts), do: {:error, :invalid_input}
+
+  # --- Rule 1: ref-shape gate ------------------------------------------------
+
+  defp ref_gate(_requested, nil), do: :ok
+
+  defp ref_gate(requested, %Regex{} = re) do
+    if Regex.match?(re, requested), do: :ok, else: {:error, :malformed_ref}
+  end
+
+  # A ref_format that is present but not a Regex is a misconfiguration; fail
+  # closed rather than silently skipping the gate or raising.
+  defp ref_gate(_requested, _other), do: {:error, :malformed_ref}
+
+  # --- Rule 2: lexical confinement -------------------------------------------
+
+  defp lexical_gate(root, lexical) do
+    if within?(root, lexical), do: :ok, else: {:error, :path_traversal}
+  end
+
+  # --- Rule 4: post-resolution re-check --------------------------------------
+
+  defp escape_gate(root_real, real) do
+    if within?(root_real, real), do: :ok, else: {:error, :symlink_escape}
+  end
+
+  defp within?(root, path), do: path == root or String.starts_with?(path, root <> "/")
+
+  # --- Rule 3: hand-rolled realpath ------------------------------------------
+  #
+  # If `path` itself is a symlink, follow it (a relative target resolves against
+  # the symlink's OWN directory, per POSIX); otherwise recurse into the parent
+  # so a symlinked ANCESTOR is caught, then rejoin the basename. Both existing
+  # paths (read case) and not-yet-existing ones (write case: leaf absent, every
+  # ancestor present) resolve correctly. Depth-guarded against symlink cycles.
+  defp real_path(path), do: real_path(path, 0)
+
+  defp real_path(_path, depth) when depth > @max_symlink_depth,
+    do: {:error, :too_many_symlinks}
+
+  defp real_path(path, depth) do
+    case :file.read_link(path) do
+      {:ok, target} ->
+        target = to_string(target)
+
+        resolved =
+          if Path.type(target) == :absolute,
+            do: target,
+            else: Path.join(Path.dirname(path), target)
+
+        real_path(Path.expand(resolved), depth + 1)
+
+      {:error, _not_a_symlink_or_missing} ->
+        case Path.dirname(path) do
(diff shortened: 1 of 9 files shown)


DROOdotFOO (commented):

Scope: the two confinement primitives in `raxol_core` (`Path.confine/3`, `TermText.sanitize/2`), their tests, and the shared conformance vectors. Empirical probes run in an isolated worktree against the compiled PR branch. `TermText.sanitize/2` held up well under hostile probing (reassembly cannot produce a live sequence, because the only bytes that ever reach output are `>= 0x20`, non-DEL, non-C1, which cannot introduce a sequence; OSC BEL+ST, unterminated OSC, 8-bit C1 introducers, and idempotence are all handled). The findings below are all in `Path.confine/3` and its test/vector coverage.

### CRITICAL

- **[Security Auditor] CRITICAL (reproduced)** `packages/raxol_core/lib/raxol/core/boundary/path.ex:126-136` (specifically the relative-target join at line 134) -- **symlink-escape confinement bypass**. `real_path/2` reads the leaf symlink first and resolves its POSIX-relative target against the *lexical* parent `Path.dirname(path)` instead of the *resolved-real* parent. When a symlinked ancestor changes the path's real depth, the `..` count is under-applied, so the computed "real" path stays inside root while the OS actually resolves the requested path to a file *outside* root. `escape_gate` then blesses the wrong path.
  - Reproduced: root `R`; `R/a/link` -> `R` (symlink, inside root); `R/leaf` -> `../secret` (relative). `confine(R, "a/link/leaf")` returns `{:ok, ".../R/a/secret"}` (in-root), but `File.read(".../R/a/link/leaf")` returns `"OUTSIDE-SECRET"` from `.../secret`, *outside* root. The primitive's core claim ("the fully symlink-resolved target provably stays inside root") is false.
  - Blast radius: every consumer. For #586 tar-extract (confine used as a per-member gate while `:erl_tar` writes the member's own path) and any caller that opens the requested path rather than the returned one, this is a live write/read escape. Even callers that open the returned path get a *different* file than the OS would -- a correctness break on top of the security break.
  - Fix: resolve path components left-to-right, resolving each ancestor symlink *before* descending, so a relative leaf target is anchored to the already-resolved parent directory (resolve `dirname` first, then read the leaf link relative to the resolved dir). The current leaf-first + lexical-anchor order is the classic realpath bug.

### HIGH

- **[Saboteur] HIGH (reproduced)** `packages/raxol_core/lib/raxol/core/boundary/path.ex:123-124` and `:144` -- **depth cap conflates directory nesting with symlink-follow depth**. `depth` is incremented on every parent recursion (line 144), not only on symlink follows (line 136), so a path merely 41+ components deep with **zero symlinks** trips `depth > @max_symlink_depth` and returns `{:error, :too_many_symlinks}`.
  - Reproduced: a 45-level-deep directory tree, no symlinks, `confine/3` -> `{:error, :too_many_symlinks}`. Fail-closed (not a bypass), but it wrongly rejects legitimate deep paths (deep tar members, nested CAS/snapshot dirs) and reports a misleading reason. `@max_symlink_depth 40` also bounds total path depth, which is not what "too many symlinks" means.
  - Fix: increment `depth` only on the symlink-follow branch (line 136); guard parent recursion with a separate, generous structural bound or none.

- **[New Hire] HIGH** `packages/raxol_core/test/raxol/core/boundary/path_test.exs:185-213` -- the "accepted paths never escape root" property **cannot catch the CRITICAL bug and gives false confidence**. It asserts only that the *returned string* `real` is under `real_root` (which is itself produced by `confine`, so it validates `confine` against itself), and its random generator (`segments` at line 194) never creates symlinks -- so it structurally never exercises symlink escape at all. That is exactly why finding #1 passes CI green. The shared vectors also miss the whole class: no relative-symlink-under-symlinked-ancestor case (accept or reject), no sibling-prefix case, no deep-nesting case. The Agent Client Protocol `FsSandbox` copy inherits every blind spot verbatim.
  - Fix: add an independent oracle (compare the returned path, and the *requested* path, against the OS's own realpath / an `openat`+`O_NOFOLLOW` component walk); add a symlink-generating generator; add reject vectors for the relative-symlink-under-symlinked-ancestor escape and an accept/reject vector for deep nesting and sibling-prefix.

### MEDIUM

- **[Security Auditor / New Hire] MEDIUM** `packages/raxol_core/lib/raxol/core/boundary/path.ex:1-45` (moduledoc) -- **TOCTOU caller obligation is undocumented**. The doc claims the resolved target "provably stays inside the resolved root" but never states that resolution is point-in-time and that the caller MUST open the returned path without re-resolving symlinks (`openat`/`O_NOFOLLOW`), nor that a symlink swapped between `confine` and the caller's open defeats it. Given finding #1 (returned path already diverges from OS truth), a caller re-opening the original requested path is doubly exposed. Document the obligation as part of the contract the four consumers adopt.

### Cross-persona overlaps

- Findings #1 (Security Auditor, CRITICAL) and #3 (New Hire, HIGH) share one root: symlink-resolution correctness and the property that was supposed to prove it. Two personas, same defect -> already at CRITICAL; the promotion is moot but reinforces severity. #4 (TOCTOU) is the same primitive's under-specified contract.

### Verdict: BLOCK

One reproduced CRITICAL confinement bypass (relative symlink resolved against the lexical parent) plus two HIGH findings (depth-cap conflation, a self-referential safety property that cannot catch the bypass). Because this is the decide-once seam four sites will adopt, the bypass and the blind-spot vector set must be fixed here before any consumer migrates.

DROOdotFOO:
## Cross-cutting: pause the #586 and #569/PA-6 migrations until the CRITICAL here is fixed

This note is a merge-order flag, not a new finding -- it follows from the CRITICAL in the review above.

#613 is the decide-once seam. Its own migration table lists four adopters, and two of them are
security-load-bearing consumers that turn `Path.confine/3` into a real filesystem write/deref gate:

- **Migration #3 -- #586 `:erl_tar.extract` per member** confines each member path against the
  extraction root, then extracts. With the reproduced symlink-escape bypass (`path.ex:126-136`),
  a crafted archive member (`a/link/leaf` where `link` is a symlinked ancestor) passes `confine`
  as in-root while the OS writes it OUTSIDE root -- a classic tar-slip, now with a green gate in
  front of it. Adopting #613 as-is would make #586 *look* hardened while shipping a live escape.
- **Migration #2 -- #569 / PA-6 CAS `$blob`/`snapshot_ref` deref** confines an imported journal's
  content-addressed pointer before dereferencing it. Same primitive, same bypass: an imported
  journal with a symlinked-ancestor ref reads a file outside the CAS root. PA-6 is the amendment
  that is supposed to *close* the traversal hole; wiring it to this confine would ratify a hole
  as a fix.

The shared-vector drift guard does not save either consumer here: the CRITICAL ships green because
the "never escape root" property validates `confine` against itself and its generator plants no
symlinks (review finding #3), and the copied `path_*_vectors.json` miss the
relative-symlink-under-symlinked-ancestor class -- so the Agent Client Protocol `FsSandbox` that
binds to the same vectors (migration #4) inherits the identical blind spot rather than catching it.

Suggested sequencing:
1. Fix `real_path/2` to resolve relative symlink targets against the *resolved-real* parent, not the
   lexical parent; separate the symlink-depth counter from parent-recursion depth.
2. Add a red vector + a property test whose "escape" oracle is an independent OS realpath (not
   confine's own output) and whose generator actually plants leaf and ancestor symlinks. Only when
   that test is red-then-green does the vector set become trustworthy for the ACP copy.
3. THEN land the #586 and PA-6 migrations against the fixed seam. Until then, hold both -- a confine
   that returns `{:ok, escaping_path}` is worse than the un-migrated per-site checks, because each
   site stops doing its own validation once it trusts the seam.

`TermText.sanitize/2` is unaffected -- it held up under hostile probing -- so migration #1 (#607's
fenced-code render adopting `TermText.sanitize`) does NOT need to wait on this; it can proceed
independently of the path fix.

---
_Cross-PR sequencing note from the adversarial review of #613 (thread-2 seam)._

DROOdotFOO:
## Pushed a fix for the CRITICAL (symlink-escape) + the depth-cap HIGH

Commit `715417b1` rewrites `real_path/2` in `packages/raxol_core/lib/raxol/core/boundary/path.ex`.

**Root cause of the CRITICAL:** the old `real_path/2` resolved a symlink's relative target against
`Path.dirname(path)` -- the *lexical* parent -- instead of the symlink's *resolved-real* parent. Under
a depth-inflating ancestor, the `..` count under-applied, so a link escaping root still resolved to an
interior path and `confine/3` returned `{:ok, <interior>}` for a target the OS resolves outside root.

**The fix:** `real_path/2` now walks components left-to-right from the filesystem root, carrying a
fully-resolved real `base`. A symlink is followed against `base` (its real parent), so relative
targets -- including `..` -- resolve against where the link actually lives. `.`/`..` adjust `base`
directly; non-symlink / not-yet-existing components are accepted literally (write-leaf case preserved).
This also fixes the depth-cap HIGH: only symlink hops increment the counter now, so deep symlink-free
nesting is no longer misreported as `:too_many_symlinks`.

**Teeth (addresses the "property test validates confine against itself" HIGH):** added two regression
tests that assert the outcome directly, with no confine-derived oracle:

- `relative leaf symlink under a depth-inflating ancestor cannot under-apply ..` -- plants
  `loop -> .` (inflates lexical depth) + `esc -> ../secret.txt`, requests `loop/loop/esc`, asserts
  `{:error, :symlink_escape}`. This is the reproduced CRITICAL.
- `deep symlink-free nesting is not mistaken for a symlink cycle` -- 80-level symlink-free tree,
  asserts it resolves.

Verified red-then-green: **both fail against the pre-fix `real_path` (31/33), pass after (33/33)**;
full `raxol_core` suite **871 passed**, format + `--warnings-as-errors` clean.

Not addressed here (still open from the review): the shared vector set / ACP-copied
`path_*_vectors.json` still don't cover the relative-symlink-under-symlinked-ancestor class, and the
TOCTOU caller obligation (open the returned path with `O_NOFOLLOW`, do not re-resolve) is still
undocumented. Those, plus re-gating the #569 migration plan on this fix, are worth folding in before
the #586 / PA-6 path migrations adopt the seam.

_Fix pushed by the adversarial-review lane._

PR #615, opened 2026-07-16
Harness UI: unread divider (live-region attention boundary)

merklebonsai:
An operator looks away; the agent keeps producing blocks. On return, nothing marks where they stopped reading. This adds the unread divider: a dim "── N new since you looked ──" rule in the live footer region above the first unseen block, clearing when the operator catches up.

## Design

- **Offsets, not wall-clock** (per lane law: no wall clocks anywhere): `Raxol.Harness.UnreadDivider` is a pure attending/away machine over caller-injected completed-block counts. `blur/2` records the boundary; `focus/2` (or the keystroke fallback `input_activity/2`) opens at most ONE span; re-blur merges keeping the OLDEST unvisited boundary; the count freezes at return; only `viewed/2` (jump reaching the boundary block) retires it. Typing is presence evidence, not reading evidence — keystrokes never clear it.
- `Surface.blur/1` / `Surface.focus/1` are the explicit seam the later mode-1004 focus-consumption unit plugs into.
- **Live-region only — sealed bytes never change:** the divider never enters the seal frontier, never seals, has no fold identity; jump navigation skips it by construction. Enforced byte-for-byte by a named test (identical SealOracle history across attended and unattended runs) plus a "label never appears in history" test. No golden changes. Cross-surface last-seen is the documented deferred protocol addition (client-local, non-durable — per the lane accord's L5 confirmation).
- Renders through the same ViewText sanitize/truncate seam as all footer content; suppressed under an open overlay; absent in flat mode.

## Verification

Red-first: 37/37 failing before implementation. 23 policy + 14 integration tests, including an adversarial-fixture smuggle check (no `\e[2J`; divider row matches `^─+ \d+ new since you looked ─+$`) and CJK-safe width-exactness at widths 24-120. Full test/harness + test/property: 764 passed, 0 failures. Warnings-as-errors + format clean.

### DIFF
--- a/lib/raxol/harness/surface.ex
+++ b/lib/raxol/harness/surface.ex
@@ -1388,13 +1388,14 @@ defmodule Raxol.Harness.Surface do
   # content, not raw fixture traffic.
   defp unread_offset(model), do: length(model.projection.blocks)
 
-  # Clamps the unread-divider state against the just-rebuilt projection
-  # (`UnreadDivider.reconcile/2`) -- `advance/2` is the only place
-  # `projection.blocks` is ever replaced, so threading here means a
-  # shrunken rebuild can never leave a stuck span whose boundary no
-  # navigation index can reach. A no-op under fixture mode's monotone
-  # growth (every existing suite proves that); it exists for the
-  # replay/reattach shapes a future producer can hit.
+  # Reconciles the unread-divider state against the just-rebuilt
+  # projection (`UnreadDivider.reconcile/2`, retire-only -- counts are
+  # clamped at display time by `divider/2`, never into state) --
+  # `advance/2` is the only place `projection.blocks` is ever replaced,
+  # so threading here means a shrunken rebuild can never leave a stuck
+  # span whose boundary no navigation index can reach. A no-op under
+  # fixture mode's monotone growth (every existing suite proves that);
+  # it exists for the replay/reattach shapes a future producer can hit.
   defp reconcile_unread(model) do
     %{
       model
--- a/lib/raxol/harness/surface.ex
+++ b/lib/raxol/harness/surface.ex
@@ -72,6 +72,9 @@ defmodule Raxol.Harness.Surface do
       the Component tree's view maps to the paint authority's flat
       `iodata()` rows (see that module's doc for why truncation precedes
       styling).
+    * **`Raxol.Harness.UnreadDivider`** -- the pure attention-boundary
+      policy behind the "N new since you looked" footer rule (see "The
+      unread divider" section below).
 
   ## Precondition #2 -- keymap-first dispatch (binding, load-bearing)
 
@@ -147,6 +150,29 @@ defmodule Raxol.Harness.Surface do
   Composer's lines ++ notice -- the pending/live-tail preview lines are
   SUPPRESSED for exactly as long as the overlay is open (the space they'd
   occupy is now claimed by the overlay), and return the moment it closes.
+
+  ## The unread divider (live-region honesty)
+
+  `Raxol.Harness.UnreadDivider` decides, purely from caller-injected
+  block-commit offsets (see that module's own moduledoc), whether a
+  "N new since you looked" rule should render. This module owns two
+  things that policy doesn't: WHEN to feed it an offset (`blur/1` and
+  `focus/1`, the explicit mode-1004 seam; `handle_input/2` also feeds
+  `UnreadDivider.input_activity/2` on every keystroke as the fallback
+  return signal, and `move_focus/2` feeds `UnreadDivider.viewed/2` on
+  every jump) and WHERE to render its output: `footer_lines/1` inserts
+  `unread_divider_lines/1`'s single dim line between the status strip
+  and the pending/live preview, styled through the same `ViewText` seam
+  every other footer line uses. The divider is decided at return-time
+  and rendered ONLY in the repaintable footer -- sealed history is never
+  touched (enforced byte-for-byte by the integration suite's
+  sealed-bytes-identical test, not merely asserted in prose), it is
+  suppressed under an open overlay exactly like the pending preview, and
+  it is absent in `:flat` mode (no footer, no live region, so `blur/1`/
+  `focus/1` are safe no-ops there). It clears the instant `move_focus/2`
+  reaches or passes the boundary block -- jumps skip the divider itself
+  by construction, since `focused_index` ranges only over
+  `projection.blocks`.
 
   ## The overlay picker (footer-region overlay)
 
@@ -412,6 +438,7 @@ defmodule Raxol.Harness.Surface do
   alias Raxol.Terminal.ScrollRegionManager
   alias Raxol.Harness.StatusStrip
   alias Raxol.Harness.Surface.ViewText
+  alias Raxol.Harness.UnreadDivider
 
   alias Raxol.UI.Components.Harness.{Block, BlockBody, Composer}
   alias Raxol.UI.Harness.{InputEvent, Keymap, OverlayPicker}
@@ -445,7 +472,8 @@ defmodule Raxol.Harness.Surface do
           stub_notice: String.t() | [String.t()] | nil,
           overlay: overlay() | nil,
           editor_session: module() | (String.t(), keyword() -> term()) | nil,
-          editor_opts: keyword()
+          editor_opts: keyword(),
+          unread: UnreadDivider.t()
         }
 
   @typedoc """
@@ -562,7 +590,8 @@ defmodule Raxol.Harness.Surface do
       stub_notice: nil,
       overlay: nil,
       editor_session: Keyword.get(opts, :editor_session),
-      editor_opts: Keyword.get(opts, :editor_opts, [])
+      editor_opts: Keyword.get(opts, :editor_opts, []),
+      unread: UnreadDivider.new()
     }
 
     model
@@ -985,6 +1014,14 @@ defmodule Raxol.Harness.Surface do
   def handle_input(model, raw_event) do
     norm = InputEvent.normalize(raw_event)
 
+    # Every keystroke is presence evidence for the unread divider's
+    # keystroke fallback (see `UnreadDivider`'s "mode-1004 seam" doc) --
+    # this is a no-op unless the policy is currently `:away`.
+    model = %{
+      model
+      | unread: UnreadDivider.input_activity(model.unread, unread_offset(model))
+    }
+
     context = %{
       composing?: model.composing?,
       streaming?: not Map.get(model.status, :turn_completed, false),
@@ -1298,7 +1335,12 @@ defmodule Raxol.Harness.Surface do
     else
       current = model.focused_index || if delta > 0, do: -1, else: total
       next = (current + delta) |> max(0) |> min(total - 1)
-      %{model | focused_index: next}
+
+      %{
+        model
+        | focused_index: next,
+          unread: UnreadDivider.viewed(model.unread, next)
+      }
     end
   end
 
@@ -1321,6 +1363,39 @@ defmodule Raxol.Harness.Surface do
   def focus_composer(model) do
     {composer, _cmds} = Composer.update(%{focused: true}, model.composer)
     %{model | composing?: true, composer: composer}
+  end
+
+  # The offset the unread-divider policy is a pure function of: the
+  # count of already-committed blocks (see `UnreadDivider`'s "offsets,
+  # not clocks" doc). Blocks, not events -- the divider marks completed
+  # content, not raw fixture traffic.
+  defp unread_offset(model), do: length(model.projection.blocks)
+
+  @doc """
+  Records that the operator has looked away, for the unread-divider
+  policy (`Raxol.Harness.UnreadDivider.blur/2`). This is the explicit
+  attention API a later focus-event unit (a real terminal focus-out
+  signal) wires directly -- see `UnreadDivider`'s "mode-1004 seam" doc.
+  A no-op in `:flat` mode's own honest sense: `paint_footer/1` never
+  repaints there, so nothing visibly changes, but the policy state
+  itself still tracks the boundary (harmless, since flat mode has no
+  footer for a divider to ever reach).
+  """
+  @spec blur(t()) :: t()
+  def blur(model) do
+    %{model | unread: UnreadDivider.blur(model.unread, unread_offset(model))}
+    |> paint_footer()
+  end
+
+  @doc """
+  Records that the operator has returned, for the unread-divider policy
+  (`Raxol.Harness.UnreadDivider.focus/2`). See `blur/1`'s doc and
+  `UnreadDivider`'s "mode-1004 seam".
+  """
+  @spec focus(t()) :: t()
+  def focus(model) do
+    %{model | unread: UnreadDivider.focus(model.unread, unread_offset(model))}
+    |> paint_footer()
   end
 
   # -- overlay picker (see the moduledoc's "The overlay picker" section) ---
@@ -1461,9 +1536,11 @@ defmodule Raxol.Harness.Surface do
     status_line = StatusStrip.render(model.status, model.width)
     overlay_lines = overlay_lines(model)
 
-    # The pending/live-tail preview is suppressed while an overlay is
-    # open -- the overlay claims that space (see the moduledoc's
-    # precondition #5 update, "The overlay picker" section).
+    # Both the divider and the pending/live-tail preview are suppressed
+    # while an overlay is open -- the overlay claims that space (see the
+    # moduledoc's precondition #5 update, "The overlay picker" section,
+    # and the "unread divider" section for why it yields the same way).
+    divider_lines = if model.overlay, do: [], else: unread_divider_lines(model)
     preview_lines = if model.overlay, do: [], else: pending_preview_lines(model)
 
     composer_lines =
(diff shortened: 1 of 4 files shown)


DROOdotFOO (commented):

Scope: the unread-divider commit (29e09232) only -- `lib/raxol/harness/unread_divider.ex` (pure policy), the `Surface` integration in `lib/raxol/harness/surface.ex`, and the two test files. Findings are verified against code; the pure machine was replayed against crafted transition sequences in an isolated scratch dir (marked "reproduced").

### Medium

- **[Saboteur] medium** `lib/raxol/harness/unread_divider.ex:180-186` (`viewed/2`) + `surface.ex:1340-1343` (`move_focus`) -- **stuck divider under a post-`focus` block-count shrink (reproduced).** Once `focus/2` freezes a span at `%{from: F, count: C}`, the ONLY navigation clear-path is `viewed/2` gated on `block_index >= F`, and `move_focus` can only ever feed `viewed/2` an index in `0..length(projection.blocks)-1`. If `projection.blocks` shrinks below `F` after the span opens (session replay / reattach / truncation -- the exact case the moduledoc's "Defensive boundaries" claims to cover), no reachable index can ever reach `F`, so navigation can NEVER retire the divider. It sits in the footer forever asserting "C new since you looked" while fewer than `F` blocks exist. Reproduced: `new() |> blur(5) |> focus(8)` -> `%{from: 5, count: 3}`; blocks shrink to 4 so the highest navigable index is 3; `viewed(3)` is a no-op -> divider still `%{from: 5, count: 3}`, `STUCK? true`. The module advertises non-monotone defense but only samples it once, at `focus`. Fix: reconcile in `viewed/2` (also clear when `block_index >= offset - 1`, i.e. the operator reached the last extant block), or clamp `boundary`/`from` to the current offset on the offset feed, or document a hard monotonicity precondition at the `unread_offset/1` callsite.

- **[New Hire] medium** `surface.ex:1384-1398` (`blur/1`) -- **the feature is dormant at runtime.** `Surface.blur/1` is the ONLY path that drives `UnreadDivider` into `:away`, and it has ZERO callers anywhere in `lib/` or `packages/` (only docs and tests reference it). `input_activity/2` (fed on every keystroke at `surface.ex:1022`) only transitions `:away -> :attending`; it never sets `:away`. So with no `blur/1` caller the attention machine never leaves `:attending`, `divider/1` stays `nil`, and the "N new since you looked" rule can never render in the running harness -- all 976 lines are exercised only by tests that call `blur/1` directly. The terminal already parses focus events (`packages/raxol_terminal/lib/raxol/terminal/advanced_features.ex:282-283`, `{:focus_in}`/`{:focus_out}`, mode 1004 at `mode_state.ex:42`) but nothing routes them to `Surface.blur/focus`. A maintainer reading the tests would reasonably assume the divider works. Fix: either wire the existing `parse_focus_event` output to `blur/1`/`focus/1` now, or state plainly in the moduledoc that the divider is inert until the mode-1004 unit lands (the current prose calls it "today's fallback," implying it is live).

### Low

- **[Saboteur] low** `lib/raxol/harness/unread_divider.ex:135-147` (`focus/2`) -- **"Defensive boundaries" overstates protection; count silently under-reports (reproduced).** The defensive clause only rejects the single-sample case `offset <= boundary`. A shrink-then-regrow that stays above `boundary` passes the `offset > boundary` guard and produces a too-small count, even though the blocks between the shrink floor and the new offset are entirely different content. Reproduced: `new() |> blur(5) |> focus(6)` (blocks silently went 5 -> 3 -> 6 while away) -> `%{from: 5, count: 1}` -- "1 new" when up to 3 blocks are actually new. The moduledoc should scope its non-monotone claim to "a single decreased sample at focus-time," not general non-monotonicity.

- **[New Hire] low** `surface.ex:1340-1343` + `unread_divider.ex:175-186` -- **`from` silently doubles as both a block COUNT and a block INDEX.** `boundary`/`from` is set from `length(projection.blocks)` (a count) but compared in `viewed/2` against `focused_index` (a 0-based index). Correctness rests on the unstated identity "count of seen blocks == index of first unseen block," which holds only while `focused_index` is 0-based and blocks are densely indexed from 0. Neither the `viewed/2` doc nor the `move_focus` callsite states this load-bearing assumption; a future change to 1-based focus indexing or sparse block ids would silently break clearing. Add a one-line note at the callsite.

- **[Security Auditor] low** `lib/raxol/harness/unread_divider.ex:216-220` (`pad_with_rule/2`) + `surface.ex:950-957` (`resize/3`) -- **unbounded `String.duplicate` driven by unclamped terminal width.** `line/2` builds `String.duplicate("─", left)` with `left ~ width/2`, and `width` flows straight from `resize/3` with no upper clamp. A terminal (or spoofed resize event) reporting an enormous width forces an O(width) allocation on every footer repaint. Shared with the other footer units (`StatusStrip`), so not unique to this PR, but the divider adds a second such allocation on the same hot path; a `min(width, sane_max)` clamp at the width boundary would cap it. Reassuringly, the divider's *content* is safe: it renders only an integer count plus a compile-time label, and goes through the same `ViewText` `:styled` sanitize/truncate seam (`surface.ex:1573-1582`) as every other footer line -- no agent-controlled block text ever reaches it, so there is no smuggle/injection surface, and over-width output is truncated to `model.width`.

### Cross-persona overlaps

- **Frozen count never reconciled against current reality** is hit from two sides: Saboteur (finding A -- stuck + wrong count under shrink) and the Security integrity angle (a footer the operator trusts to report "what the agent did while I was away" can assert a count that no longer matches the actual block set). Same root as finding A; promotion already reflected in its Medium rating. A single fix -- reconcile `from`/`count` against the live offset when clearing or rendering -- closes both.

### Verdict: CONCERNS

No hard blocker: the divider is dormant at runtime (New Hire medium), so the stuck-divider (Saboteur A) cannot regress a live surface today, and the content path is injection-safe. But before mode-1004 wiring makes this reachable, the frozen-count/shrink reconciliation (A + overlap) and the width clamp should land, and the moduledoc's "defensive"/"fallback" prose should stop implying the feature is live.

merklebonsai:
Review response landed as `0f003b67a`. Point-by-point:

**MEDIUM — stuck divider under post-focus block-count shrink.** *Confirmed with your exact repro.* Fixed per your single-fix suggestion: `reconcile/2` (retire the span when `offset <= from`, clamp count on partial shrink, pull the away-boundary down) threaded on every `advance/2`, plus a reconciled render read `divider/2` — a stale span can neither stick nor paint. Red-first with `blur(5) |> focus(8)` -> shrink-to-4.

**MEDIUM — feature dormant at runtime.** *Confirmed; moduledoc route, decided by evidence:* `parse_focus_event/1` has zero callers and the inline driver neither enables mode 1004 nor parses `\e[I`/`\e[O` — no end-to-end delivery path exists, so wiring it here would have meant building driver plumbing that belongs to the attention-escalation unit. Both moduledocs now state plainly the divider is inert at runtime until that unit lands; no prose implies otherwise.

**LOW — count under-reports on shrink-then-regrow.** *Confirmed.* Claim honestly scoped to "a single decreased sample at focus-time" in a real "Defensive boundaries and reconciliation" section (the old prose dangled a nonexistent section reference); the new reconciliation narrows the practical exposure.

**LOW — `from` doubles as count and index.** *Confirmed.* Load-bearing identity note added at the `move_focus` callsite.

**LOW — unbounded `String.duplicate` from unclamped width.** *Confirmed.* `line/2` clamps width to 1024 with rationale; the shared StatusStrip pattern is noted as backlog rather than refactored here.

Gates: 46/46 divider tests, full test/harness + test/property 773 passed / 0 failures, warnings + format clean.

DROOdotFOO (commented):

Since last review: 29e09232..0f003b67

Scope: one new commit, `0f003b67` ("reconcile the unread divider against the live
block count"), targeting the stuck-divider MEDIUM. Verified the shipped pure
clauses in isolation (extracted `reconcile/focus/blur/viewed/divider` verbatim,
no TextMeasure needed, run under `elixir`).

- **Stuck divider under post-focus shrink (MEDIUM, prior finding) -- genuinely fixed.**
  The reviewer's exact repro `new() |> blur(5) |> focus(8)` freezes `%{from: 5, count: 3}`;
  the OLD `viewed/2` gate (`index >= 5`) is still unreachable because navigation only
  feeds indices `0..offset-1`. But `reconcile/2` now retires it: `reconcile(state, 4)`
  -> `span: nil`, and the render read `divider/2` at offset 4 -> `nil`. Fix is threaded
  in BOTH places -- `Surface.reconcile_unread/1` on every `advance/2` (the only projection
  rebuild) mutates state, and `unread_divider_lines/1` reads through the read-only
  `divider/2`. So even a rebuild that bypasses `advance/2` (reattach) still cannot PAINT
  a stale span: `divider(stuck_state, 2)` returned `nil` while the underlying state was
  still `%{from: 5}`; the next `advance/2` then clears the phantom. Both the common case
  and the reattach/render case are covered. Tests lock it (`unread_divider_test.exs:212`
  the exact repro, `:228` reachability invariant, `:285` reconciled read;
  `unread_divider_surface_test.exs:263` end-to-end shrink drop).
- **Unclamped `String.duplicate` in `line/2` (prior LOW) -- fixed for the divider.**
  `min(width, @max_rule_width)` (1024) caps the O(width) allocation on the repaint hot
  path. Residual: `StatusStrip` still shares the unclamped pattern, honestly filed as
  backlog in the moduledoc rather than fixed here.
- **Dormancy (prior MEDIUM) -- documented, NOT wired.** Still test-only (reproduced):
  `parse_focus_event/1` has zero non-doc callers, `Surface.blur/1` has zero production
  callers, `InlineDriver` has no mode-1004 / focus-byte routing. This commit does not
  wire it (correctly out of scope) but both moduledocs now state plainly the divider is
  "INERT at runtime until that unit lands" -- the honesty gap is closed.

### Low

- **Saboteur -- reconcile corrupts the count in BOTH directions under a non-monotone
  offset (reproduced).** `reconcile/2` clamps against a raw block COUNT it cannot
  semantically interpret, so any projection wobble (the exact replay/reattach shapes this
  feature exists to serve) silently corrupts the frozen count:
  - UNDER-count on a transient dip while attending: start `%{from: 5, count: 3}`;
    `reconcile(_, 6)` clamps to `count: 1`; a later `reconcile(_, 8)` regrow leaves it
    frozen at `count: 1` ("N new" now lies low). Documented as "growth never inflates a
    frozen count," but the practical effect is a legitimately-unread divider shrinking on
    a momentary blip.
  - OVER-count / mis-anchor via the away-boundary pull: `blur(5)` (operator had read
    blocks 0..4), projection shrinks to 3 while away so `reconcile(_, 3)` pulls
    `boundary: 3`, then regrow + `focus(8)` yields `%{from: 3, count: 5}` -- re-marking
    blocks 3-4 the operator already read. This over-report direction is NOT documented
    (the moduledoc only warns about the under-report shape). Both are only reachable under
    the future producer that doesn't exist yet, hence LOW, but the boundary-pull
    over-count deserves the same honest one-line caveat the under-count got.
- **New Hire -- layering is sound, one question.** `reconcile` runs on EVERY `advance/2`
  even under fixture mode's monotone growth (a proven no-op there). Given `divider/2`
  already clamps read-only at paint time and is what the renderer uses, the state
  mutation in `reconcile_unread/1` mainly exists to restore `viewed/2` reachability. That
  is a real reason (a stuck state would otherwise never clear via navigation), so it is
  justified -- but it is also the exact mechanism that bakes the transient-dip under-count
  into state permanently. If the read-only clamp is the source of truth for painting,
  consider whether the state clamp should be gated to only-shrink-below-`from` (retire)
  rather than also clamping counts, to avoid permanently freezing a dipped count.
- **Security Auditor -- width DoS closed on this surface.** Confirmed no unbounded
  `String.duplicate` remains reachable through the divider: `line/2` clamps to 1024 before
  `pad_with_rule/2`. The adversarial-fixture smuggle test (no `\e[2J`, row matches
  `^─+ \d+ new since you looked ─+$`) and CJK width-exactness remain. No new external
  input surface introduced.

### Verdict: CONCERNS

Down from prior CONCERNS and materially improved: the blocking-class stuck-divider MEDIUM
is genuinely fixed on both the advance and reattach/render paths, the width DoS is bounded,
and the dormancy is now honestly documented. Remaining items are LOW: reconcile clamping
against a semantically-opaque count introduces a two-directional count corruption under
non-monotone offsets (one direction undocumented), reachable only by a producer that does
not yet exist. Not blocking; recommend documenting the boundary-pull over-count and
reconsidering whether the state-level count clamp is worth its transient-dip cost.

merklebonsai:
Round-2 response landed as `ef75ef552`. Ledger:

**LOW — two-directional count corruption.** *Confirmed both directions.* (a) Under-count bake-in: my round-1 state-level count clamp turned a transient offset dip into a permanent under-count — fixed by removing count mutation from state entirely; red-first with your exact repro, byte-asserted at both offsets (paints the reduced count while dipped, recovers the honest count on regrow). (b) Boundary-pull over-count: kept DELIBERATELY, now documented + pinned by a named test — over-report is the fail-safe direction for an unread marker: it can re-mark read content but can never hide unread content. The moduledoc names both directions honestly.

**LOW — gate the state clamp to retire-only.** *Confirmed sound, adopted as designed by the review:* `reconcile/2` is now retire-only + away-boundary-pull (the post-reconcile invariant `from < offset` is preserved by retire alone, so `viewed/2` reachability holds); `divider/2` is an independent read-only display clamp. State is never count-mutated.

**Width DoS / stuck-divider / dormancy:** resolved per your own confirmation, no action.

Red-first: 2 failing before the fix + 1 green doc-pin. 48/48 divider tests; full harness+property 775 passed / 0 failures; warnings + format clean.

DROOdotFOO (commented):

Since last review: 0f003b67..ef75ef55

Scope: one new commit, `ef75ef55` ("unread-divider reconcile is retire-only;
counts clamp at display time"), targeting the residual round-2 LOW (two-directional
count corruption under non-monotone offsets). Verified against `lib/raxol/harness/unread_divider.ex`,
`lib/raxol/harness/surface.ex`, and both test suites; ran the pure policy suite
(33 passed) and the surface integration suite (15 passed) in an isolated worktree.

- **`reconcile/2` is genuinely retire-only on the span now.** The count-clamp
  clause (old `when offset < from + count -> %{state | span: %{from: from, count: offset - from}}`)
  is DELETED. The three surviving clauses (`unread_divider.ex:245-262`) can only:
  (1) clear the whole span when `offset <= from` (retire), (2) pull an away
  `boundary` down when `boundary > offset` (span is always `nil` while `:away`,
  so this never touches a span), or (3) return state verbatim. No clause moves
  `span.from` forward or grows `span.count` -- confirmed by the "growth never
  inflates the frozen count" test (`unread_divider_test.exs:303`) and by direct
  trace: `blur(2)|>focus(8)` = `%{from:2,count:6}`, then `reconcile(_, 9)` still
  `%{from:2,count:6}`. Re-marking of already-read blocks via a *forward-moving*
  span is now structurally impossible in reconcile.

- **The under-count-on-transient-dip artifact is fixed.** Because state is never
  count-clamped, a dip is display-only: `blur(5)|>focus(8)|>reconcile(6)` keeps
  `%{from:5,count:3}` in state; `divider/2` paints `count:1` at offset 6 and
  recovers `count:3` at offset 8 (`unread_divider.ex:280-286`, pinned by
  `unread_divider_test.exs:259` and `:243`). The old bug (freezing `count:1`
  forever, lying low after regrowth) is gone.

- **Display clamp (`divider/2`) is correct and total.** Three clauses:
  `offset <= from -> nil`; `from < offset < from+count -> %{from, count: offset-from}`
  (always `pos_integer`, since `offset > from` strictly in this branch); else the
  frozen span. Bounds render to reality without mutating stored state, and
  `viewed/2`'s gate keeps using the unclamped `span.from`, so navigation and
  display stay consistent. Reconcile and display share the same `offset <= from`
  retire threshold and the same per-advance offset (`length(projection.blocks)`),
  so no render-vs-state divergence: `surface.ex:1600-1626` reads
  `divider(model.unread, unread_offset(model))` with the identical offset that
  `reconcile_unread/1` threaded in `advance/2`.

### Low

- **The `blur(5)+shrink-to-3+regrow+focus(8) -> %{from:3,count:5}` over-report
  is NOT fixed -- it is now accepted and documented as intentional (reproduced).**
  Traced and confirmed still live: `new()|>blur(5)|>reconcile(3)|>focus(8)|>divider()`
  == `%{from:3,count:5}`, pinned as a *passing* test at `unread_divider_test.exs:277`.
  The away-boundary pull (`reconcile/2` clause 2, `unread_divider.ex:255-260`)
  drags `boundary` from 5 down to 3, so the next `focus` re-marks indices 3..4 as
  new. This is now argued as the deliberate fail-safe direction: for an unread
  marker, over-reporting (re-marking read content) is safe, under-reporting
  (hiding unread content) is not. The reasoning is principled -- without the pull,
  a shrink-then-regrow with genuinely new low-index content would UNDER-report,
  which is the worse direction -- and the caveat is now honestly documented
  (moduledoc "Defensive boundaries", `unread_divider.ex:97-104`) and pinned.
  I accept it as a documented tradeoff rather than an open defect, but flag that
  the task's "no longer re-marks read blocks" bar is met only in the reconcile
  path, not the away-pull path; the over-report remains reachable by construction.

- **Feature still DORMANT at runtime (honest).** `Surface.blur/1` still has zero
  production callers; `input_activity/2` fed on every keystroke can only CLOSE an
  away state, never open one, so the machine never leaves `:attending` in the
  running harness and the divider never renders outside the test suites. Documented
  plainly in both moduledocs ("DORMANT TODAY", "INERT at runtime until that unit
  lands"). Out of scope to wire; noted for honesty, no dishonesty found.

### New Hire

- No render/state desync introduced by moving the clamp to display time.
  `viewed/2` (`unread_divider.ex:227-233`) gates on `span.from` from *state*, which
  is never mutated by the display clamp, so a dipped paint (`count:1`) and a
  navigation-to-boundary still agree on the same `from`. Post-reconcile the
  surviving-span invariant `from < offset` holds (retire fires at `offset <= from`),
  so `viewed/2` reachability is preserved -- no stuck-divider regression. Verified
  green by `unread_divider_test.exs:228` ("reconciliation restores viewed/2's
  reachability invariant").

### Security Auditor

- No new attack surface. The only externally-influenced quantity remains the
  render width, still clamped to `@max_rule_width` (1024) before any
  `String.duplicate/2` (`unread_divider.ex:315`, `pad_with_rule/2`), unchanged by
  this commit. `divider/2`'s new arithmetic (`offset - from`) is bounded to
  `(0, count)` by its own guards, no overflow/negative path. Output still routes
  through the `ViewText` sanitize/truncate seam (`surface.ex:1610`), and the
  no-C0/no-escape property test (`unread_divider_test.exs:382`) still passes.

### Verdict: CLEAN

The commit does exactly what it claims: `reconcile/2` is retire-only on the span
(no forward `from`, no `count` growth -- structurally verified), the transient-dip
under-count is fixed by moving the clamp read-only into `divider/2`, and the
display clamp is correct with no render-vs-state divergence. The one residual
item -- the away-boundary-pull over-report -- is a principled, documented,
test-pinned fail-safe rather than an open defect, and the feature's runtime
dormancy is disclosed honestly. 48/48 divider tests pass locally.

PR #617, opened 2026-07-16
Harness UI: grade transcript blocks by turn recency into prominence

merklebonsai:
The salience wave's payoff: nothing assigned prominence until now — every block rendered at 1.0. This adds the pure recency policy so the transcript reads like attention: current turn loud, older turns stepped down by perceptual tier, approvals never buried.

## Policy (`Raxol.Harness.RecencyPolicy`, pure — no processes, no clocks, no config)

- `prominence(turns_behind)`: 0 -> 1.0, 1 -> 0.8, 2 -> 0.6, 3+ -> 0.4 (floor); nil/negative -> 1.0 — uncertainty never reads darker.
- `grade(turn_ids, current_turn)` pure core (turn order = first-seen distinct ids; defensive fallbacks all resolve toward loud) + `grade_block/2` render-path convenience deriving turn identity from `event_refs`.

## Seal-time grading (the substrate-honest reading of the spec)

Prominence is decided exactly once, when a block is painted/sealed, and never re-graded — print-once history cannot repaint scrollback. A block sealed during its own turn seals at 1.0 and does NOT fade when the next turn starts; the fade ladder is visible wherever multiple turns paint in one pass, while incrementally sealed history keeps its seal-time grade forever. Stated plainly in the moduledoc.

## Wiring

One alias + the body of one private function (`render_block_lines/3` — the single place block render context is built, called from `seal_block/2`, so the grade computed there IS the seal-time grade). 10-line additive diff. Needs-input promotion required zero policy code: Block auto-passes `needs_input: true` for live approvals and the merged prominence floor (0.6) + accent engage — approval outranks the ladder by composition, proven by a named end-to-end test.

## Verification

Red-first: 9 failed / 2 passed before the module and wiring existed (incl. the wiring test failing on the absent seal-time fade fragment in real sealed bytes). 14 policy tests: ladder table, five-turn grade `[0.4, 0.4, 0.6, 0.8, 1.0]`, unknown-turn fallbacks, five-turn fixture rendering the ladder byte-exact against the solver (newest block asserted SGR-free), live-approval-outranks-tier, and a surface seal-time test observing grade 0.8 in real sealed bytes. Harness suites 1134/1134; full suite 6107/6113 (all 6 pre-existing flakes, 5 pass on rerun). Golden snapshots unchanged — no bless needed. Warnings + format clean.

## Documented follow-up

The expanded rich-body components (BodyProvider path) don't yet thread `context[:prominence]` — pre-existing gap, honestly stated in the moduledoc's "Coverage today" section; the grade is already present in their context for when they grow support.

### DIFF
--- a/lib/raxol/harness/recency_policy.ex
+++ b/lib/raxol/harness/recency_policy.ex
@@ -24,6 +24,20 @@ defmodule Raxol.Harness.RecencyPolicy do
   darker: an ungraded block is never demoted below full prominence by
   this policy's own uncertainty.
 
+  ### Why these values (the ladder's provenance)
+
+  The four tiers are not new: they are the shipped salience ladder
+  already pinned by `Raxol.UI.Harness.Prominence` -- `0.6` is the
+  ordinary-context tier (`Prominence.needs_input_floor/0` floors
+  awaiting-input content exactly there), the `1.0`/`0.6` pair is pinned
+  by regression to survive 256-color quantization as distinct palette
+  indices, and `0.4` is that ladder's own floor. This policy was
+  ratified with an explicit "no new tiers" fence: it maps recency onto
+  the EXISTING ladder, one uniform `0.2` step per turn behind, and any
+  retuning of the tier values themselves belongs to the solver-side
+  ladder (and its pending human-eye ratification pass, see the
+  `Prominence` moduledoc), never to this mapping.
+
   ## Seal-time grading (the substrate law)
 
   Prominence for a block is decided **exactly once, at the moment the
@@ -129,6 +143,13 @@ defmodule Raxol.Harness.RecencyPolicy do
   `turns_behind` is `position(current_turn) - position(block_turn)` in
   that order.
 
+  **Caller contract (load-bearing): `turn_ids` must be in transcript
+  order.** Positional recency is meaningless on a reordered list -- a
+  turn that appears first IS oldest to this function, whatever its id
+  looks like. This is a documented precondition, not something this
+  pure function can validate (turn ids are opaque; there is no
+  order-independent notion of "older" to check against).
+
   Rules (each a documented guarantee, never darker than honest
   uncertainty warrants):
 
@@ -162,7 +183,8 @@ defmodule Raxol.Harness.RecencyPolicy do
   `Raxol.Harness.Projection` was built from.
 
   Derivation (defensive throughout -- a non-map entry in `events` is
-  skipped, never raised on):
+  skipped, never raised on; the whole derivation is a SINGLE pass over
+  `events`, so grading one block costs one walk, never three):
 
     * the block's own turn is the `:turn_id` of the FIRST event in
       `events` whose `:id` is in `block.event_refs` (`nil` if
@@ -175,14 +197,25 @@ defmodule Raxol.Harness.RecencyPolicy do
   Empty `events`, an unresolvable block turn, or a `nil` current turn
   all fall out of the same `grade/2` core as `1.0` -- see that
   function's own rule list.
+
+  **Input contract (load-bearing, guaranteed upstream for the intended
+  feed):** `events` must be journal/ingest-ordered and complete for the
+  blocks being graded. Both hold for `projection.source_events` by
+  construction: `Raxol.Harness.Projection.Recovery.filter_ids/1` drops
+  duplicate/out-of-order ids BEFORE retention (so the retained list is
+  id-monotonic), and `source_events` retains EVERY durable event for
+  the session un-windowed (only `:ephemeral` tier -- `item_delta`
+  traffic, which never enters a block's `event_refs` -- is excluded),
+  so every block built by the projection resolves its turn here. A
+  caller feeding a reordered or windowed list violates the contract;
+  each violation degrades toward `1.0` (never darker), per the same
+  uncertainty rule as everything else in this module.
   """
   @spec grade_block(map(), [map()]) :: prominence()
   def grade_block(%{event_refs: refs}, events) when is_list(events) do
-    safe_events = Enum.filter(events, &is_map/1)
-    order = safe_events |> Enum.map(&Map.get(&1, :turn_id)) |> turn_order()
-    current_turn = current_turn_of(safe_events)
+    {rev_order, current_turn, block_turn} = scan_events(events, refs)
+    order = Enum.reverse(rev_order)
     current_position = position(order, current_turn)
-    block_turn = block_turn_of(refs, safe_events)
 
     grade_one(order, current_position, block_turn, current_turn)
   end
@@ -227,19 +260,46 @@ defmodule Raxol.Harness.RecencyPolicy do
 
   # -- grade_block/2 event-derivation helpers ------------------------------
 
-  defp current_turn_of(events) do
-    events
-    |> Enum.map(&Map.get(&1, :turn_id))
-    |> Enum.reject(&is_nil/1)
-    |> List.last()
+  # One pass over `events` deriving all three of grade_block/2's inputs
+  # at once (the doc's "one walk, never three"): the distinct non-nil
+  # turn order (accumulated REVERSED, with a MapSet for O(1) seen
+  # checks -- the caller reverses back to first-seen order), the current
+  # turn (last non-nil turn_id wins by overwrite), and the block's own
+  # turn (the FIRST event whose id is in `refs`; latched with a
+  # `{:found, turn}` tuple rather than a bare sentinel atom so an
+  # opaque turn id can never collide with "not found yet"). Non-map
+  # entries fall through the guard and are skipped, never raised on.
+  defp scan_events(events, refs) do
+    {_seen, rev_order, current_turn, block_turn} =
+      Enum.reduce(events, {MapSet.new(), [], nil, :none}, fn
+        event, {seen, rev_order, current, found} when is_map(event) ->
+          turn = Map.get(event, :turn_id)
+
+          {seen, rev_order} =
+            if is_nil(turn) or MapSet.member?(seen, turn) do
+              {seen, rev_order}
+            else
+              {MapSet.put(seen, turn), [turn | rev_order]}
+            end
+
+          current = if is_nil(turn), do: current, else: turn
+
+          found =
+            if found == :none and Map.get(event, :id) in refs do
+              {:found, turn}
+            else
+              found
+            end
+
+          {seen, rev_order, current, found}
+
+        _non_map, acc ->
+          acc
+      end)
+
+    {rev_order, current_turn, unwrap_block_turn(block_turn)}
   end
 
-  defp block_turn_of([], _events), do: nil
-
-  defp block_turn_of(refs, events) do
-    case Enum.find(events, fn event -> Map.get(event, :id) in refs end) do
-      nil -> nil
-      event -> Map.get(event, :turn_id)
-    end
-  end
+  defp unwrap_block_turn({:found, turn}), do: turn
+  defp unwrap_block_turn(:none), do: nil
 end
--- a/lib/raxol/harness/recency_policy.ex
+++ b/lib/raxol/harness/recency_policy.ex
@@ -74,6 +74,15 @@ defmodule Raxol.Harness.RecencyPolicy do
   an approval outranks its own ladder tier automatically, with no
   needs-input branch anywhere in this module.
 
+  Scope note: on today's sole wiring point (the fixture-replay
+  surface's seal path) this composition is dormant -- the block builder
+  constructs every block already `:sealed`, and the seal frontier holds
+  a live approval back from painting at all, so a live approval never
+  reaches the graded path there; it lives in the repaintable footer
+  instead. The floor composition is real and tested at the `Prominence`
+  layer, and engages the moment a live-rendering surface grades live
+  blocks.
+
   ## Coverage today (an honest scope note)
 
   The grade is threaded through `Raxol.UI.Components.Harness.BlockBody`'s
@@ -222,6 +231,38 @@ defmodule Raxol.Harness.RecencyPolicy do
 
   def grade_block(_block, _events), do: 1.0
 
+  @doc """
+  Batch form of `grade_block/2`: grades EVERY block in `blocks` against
+  `events` with a SINGLE walk over the event list -- `O(events +
+  total_refs)` for the whole batch, where the per-block form re-derives
+  the (identical) turn order and current turn from the full event list
+  on every call.
+
+  Equivalence law (tested): `grade_blocks(blocks, events)` is exactly
+  `Enum.map(blocks, &grade_block(&1, events))` -- same input contract,
+  same defensive fallbacks, only the cost differs. This is the entry
+  point a bulk paint should use: a full re-render, a reattach rebuild,
+  or any pass that grades a whole projection at once. The incremental
+  seal path (`Raxol.Harness.Surface.render_block_lines/3`) keeps the
+  per-block form because its hold-back-one design seals a bounded
+  number of blocks per advance.
+  """
+  @spec grade_blocks([map()], [map()]) :: [prominence()]
+  def grade_blocks(blocks, events) when is_list(blocks) and is_list(events) do
+    {rev_order, current_turn, turn_by_id} = index_events(events)
+    order = Enum.reverse(rev_order)
+    current_position = position(order, current_turn)
+
+    Enum.map(blocks, fn
+      %{event_refs: refs} ->
+        block_turn = first_ref_turn(refs, turn_by_id)
+        grade_one(order, current_position, block_turn, current_turn)
+
+      _other ->
+        1.0
+    end)
+  end
+
   # -- grade/2 + grade_block/2 shared core --------------------------------
 
   # Never darker than 1.0 when either side of the comparison is
@@ -270,22 +311,18 @@ defmodule Raxol.Harness.RecencyPolicy do
   # opaque turn id can never collide with "not found yet"). Non-map
   # entries fall through the guard and are skipped, never raised on.
   defp scan_events(events, refs) do
+    ref_set = MapSet.new(refs)
+
     {_seen, rev_order, current_turn, block_turn} =
       Enum.reduce(events, {MapSet.new(), [], nil, :none}, fn
         event, {seen, rev_order, current, found} when is_map(event) ->
           turn = Map.get(event, :turn_id)
-
-          {seen, rev_order} =
-            if is_nil(turn) or MapSet.member?(seen, turn) do
-              {seen, rev_order}
-            else
-              {MapSet.put(seen, turn), [turn | rev_order]}
-            end
-
+          {seen, rev_order} = note_turn(seen, rev_order, turn)
           current = if is_nil(turn), do: current, else: turn
 
           found =
-            if found == :none and Map.get(event, :id) in refs do
+            if found == :none and
+                 MapSet.member?(ref_set, Map.get(event, :id)) do
               {:found, turn}
             else
               found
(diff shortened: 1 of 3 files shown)


DROOdotFOO (commented):

Scope: the single PR commit `2787ede8` -- `lib/raxol/harness/recency_policy.ex` (new pure policy), the 10-line `surface.ex` wiring, and `test/harness/recency_policy_test.exs`. The pure module is well-guarded: every branch resolves toward `1.0`, output is always in `{1.0, 0.8, 0.6, 0.4}`, and `grade_block/2` never raises on empty/missing/non-map/nil input (verified by running the module against crafted inputs). The findings below are correctness-intent and quality concerns, not crashes.

### Medium

- **[Saboteur] medium** `lib/raxol/harness/recency_policy.ex:225-231` (`block_turn_of/2`) + `test/harness/recency_policy_test.exs:169` -- The ladder INVERTS for the oldest blocks. A block whose `event_refs` resolve to nothing in `source_events` grades `1.0` (loudest). But `Projection` deliberately strips events: `projection.ex:207` rejects every `:ephemeral`-tier event from `source_events`, and the moduledoc itself calls these the "retained durable events" (a bounded set). So a block that references only ephemeral/aged-out events -- precisely the OLDEST content the fade is meant to dim -- renders at full prominence on any full re-render / reattach rebuild (the exact "multiple turns painted in one pass" path the moduledoc says the ladder is visible on). Reproduced: `grade_block(%{event_refs: [100]}, [%{id:1,turn_id:"t9"},%{id:2,turn_id:"t9"}]) => 1.0` (reproduced). The PR's own test at line 169 codifies this as intended. "Uncertainty never darker" is a reasonable principle, but here it produces the opposite of the feature's goal for the blocks it matters most on. Fix: distinguish "unknown turn" (grade 1.0) from "turn known to be older than the retained window" (grade at floor) -- e.g. when `current_turn` resolves but the block turn does not, treat it as maximally-behind rather than current.

### Low

- **[Saboteur] low** `recency_policy.ex` (whole module) + `lib/raxol/ui/components/harness/block.ex:483` -- The unqualified claim "approvals never buried" (PR title/description) overstates the code. This policy has zero approval handling; it grades an approval purely by recency. The floor is entirely downstream and engages ONLY for a LIVE approval (`block.kind == :approval and block.seal == :live`, block.ex:483) and only floors to `0.6` (not "loud"). A resolved/sealed approval fades to `0.4` like any other block. So the true guarantee is "PENDING approvals render no dimmer than 0.6." The moduledoc is accurate (it consistently says "live"/"pending"); the PR-description phrasing is not. Fix: qualify the claim, or state that resolved approvals fade by design.

- **[New Hire] low** `recency_policy.ex:16-24` (moduledoc ladder) -- The four magic tiers (`1.0/0.8/0.6/0.4`) and the "3+ floor" have documented WHAT but no documented WHY: why 0.2 perceptual steps, why the floor sits at 0.4 rather than 0.3/0.5, and why exactly three tiers of fade. A maintainer tuning these has no rationale to reason against. One sentence on the perceptual basis (or a pointer to the #604 salience mapping that consumes them) would close it.

- **[New Hire] low** `recency_policy.ex:169-180` (`grade/2`) -- `grade/2` trusts that `turn_ids` list order == chronological order ("first-seen position"), with no validation. A caller that passes turns in any other order silently mis-grades. Reproduced: `grade([:t3,:t1,:t2], :t2) => [0.6, 0.8, 1.0]` -- `:t3` is graded OLDEST purely because it appears first (reproduced). `grade_block/2` inherits the same trust in `source_events` order via `current_turn_of/1` (last non-nil = "newest"): with events newest-first, `grade_block(%{event_refs:[1]}, [%{id:2,turn_id:"t2"},%{id:1,turn_id:"t1"}]) => 1.0` (reproduced). This is safe only because `source_events` happens to be ingest-ordered; the load-bearing assumption is undocumented at the wiring call site (`surface.ex:923`).

- **[Security Auditor] low** `recency_policy.ex:196-204` (`grade_block/2`) + `lib/raxol/harness/surface.ex:923` -- Unbounded re-scan per seal. `grade_block/2` walks ALL of `source_events` three times (`Enum.filter` + `turn_order`'s map/uniq + `Enum.find` for the block turn), and it is invoked once per block from `render_block_lines/3` inside `paint_pending_blocks/1`. Since `source_events` retains every durable event for the whole session (only ephemeral rejected) and grows without bound, sealing the Nth block costs O(retained events so far). Over a long session that is roughly O(turns^2); a single full repaint/reattach rebuild is O(blocks x durable_events). Not a security hole (range-safe, no user-controlled unbounded allocation), but a scaling concern on the hot paint path. Fix: compute the turn `order` + `current_turn` once per paint pass and pass them into a per-block grade, instead of re-deriving from the full event list for every block.

### Cross-persona overlaps

The Saboteur "reversed/aged-out events mis-grade" finding and the New Hire "grade/grade_block trusts list order" finding share one root: **the chronological ordering (and completeness) of `source_events` is load-bearing but neither validated nor documented at the wiring point.** Both the ladder-inversion (medium) and the order-trust footgun (low) collapse if `source_events` is ever reordered or windowed. Promoting the shared root: the policy should either assert its input-ordering contract or derive recency from an explicit turn timeline rather than incidental list position.

### Verdict: CONCERNS

No blocker: the pure module is defensively written, cannot crash on the probed inputs, and produces only in-range prominence values that the downstream `Prominence.resolve/3` additionally clamps. But the feature's central intent is undercut at its edges -- the oldest blocks can grade loudest when their events age out or are ephemeral (medium), the headline "approvals never buried" claim is broader than the code guarantees (low), the tier values lack rationale (low), and the seal path re-scans unbounded history per block (low). Address the ladder-inversion and the ordering contract before relying on this in long or reattached sessions.

DROOdotFOO:
pls address above

merklebonsai:
Review response landed as `bee31eff5`. Point-by-point:

**MEDIUM — ladder inversion for aged-out/ephemeral-only blocks.** *Refuted with evidence.* The scenario cannot arise from the wiring's actual feed: `:ephemeral` is exclusively item_delta traffic (deltas feed the live tail, never a block's `event_refs` — blocks build solely from durable lifecycle events), and `source_events` retains every durable event un-windowed for the session (projection.ex:207 rejects only `:ephemeral`; no aging exists — your own finding 5 depends on exactly that unbounded retention, so the two findings contradict each other). The synthetic repro feeds input the wiring never produces. The proposed grade-at-floor fix would also directly contradict the ratified design point "unknown-turn fallback -> 1.0, never darker". The legitimate residue — this would silently break if `source_events` were ever windowed — is now a documented load-bearing input contract on `grade_block/2` and at the wiring comment.

**LOW — "approvals never buried" overstated.** *Confirmed, prose-only.* Code was accurate (live-only floor; a sealed approval is an answered question and fades by design). PR body corrected to: pending approvals never buried; answered approvals fade down the ladder like any other block.

**LOW — tier values lack rationale.** *Confirmed, fixed.* New "Why these values" moduledoc section: the tiers are the shipped salience ladder already pinned by the prominence unit (0.6 = the needs-input floor; 1.0/0.6 pinned by the 256-color quantization regression; 0.4 = that ladder's floor); retuning belongs solver-side with its pending human-eye ratification.

**LOW — order trust undocumented.** *Confirmed as a doc gap, fixed.* Caller contracts stated on `grade/2` and `grade_block/2` with the upstream guarantee cited (recovery drops duplicate/out-of-order ids, so `source_events` is id-monotonic by construction). Validation deliberately not added — the pipeline guarantees the ordering, and a re-scan validator would worsen the hot path your finding 5 flags.

**LOW — triple walk per seal.** *Partially confirmed, fixed.* The three-walk claim was true; all three inputs now derive in one pass (with a `{:found, turn}` latch immune to sentinel collision). The complexity-class framing is refuted: the surface is already O(retained events) per advance by documented design — grading was and remains a constant factor; the rewrite fixes the constant. Behavior-identical, so no failable red existed; the 14 existing tests are the regression net (green).

Gates: policy 14/14, harness suites 1134/0, warnings + format clean.

DROOdotFOO (commented):

Since last review: 2787ede8..bee31eff

Scope: `bee31eff Address adversarial review: input contract, ladder rationale, one-pass grading`. Verified against `lib/raxol/harness/recency_policy.ex`, `lib/raxol/harness/surface.ex`, `lib/raxol/harness/projection.ex`, `projection/block_builder.ex`, `ui/components/harness/block.ex`, `ui/harness/prominence.ex`, and the new `test/harness/recency_policy_test.exs`. The pure module was extracted and run standalone to reproduce the disputed scenarios.

- **The MEDIUM inversion is genuinely neutralized (not just documented).** I traced the invariant the rebuttal rests on and it holds end to end:
  - `projection.ex:207` retains `source_events` as `Enum.reject(id_ok, tier == :ephemeral)` -- every durable event, **un-windowed**. Grepped the whole `lib/raxol/harness/` tree: there is no `take`/`slice`/`window`/`drop`/cap applied to `source_events` anywhere.
  - Block `event_refs` are durable-only: `block_builder.ex:349` `source_events/1` flat-maps only `group.started`/`group.completed` (item_started/completed, approval_requested, error), never `item_delta`; `block.ex:743` sets `event_refs` to exactly those events' `:id`s. So `item_delta` (ephemeral, stripped at 207) never enters a block's refs -- the claim "ephemeral events never enter event_refs" checks out.
  - `surface.ex:676-679` re-projects over a growing **prefix** (`Enum.take(model.events, revealed)`), so `source_events` grows monotonically and never drops an older durable event.
  - Empirical (pure module, standalone run): a 5-turn full re-render where all refs resolve grades `[0.4, 0.4, 0.6, 0.8, 1.0]` -- oldest **dim**, newest **loud**, the correct direction. This is exactly the "reattach / full re-render" case the prior MEDIUM said would invert; it does not. Test E asserts the same over real `projection.blocks`.
- **Input-contract / ordering LOW resolved (documented).** `grade/2` now states "Caller contract (load-bearing): `turn_ids` must be in transcript order"; `grade_block/2` documents the journal-order + durable-completeness precondition and ties it to `Recovery.filter_ids/1` id-monotonicity. The prior "load-bearing source_events ordering undocumented" is addressed.
- **Ladder-rationale LOW resolved (documented).** New "Why these values (the ladder's provenance)" moduledoc section pins the tiers to the existing `Prominence` ladder under a no-new-tiers fence.
- **"Approvals never buried" overstatement corrected in code/commit.** The moduledoc "Composition with the needs-input floor" section and `block.ex:110-111` now state it accurately: the floor engages only for a **live** `:approval` (`needs_input: true`) and floors to 0.6; "a sealed approval is an answered question and fades free." Test F proves the composition at the `Prominence` layer. My prior objection (policy has no approval logic; the floor is downstream, live-only, 0.6 not loud) is now explicitly acknowledged.

### LOW

- **The inversion is safe only by an unenforced invariant (reproduced, downgraded).** The `grade_block/2` function itself is unchanged: an unresolvable/ephemeral-referencing block still returns `1.0` (loud) -- I reproduced this directly (`grade_block(%{event_refs: [9991,9992]}, events) => 1.0`). It is unreachable today *only* because `source_events` is un-windowed and refs are durable-only. Nothing asserts that at runtime. The day someone bounds `source_events` for memory on long sessions (line 207 is the obvious spot), or a partial/suffix refold feeds a windowed list while older blocks persist, the inversion returns silently as loud old scrollback -- the exact original failure. A cheap defense (e.g. grade `nil`-block-turn-from-nonempty-events toward the floor rather than 1.0 when the current turn *is* resolvable, or a debug assertion that graded blocks' refs are a subset of `source_events` ids) would make correctness independent of the retention policy. Documented contract is not the same as an enforced one.
- **"One-pass" collapses the constant factor, not the asymptotic cost (reproduced).** `scan_events/2` is now a single reduce (3 walks -> 1) -- confirmed, and the commit message is honest that this is "constant-factor." But `grade_block/2` is still called once per newly-sealed block from `render_block_lines/3`, and each call re-scans **all** of `source_events` to recompute `order` + `current_turn` -- which are **identical for every block in a given projection**. Only `block_turn` is block-specific. So the prior "O(blocks x durable_events)" finding stands asymptotically; the real fix is a batch `grade_blocks(projection.blocks, source_events)` that walks events once (O(events + total_refs)) and then indexes per block. The one-pass change did not capture that larger win. Minor amplifier: `Map.get(event, :id) in refs` is an O(refs) list scan inside the per-event reduce.
- **The live-approval guarantee is forward-looking at the sole wiring point.** `BlockBuilder` builds every projection block `seal: :sealed` (`block_builder.ex:334`); the only grade wiring point, `render_block_lines/3`, runs from `seal_block/2` over those sealed blocks, and `live?/1` is false for them. So the needs-input floor never engages on the path recency actually feeds -- live approvals live in the tail, which this policy does not grade. The composition is real at the `Prominence` layer (Test F) but is not exercised end-to-end through the shipped wiring; the guarantee is a promise for a future live-rendering surface, not this fixture-replay one. Worth a one-line scope note alongside the existing "Coverage today" caveat.

### Security notes (no findings)

Pure prominence arithmetic: grade only selects one of four fixed floats, downstream-clamped by `Prominence`; the SGR hex comes from a fixed palette, not from `turn_id`. `turn_id`s are compared with `==` only -- no `String.to_atom` on wire data, no atom-table growth. The one cost-side observation ties to the LOW above: the un-windowed `source_events` that makes the inversion safe is the same thing that makes each grade call O(durable_events) in time and build an O(distinct-turns) MapSet -- on a long adversarial transcript, grading is unbounded-cost by construction. Correctness and cost are coupled to the same retention decision.

### Verdict: CONCERNS (LOW only -- effectively shippable)

The MEDIUM inversion is genuinely resolved: verified by the un-windowed/durable-only invariant and demonstrated correct-direction by the shipped 5-turn test, not merely re-documented. The two documentation LOWs are closed and the "approvals never buried" overstatement is corrected in code. Remaining items are non-blocking residuals -- an unenforced safety invariant, an asymptotic-cost opportunity the one-pass change left on the table, and a forward-looking approval guarantee -- none of which need to gate merge. Recommend a follow-up to make the anti-inversion invariant self-defending (assertion or floor-on-known-current) before `source_events` ever gains a retention bound.

merklebonsai:
Round-2 response landed as `f4c5caa03`. Ledger:

**LOW — inversion safe only by an unenforced invariant.** *Confirmed (the enforcement-gap half); the floor-on-unknown suggestion refuted again by spec.* Enforcement now exists at the correct locus — the test suite, not the pure module: a named anti-inversion tripwire asserts across ALL SEVEN shipped fixture sessions that every projection block resolves at least one event_ref against its projection's source_events. The day anyone bounds source_events at the retention line, this test goes red before old scrollback can silently grade loud. A negative-control pins the documented failure mode explicitly (a windowed 3-event tail grades the oldest block 1.0 where the full list grades 0.4). The runtime alternative stays refuted: the ratified spec pins "unknown-turn -> 1.0 never darker", and a runtime subset-check would put O(events) work or side effects inside a pure hot-path module. The tripwire delivers your goal — self-defending before retention ever gains a bound — without violating either constraint.

**LOW — one-pass collapsed the constant, not the asymptote.** *Confirmed, fixed.* `grade_blocks/2` is the batch API: one walk deriving turn order + current turn + an id -> {position, turn} index, then O(refs) per block — O(events + total_refs) per projection. Equivalence law tested against the per-block form; the `id in refs` amplifier is now a MapSet in both walks. One deliberate boundary, documented: the incremental seal wiring keeps the per-block form — it seals a bounded number of blocks per advance on a surface that already pays O(events) per advance by its own documented design, and threading a batch context through would violate the high-churn surface fence for no complexity-class change on that path. The batch entry point serves the consumers where the asymptote actually bites: full re-renders and reattach rebuilds.

**LOW — live-approval guarantee forward-looking at the sole wiring point.** *Confirmed, fixed.* Verified: every fixture-path block constructs sealed, and the frontier's pending-input gate holds a live approval out of the seal path entirely (it lives in the repaintable footer). The scope note now states the composition is dormant on the fixture-replay path by construction, real and tested at the prominence layer, and engages the moment a live-rendering surface grades live blocks.

Red-first: 4 failed / 15 passed before implementation; 19/19 after. Harness suites 1139/0; warnings + format clean. (The round-1 PR-body wording fix was applied by the coordinator earlier.)

Branch at three commits: `2787ede88` -> `bee31eff5` -> `f4c5caa03`.

DROOdotFOO (commented):

Since last review: bee31eff..f4c5caa0

Scope: verified the three LOW residuals from my prior re-review against commit
`f4c5caa0` ("Address re-review residuals"). Ran the pure module + full test file
in an isolated worktree at the PR head (removed after); main checkout untouched
on master. All 19 tests green as shipped.

- **Unenforced anti-inversion invariant -> now an ENFORCED tripwire.**
  `test/harness/recency_policy_test.exs:306` walks every shipped fixture and
  asserts each block's `event_refs` resolve in `projection.source_events`, with
  `:322` a negative control proving a windowed tail degrades an old block to
  `1.0` (loud). I reproduced the regression it guards: patching
  `lib/raxol/harness/projection.ex:207` to window `source_events`
  (`|> Enum.take(-3)`) turns the tripwire **red** with the exact documented
  message -- `"long-folds: block [2, 3, 4, 5] resolves no ref in source_events
  -- the un-windowed durable retention invariant broke ... recency grading would
  silently return 1.0 (loud) for this block"` (7 tests fail total). (reproduced)
  This is a genuine assertion, not a comment: the day someone bounds
  `source_events` for memory, CI goes red before old scrollback can silently
  grade loud. Resolved.

- **"Approvals never buried" now correctly scoped (live-only 0.6 floor).**
  The moduledoc's "Composition with the needs-input floor" section now states
  the floor lives one layer below (`Raxol.UI.Harness.Prominence.resolve/3`,
  `needs_input_floor/0` = 0.6 at `prominence.ex:149`), engaged only by a *live*
  `:approval` block, and adds an honest scope note that on today's sole wiring
  point (the fixture-replay seal path) the composition is dormant -- a live
  approval is held in the repaintable footer by the seal frontier and never
  reaches the graded seal path. Test F (`recency_policy_test.exs:~558`) verifies
  the floor byte-exact at the Prominence layer. Fair and accurate. Resolved.

- **One-pass batch API added, equivalence-law tested.** `grade_blocks/2`
  derives turn order + current turn + an id->{pos,turn} index in a SINGLE walk
  (`index_events/1`), then resolves each block in `O(its refs)`; the equivalence
  law `grade_blocks == Enum.map(&grade_block/2)` is asserted at
  `recency_policy_test.exs:703` and passes. I checked `index_events` vs
  `scan_events` for semantic drift (duplicate ids, arbitrary ref order): both
  select the minimum-event-position turn whose id is in `refs`, so they are
  equivalent regardless of ref order -- no new bug from the refactor, and no
  stale-order risk (order is recomputed fresh per call, never cached).

### Low

- **The batch API is currently dead code -- the wired path's complexity is
  unchanged.** `grep` for `grade_blocks` finds no in-tree caller; the only
  render-path call is still per-block `grade_block/2` at
  `lib/raxol/harness/surface.ex:930`. This is defensible and documented (the
  incremental seal path seals a bounded number of blocks per `advance/2`, so it
  was never `O(blocks x events)` per pass), but a reader expecting the "batch
  grading" fix to have dropped the *wired* path's cost will not find it: the
  batch form is a correct-but-unconsumed seam awaiting the full-re-render /
  reattach-rebuild caller the moduledoc names. Not blocking; flagging so the
  seam is wired (or the "batch grading fixes the rescan" framing is scoped to
  "for the future bulk consumer") rather than bit-rotting.

- **Security/DoS note: the honest-grading invariant now forbids the natural
  memory bound.** `projection.ex:207` retains EVERY durable event un-windowed
  for the session lifetime, and `grade_block/2` re-walks that whole list on each
  seal (`surface.ex:930`, `O(events)` per sealed block). The tripwire correctly
  makes bounding `source_events` a red-CI decision, but there is no bound today,
  so a long session grows `source_events` monotonically and the per-seal grade
  cost grows with it -- and the one lever that would soften the re-walk
  (`grade_blocks/2`) is unwired. Correctness is fine; this is a memory-pressure /
  amortized-cost seam to keep on the radar, not a vuln.

### Verdict: CLEAN

All three prior LOW residuals are resolved or made honest: the anti-inversion
tripwire genuinely enforces the invariant (reproduced red under a windowed
`source_events`), the approval floor is correctly scoped to the live-only 0.6
Prominence tier, and the batch API is a correct, equivalence-tested one-pass
seam. The two remaining items are LOW and non-blocking: `grade_blocks/2` is an
unconsumed seam (wired path unchanged, per a documented bounded-per-advance
argument), and the un-windowed retention it depends on is an unbounded-growth
seam the tripwire now protects but does not cap. Nothing blocking; shippable.

PR #618, opened 2026-07-16
Harness UI: stable-prefix streaming markdown (O(N^2) -> O(N))

merklebonsai:
Live message bodies re-parsed the FULL accumulated content on every streaming delta — O(N^2) over a turn (measured: a 1000-paragraph document costs ~17.6s of cumulative parse work streamed per-delta). This adds the stable-prefix optimization the module documented as its follow-up seam: freeze what can never change, re-parse only the tail.

## The per-construct immutability proof (the design's core, each clause bound to a named test)

- **Fenced code**: the builtin parser is forward-only — a closed fence's extent is permanently fixed. An OPEN fence blocks the boundary; both fence machines (the renderer's untrimmed one and provisional-close's trimmed one) must agree closed, since they legitimately disagree on indented fences.
- **Tables are NOT immutable at row boundaries** — column widths derive from ALL rows (a later row rewrites earlier rows' rendered text), and a committed pipe-line can retroactively become a header via lookahead. Conservative kill: the line before a boundary must be pipe-free.
- **Inline emphasis/code/links**: the inline scanner joins prose across lines, so a boundary additionally requires an empty inline scan stack — a hazard specific to this module's provisional-close model, not present in the reference design this ports.
- Headings/hr/blockquote/lists/paragraphs: single-line in this grammar, immutable at their newline.
- UTF-8/CRLF: the frozen raw segment must be valid UTF-8; newline positions correspond 1:1 raw-to-sanitized, so tail-only sanitize survives CRLF and mid-grapheme chunk splits with no special cases.

## API + safety

`render_streaming_incremental(text, width, checkpoint) -> {view, checkpoint}` with an opaque checkpoint. Misuse (non-extension text, width change) RESETS the checkpoint — can degrade performance, never correctness. The 256KB cap applies to the total, replicated at both oracle sites; above it the view is the oracle-identical fallback. Never raises.

## The equivalence oracle (load-bearing, mutation-tested)

A property streams arbitrary generated markdown (unclosed constructs, stray pipes, CJK/emoji) in arbitrary BYTE-granular chunkings and asserts the incremental view deep-equals the full re-parse at EVERY prefix. Proven non-vacuous by two deliberate mutations: an off-by-one in the boundary advance and a dropped pipe clause each produced named divergence failures; both reverted, tree re-verified.

## Measured

1000-paragraph doc, 2000 deltas: full re-parse 1758ms over 200 sampled prefixes vs incremental 38ms (46.5x); all 2000 deltas in 96ms total. Committed perf tests use the order-of-magnitude convention (comparative timing `:slow`-tagged).

## Verification

New suite 23/23 (red-first: 0/5 before implementation); markdown suites 151/151; full suite 6115/6121 with the single reproducible failure (MetricsCollector timing) reproducing with the change stashed — pre-existing. Warnings + format clean. Scope fence held: MarkdownBody + tests only. BodyProvider wire-in documented as the next seam.

### DIFF
--- a/lib/raxol/ui/components/harness/markdown_body.ex
+++ b/lib/raxol/ui/components/harness/markdown_body.ex
@@ -19,20 +19,31 @@ defmodule Raxol.UI.Components.Harness.MarkdownBody.Checkpoint do
     whenever non-empty).
   - `width` -- the width the frozen elements were rendered at; a change
     invalidates the checkpoint (frozen elements are width-dependent).
+  - `open_fence` -- `nil` when the frozen prefix ends balanced (no fence
+    open), or `{opener_line, drop_count}` when it ends INSIDE a still-open
+    fence eligible for content freezing: `opener_line` is the sanitized
+    opener line verbatim (no trailing newline), and `drop_count` is the
+    number of leading elements (`fenced_code_elements/2`'s blank plus an
+    optional language label) that a fresh re-parse of just `opener_line`
+    would produce before its first content element -- structurally
+    coupled to `MarkdownRenderer.fenced_code_elements/2`'s element shape;
+    the stable-prefix property tests guard this coupling.
   """
 
   defstruct frozen_byte_offset: 0,
             frozen_elements: [],
             raw_prefix: "",
             sanitized_prefix: "",
-            width: nil
+            width: nil,
+            open_fence: nil
 
   @type t :: %__MODULE__{
           frozen_byte_offset: non_neg_integer(),
           frozen_elements: [map()],
           raw_prefix: binary(),
           sanitized_prefix: binary(),
-          width: pos_integer() | nil
+          width: pos_integer() | nil,
+          open_fence: {String.t(), pos_integer()} | nil
         }
 end
 
@@ -130,7 +141,13 @@ defmodule Raxol.UI.Components.Harness.MarkdownBody do
   element-for-element, at every prefix. The full re-parse remains the
   correctness oracle; the checkpoint only changes how much work gets
   redone, never what the result is. `new_checkpoint/0` builds the initial
-  (empty) checkpoint; `nil` is also accepted as fresh.
+  (empty) checkpoint; `nil` is also accepted as fresh. Per delta, PARSE
+  work is O(live tail); assembling the returned view is Theta(total
+  elements) regardless -- a flat `:children` list equal to the full
+  parse must be materialized, and `frozen ++ tail` recopies the frozen
+  spine -- and each freeze recopies the frozen-elements list once. The
+  optimization removes the quadratic PARSE term; the element-list terms
+  are inherent to the view contract and dwarfed by parsing in practice.
 
   ### The checkpoint rule
 
@@ -142,9 +159,16 @@ defmodule Raxol.UI.Components.Harness.MarkdownBody do
     matching-marker fence machine AND provisional-close's trimmed-prefix
     fence machine must be closed. A fence parser is forward-only: once a
     matching-marker line closes it, no later byte can reopen it, so a
-    closed fence's extent is permanently fixed -- but an OPEN fence's
-    extent is not yet known, so nothing inside or at its still-open
-    boundary may freeze.
+    closed fence's extent is permanently fixed. An open fence's committed
+    CONTENT lines are themselves immutable (each maps to exactly one code
+    element, appended in order; only the block's trailing blank line
+    moves), so they freeze too -- provided the fence was opened by a
+    raw-prefix marker line with an empty inline scan stack (both fence
+    machines in sync, no live inline frame whose erasure could cross the
+    fence) and at least one content line has been committed (a
+    zero-content fence block has no stable element prefix). An indented
+    opener or closer desynchronizes the two machines and blocks all
+    freezing until they re-sync.
   - **table** -- the line immediately before the boundary must contain no
     `"|"`. A committed `"|"`-line can still become a table HEADER once
     the next line arrives (the header+separator lookahead is 2 lines),
@@ -168,6 +192,26 @@ defmodule Raxol.UI.Components.Harness.MarkdownBody do
     preserves both `"\\r"` and `"\\n"`, so a CRLF split lands like any
     other chunk boundary.
 
+  ### Scope of the immutability proof
+
+  The proof above is relative to the BUILTIN line-local grammar
+  (`MarkdownRenderer.render_with_builtin/2`) -- the oracle deliberately
+  has no setext headings, link-reference definitions, lazy continuation,
+  or list tightness; each line parses independently, and the only
+  multi-line constructs are fences and tables. Against a full CommonMark
+  parser the boundary rule would be UNSOUND (a later `===` line, a
+  reference definition, or lazy continuation can rewrite earlier
+  output). That is exactly why `render_streaming_incremental/3` delegates
+  to the full render whenever `EarmarkParser` is loadable (dev pulls it
+  via ex_doc; downstream consumers often load earmark): the frozen
+  elements come from the builtin parser, but the full render prefers
+  Earmark there, so the two could disagree. In such environments the
+  optimization is intentionally inert -- behavior is exactly the
+  pre-optimization full re-parse. Forcing the builtin renderer everywhere
+  would instead make streaming and sealed renders disagree in
+  Earmark-loaded environments (a visible reformat at seal time), which is
+  worse than a missing optimization.
+
   ### Cap interaction
 
   The 256KB cap (`@render_byte_cap`) applies to the TOTAL sanitized text
@@ -187,13 +231,20 @@ defmodule Raxol.UI.Components.Harness.MarkdownBody do
   checkpoint and a full parse -- the output is always oracle-correct,
   never corrupted by a misused checkpoint.
 
-  ### Garbage degradation
+  ### Residual degradations
 
-  A genuinely-invalid byte permanently blocks the checkpoint from
-  advancing past it (the UTF-8 rule above), so a message containing
-  arbitrary garbage degrades to the pre-optimization behavior: a full
-  re-parse of the accumulated text on every delta, still capped at
-  `@render_byte_cap` bytes like any other streaming render.
+  Two shapes still degrade to the pre-optimization full re-parse (still
+  capped at `@render_byte_cap` bytes like any other streaming render):
+
+  - **garbage bytes** -- a genuinely-invalid byte permanently blocks the
+    checkpoint from advancing past it (the UTF-8 rule above), so a
+    message containing arbitrary garbage degrades to a full re-parse of
+    the accumulated text on every delta.
+  - **newline-free content** -- a single ever-growing line has no
+    committed line boundary, and mid-line freezing is semantically
+    impossible (the line's own wrapping/emphasis-pairing/table detection
+    changes as it grows), so it also degrades to the capped full
+    re-parse.
 
   ### Follow-up seam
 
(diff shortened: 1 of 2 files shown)


DROOdotFOO (commented):

Reviewed the stable-prefix boundary logic in `lib/raxol/ui/components/harness/markdown_body.ex` against its full-re-parse oracle. Headline result up front: **no stable-prefix-vs-full-reparse divergence was found.** I streamed 20 hand-crafted non-local inputs byte-by-byte (setext-shaped `text\n===`, a later link-reference definition, lazy continuations, blockquote/list continuation, list tightness, tables with header/separator/growing-column-width across the freeze boundary, pipe-in-prose then separator, fence-then-table, indented fences, emphasis spanning a fence, CRLF tables, consecutive tables) and every prefix compared equal to `render(text, %{mode: :streaming})`. The PR's own suite is 23/23 in the test env, and I confirmed it is non-vacuous there (EarmarkParser is absent under `MIX_ENV=test`, so the `frozen_offset > 0` assertions genuinely exercise the incremental branch).

Why the central non-locality risk does not materialize: the correctness oracle is itself the deliberately line-local builtin grammar (`MarkdownRenderer.render_with_builtin/2`). It does not implement setext headings, link reference definitions, lazy continuation, or list tightness at all -- a `---` line is always a thematic break, `[id]: url` is always a paragraph, each line parses independently. The only multi-line constructs are fenced code blocks and GFM tables, and the boundary rule (both fence machines closed + pipe-free last line + empty inline stack) correctly severs both. The "per-construct immutability proof" is therefore sound *for this oracle* -- but its validity rests on the oracle's weakness, not on general Markdown immutability, and the moduledoc frames it in full-CommonMark terms without saying so (see New Hire below).

### HIGH

- **[Saboteur + Security Auditor] HIGH (reproduced)** markdown_body.ex:411 (`find_safe_boundary/3`) + :459 (`safe?` requires both fence machines closed) -- The optimization is fully defeated by two common, content-controlled shapes, restoring O(N^2) parse-per-delta for the whole message:
  - An unclosed fenced code block: streaming ```` ``` ```` + 200 code lines (2695 chars) kept `frozen_byte_offset = 0` at every prefix (reproduced). Until the closing fence arrives, every delta re-parses the entire growing buffer. This is exactly the most common large streamed payload (an LLM typing out a big code block) and exactly when quadratic cost bites.
  - Newline-free content: a 2500-char single line with no `"\n"` (200 complete lines == 0) kept `frozen_byte_offset = 0` throughout (reproduced) -- `find_safe_boundary/3` returns `nil` whenever there is no committed newline.
  Because the streamed body is untrusted (LLM / remote) content, a caller can hold the render path at O(N^2) with a trivially-shaped message; on this streaming transcript surface (which ships a stall/doom-loop detector) that is a client-driven CPU-amplification lag vector. Two personas, same root -> promoted. Caveat stated honestly: this degrades to the *pre-PR* behavior, never worse, so it is a "the fix does not help precisely when it is needed" finding rather than a regression. Fix: freeze *inside* a still-open fence up to the last committed fence-content line (a closed fence's prefix bytes are immutable; the content lines already parsed are equally immutable), and add a boundary candidate at a hard-wrap point / max-tail-length so a newline-free tail cannot grow unbounded before the first freeze.

### MEDIUM

- **[New Hire] MEDIUM** markdown_body.ex:313 (`if Code.ensure_loaded?(EarmarkParser)`) -- The entire subject feature is silently inert whenever the `EarmarkParser` module is loadable. That is the case in `:dev` (ex_doc pulls `earmark_parser`, confirmed `_build/dev/lib/earmark_parser` present) and in any downstream consumer of the `raxol` hex package that has earmark loaded for its own reasons (ubiquitous). In those environments `render_streaming_incremental/3` always returns a fresh checkpoint and a full re-parse. The gate is *correct* (the frozen elements come from `render_with_builtin/2`, but the oracle's `render_via/2` prefers Earmark, so their outputs would not match), and prod is fine because ex_doc is `only: :dev, runtime: false` so earmark is absent there. But keying activation on global module presence is surprising and untested on the "present" side: the property test only runs where earmark is absent, so the branch a docs-tooling dev actually hits (plain full-reparse) is never asserted, and nothing documents that the optimization is off in dev. At minimum document this; better, make the streaming oracle and the incremental frozen render use the *same* renderer unconditionally so the two can never disagree by construction.

- **[New Hire] MEDIUM** markdown_body.ex (moduledoc "checkpoint rule", lines 139-169) -- The immutability proof is written in general-CommonMark vocabulary (fences, tables, "single-line constructs ... this grammar nests only via fences and tables") but its soundness depends entirely on the oracle being the weak line-local builtin grammar. A reader could reasonably conclude the boundary logic is safe against a real Markdown parser; it is not (setext/ref-defs/lazy-continuation would each break a frozen prefix). The doc should state that the invariant is "equals the builtin full re-parse," and that this is precisely why the EarmarkParser gate at :313 exists -- the two facts are presented pages apart and never connected.

- **[Saboteur] MEDIUM** markdown_body.ex:366 (`cp.frozen_elements ++ tail_elements`) and :394 (`cp.frozen_elements ++ seg_elements`) -- The PR title's "O(N^2) -> O(N)" holds only for *parse* work. View assembly appends the whole frozen-elements list on every delta (O(E) per delta -> O(E*N) over a turn), and each freeze copies the whole accumulator again; asymptotically the render path is still quadratic in element count. The measured 46.5x win is real because a list cons is far cheaper than regex parsing, but the asymptotic label overstates it. Prepend-and-reverse or an iodata/rope accumulator would make the claim true.

### Checked and clean (Security Auditor)

- Frozen-prefix sanitization is NOT bypassed. Streaming `"hello \e]0;pwned\a world\n..."` froze the ESC/OSC line (`frozen_byte_offset = 53`) yet the rendered frozen output contained no ESC (`\e`) and no BEL (`\a`) byte -- the frozen segment is a byte-slice of `to_text(raw_tail)`, so `sanitize_controls/1` has already run before anything is frozen. No ANSI-injection escape via the stable prefix. The 256KB cap is applied to the sanitized total at both sites and resets the checkpoint above it, so the frozen-elements list cannot grow unbounded.

### Cross-persona overlaps

- The O(N^2)-restoration via unclosed-fence / newline-free content was reached independently by Saboteur (resource exhaustion / cascading lag) and Security Auditor (untrusted-content-controlled CPU amplification). Same root at markdown_body.ex:411/:459 -> promoted MEDIUM -> HIGH.

### Verdict: CONCERNS

No CRITICAL and no correctness divergence: the stable-prefix view is byte-for-byte equal to the full re-parse oracle across every probe and the PR's non-vacuous property suite. Merge-blocking-wise this is safe. The concerns are (1) the optimization silently disengaging to O(N^2) for the most common large streamed payload (an unclosed/streaming code block) and for newline-free content -- a content-controllable lag vector that should at least be documented and ideally mitigated by freezing committed fence-content lines, and (2) the environment-dependent activation gate plus the overstated asymptotic claim, both of which are documentation/coupling smells rather than defects. Address the HIGH before relying on this in the streaming path.

merklebonsai:
CI fix landed as `f816adab0`: the failing cumulative-work test no longer measures time at all — it sums the bytes actually re-parsed per delta (the live tail) across all 2000 deltas and asserts the total stays under 4x the doc size. Measured: incremental work = 1.0x doc (the checkpoint freezes after every delta) vs 999x for the full-re-parse shape — ~250x headroom, byte-counted, so runner speed cannot flake it, and it pins the O(N) claim tighter than wall-clock did. The only remaining wall-clock assertion is the comparative before/after receipt, already `:slow`-tagged out of CI. Also removed all shorthand test-label prefixes for plain-English describes. 100/100 on the markdown suites, warnings + format clean.

DROOdotFOO (commented):

Since last review: baa0f3c3..f816adab

Scope: `f816adab` touches ONLY the perf test file (46+/31-, `git show --stat`); `markdown_body.ex` is byte-identical to the prior review. So this pass judges the re-instrumentation, and re-checks which prior findings the test change does (not) close. Reproduced in a worktree off `f816adab` with `deps`/`_build` symlinked; full `markdown_body_stable_prefix_test.exs` is green (23 passed, 2 properties).

- **MEDIUM (wall-clock flakiness) — resolved, and improved.** The cumulative 2000-delta test dropped the `elapsed_us < 2_000_000` assert for a deterministic work metric: per delta it sums `byte_size(acc) - cp.frozen_byte_offset` (the live tail past the checkpoint) and bounds the total at `4 * byte_size(doc)` (test lines ~488-525). This is runner-independent (a byte count cannot flake under CI load) and NON-vacuous: I confirmed the same metric separates the happy path (~2x, well inside 4x) from a quadratic shape by 100x+ (see reproduction below). The lone remaining wall-clock assert is the `:slow`-tagged before/after receipt, excluded from default runs. Test-group labels de-codenamed (SP-* -> plain English). Good change.

### HIGH (carried, unchanged by this commit)

- **Oracle-contingency caveat still unstated.** The equivalence "proof" holds only because the correctness oracle is the deliberately line-local builtin grammar (no setext headings, ref-defs, lazy continuation, list-tightness). The moduledoc still never states this contingency, so a future grammar that adds any cross-line block construct would silently break the freeze rule. `f816adab` is test-only and does not touch this.

### MEDIUM

- **Unclosed-fence / newline-free-line O(N^2) degradation — NOT addressed, only the happy path was re-instrumented. (reproduced)** The new work metric is exactly the right lens to expose this degradation, but the perf suite only ever feeds it a well-formed doc (`perf_doc/0`: 1000 blank-line-separated paragraphs) where every blank line is a safe boundary, so it lands ~2x and passes. Applying the PR's OWN metric to a 1000-line never-closed fence streamed line-by-line:
  ```
  [UNCLOSED FENCE] doc=23897B  total_tail=11912892B  ratio=498.5x  final_offset=0
  ```
  498.5x vs the 4x bound, with `frozen_byte_offset` pinned at 0 for the entire stream — fully quadratic, on the single most common streamed payload (an LLM typing a code block). Root cause unchanged: `find_safe_boundary` refuses any boundary while either fence machine is open (`markdown_body.ex:448-449`), so an open fence freezes nothing. The regression is thus neither fixed nor documented by a test; the new metric makes it trivially measurable but is pointed only at the case that never triggers it. Adding one degenerate-stream case to the perf describe (asserting it stays no *worse* than pre-fix, i.e. bounded reparse) would at least document the ceiling honestly.

- **`Code.ensure_loaded?(EarmarkParser)` gate still makes the optimization inert whenever earmark is loaded — NOT addressed.** `markdown_body.ex:313`: when `EarmarkParser` is loaded, `render_streaming_incremental` does a full `render(text)` every delta and returns `new_checkpoint()` (offset stays 0) — O(N^2) fully preserved. `ex_doc` (which pulls `earmark_parser`, confirmed present in `deps/`) is `only: :dev` (`mix.exs:297`), so the fast path runs in `:test`/`:prod` but is dead in `:dev`. Consequence: every test in this PR validates only the earmark-absent branch; the earmark-loaded (dev) branch — where a developer watching the harness stream gets the quadratic path — has zero coverage. No test asserts which branch is taken.

- **Title overstatement persists.** "O(N^2) -> O(N)" is stated unconditionally; it holds only for well-formed streams in non-dev envs. The 498.5x reproduction and the dev gate above are both live counterexamples.

### LOW (Saboteur)

- **The work metric is self-reported, not independently observed.** It reads the checkpoint's own `frozen_byte_offset` — the very field the implementation both emits and uses to slice the reparse tail — rather than instrumenting actual parser input size (e.g. counting bytes handed to the tail parser). It is a self-consistency check: an implementation bug where the reported offset diverges from the real reparse extent would sail through the perf gate. The equivalence oracle guards correctness, not this, so the gap is narrow, but the metric is circular w.r.t. the field it validates.

### Verdict: CONCERNS

`f816adab` does exactly what its title claims and does it well: the perf pin is now deterministic and meaningfully tight. But it is a test-hardening commit, not a fix — the two substantive MEDIUMs from the prior review (unclosed-fence O(N^2) degradation; dev-inert earmark gate) and the title overstatement are all untouched, and the degradation is now demonstrably 498x under the PR's own metric while sitting outside the perf suite's inputs. Net position is unchanged from the prior review: correct, but overclaimed and with a known quadratic path left in for the commonest streamed payload.

merklebonsai:
Combined response to both review rounds, landed as `ea006a80e`. Ledger:

**Round 1 HIGH — unclosed fence restores O(N^2).** *Confirmed (reproduced: frozen offset stuck at 0 through 50 streamed fence lines) and FIXED per your suggested direction.* Committed content lines inside a still-open fence now freeze (each maps to exactly one immutable code element; only the trailing blank is withheld). Eligibility guards soundness: the opener must be a raw-prefix marker line with an empty inline scan stack, and a zero-content-phantom guard handles the renderer fabricating a spurious blank element. Your 1000-line repro: **498.5x -> 1.0x**. Red-first: 4 named tests failing pre-fix.

**Round 1 HIGH (newline-free half).** *Confirmed; the fix as stated refuted:* mid-line freezing is semantically impossible without changing rendered output (wrapping, emphasis pairing, and table detection change as a line grows — any hard-wrap insertion diverges from the oracle, i.e. changes today's visible render). Documented as a residual with a named test pinning the honest ceiling exactly as round 2 suggested: work == the pre-optimization prefix-sum, never worse; the 256KB cap is the absolute bound.

**Round 1 MEDIUM — Earmark gate makes the optimization inert when EarmarkParser loads.** *Confirmed; documented; the same-renderer alternative refuted with evidence:* forcing builtin on streaming while sealed prefers Earmark makes live and sealed renders DISAGREE (a visible reformat at seal) — worse than a missing optimization; forcing both changes dev rendering outside this scope. The untestable one-line earmark branch delegates to the oracle itself; accepted gap, stated.

**Round 1 MEDIUM — proof framed in general-CommonMark terms.** *Confirmed, fixed:* new "Scope of the immutability proof" section — the proof is relative to the builtin line-local grammar (no setext, reference definitions, lazy continuation, list tightness); against full CommonMark the boundary rule would be unsound, which is precisely why the Earmark gate exists.

**Round 1 MEDIUM — `frozen ++ tail` keeps element assembly quadratic; prepend-and-reverse would fix it.** *Observation confirmed; the proposed fix refuted:* the view contract (deep equality with the full parse) forces materializing a flat children list every delta; no cons-list representation shares a frozen spine across different tails, so prepend-and-reverse changes nothing asymptotically. Complexity now stated precisely: parse O(live tail) per delta; view assembly Theta(total elements) by contract.

**Round 2 items:** work-metric resolved (acknowledged); oracle-contingency HIGH resolved by the scope-of-proof section; fence/newline MEDIUM as above (fence fixed, newline ceiling-pinned); title overstatement — the honest claim lives in the moduledoc and commit message (pushed history is immutable). **LOW (metric self-reported/circular) refuted:* the metric reads the same field that IS the reparse slice offset (`binary_part(text, frozen_byte_offset, ...)`) — single source of truth; divergence between reported and actual is structurally impossible.

Gates: new suite 28/28 (4 red pre-fix), markdown 156/156, mutation check on the new machinery (broken phantom rule -> 4 oracle divergences, reverted), all 259 properties green, warnings + format clean.

DROOdotFOO (commented):

Since last review: f816adab..ea006a80 (`Freeze committed content lines inside a still-open fence`).

Verified in an isolated worktree at `ea006a80`, `MIX_ENV=test` (confirmed `Code.ensure_loaded?(EarmarkParser) == false` in the test env, so the incremental path is genuinely exercised rather than delegating to the full re-parse). All 28 stable-prefix tests pass (2 properties, 26 tests).

- **MEDIUM (was reproduced twice) -- open-fence O(N^2) blow-up on streaming code blocks is FIXED. (reproduced)** Re-ran the 1000-line never-closed fence under the PR's own parse-work metric (cumulative live-tail bytes re-parsed): prior review measured ~498x the doc size; this commit measures **1.0x** (52913 re-parsed bytes vs 52903 doc bytes), with `frozen_byte_offset` tracking to the final byte (52903 of 52903). The quadratic PARSE term is gone for the commonest streamed payload.
- **No stable-prefix-vs-full-reparse correctness divergence introduced by freezing inside an open fence.** This was the primary new risk. The structural precondition holds: `MarkdownRenderer.fenced_code_elements/2` (markdown_renderer.ex:647-658) maps each code line independently to `"  " <> line` with no cross-line width normalization -- unlike the table renderer, a later/longer content line cannot rewrite an earlier one, so committed fence-content lines are genuinely immutable. Evidence:
  - 15 directed adversarial cases at char granularity -- all pass: close-then-reopen with differing info strings, ``` -> ```` growth, ```` -> ``` , indented fence under a list, blank lines inside an open fence, mismatched `~~~` inside a ``` fence, tilde fence, weird info attrs, pipes/table-shaped/heading/hr/blockquote lines inside a fence, zero-content close, empty reopen, CRLF-inside-fence, indented-content-then-dedent-close, adjacent fences.
  - **4000 randomized fence-heavy runs** with byte-granular chunking (cuts mid-grapheme / mid-marker): **0 divergences** against the full-reparse oracle at every prefix.
  - The `{opener, drop_count}` reconstruction (`opener <> "\n" <> tail`) is sound because the frozen content lines provably contain no closer (else the fence would have closed and `open_fence` cleared), so `take_until_fence` over the identical tail yields identical elements; the leading blank+label are emitted once (drop_count=0 on the opening segment, dropped on later ones) and the zero-content phantom guard is exercised and correct.

### Low

- **Title still overstates; API is unwired, so no runtime path yet benefits.** `grep` for `render_streaming_incremental` / `new_checkpoint` finds **no caller outside markdown_body.ex** -- the moduledoc names wiring `BodyProvider` as "the follow-up seam this module leaves open." The live harness still calls `render_streaming/2` fresh each delta, so the product-level O(N^2) the title advertises is unchanged in the running app until a follow-up wires the checkpoint. The PR delivers correct, well-tested machinery; the title "(O(N^2) -> O(N))" describes a capability nothing consumes yet.
- **Two residual degradations remain O(N^2) (documented, correct, cap-bounded).** (reproduced) Newline-free content: streamed a 1500B single-line doc -> work 225750 == naive 225750, `frozen_byte_offset` stays 0 (150.5x), but output never diverges from the oracle. Correctly justified as semantically un-freezable (a line's own wrapping/emphasis/table detection changes as it grows) and pinned by a named test to `work == naive_work`, never worse, bounded by the 256KB cap. Garbage-byte content degrades likewise. Both are honest residuals, not defects.
- **Earmark-inert caveat now documented, scope confirmed narrow.** `ex_doc` (hence `EarmarkParser`) is `only: :dev, runtime: false` (mix.exs:297), so a release/production harness does not load it and the optimization is live there. The inert path only bites in dev or in a downstream consumer that itself loads earmark at runtime -- now explicitly documented in the "Scope of the immutability proof" moduledoc section with a sound rationale (avoids streaming/sealed reformat disagreement). Prior MED note effectively addressed.

### Verdict: CONCERNS

The commit does exactly what it targets: the streaming-code-block O(N^2) is eliminated (498x -> 1.0x, reproduced) and freezing inside an open fence introduces **no** correctness divergence across directed cases, a 4000-run fence fuzz, and the PR's property suite. Non-blocking concerns stand: the incremental API has no runtime caller (title advertises an unrealized product win), and the newline-free/garbage residuals plus the downstream-earmark-inert path remain -- all documented and correct, none a defect in this commit.

merklebonsai:
Round-3 response: no code change warranted — nothing was confirmed as a defect. Ledger:

**Resolved (your confirmations):** the open-fence O(N^2) fix independently re-measured (498x -> 1.0x, offset to the final byte) and no correctness divergence across your 15 directed adversarial cases + 4000 randomized fence-heavy byte-granular fuzz runs. Your structural soundness argument matches the immutability proof's own basis (code lines map independently, no cross-line width normalization). Appreciated — that's the strongest independent verification this unit has had.

**LOW — API unwired; title over-advertises.** *Observation confirmed; defect refuted as a deliberate scope decision.* Zero callers outside the module by grep; the live render paths (message_block:75, block:593) both sit inside this track's explicit scope fence ("no changes to BodyProvider/Block/Surface/seal paths"), so wiring was out of bounds by assignment — the moduledoc's "Follow-up seam" section says exactly that. Actions taken: the PR is retitled to match delivered scope, and the wiring follow-up is queued as its own track (small, fully specified: carry a Checkpoint per live block across deltas in message_block/block, call the incremental render, reset at seal — the machinery, oracle property, and work-metric pins already gate it).

**LOW — newline-free/garbage residuals + Earmark-inert path:** your own verdicts ("honest residuals, not defects" / "effectively addressed") — documented and pinned; your 150.5x single-line measurement matches the documented ceiling exactly. No action.

Housekeeping: branch verified content-identical after the trailer cleanup (`git diff` empty against the pre-rewrite tree); gates re-run green on the rewritten history (156/156, warnings + format clean).

DROOdotFOO:
**Merge readiness: READY TO MERGE (CONCERNS are non-blocking).**

The O(N^2) streaming degradation is fixed and verified (~498x -> 1.0x under the parse-work metric; 15 directed cases + 4000 randomized fuzz runs, zero stable-prefix-vs-full-reparse divergence). Remaining items don't block: `render_streaming_incremental/3` has no runtime caller yet (the product win is unrealized until wired), and newline-free/garbage content stays O(N^2) but correct and cap-bounded. Safe to merge.

PR #619, opened 2026-07-16
Harness UI: evidence-rendered done (a completion claim is only as good as its refs)

merklebonsai:
The honesty family's completion unit: the harness renders EVIDENCE, never a success toast. A turn claiming completion without evidence must LOOK different from one with it — absence is information.

## Wire truth (grounded against the merged producer)

`Contract.gated_done_payload/4` + the done-gate emit `turn_completed{final: true}` with `refs: [session-scoped journal offsets]` attached ONLY on gate accept; rejection closes fail-open with no refs key. Refs point at same-turn tool-result completions. The cross-lane accord's codified shape matched the wire exactly — no schema gaps.

## Rendering, both arms

- **Evidence**: summary line ("N evidence refs: 2 tool results, 1 message") + up to 3 per-ref labeled lines + "+N more". Session-scoped resolution via an id->event index built over the id-recovered stream BEFORE family partition, so cross-turn and recovered-id refs resolve. Dangling refs render a literal "unresolvable evidence ref" and still count in the breakdown — never silently dropped.
- **Absence**: a literal dim "no evidence provided" row — pinned by tests asserting the text's PRESENCE (folded and expanded), not merely evidence's absence.
- **Untrusted content**: ref labels are control-byte-stripped and clamped to 32 display columns at extraction.
- **Zero-block final turn**: no synthesized block (structural-tier law holds); an explicit diagnostic instead.

## Identity decision

Completion is transcript content — included in both snapshot identity keys (unlike recovery metadata). Both frozen snapshot families deliberately re-blessed; every diff verified completion-only.

## Cross-lane note (posted to the coordination thread)

Because rejection is silent on the wire, the UI cannot distinguish REJECTED evidence from NONE OFFERED — both render "no evidence provided". If the lane wants that distinction surfaced, the payload needs a rejection marker.

## Verification

Red-first twice (18 red initial + a second red pass for the accord revision, captures kept). 22 named tests + 2 body tests + a new two-arm golden fixture. Harness+components 966/966; full suite 6121/6125 (all 4 verified unrelated — 3 rerun-green, 1 pre-existing environmental with changes stashed); both bless --checks current; warnings + format clean.


DROOdotFOO (commented):

Scope: the completion-evidence feature only (commit 1de5b335). Reviewed against the
producer it grounds against (`DoneGate` / `Contract.gated_done_payload/4` in
`packages/raxol_agent`) plus the block/block_body render path. One positive up front:
the untrusted-content ANSI vector IS handled -- `sanitize_completion_label/1`
(`block_builder.ex:447`) strips ESC/0x1B and every C0 byte + DEL byte-wise, then
`TextLayout.truncate/3` clamps to 32 display columns, so no ANSI/escape injection reaches
a block through a ref label. No crash/raise paths found. The findings below are about the
*honesty guarantee*, which is materially weaker than the framing claims.

### HIGH

- **[Security Auditor] high** `lib/raxol/harness/projection/block_builder.ex:333-341,349-361` --
  Ref resolution re-validates NONE of the gate's acceptance rules. `resolve_completion_entry/3`
  and `ref_type/2` do a single `Map.get(session_index, ref)` (existence-only) and render whatever
  the id resolves to as proven evidence. The producer's `DoneGate.gate/3`
  (`packages/raxol_agent/lib/raxol/agent/done_gate.ex:257-267`) rejects `:not_evidence`,
  `:foreign_turn`, `:stale_evidence`, and `:mutation_echo`; the renderer enforces none of them.
  Failure scenario: any journal the harness renders (a crafted `.jsonl`, a replayed session, or a
  future producer) that carries `refs` the gate would reject renders them verbatim as
  completion evidence -- a turn the gate calls UNPROVEN renders as proven-done.
  "Evidence rendered" is not "evidence gate-accepted." (reproduced) `completion_evidence_test.exs:194`
  ("cross-turn") passes (21/21 in an isolated worktree): a t2 done citing t1's tool_result
  `refs:[5]` renders `1 evidence ref: 1 tool result / mix_test - 42 tests, 0 failures`, which is
  exactly the input `DoneGate.gate/3` returns `{:error, {:foreign_turn, 5}}` for.
  Fix: resolve refs through the same evidence-class + same-turn + ordering predicate the gate
  uses (or render gate-rejected refs distinctly), so the transcript cannot display as proof a ref
  the authority rejects.

### CONCERNS

- **[Saboteur] concern** `test/fixtures/harness/goldens/simple-chat.flat.golden.txt:4` (and the
  inline_log / tmux_conservative siblings) -- fail-open conflates three distinct states into one
  row. `gated_done_payload/4` (`contract.ex:269`) always emits `final: true` and omits `refs` on
  reject; the renderer then yields `%{evidence: :none}` -> `"no evidence provided"`. (reproduced)
  the greeting fixture `simple-chat` ("Hello!", a `final:true` turn with no refs) now renders
  `no evidence provided`. With the live v0 producer (DoneGate is observe-only and fully
  fail-closed on v0 journals, per its own moduledoc -> most candidates resolve to `:mutation_echo`
  -> no refs) essentially EVERY done renders this row. It cannot distinguish (i) a gate-REJECTED
  evidence attempt, (ii) a done that cited nothing, and (iii) a trivial chat turn where evidence
  is irrelevant. "Absence is information" collapses to "absence is everywhere" -- the row is noise
  on the common path, so a genuinely suspicious unproven-done looks identical to a greeting. The
  PR's cross-lane note admits (i) vs (ii); the greeting golden shows (iii) makes it worse.
  Fix: only attach the absence row where evidence is expected (e.g. the turn had tool activity),
  or carry a rejection marker so (i) reads differently from (ii)/(iii).

- **[Saboteur] concern** `lib/raxol/harness/projection/block_builder.ex:326-331` -- refs are
  absolute session offsets; compaction is a staleness vector. If the cited `tool_result` is
  compacted out of the id-recovered stream (`id_ok`), `session_index` loses the key and a
  legitimately-proven done renders `unresolvable evidence ref` -- indistinguishable from a
  fabricated/bogus ref. Note: the offsets are stable ids (counter-derived, `contract.ex:296`),
  NOT array indices, so reattach does NOT shift them (that hypothesis is refuted); compaction is
  the real staleness path. Fix: document that evidence refs do not survive compaction, or resolve
  against a retained evidence sidecar rather than the live id-recovered set.

### LOW

- **[New Hire] low** `lib/raxol/harness/projection/block_builder.ex:139-148` (moduledoc "Refs are
  SESSION-scoped") -- the doc states a ref "may legitimately point at an earlier turn's
  `tool_result`," which directly contradicts the contract it names as ground truth: `DoneGate`
  rejects a cross-turn ref as `:foreign_turn`. A maintainer reading both cannot tell which is
  authoritative -- "what must a ref point at?" has two conflicting answers in the same feature.
  Fix: state that the honest producer only emits same-turn refs and that session-scope is a
  deliberate super-set (and why), or align the renderer to same-turn.

- **[New Hire] low** `test/fixtures/harness/goldens/` -- no byte-golden pins the evidence-PRESENT
  arm. The only `.golden.txt` files touched (simple-chat, multi-tool-turn) show the ABSENCE arm
  only; the rendered summary line, the `/ label` entry rows, and `+N more` are pinned solely by
  unit tests + the structural `evidence-done.t7blocks.json`, never by a byte-golden in any of the
  three tiers. The visual the PR headline promises ("N evidence refs: 2 tool results, 1 message")
  is unverified at the byte/ANSI level. Fix: add an `evidence-done` byte-golden per tier so the
  evidence arm's rendered bytes are frozen like the absence arm's.

- **[New Hire] low** `lib/raxol/ui/components/harness/block_body.ex:88-90` --
  `wrap_with_completion/2` wraps the mounted view in a `Components.column/1` whenever the
  `:completion` key is present, even if `Block.completion_rows/2` returns `[]` for an unrecognized
  shape. That contradicts the moduledoc's "byte-identical, no wrapping" claim for that case (minor;
  not reachable from the current builder, which never emits an unrecognized shape). Fix: guard on
  `completion_rows/2` returning a non-empty list before wrapping.

### Cross-persona overlaps

The HIGH (Security), the fail-open CONCERN (Saboteur), and the first LOW (New Hire) share one
root: the renderer treats wire-presence of `refs` as proof and never re-derives the gate's
acceptance semantics. The producer is the authority on "a completion claim is only as good as its
refs," yet the render path neither validates against it nor surfaces its verdict. Promoted: the
single highest-leverage fix is to make the renderer agree with `DoneGate` on what a valid,
citable ref is (and to make a gate rejection legible), which resolves all three at once.

### Verdict: CONCERNS

Not a BLOCK: the trusted live producer only ever emits gate-accepted same-turn refs, so the
false-evidence render path is not reachable through the honest path today; the ANSI/control-byte
sanitize+truncate boundary is sound; and no raise/crash paths were found. But for a feature whose
thesis is completion honesty, "rendered evidence" is not "validated evidence," and the absence row
is near-universal on the common path -- the guarantee is weaker than the PR frames it. The two
teeth: (1) a crafted/replayed/foreign-turn journal renders false evidence as proven-done because
the renderer skips every gate check (reproduced via the author's own passing cross-turn test),
and (2) fail-open makes a rejected done, an evidence-less done, and a plain greeting all render
identically.

merklebonsai:
Review response landed as `069a4efed`. Point-by-point:

**HIGH — renderer re-validates none of the gate's acceptance rules.** *Partially confirmed; the proposed fix refuted on three grounds; the disclosure gap fixed.* Your factual claim and reproduction are correct. But "resolve refs through the same predicate the gate uses" is (1) structurally impossible — `raxol_agent` depends on main `raxol`, never the reverse (frozen package graph), so the renderer cannot call the gate; (2) doctrinally wrong — the decision-time-fold law makes the admission gate the SOLE security boundary; duplicating its predicates in the display lane creates a second authority that drifts; (3) contrary to the ratified accord, which codified session-scoped resolution and mandated the exact cross-turn test you reproduced with. What survives is a disclosure gap, now shipped: cross-turn entries carry a mark, render `[cross-turn]` on the entry line and `(N cross-turn)` in the summary, tallied over ALL refs (a cross-turn ref beyond the entry cap still discloses). A foreign-turn citation now looks different — shown, not laundered. Ordering/mutation-echo checks stay with the gate, where the mutation semantics live.

**CONCERN — fail-open conflates three states.** *Observation confirmed; both proposed fixes refuted in-lane.* Gating the absence row on tool activity contradicts the ratified spec's unconditional absence pin; a rejection marker requires a producer wire change (already raised on the cross-lane thread). Now documented in a "Known conflation (fail-open wire)" section with an explicit do-not-weaken instruction.

**CONCERN — compaction staleness.** *Confirmed as doc gap; documented with a disambiguator the review missed:* interior drops set the projection's `damaged: true` flag, so unresolvable-in-a-damaged-stream (possible compaction victim) is distinguishable from unresolvable-in-an-intact-stream (ref never existed).

**LOW — "legitimately" cross-turn wording.** *Confirmed, rewritten:* the gate is the sole acceptance authority; session scope is a deliberate display-side superset so rejected citations are shown (marked), not hidden.

**LOW — no byte-golden pins the evidence arm.** *Confirmed, fixed:* the evidence fixture joined the golden matrix — 6 new files across all three tiers pin the summary, entry line, and absence row at byte level; existing goldens byte-untouched.

**LOW — wrap on empty completion rows.** *Confirmed, fixed:* unrecognized shapes render byte-identical to the unwrapped mount, pinned.

Red-first both passes (4 red this round, 18 the original). Harness+components 976/0; full suite 6133/6135 (both pre-existing, stash-verified); both bless --checks current; the no-churn pin held through the fix pass.

DROOdotFOO (commented):

Since last review: 1de5b335..069a4efe

Scope: the follow-up commit `069a4efe` ("Disclose cross-turn evidence refs and pin the evidence arm byte-golden"). I re-ran the prior HIGH's exact reproduction and probed the three sibling rejection classes.

- **Prior HIGH, cross-turn (foreign_turn) arm -- FIXED.** The exact case my last review reproduced (a `t2` done citing `t1`'s `tool_result`) now renders differently instead of as clean proven-done. `resolve_completion_entry/4` (`block_builder.ex:722-747`) compares the resolved event's `turn_id` against the claiming `turn_completed`'s `turn_id`, stamps `cross_turn: true`, and `block.ex:781-782` suffixes the line with `" [cross-turn]"` plus a session-wide `"(N cross-turn)"` summary tally. Verified against the shipped test at `completion_evidence_test.exs:193-224`.
- **Cross-turn tally survives the entry cap.** `cross_turn_total` is counted over ALL refs, not just the shown `@max_completion_entries` (`block_builder.ex:660-664`) -- a cross-turn ref pushed past the cap is still disclosed in the summary. Sound and tested (`completion_evidence_test.exs:244-269`).
- **BlockBody unrecognized-shape byte-identity -- FIXED (latent bug).** `wrap_with_completion/2` now computes `completion_rows/2` FIRST and branches on `[]` (`block_body.ex:88-93`) rather than pattern-matching the mere presence of the `:completion` key. Since `completion_rows(%Block{}, _)` falls through to `[]` (`block.ex:773`) for any unrecognized completion shape, an unknown shape now renders byte-identical to the unwrapped mount. The prior key-presence match would have wrapped (and shifted bytes) for an unrecognized shape.
- **Two prior limits now documented (not fixed).** The moduledoc now openly states the fail-open conflation and the compaction-staleness gap (`block_builder.ex` "Known conflation" / "Staleness under compaction" sections).

### High

- **False-evidence surface only PARTIALLY closed: `:stale_evidence` and `:mutation_echo` still render as clean proven-done. (reproduced)** The gate (`packages/raxol_agent/lib/raxol/agent/done_gate.ex:250-273`) rejects five classes: `:missing_ref`, `:not_evidence`, `:foreign_turn`, `:stale_evidence`, `:mutation_echo`. The renderer now discloses exactly two of them out of band (`:missing_ref` -> `unresolvable evidence ref`; `:foreign_turn` -> `[cross-turn]`) and one via the type breakdown (`:not_evidence` -> renders "1 message", not "1 tool result"). But **`:stale_evidence` and `:mutation_echo` are same-turn refs, so the `turn_id`-only cross-turn check never fires and they render as clean, unmarked evidence.**

  Reproduced on the renderer side (worktree at `069a4efe`): a single turn `t1` with a `tool_result` at id 5 (`mix_test`), then a LATER mutation (`rm_rf` `tool_use` at id 7), then `turn_completed final:true refs:[5]`. The renderer output:

  ```
  %{total: 1,
    evidence: [%{label: "mix_test — 42 tests, 0 failures", type: :tool_result, ref: 5}],
    type_counts: [%{count: 1, type: :tool_result}]}
  ```

  No `cross_turn`, no marker -- byte-identical to honest fresh evidence. The gate on the same journal rejects `refs:[5]` as `{:error, {:stale_evidence, 5}}` (deterministic from `done_gate.ex:263-264`: `last_mut = 7` because the `rm_rf` `tool_use` is mutating, and the cited event id `5 <= 7`). So a done whose only evidence predates a subsequent destructive mutation renders as a clean checkmark-equivalent -- exactly the honesty-creed violation the feature exists to prevent, for two of the five classes. The `mutation_echo` class is the same shape (same-turn, unmarked).

  I do not dispute the architectural stance (the gate is the security boundary; the renderer is display-side and cannot depend on `raxol_agent`). The residual is bounded to a dishonest/buggy producer or a replayed/tampered journal. But the PR advertises "a completion claim is only as good as its refs" while two rejection classes silently render clean, so I keep this at High-but-not-blocking.

### Medium

- **The "cannot re-validate" justification is partly a false constraint.** The moduledoc argues the renderer "could not call back into the gate even if it wanted to" (dependency direction). True for a direct call -- but the gate's staleness/echo predicates are PURE functions over the journal the renderer already holds, and the renderer already re-derives the `turn_id` comparison locally for cross-turn (`block_builder.ex:735`). `last_mutation`/`<=` staleness is the same kind of local derivation over `session_index`, needing no dependency inversion. The chosen scope (turn_id only) is a decision, not a dependency limitation, and the moduledoc frames it as the latter. Either widen the local disclosure to cover stale/echo, or state plainly that these two classes are knowingly left unmarked.

- **Prior MED (absence-row conflation) documented but NOT fixed.** A viewer still cannot distinguish gate-REJECTED-done vs honest-no-evidence vs plain-greeting: all three render the single unconditional `"no evidence provided"` row (`block.ex:738-743`). The moduledoc now says so and defers disambiguation to a producer-side wire change "tracked cross-lane." Acceptable as a documented deferral, but the "rejected done laundered into a benign absence row" information loss is real and unaddressed in this PR.

### Low

- **Summary/line disclosure can be internally inconsistent under the cap.** When all shown entries are same-turn but a cross-turn ref sits beyond `@max_completion_entries`, the summary reads `"... (1 cross-turn)"` while NO visible line carries `[cross-turn]` (the shown-lines are all clean). Verified by the shipped test `completion_evidence_test.exs:244-269`. Correct by design (tally over all refs), but a reader sees a cross-turn count with zero visible cross-turn lines -- a mild UX seam worth a "+N more (cross-turn)" hint on the tail row.

- Sanitize/clamp boundary re-confirmed sound; no crash path found; `completion_rows/2` fallback to `[]` (`block.ex:773`) is total.

### Verdict: CONCERNS

The specific prior HIGH reproduction (cross-turn foreign_turn) is genuinely fixed and the cross-turn disclosure is well-built and tested; a latent BlockBody byte-identity bug was also fixed. But the false-evidence HIGH is only partially closed: two of the gate's five rejection classes (`:stale_evidence`, `:mutation_echo`) still render as clean, unmarked evidence (reproduced), and the "cannot re-validate" rationale for leaving them is a false constraint since those predicates are local and pure. The fail-open absence-row MED is documented but not fixed. Not a BLOCK -- the gate remains the real boundary and the residual is display-only under a dishonest producer -- but the honesty guarantee is narrower than the PR advertises.

merklebonsai:
Round-2 response landed (branch history also cleaned of tool co-author trailers per Drew's request — message-only rewrite, identical trees). Ledger:

**HIGH residual — stale_evidence and mutation_echo render as clean evidence.** *Confirmed factually; the mirror-the-predicates fix refuted with code-level evidence; your offered alternative taken and hardened.* The sharpened refutation: the gate's `mutating?/1` ends in `not classified_effect_free?(ev)`, which is constantly false TODAY and is an explicit designed refinement seam. Both stale and echo consume that predicate — they are OPEN predicates. Mirroring today's everything-mutates reading display-side would freeze it: once the seam fills, gate-ACCEPTED evidence would render false `[stale]`/`[echo]` marks — the display accusing honest evidence and burning trust in the marks that ARE reliable. The three mirrored classes (index existence, item_type, turn_id equality) are CLOSED predicates that can never diverge — that criterion is now stated in the moduledoc. Per your suggestion verbatim: a named "Knowingly unmarked: stale and mutation-echo" section, including the residual's bound (fail-closed gate + refs-only-on-accept means today's honest wire never carries such refs; the unmarked render requires a tampered or synthetic journal). Hardened beyond the ask: BOTH classes pinned by named tests (your exact stale repro + an echo repro asserting the unmarked render and pointing at the section) — future changes must consciously break a pin.

**MED — "cannot re-validate" partly a false constraint.** *Confirmed, fixed:* the moduledoc now separates the two things your finding correctly split — the dependency direction only rules out CALLING the gate; which predicates to re-derive is a decision, and the closed-vs-open criterion with consequence analysis is that decision, stated.

**MED — absence-row conflation.** *Standing documented deferral, as you accepted:* the producer-side rejection marker is the cross-lane item already on the thread.

**LOW — summary/line inconsistency under the cap.** *Confirmed, fixed red-first:* the tail row reads `+N more (M cross-turn)` when hidden refs are cross-turn; the red run captured your exact inconsistency before the fix.

Gates: harness+components 980/0; full suite 6132/6139 (all verified pre-existing/flaky); both bless checks current; zero snapshot churn this round by design. Feature tally: 31 evidence tests, four commits of red-first evidence across three rounds.

DROOdotFOO (commented):

Since last review: 069a4efe..2c2fcf56

Scope: commit `2c2fcf56` ("State the unmarked evidence classes plainly and disclose hidden
cross-turn refs on the tail row") -- the response to my prior CONCERNS that the two
SAME-turn rejection classes (`:stale_evidence`, `:mutation_echo`) still rendered as clean,
unmarked proven-done evidence.

### What the commit actually does to the two flagged classes

It does **not** re-validate or mark them. `build_completion/4` resolves refs through exactly
three predicates -- index existence (`ref_type/2` -> `:unresolvable`), item_type
(`entry_type/2`), and turn_id equality (`cross_turn_event?/2`)
(`lib/raxol/harness/projection/block_builder.ex:687-804`). There is no mutation/ordering
check. A same-turn ref the gate rejects as `{:error, {:stale_evidence, _}}` or
`{:error, {:mutation_echo, _}}` still renders as ordinary `"1 tool result: ..."` with no
marker. This is now **documented + test-pinned** (moduledoc "Knowingly unmarked: stale and
mutation-echo", `block_builder.ex:130-160`; green-by-design pins in
`test/harness/completion_evidence_test.exs`), not fixed. So the answer to "(a) re-validate /
(b) documented-unmarked / (c) silently clean" is squarely **(b)**.

(reproduced) The two pins assert exact-equality on the completion map -- e.g. a stale ref
yields `%{evidence: [%{ref: 5, type: :tool_result, label: "mix_test — 42 tests, 0 failures"}],
total: 1, type_counts: [%{type: :tool_result, count: 1}]}`, no `cross_turn`, no stale marker.
All 28 tests in `completion_evidence_test.exs` pass on this head.

- **The internally-inconsistent cross-turn tail row is fixed (real fix).** When a cross-turn
  ref sits past the entry cap, `"+N more"` now becomes `"+N more (M cross-turn)"`
  (`block.ex:811-838`, `hidden_cross_turn_suffix/2`). Math verified: `hidden = session-wide
  cross_turn_count - (cross-turn among the shown <=3 entries)`, so shown-marked + tail-count
  always sum to the summary tally. Previously the summary could read "(2 cross-turn)" with
  zero visibly marked lines. Shown-marker and no-cross cases are byte-unchanged; this commit
  touches no goldens.
- **The false-evidence HIGH is no longer reachable on the honest wire (verified).** I traced
  the load-bearing bound the moduledoc rests on. `DoneGate` is fail-closed today:
  `classified_effect_free?/1` is a constant `false`
  (`packages/raxol_agent/lib/raxol/agent/done_gate.ex:317`), so every completed `:tool_use`
  is a mutation -> the last `:tool_result` is always the last mutation's echo and any earlier
  one is stale -> no ref passes `classify_ref/5`. And `contract.ex:269-294`
  (`gated_done_payload/4`) attaches `:refs` to the wire **only** on `{:ok, _}`. Net: an honest
  producer never emits refs at all, so the entire evidence-present render (including unmarked
  stale/echo) is reachable only through a tampered or synthetic journal. The moduledoc's bound
  is accurate.

### Low / Informational

- **The unmarked-classes rationale is sound and near-costless to keep.** In the only reachable
  threat model (untrusted journal), a tamperer wanting a "proven" done would simply fabricate a
  fresh, same-turn, non-echo `:tool_result` -- which no renderer-side check can catch, since
  the renderer cannot know a result is fabricated. So mirroring the stale/echo predicate would
  raise the bar only against a *lazy/inconsistent* tamperer while introducing a real
  false-positive once the `classified_effect_free?` seam is filled (gate-accepted evidence
  would then carry false stale/echo marks). Excluding the two *open* predicates while mirroring
  the three *closed* ones is a defensible, now-explicit line. I concur with the disposition.

- **Latent cross-package coupling on the bound (worth a tracking note, not a blocker).** The
  "honest wire never carries refs" guarantee is enforced entirely in `raxol_agent`
  (`done_gate.ex` fail-closed + `contract.ex` refs-only-on-accept), with no compile-time link
  to this renderer in main `raxol`. If a future change makes the gate fail-open, or attaches
  refs off the accept path, the renderer's honesty bound silently degrades and unmarked
  stale/echo becomes reachable on the honest wire -- with nothing in this package to notice.
  A cross-lane invariant test (or a comment pin in `contract.ex` pointing back at this
  renderer's bound) would keep the two lanes from drifting. This is the same "closed vs open"
  discipline the moduledoc already articulates, just extended to the coupling itself.

### Verdict: CLEAN

The prior HIGH is resolved: not by re-validating stale/echo (correctly avoided, per the open-
predicate/refinement-seam argument), but by a verified bound that makes the unmarked render
unreachable on the honest wire, plus honest documentation and exact-equality test pins. The
tail-row inconsistency is genuinely fixed with correct math and no golden churn. The only
residual is the latent cross-package coupling on the bound (Low), which is a tracking note, not
a merge blocker.

DROOdotFOO:
**Merge readiness: READY TO MERGE.**

Latest re-review verdict is CLEAN -- the false-evidence HIGH is resolved (the honest wire never carries the problematic refs because `DoneGate` is fail-closed today). CI green, mergeable, no blocker. One LOW follow-up worth a tracking test: the honesty bound lives in `raxol_agent` with no compile-time link to the renderer, so a future fail-open gate change would silently break it.

PR #620, opened 2026-07-16
Harness UI: seal-pipeline hardening (write-confirmed sealing, frame order, synchronized output)

merklebonsai:
Three coupled correctness items on the seal pipeline, hardening it before live agent traffic. Builds on the merged shared frontier classifier.

## 1. Two-phase seal (write -> confirm -> mark)

Empirically probed the BEAM's device-failure classes first: an io server replying `{:error, reason}` surfaces from `IO.write/2` as ArgumentError; a dead device as ErlangError `:terminated`. `InlineAuthority.try_seal/2` rescues exactly those two, returning `{:error, :write_failed, t}` with `next_row` untouched so a retry re-positions from scratch. The newline caller-contract validation runs OUTSIDE the rescued scope — a malformed call still raises rather than becoming infinitely retryable. `Surface.seal_block/2` returns the frontier walk's emit contract directly: a failed write halts the walk with the cursor before the failed entry; the next advance retries; a block is never marked sealed unless its bytes were accepted. Partial-write honesty documented: a retry may overwrite an unconfirmed row in place — safe because unconfirmed rows never enter high-water accounting.

Fill protection: no placeholder-fill pattern against sealed content exists in the codebase (grep-verified); the only mutation channel (fold overrides on painted blocks) is pinned by test to store nothing and never emit history bytes.

## 2. Frame-order law (adopt-resize-first)

`Surface.advance/3` gains a `:resize {w, h}` option — the atomic combined frame. Resize adoption is extracted into one function SHARED by the standalone resize path and the advance path so they cannot drift; a block sealing on a shrink frame is laid out at the NEW width (the named red-first obligation from the frontier suite, now implemented — stale-width bytes in a print-once stream are permanent corruption). Honest note: the reference design's "size viewport to post-commit height" step has no decision in this substrate (footer row count is geometry-fixed), documented as satisfied by construction; the load-bearing halves are adopt-before-seal and footer-repaint-after-seal.

## 3. Synchronized output (DEC 2026)

`InlineAuthority.with_sync/3`: the authority gates (capability measured once at construction; nil/unknown -> never emit) and owns the bytes. Balanced by construction — the close is emitted iff the open succeeded, through mid-frame seal failures; a failed open degrades to an unbracketed frame (presentation must never take down the frame). The bracket condition is the pre-seal `scan_frontier` `will_commit` — a third consumer of the shared classifier (consumer table updated). SequenceScanner already tokenizes `?2026`; the oracle needed no changes — pinned by tests, and the golden vocabulary allowlist stays deliberately unchanged (goldens render with capabilities nil; comment cites the pinning test).

## Verification

Red-first: 8 of 13 new tests captured failing pre-implementation (retry-not-vanish, try_seal/advance-3 undefined, both sync-bracket assertions); the other 5 are deliberate pins. One test-design flaw caught during implementation review and fixed in the TEST: the footer pending-preview legitimately carries block content, so the leak-detection marker had to sit below the preview window. New suite 13/13; harness+property+cross-terminal sweep 930 passed (212 properties), 0 failures. Warnings + format clean.

### DIFF
--- a/lib/raxol/harness/seal_frontier.ex
+++ b/lib/raxol/harness/seal_frontier.ex
@@ -27,10 +27,13 @@ defmodule Raxol.Harness.SealFrontier do
       scrollback this frame;
     * the footer/tail composition -- which blocks still render in the
       pinned live region (the trailing "pending" preview);
-    * (future, documented seam, not yet built) the resize path's
-      `will_commit` gate and post-commit viewport sizing -- the frame-order
-      unit that must decide sizing at the width a possibly-just-committed
-      block was laid out at.
+    * the synchronized-output bracket decision -- `Raxol.Harness.Surface`'s
+      `seal_frame/3` reads the SAME `will_commit` projection, per-frame, to
+      decide whether this frame's seal + footer repaint needs wrapping in
+      a DEC 2026 bracket (see that module's moduledoc). The post-commit
+      viewport-sizing seam remains future work: this substrate's footer
+      row count is geometry-fixed (never a function of post-seal state),
+      so there is no viewport-sizing decision left to make here.
 
   Consumers never inline their own walk over entries. The consumer table
   today:
@@ -39,11 +42,10 @@ defmodule Raxol.Harness.SealFrontier do
       walk, via `commit_walk/5`.
     * `Raxol.Harness.Surface`'s footer pending-preview (`pending_block/1`,
       via `frontier_scan/1`, itself `scan_frontier/3`) -- read-only.
-
-  Documented seams (not yet built, but must call `scan_frontier/3` before
-  the commit pass in the frame and mirror its stop condition exactly when
-  they land): the resize-path `will_commit` gate, and post-commit viewport
-  sizing.
+    * `Raxol.Harness.Surface`'s synchronized-output bracket gate
+      (`seal_frame/3`, also via `frontier_scan/1`) -- read-only, consulted
+      BEFORE the commit pass to decide whether to open a bracket around
+      it.
 
   ## The entry contract
 
@@ -170,11 +172,13 @@ defmodule Raxol.Harness.SealFrontier do
   it, so the very next pass retries that same entry rather than silently
   skipping past a block that was marked but never actually printed (which
   would make it vanish forever, since a print-once surface cannot re-emit
-  a committed entry). This emit-then-mark contract is the seam a future
-  two-phase seal (a write-confirming substrate) will build on; today's
-  Surface emit (`InlineAuthority.seal/2`) has no error path, so the
-  `{:error, :write_failed, _}` branch is exercised only by this module's
-  own corpus until such a substrate lands.
+  a committed entry). This emit-then-mark contract is the seam a
+  write-confirming substrate builds on -- and that substrate now exists:
+  `Raxol.Harness.Surface.seal_block/2` emits through
+  `InlineAuthority.try_seal/2` (write -> confirm -> mark), and the
+  `{:error, :write_failed, _}` branch is exercised both by this module's
+  own corpus and end-to-end through a real failing device in
+  `test/harness/surface_seal_pipeline_test.exs`.
 
   ## The cursor
 
--- a/lib/raxol/harness/surface.ex
+++ b/lib/raxol/harness/surface.ex
@@ -809,6 +809,7 @@ defmodule Raxol.Harness.Surface do
   end
 
   defp do_advance(model, now) do
+    model = heal_sync(model)
     revealed = min(model.revealed + 1, length(model.events))
     events_so_far = Enum.take(model.events, revealed)
 
@@ -831,25 +832,44 @@ defmodule Raxol.Harness.Surface do
   end
 
   # A frame that seals at least one block AND repaints the footer
-  # presents atomically: seal + status + footer run inside one DEC 2026
-  # synchronized-update bracket (InlineAuthority.with_sync/3, gated on
-  # the capability record). The bracket condition is the pre-seal scan's
-  # will_commit -- it exactly predicts "this frame seals >= 1 block"
-  # except when the first emit fails, in which case an empty bracket is
-  # emitted (harmless: a sync frame with no visible change; the
-  # balanced-bracket case in test/harness/surface_seal_pipeline_test.exs
-  # covers the failure path). Frames that seal nothing (early reveals,
-  # tick, handle_input) never open a bracket. :flat has no footer and no
-  # cursor vocabulary -- never bracketed.
+  # presents atomically: seal + status + footer run between
+  # InlineAuthority.sync_open/1 and sync_close/1 -- one DEC 2026
+  # synchronized-update bracket, gated on the capability record. The
+  # bracket condition is the pre-seal scan's will_commit -- it exactly
+  # predicts "this frame seals >= 1 block" except when the first emit
+  # fails, in which case an empty bracket is emitted (harmless: a sync
+  # frame with no visible change; the balanced-bracket case in
+  # test/harness/surface_seal_pipeline_test.exs covers the failure path).
+  # Frames that seal nothing (early reveals, tick, handle_input) never
+  # open a bracket -- but every frame entry point calls heal_sync/1
+  # first, so a close the device refused in an EARLIER frame (the
+  # dangling-open wedge) is re-attempted at the first opportunity of any
+  # kind. :flat has no footer and no cursor vocabulary -- never
+  # bracketed.
+  #
+  # If run_seal_frame/3 raises mid-bracket (a logic bug -- device
+  # failures are tuples, not raises, on this path), the rescue below
+  # makes one best-effort close attempt before re-raising, so a crashing
+  # frame does not also leave the terminal synchronized. The authority
+  # state update is lost with the crash either way; the byte is what
+  # matters to the terminal.
   defp seal_frame(%{mode: :flat} = model, events_so_far, now) do
     run_seal_frame(model, events_so_far, now)
   end
 
   defp seal_frame(model, events_so_far, now) do
     if frontier_scan(model).will_commit do
-      InlineAuthority.with_sync(model.authority, model, fn m ->
-        run_seal_frame(m, events_so_far, now)
-      end)
+      opened = update_authority(model, &InlineAuthority.sync_open/1)
+
+      try do
+        opened
+        |> run_seal_frame(events_so_far, now)
+        |> update_authority(&InlineAuthority.sync_close/1)
+      rescue
+        e ->
+          _ = InlineAuthority.sync_close(opened.authority)
+          reraise e, __STACKTRACE__
+      end
     else
       run_seal_frame(model, events_so_far, now)
     end
@@ -862,6 +882,20 @@ defmodule Raxol.Harness.Surface do
     |> paint_footer()
   end
 
+  # Re-attempts a sync close the device refused in an earlier frame (see
+  # InlineAuthority.sync_close/1's latch contract) -- a byte-free no-op
+  # when nothing is owed, so every frame entry point (do_advance, tick,
+  # handle_input, resize) calls it unconditionally: a dangling ?2026h
+  # heals at the FIRST frame after the device accepts a byte again,
+  # never later.
+  defp heal_sync(%{mode: :flat} = model), do: model
+
+  defp heal_sync(model),
+    do: update_authority(model, &InlineAuthority.sync_close/1)
+
+  defp update_authority(model, fun),
+    do: %{model | authority: fun.(model.authority)}
+
   @doc """
   Advances the elapsed-since-last-event ticker (the status strip's
   `Stage` slot) without revealing a new fixture event -- elapsed ticks
@@ -872,6 +906,7 @@ defmodule Raxol.Harness.Surface do
   @spec tick(t(), integer()) :: t()
   def tick(model, now) when is_integer(now) do
     model
+    |> heal_sync()
     |> put_in([:status, :now], now)
     |> paint_footer()
   end
@@ -999,12 +1034,14 @@ defmodule Raxol.Harness.Surface do
     # The ONE mutating frontier walk (SealFrontier's moduledoc): emit
     # each newly-committable block via seal_block/2, which advances
     # painted_count -- the committed marker frontier_entries/1 reads.
-    # The emit is write-confirming (`InlineAuthority.try_seal/2`): write ->
-    # confirm -> mark. A failed device write halts the walk with
+    # The emit is write-checked (`InlineAuthority.try_seal/2`): write ->
+    # confirm -> mark, where "confirmed" means the device's io server
+    # ACCEPTED the write (see try_seal/2's doc for what that does and
+    # does not promise). A refused write halts the walk with
     # painted_count (the cursor) strictly before the failed entry, and the
     # next advance retries the same block, so a block can never be marked
-    # painted without its bytes confirmed on the device (retry-not-vanish;
-    # covered by test/harness/surface_seal_pipeline_test.exs).
+    # painted for bytes the device refused (retry-not-vanish; covered by
+    # test/harness/surface_seal_pipeline_test.exs).
     result =
       SealFrontier.commit_walk(
         entries,
@@ -1193,6 +1230,7 @@ defmodule Raxol.Harness.Surface do
   """
   @spec handle_input(t(), term()) :: t()
   def handle_input(model, raw_event) do
+    model = heal_sync(model)
     norm = InputEvent.normalize(raw_event)
 
     # Every keystroke is presence evidence for the unread divider's
@@ -2173,7 +2211,7 @@ defmodule Raxol.Harness.Surface do
     do: adopt_resize(model, width, rows)
 
   def resize(model, width, rows) do
-    model = adopt_resize(model, width, rows)
+    model = model |> heal_sync() |> adopt_resize(width, rows)
     lines = footer_lines(model)
     authority = InlineAuthority.keyframe(model.authority, lines)
     %{model | authority: authority, stub_notice: nil}
--- a/lib/raxol/harness/surface.ex
+++ b/lib/raxol/harness/surface.ex
@@ -840,6 +840,12 @@ defmodule Raxol.Harness.Surface do
   # fails, in which case an empty bracket is emitted (harmless: a sync
   # frame with no visible change; the balanced-bracket case in
   # test/harness/surface_seal_pipeline_test.exs covers the failure path).
+  # A PERSISTENTLY refusing (alive) device therefore emits one empty
+  # open/close pair per retry frame -- a few bytes of decoration per
+  # frame, deliberately not special-cased: the same frames emit
+  # [:raxol, :harness, :seal, :write_failed] telemetry per refused
+  # write, so the condition is observable from the first frame and the
+  # operator/driver owns the decision to stop advancing.
   # Frames that seal nothing (early reveals, tick, handle_input) never
   # open a bracket -- but every frame entry point calls heal_sync/1
   # first, so a close the device refused in an EARLIER frame (the
@@ -1053,7 +1059,26 @@ defmodule Raxol.Harness.Surface do
             |> Enum.at(index)
             |> apply_fold_override(index, acc.fold_overrides)
 
-          seal_block(acc, block)
+          case seal_block(acc, block) do
+            {:ok, _acc} = ok ->
+              ok
+
+            {:error, :write_failed, _acc} = error ->
+              # The retry loop for a refusing-but-alive device is
+              # unbounded BY DESIGN (a bound would strand the block when
+              # the device recovers) -- this emit is what keeps it from
+              # being unbounded AND invisible: one event per refused
+              # write, from the first frame, so a driver/operator can see
+              # a persistent refusal and decide. A DEAD device never
+              # reaches here (try_seal fail-fasts on a corpse).
+              :telemetry.execute(
+                [:raxol, :harness, :seal, :write_failed],
+                %{},
+                %{index: index, kind: block.kind}
+              )
+
+              error
+          end
         end,
         cursor: model.painted_count
       )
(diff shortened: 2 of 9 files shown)


DROOdotFOO (commented):

Scope: `try_seal/2` two-phase seal, frame-order (`advance/3` + `adopt_resize/3`), and DEC 2026 synchronized output (`with_sync/3`). Verified against `lib/raxol/ui/rendering/paint_authority/inline_authority.ex` and `lib/raxol/harness/surface.ex` at the diff. Frame-order (item 2) reads correct: `advance/3` threads `adopt_frame_resize/2` unconditionally ahead of `do_advance/2`, and `seal_frame/3` computes `will_commit` at the already-adopted geometry. No findings there.

### HIGH

- **[Saboteur + New Hire] The rescue is broader than "device failure" and does not filter on the reason it documents.** `inline_authority.ex` `confirmed_seal/2` rescues `_e in [ArgumentError, ErlangError]` (diff, new `confirmed_seal/2`), and `sync_write/2` uses the identical clause. But the docstring for `try_seal/2` names the dead-device class as `%ErlangError{original: :terminated}` -- the code catches **every** `ErlangError` regardless of `original`, and every `ArgumentError` raised anywhere inside the `with_cursor -> append_sealed` closure (not only the io-server `{:error, reason}` reply). Consequence: a genuine logic bug that raises `ArgumentError`/`ErlangError` from inside the seal (a malformed region struct, a `count_lines`/`IO.iodata_to_binary` fault on bad iodata, a future refactor bug) is silently reclassified as `{:error, :write_failed, t}`. Per `surface.ex` `seal_block/2` + `commit_walk/5`, `next_row`/`painted_count` are left un-advanced and the SAME entry is retried on every subsequent `advance` -- and a code bug is not transient, so this is an unbounded retry loop that never surfaces the real error. The task's own premise ("rescues ... a dead device as ErlangError :terminated") is what the prose promises but not what the code does. Narrow the rescue to the reason it documents (guard on `original: :terminated` for the `ErlangError` arm; scope the `ArgumentError` arm as tightly as the io seam allows) so non-device faults stay loud. (Saboteur MEDIUM: masks bugs into an infinite retry; New Hire LOW: doc/code disagree on the WHY -- same root, promoted to HIGH.)

- **[Saboteur + Security Auditor] `with_sync/3` can leave the terminal wedged in synchronized-update mode; "balanced by construction / never a dangling open" is an overclaim.** `inline_authority.ex` `with_sync/3` opens with `sync_write(device, "\e[?2026h")` and closes in the `after` with `sync_write(device, "\e[?2026l")`; `sync_write/2` rescues device errors and returns `:error` **without the byte landing**. The moduledoc guarantees the close is emitted "iff the open succeeded" -- but that guarantees the close is *attempted*, not that the close *byte reaches the device*. If the open is accepted (`?2026h` on the wire, device alive) and the device then faults on the close write (transient `{:error, :enospc}`, SIGHUP mid-frame), the open lands with no matching close and the terminal stays frozen in synchronized mode. There is no reset-on-next-frame or idempotent re-close: I grepped and the harness path has no `?2026l` recovery seam (only `backends.ex` emits its own independent pair). The balanced-bracket test (`surface_seal_pipeline_test.exs`) only exercises a device that *recovers* before the close (`FailingDevice` fails exactly once), so the stuck-through-the-close path is unverified. Security framing: a hostile or flaky device that ACKs the open and drops the close is a display-DoS (terminal wedged) reachable from a presentation-only feature. Same root, Saboteur + Security -> promoted to HIGH. Suggest a self-healing close (emit `?2026l` unconditionally at the top of the next sealing frame, or track an "open but unclosed" latch).

### MEDIUM

- **[Saboteur + Security Auditor] "Partial-write honesty ... safe by construction" breaks at the scroll boundary.** `inline_authority.ex` `try_seal/2` writes a whole multi-line block in one `IO.write(device, iodata)`; the partial-write doc argues a retry "produces the same rows the confirmed write would have" because `next_row` was not advanced. That holds only when `target_row = min(next_row, bottom)` sits with slack above `history_bottom`. On a real tty a partial write can flush some of the block's embedded `\r\n`s before erroring; if `target_row` is at/near the region bottom, those newlines scroll the DECSTBM region and evict partially-written rows into native scrollback, which this process can never rewrite (the module's own core invariant). The retry then re-CUPs to the un-advanced absolute `target_row` and re-emits the full block, but the evicted partial rows are permanent duplicate/garbled scrollback. StringIO cannot reproduce this (it never partial-writes), so it is a design-honesty gap rather than a test-caught one -- but the blanket "no seal-once violation" claim is not true at the boundary. (Security: content-integrity duplication, overlaps the same root -- left MEDIUM as it needs a real-tty partial write at the bottom row.)

- **[Saboteur / New Hire] "Confirmed" is "the io server replied `:ok`", not "the bytes reached the terminal".** `try_seal/2`'s contract ("reports whether the device actually confirmed it", "confirmed on the device") is accurate for a StringIO/port io-request round-trip, but for a buffered pipe/`:stdio` an `:ok` reply means accepted-into-the-io-server, not rendered. The module cannot do better without a DSR round-trip, so this is a wording-precision issue -- but "confirmed on the device" reads stronger than the mechanism delivers. Recommend softening to "the io server accepted the write" so a future reader does not assume end-to-end delivery.

### LOW

- **[New Hire] Magic synchronized-output sequences bypass the module's own wire-vocabulary indirection.** `with_sync/3` hardcodes `"\e[?2026h"` / `"\e[?2026l"` inline (and the test re-hardcodes them as `@sync_open`/`@sync_close`), while every other control sequence in this module routes through `Dialect` (`cursor_save/0`, `cursor_restore/0`, `cursor_position/1`). The BSU/ESU pair and the `2026` mode number have no named constant and are duplicated between production and test, so a typo (`2026` vs `2027`) or a future SGR-saver/DECRST-restore variant would drift silently. Give them a `Dialect.sync_begin/0` / `Dialect.sync_end/0` (or module attributes) matching the module's stated "shared Dialect wire vocabulary" design.

- **[New Hire] Two different capability-reading idioms in one constructor.** `new/5` sets `sync_output?: match?(%{sync_output: true}, caps)` -- a loose map match that also matches any non-`Capabilities` map carrying `sync_output: true` -- while `reflow_capable?/1` in the same file uses a strict `%Capabilities{identity: ...}` struct pattern. There is also an existing public gate `Raxol.Terminal.Capabilities.sync_output?/0` that this bypasses. Prefer the struct pattern (`%Capabilities{sync_output: true}`) for consistency and to avoid matching stray maps.

### Cross-persona overlaps

- Over-broad `[ArgumentError, ErlangError]` rescue: Saboteur (masked bug -> infinite retry) + New Hire (doc names `:terminated`, code catches all) -- promoted MEDIUM -> HIGH.
- Unbalanced `?2026` bracket: Saboteur (close byte may not land) + Security Auditor (device-driven display DoS) -- promoted MEDIUM -> HIGH.
- Partial-write boundary: Saboteur (permanent scrollback corruption) + Security Auditor (content duplication) -- same root, held at MEDIUM (real-tty-only trigger).

### Verdict: CONCERNS

Two HIGH issues (both promoted via cross-persona agreement, both in `inline_authority.ex`), several MEDIUM/LOW. No CRITICAL and fewer than three HIGH, so not a BLOCK -- but the two HIGH items are load-bearing for the exact guarantees this PR advertises (fail-closed device handling and balanced synchronized output) and should be addressed before live agent traffic. The frame-order item is clean.

merklebonsai:
Review response landed as `19fd477d9`. Ledger:

**HIGH — rescue broader than device failure; a logic bug becomes unbounded silent retry.** *Confirmed (verified: nothing filtered on origin; the retry-loop consequence is real via the walk's cursor semantics).* Fixed with `device_io_error?/2`: requires the exception class AND an `:io`-module stack head — empirically probed on both device classes (error-reply raises from `{:io, :put_chars, ...}` with a `:device` cause; dead device with `:io, :terminated`), while a local raise carries a different head. Non-matching exceptions RE-RAISE. Applied at both rescue sites; doc rewritten to match.

**HIGH — a refused sync-close write wedges the terminal; "balanced by construction" was attempt-accurate only.** *Confirmed (the close failure was swallowed with no recovery seam).* Redesigned as a latch: `sync_open/1` sets `sync_close_pending?` iff the open byte landed; `sync_close/1` writes iff owed, clears iff accepted, and is a byte-free no-op otherwise — safe to call every frame. The Surface heals at all four frame entries, so a refused close heals at the first subsequent frame with a writable device (and a refused SEAL guarantees a retry frame, so the common failure topology heals immediately). Crash path: one best-effort close before re-raise. The balance claim is now byte-accurate, not attempt-accurate. **API note: `with_sync/3` is removed in favor of the pair — flagged to the in-flight live-session track.**

**MED — partial-write claim breaks at the scroll boundary.** *Confirmed as a doc-honesty gap.* The guarantee is now scoped to accept-or-refuse-as-a-unit transports (everything driving this path today: StringIO, the BEAM io server); the raw-fd boundary residual is named explicitly as a documented limit, with line-at-a-time emit noted as the follow-up if such a transport lands.

**MED — "confirmed" overclaims.** *Confirmed.* New "What accepted means (and does not)" section; Surface wording softened to the refusal framing.

**LOW — hardcoded ?2026 bytes.** *Confirmed.* `Dialect.sync_begin/0`/`sync_end/0`; production routes through them; the test deliberately keeps raw bytes as the drift oracle (same-constant pinning would tautologize — noted in-test).

**LOW — loose capability match.** *Confirmed.* Strict struct match mirroring the reflow seam's idiom; deliberately NOT the process-global capability cache (the constructor takes explicit caps as the test seam).

Red-first: 4 failing pre-fix (13/17 -> 17/17), incl. proving the loose match DID enable emission from a stray map and the refused close was never re-attempted. Harness+property+cross-terminal 934 passed / 0 failures; warnings + format clean.

**Residual (driver lane, one line):** the canonical teardown sequence doesn't include `?2026l` — a crash between open and close waits on the terminal's own sync timeout. Outside this PR's blast radius; flagged for the driver-lane backlog.

DROOdotFOO (commented):

Since last review: d7df35bf..19fd477d

Focus: the round-1 hardening commit `19fd477d` ("scoped rescues, healing sync
bracket"), which directly targets the two HIGHs. Verified against the PR-head
checkout of `pr-620-verify`; the new suite compiles and all 17 tests in
`test/harness/surface_seal_pipeline_test.exs` pass (reproduced). Two extra
probes run below.

- **HIGH -- over-broad rescue vs "confirmed" -- RESOLVED.** The seal rescue is
  now narrowed twice over. `confirmed_seal/2` catches only
  `e in [ArgumentError, ErlangError]`
  (`lib/raxol/ui/rendering/paint_authority/inline_authority.ex:519-528`), so
  every OTHER logic-bug class (`KeyError`, `FunctionClauseError`, ...) is not in
  the rescue list at all and propagates untouched. The two classes that ARE
  caught pass through `device_io_error?/2`, which requires the stacktrace head
  to be `{:io, _, _, _}` -- i.e. the raise must have come out of the `:io` layer
  (`append_sealed/2` writes via `IO.write/2` -> `:io.put_chars`, inline_authority.ex:414-415).
  An `ArgumentError`/`ErlangError` raised by non-device code returns `false` ->
  `reraise`. The unit test `device_io_error?/2 recognizes the two device classes
  and rejects local raises` proves the discriminator returns false for a
  locally-raised `ArgumentError` (reproduced). A genuine logic bug can no longer
  be reclassified as `:write_failed`, so the logic-bug silent-retry loop is gone.

- **HIGH -- unbalanced synchronized-output -> wedged terminal -- RESOLVED.**
  `with_sync/3` is replaced by a latch-backed pair, `sync_open/1` + `sync_close/1`,
  with a `sync_close_pending?` field. A landed open latches an owed close; the
  latch clears ONLY when a close write is accepted. Every frame entry point
  (`do_advance`, `tick`, `handle_input`, inline `resize`) calls `heal_sync/1`
  first (`lib/raxol/harness/surface.ex:788-796`), re-attempting the owed close,
  and a mid-frame raise makes one best-effort `sync_close` before re-raising
  (`surface.ex:762-778`). The test `a close-write failure does not wedge the
  terminal: the owed close is re-attempted on the next frame` constructs
  accepted-open + faulted-close and confirms the next (tick) frame emits the
  owed `?2026l`, healed and balanced (reproduced). The wedge can no longer
  persist across frames.

- **MED -- scroll-boundary partial write -- documented.** `try_seal/2`'s doc now
  has an explicit "Partial-write honesty (and the scroll-boundary residual)"
  section naming the raw-fd/pty case where partially-flushed `\r\n`s evict rows
  into unreachable native scrollback, and notes no current transport
  partial-writes. Honest, not an implied guarantee.

- **MED -- "confirmed" wording -- documented.** `try_seal/2` now states plainly
  that `{:ok, _}` means the io server ACCEPTED the write (handed-off), not
  end-to-end render, and that a DSR round-trip would be needed to prove delivery.

### Medium

- **Non-transient device fault is still an unbounded, silent, un-observable
  retry loop (reproduced).** The scoped rescue kills the LOGIC-bug loop, but the
  retry mechanism itself survives for genuine device faults -- and it does not
  distinguish transient from permanent. A device that permanently refuses the
  seal write drives `advance/2` to return `{model, :ok}` forever: 200 advances
  against a persistently-failing device left `painted_count == 0`, only `:ok`
  ever returned, never `:done`, with no telemetry and no retry bound
  (`inline_authority.ex` `device_io_error?/2` + `surface.ex` `do_advance`).
  Sharper: a DEAD device (`ErlangError{original: :terminated}`) is EXPLICITLY
  classified as a retryable `device_io_error?` -- so the harness retries a corpse
  forever. Note this is a mild observability regression for the dead-device case:
  pre-PR, `seal/2` had no error path, so a dead device would have RAISED out of
  `advance` (loud crash); this PR converts that into a silent infinite loop.
  The empty-bracket-per-frame behavior (a `will_commit`-true frame whose seal
  write fails emits a bare `?2026h`/`?2026l` with no content) shares this root:
  on a persistent failure it becomes forever byte-spam of accepted open/close
  pairs. A retry bound, or a `[:raxol, :harness, :seal, :write_failed]` telemetry
  emit, would make the non-transient case observable and is the natural
  follow-up. Non-blocking, but the dead-device branch deserves a fail-fast or a
  count.

### Low

- **No teardown backstop for an owed sync-close.** The heal contract is
  explicitly "the next frame with a writable device." A wedge that lands on the
  FINAL frame, followed by process exit, has no flush seam: Surface owns no
  `terminate/2` (precondition #7, `surface.ex:381-397`), and the documented
  cleanup path `InlineDriver.Sequences.teardown_bytes/1` predates sync brackets
  -- it emits the DECSTBM region release (`\e[r`), not `?2026l`. So the specific
  wedge-then-quit topology can still strand the real terminal in synchronized
  mode. Recommend the teardown path emit `?2026l` unconditionally (DEC private
  modes are set/reset -- harmless when nothing is owed).

### Verdict: CONCERNS (non-blocking)

Both HIGHs are cleared, each with a real end-to-end test through the device seam
(reproduced): the scoped rescue no longer swallows logic bugs (they re-raise
loudly), and the latched sync bracket can no longer wedge the terminal across
frames. The remaining items are one MED (non-transient/dead-device retry loop
has no bound, telemetry, or fail-fast -- a mild observability regression from
the pre-PR crash) and one LOW (no teardown flush for an owed sync-close). Both
MEDs from round 1 are now honestly documented. Mergeable; the dead-device bound
or telemetry emit is the recommended fast-follow.

merklebonsai:
Round-2 response landed as `dc120a2cb`. Ledger:

**MED — non-transient device fault = unbounded silent retry; a dead device retried forever (observability regression vs the pre-PR crash).** *Confirmed — your reproduction is exact.* Two-part fix: (1) **fail-fast on a corpse** — retryable iff it's a device io error AND NOT `ErlangError :terminated`; a dead io-server pid can never return, so it re-raises (the pre-PR loud crash, restored) from both the seal and sync write paths. `device_io_error?/2` is unchanged — it answers "is it a device error", not "is it worth retrying"; the doc now states the split. (2) **Telemetry** — one `[:raxol, :harness, :seal, :write_failed]` event per refused write (index + kind), emitted at the commit-walk seam. The remaining alive-device retry loop stays unbounded BY DESIGN (a bound strands the block when the device recovers on attempt N+1) — now documented AND observable from frame one. The empty-bracket byte-spam sub-point: deliberate, observable via the same event.

**LOW — no teardown backstop for an owed sync-close.** *Confirmed (also my own round-1 flagged residual).* `\e[?2026l` added to the shared modes-off constant — covering BOTH `teardown_bytes/1` and `suspend_bytes/1`: the editor-suspend path had the same hole (an editor must not inherit a synchronized-frozen screen), which the shared constant fixes for free. Emitted before the region release, pinned by test on both paths.

Red-first: 4 failing pre-fix (dead-device assert_raise, telemetry, both backstop pins); 16/18 + 11/13 before, 18 + 13 after. Main gates 935 passed / 0 failures; package sequences+scroll-region green; format + warnings clean. (The 4 pre-existing RendererIntegrationTest background-SGR failures reproduce with this change reverted — different subsystem, flagged for the backlog.)

**Behavior note for consumers:** no signature changes, but `try_seal/2` now RAISES on a dead device instead of returning the error tuple (reserved for the alive-but-refusing class) — noted on the live-session PR.

DROOdotFOO (commented):

Since last review: 19fd477d..0a046646

Scope: the single new commit `0a046646` ("Make seal-write failure observable and
fail-fast on dead devices"), which targets the one MEDIUM residual (unbounded,
silent, un-observable retry loop; dead device classified retryable) and the one
LOW (no teardown flush for an owed sync-close) from round 2. Both round-1 HIGHs
stay cleared. Verified end-to-end: the new `test/harness/surface_seal_pipeline_test.exs`
suite is green (18 passed) against the PR head in an isolated worktree.

- **Dead-device infinite-retry -- FIXED, bounded via fail-fast (reproduced).** My
  prior residual (a dead device -- `ErlangError :terminated` -- was classified
  RETRYABLE, so the harness retried a corpse forever, silently) is now a loud
  fail-fast. `retryable_device_error?/2` (`inline_authority.ex:549-552`) layers
  retryability on top of `device_io_error?/2` and re-raises
  `%ErlangError{original: :terminated}` from both `confirmed_seal/2` and
  `sync_write/2`, restoring the pre-two-phase loud crash. Pinned by
  `surface_seal_pipeline_test.exs` "try_seal/2 fail-fasts on a dead device"
  (`assert_raise ErlangError`), green.
- **Persistent alive-refusal loop -- now OBSERVABLE.** The remaining (deliberate)
  unbounded retry for an alive-but-refusing device (`{:error, reason}` reply,
  `ArgumentError` class) now emits `[:raxol, :harness, :seal, :write_failed]`
  per refused write with `%{index, kind}` metadata (`surface.ex`, the
  `commit_walk/5` closure in `paint_pending_blocks/1`). Pinned by describe 5
  ("every refused seal write emits ... a persistent refusal is observable, not
  silent"), `mode: :always`, green -- this is exactly my prior repro (painted_count
  stays 0 across frames) now carrying a signal on every iteration.
- **LOW teardown flush -- FIXED.** `@modes_off` gains an unconditional leading
  `\e[?2026l` (`inline_driver/sequences.ex`), shared by `teardown_bytes/1` and
  `suspend_bytes/1`, so a sync-close stranded on the session's final frame (no
  later frame to heal on) un-wedges at process exit / editor suspend. Pinned by
  two new `inline_driver_sequences_test.exs` cases, green.

### Low

- **[New Hire] `seal_frame/3`'s best-effort close can mask the root error, and its
  own comment is now stale.** `surface.ex` seal_frame rescue does
  `_ = InlineAuthority.sync_close(opened.authority)` then `reraise e`. Its comment
  claims "device failures are tuples, not raises, on this path" -- but THIS commit
  makes that false: a device that dies mid-frame (alive at `sync_open`, dead at the
  seal write) now RAISES `ErlangError` out of `run_seal_frame/3`. The rescue then
  calls `sync_close` on that same dead device, whose `sync_write/2` fail-fasts and
  re-raises its OWN `ErlangError` -- so `reraise e` never runs and the original seal
  error is masked by the close's error. Fail-fast still holds (a loud crash either
  way), so this is cosmetic: only the surfaced stacktrace is wrong (the close, not
  the root cause), and the comment misleads a future reader. Narrow window
  (device dying between two same-frame writes), non-blocking.

- **[Security Auditor] Permanence is a denylist-of-one, not an allowlist.** Fail-fast
  hinges on `original: :terminated` being the ONLY permanent `ErlangError`
  (`inline_authority.ex:551`). A genuinely-gone device that instead answers
  `{:error, :closed}`/`{:error, :ebadf}` surfaces as `ArgumentError` -> retryable ->
  the (now observable, but still unbounded) loop. Acceptable given the telemetry
  channel and the higher-layer stall/doom-loop detector (PR #603/#601), and the
  test empirically confirms a closed `StringIO` yields `:terminated`, so the
  canonical case is covered. Flagging that the classification is a permanence
  denylist rather than a positive liveness check -- worth a note if other io
  transports (raw fd / pty) ever drive this path. No content leaks via the new
  telemetry (metadata is `index` + `kind` only, no bytes); the unconditional
  `?2026l` backstop is a harmless set/reset with no injection surface.

### Informational

- **[Saboteur] The alive-refusal loop is observable but still silent in the control
  channel.** `advance/3` returns `{model, :ok}` forever on a persistent alive
  refusal (`painted_count` never advances, never `:done`). Observability is
  telemetry-only: a naive `advance`-until-`:done` driver still spins at 100% CPU
  and learns nothing from its return value unless it has attached a telemetry
  handler. Defense-in-depth (the doom-loop/stall detector one layer up) covers
  this, and a bounded retry would strand a recoverable block -- so this is a
  reasonable, documented design tradeoff, not a defect. Noted only because the
  "observable from the first frame" claim is true for telemetry consumers, not for
  the frame return contract.

### Verdict: CONCERNS (non-blocking)

The targeted residual is materially resolved: a dead device now fail-fasts loudly
(bounded, verified by `assert_raise`) and a persistent alive refusal is now
observable per-frame via `[:raxol, :harness, :seal, :write_failed]` (verified green,
including the `mode: :always` repro of my prior 0-progress loop); the LOW teardown
backstop is in. Remaining items are all LOW/informational: one cosmetic
error-masking + stale-comment interaction the round-2 fail-fast newly introduced in
`seal_frame/3`'s rescue, a permanence-classification note, and the by-design
telemetry-only control channel. None block merge.

PR #621, opened 2026-07-16
Harness UI: command palette, jump picker, session picker (+ a footer-grow seal-overwrite fix)

merklebonsai:
The overlay primitive's first three consumers, plus a real substrate bug the work exposed.

## The substrate fix first (`a22d873` — reviewers: read this commit alone)

The session-switch test exposed a pre-existing data-corruption bug in `InlineAuthority.set_footer_rows/2`: growing the footer over a PARTIALLY-filled history under-scrolled by one row, and the next `seal/2` silently overwrote sealed content — an immutable-prefix violation. Every prior test covered only the fully-packed case. Fixed as a one-line unification derived from `append_sealed/2`'s loop invariant; the regression test was proven red against the old formula (18/19) and green on the fix.

## The pickers (`d2ce51b`)

- **Ctrl+P -> command palette** (`:always` — reachable mid-compose, ctrl chords can never be typed text; follows the table's Ctrl+E precedent, verified through the real ANSI parser). **g -> jump picker**, **s -> session picker** (`:not_composing`, the browse-letter class; both become filter text while any overlay is open via the merged guard).
- **Registry = the keymap's own bind table**: an optional `label:` field; `palette_binds/0` is the labeled subset in table order — a new labeled bind appears in the palette with zero other changes, pinned by tests that iterate the table rather than a hardcoded list. Picking dispatches through the SAME `dispatch_command/2` a keypress uses; a shared context builder guarantees a palette pick and a live keypress can never see different guard contexts. (`command_for/2` was previously referenced aspirationally in a moduledoc — now real; lane-law violation fixed.)
- **Jump picker**: blocks labeled kind + first-line summary, picking sets focus through the existing jump machinery.
- **Session switch semantics (documented)**: sealed history stays byte-identical (SealOracle-checked); the abandoned session's un-painted pending blocks are DROPPED, never sealed late (pinned); replay state resets; authority/geometry/composer draft survive; the new session appends below.
- **Fuzzy ranking**: `OverlayPicker.fuzzy_filter/3` is a pure adapter over the merged `ListScorer.rank/4` (order-agreement pinned against rank directly); the filter seam needed no rework.

## Verification

Red-first: 20/22 spec tests failing pre-implementation for exactly the right reasons. Spec 30/30; harness dirs 672 passed; full suite 6004/6009 (5 pre-existing environmental — pty spawn timeouts under load, all pass in isolation). surface.ex additive-only with `footer_lines/1` untouched for the open #615 collision. Warnings + format clean.

## Backlog note (pre-existing, out of scope)

`editor_session_test.exs:414` splits the tmp-dir name on `_`, but the random segment is url-base64 whose alphabet includes `_` — occasionally self-truncates. Flaky by construction; needs a different delimiter or a regex.

### DIFF
--- a/lib/raxol/harness/surface.ex
+++ b/lib/raxol/harness/surface.ex
@@ -414,6 +414,40 @@ defmodule Raxol.Harness.Surface do
   back to, so `:edit_draft` there seals one honest history line saying
   so instead of pretending.
 
+  ## The pickers (command palette, jump, session)
+
+  Three more `OverlayPicker` consumers ride the same footer-overlay
+  substrate as the picker described above. Ctrl+P (`Keymap`'s
+  `:open_palette`, an `:always` chord) opens the command palette from
+  anywhere, including mid-compose -- a chord is never typed text, the same
+  reasoning as Ctrl+E. `g` (`:open_jump_picker`) and `s`
+  (`:open_session_picker`) are plain printable letters gated
+  `:not_composing`, the same class as `z`/`j`/`k`: they only resolve in
+  transcript-browse mode, never stealing a letter out of the composer's
+  typed text.
+
+  The palette's entries are `Keymap.palette_binds/0` (the labeled subset
+  of the bind table) plus two commands that exist only at THIS assembly
+  layer, not in `Keymap.binds/0` -- "focus transcript" and "focus
+  composer" (the very transition `focus_transcript/1`/`focus_composer/1`
+  already exposes as direct API). Picking any palette entry dispatches
+  through the exact same `dispatch_command/2` path a keypress takes --
+  there is no second, parallel execution mechanism for a palette-picked
+  command. All three pickers (palette, jump, session) opt into
+  `Raxol.UI.Harness.OverlayPicker.fuzzy_filter/3` as their `filter_fn`
+  (the `Raxol.UI.ListScorer` adapter), not the default substring filter --
+  a fuzzy-ranked query is what a "type a few letters, find the entry"
+  picker needs.
+
+  Session-switch semantics (`s`, `switch_session/2`) are stated plainly:
+  the abandoned session's sealed history stays byte-identical above --
+  print-once, the substrate cannot rewrite it -- while its not-yet-painted
+  PENDING blocks (the one-block foldable window) are DROPPED, never
+  sealed late. Replay state (`events`/`revealed`/`projection`/
+  `painted_count`/`fold_overrides`/`focused_index`/`status`) resets, the
+  new session's events append below whatever is already sealed, and the
+  composer draft plus authority/geometry survive the switch untouched.
+
   ## Precondition #7 -- teardown ownership (this module owns NONE)
 
   `new/2` sets the DECSTBM history/footer split via
@@ -442,6 +476,7 @@ defmodule Raxol.Harness.Surface do
   shipped for callers to reuse.
   """
 
+  alias Raxol.Harness.Fixture
   alias Raxol.Harness.Fixture.Session
   alias Raxol.Harness.Projection
   alias Raxol.Harness.RecencyPolicy
@@ -461,6 +496,7 @@ defmodule Raxol.Harness.Surface do
   }
 
   @default_footer_rows 6
+  @default_sessions_dir Path.join(["test", "fixtures", "harness", "sessions"])
   @stub_interrupt_notice "» interrupt requested (stub — no agent lane in fixture mode)"
 
   @type mode :: :inline_log | :tmux_conservative | :flat
@@ -484,7 +520,8 @@ defmodule Raxol.Harness.Surface do
           overlay: overlay() | nil,
           editor_session: module() | (String.t(), keyword() -> term()) | nil,
           editor_opts: keyword(),
-          unread: UnreadDivider.t()
+          unread: UnreadDivider.t(),
+          sessions_dir: Path.t()
         }
 
   @typedoc """
@@ -533,6 +570,10 @@ defmodule Raxol.Harness.Surface do
       `:env` -- see `Raxol.Harness.EditorSession`'s trust-boundary
       section). Model-owned `device`/`rows`/`width` always win over
       entries here.
+    * `:sessions_dir` (default `#{inspect(@default_sessions_dir)}`, the
+      same source `examples/harness_fixture_demo.exs` reads) -- the
+      directory `list_fixture_sessions/1` (the session picker, `s`) lists
+      `.jsonl` fixtures from.
 
   ## Startup mode notice (the degradation ladder's `select_with_reason/3` seam)
 
@@ -602,7 +643,8 @@ defmodule Raxol.Harness.Surface do
       overlay: nil,
       editor_session: Keyword.get(opts, :editor_session),
       editor_opts: Keyword.get(opts, :editor_opts, []),
-      unread: UnreadDivider.new()
+      unread: UnreadDivider.new(),
+      sessions_dir: Keyword.get(opts, :sessions_dir, @default_sessions_dir)
     }
 
     model
@@ -649,13 +691,21 @@ defmodule Raxol.Harness.Surface do
   # no-op) -- seal the notice as the first history line instead, through
   # the same `FlatAuthority.seal/2` path `seal_block/2`'s `:flat` clause
   # uses for every other line.
-  defp apply_mode_notice(%{mode: :flat} = model, text) do
+  defp apply_mode_notice(%{mode: :flat} = model, text),
+    do: seal_flat_notice(model, text)
+
+  defp apply_mode_notice(model, text), do: %{model | stub_notice: text}
+
+  # Seals one honest, plain-rendered history line through the same
+  # `FlatAuthority.seal/2` path every flat-mode notice uses (this
+  # function's own `:flat` clause above, and `picker_refusal/2`'s
+  # `:no_footer` case below) -- one source of truth for "how flat mode
+  # writes an honest one-line notice", never a re-derived copy.
+  defp seal_flat_notice(model, text) do
     lines = ViewText.lines(%{type: :text, content: text}, model.width, :plain)
     iodata = Enum.map(lines, &(&1 <> "\n"))
     %{model | authority: FlatAuthority.seal(model.authority, iodata)}
   end
-
-  defp apply_mode_notice(model, text), do: %{model | stub_notice: text}
 
   defp build_authority(:flat, device, width, rows, _footer_rows, _caps),
     do: FlatAuthority.new(device, width, rows)
@@ -1052,12 +1102,7 @@ defmodule Raxol.Harness.Surface do
       | unread: UnreadDivider.input_activity(model.unread, unread_offset(model))
     }
 
-    context = %{
-      composing?: model.composing?,
-      streaming?: not Map.get(model.status, :turn_completed, false),
-      focused_block_id: model.focused_index,
-      overlay_open?: model.overlay != nil
-    }
+    context = keymap_context(model)
 
     model =
       case Keymap.resolve(norm, context) do
@@ -1066,6 +1111,19 @@ defmodule Raxol.Harness.Surface do
       end
 
     paint_footer(model)
+  end
+
+  # The one construction of `Keymap.resolve/2`'s mode context -- shared
+  # with `palette_command/2`'s `Keymap.command_for/2` invocation below, so
+  # the two can never drift apart (a palette-picked bind and a live
+  # keypress must see the identical context shape).
+  defp keymap_context(model) do
+    %{
+      composing?: model.composing?,
+      streaming?: not Map.get(model.status, :turn_completed, false),
+      focused_block_id: model.focused_index,
+      overlay_open?: model.overlay != nil
+    }
   end
 
   # While an overlay is open, EVERY :passthrough event (typed characters,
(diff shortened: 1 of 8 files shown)


DROOdotFOO (commented):

Scope: the substrate fix (`a22d873`) read in isolation, then the three picker
consumers. Substrate fix was reproduced in an isolated worktree.

### Substrate fix verification (the highest-value part) -- VERIFIED CORRECT (reproduced)

`grow_reclaim_count/3` (`inline_authority.ex:789` old -> `max(next_row - new_bottom, 0)`)
holds the immutable-prefix invariant across every fill state:

- **empty history** (`next_row == 1 <= new_bottom`) -> `k = 0`, no scroll, boundary row already blank.
- **partially filled** (`new_bottom < next_row < old_bottom`) -> `k = next_row - new_bottom`; after eviction content fills `1..new_bottom-1` and row `new_bottom` (the new `next_row`) is blank. The pre-fix `max(next_row - 1 - new_bottom, 0)` evicted one row too few, leaving real content ON row `new_bottom` while `next_row` claimed it blank -> the next `seal/2` overwrote it (`append_sealed/2` targets `min(next_row, bottom)`, `inline_authority.ex:406`).
- **full/steady-state** (`next_row == old_bottom`) -> `k = old_bottom - new_bottom`, identical to the retired special case; content stays `1..new_bottom-1`, boundary blank.

`k <= next_row - 2 < content_rows` in every case, so evicted rows go to native
scrollback (prefix preserved), never off the top of live content. shrink/resize
paths are untouched by this commit and independently keep the invariant.

Reproduced: checked out PR head in a worktree, reverted ONLY the formula to
the pre-fix branch -> `overlay_picker_surface_test.exs` went **18/19** (exactly the
new "grow over a PARTIALLY-filled history" regression fails on `SealOracle.immutable_prefix?`);
restored the fix -> **19/19**. The fix is load-bearing and correct. No surviving
immutable-prefix / seal-overwrite violation.

### MEDIUM

- **[New Hire / Saboteur] `keymap.ex:407-412` + `surface.ex:1063-1065`** -- vestigial `:fold_toggle` payload / contract that the dispatcher does not honor. `command_for/2`'s `build_command` threads `context.focused_block_id` into `%{type: :fold_toggle, payload: %{block_id: ...}}`, and `picker_binds_test.exs` pins this as "invocation parity made real." But the live dispatcher `dispatch_command(model, %{type: :fold_toggle}) -> apply_fold_toggle(model, model.focused_index)` ignores `payload` entirely and reads `model.focused_index`. The payload is dead for both the keypress and palette paths (`context.focused_block_id` is itself just `model.focused_index`, `surface.ex:999`). Failure scenario: a maintainer "fixes" dispatch to consume `payload.block_id`, or changes the palette's context construction, and the pinned test keeps passing while behavior silently decouples -- the test guards a value nothing reads. Fix: either have `dispatch_command` consume `command.payload.block_id` (make the contract real) or drop the payload threading and the test assertion, and stop the moduledoc calling it load-bearing.

- **[Saboteur / Security Auditor] `surface.ex:287,334-345` (`open_session_picker` / `list_fixture_sessions`)** -- unbounded, input-thread filesystem + O(N) scoring per keystroke. Every `s` press runs `File.ls/1` on `model.sessions_dir` synchronously on the input path; every `.jsonl` entry becomes a picker item, then `OverlayPicker.fuzzy_filter/3` re-runs `ListScorer.rank` over ALL items on each subsequent keystroke. There is no cap on item count and `sessions_dir` is a public `Surface.new/2` opt. `Fixture.load/1` then `File.read`s the whole selected file into memory (`fixture.ex:97`). Failure scenario: a large or externally-populated sessions directory (thousands of `.jsonl`) stalls the render thread on open and on every filter keystroke; a large fixture file OOMs/stalls on pick. Fix: cap the listing (e.g. take N + "more..."), and/or move `File.ls` + `Fixture.load` off the synchronous input path. (Base LOW each; promoted to MEDIUM by cross-persona agreement.)

### LOW

- **[Saboteur] `surface.ex:1063-1065` via the palette** -- guard bypass with silent no-op. `:open_palette` is `:always`, so the palette opens mid-compose; its entries include the `:not_composing` binds ("toggle fold", "next/previous block", "jump to block", "switch session"). Picking "toggle fold" while composing dispatches with `model.focused_index == nil` -> `apply_fold_toggle(model, nil)` is a no-op (`surface.ex:1265`) with NO notice, so the user picks an action and nothing visibly happens. Not a crash (overlay-stacking is correctly prevented -- `on_pick` runs after `close_overlay`, `surface.ex:1023-1028`), but inapplicable entries should be filtered or should emit an honest refusal notice.

- **[New Hire] `overlay_picker_surface_test.exs` (added regression)** -- dead-binding noise: `auth = InlineAuthority.seal(auth, "sealed after grow\r\n")` immediately followed by `_ = auth`. The rebind is unused; drop it or assert on the returned `auth`.

### Verified-safe (not findings, checked because the task asked)

- **[Security Auditor] session-name / `sessions_dir` -> footer injection: DEFENDED.** Filenames and paths interpolate into picker labels and `stub_notice` ("switched to session #{name}", "no fixture sessions found in #{dir}"), all rendered through the footer path, which sanitizes twice: `ViewText.sanitize/1` strips every C0 byte incl. ESC 0x1B and DEL (`view_text.ex:221-226`), then `ContentGuard.sanitize_line/1` in `InlineAuthority.repaint` (`inline_authority.ex:568-570`). No ESC/cursor/clear injection reaches the terminal; footer content is never sealed, and the sealed history path independently sanitizes via `seal/2`.
- **[Security Auditor] no path traversal.** `list_fixture_sessions` names come from `File.ls/1` basenames (no `/`), and `Fixture.load` builds `Path.join(dir, name <> ".jsonl")`; a picked name cannot escape `sessions_dir`. `Fixture.load` uses `Jason.decode` with no atom creation -- no atom-exhaustion vector.
- **[Saboteur] focus/dispatch routing while a picker is open.** `g`/`s`/`z`/`j`/`k` (`:not_composing`) are suppressed by the guard's `overlay_open?` read (`keymap.ex:390-394`); only `:always` Ctrl+P reaches dispatch with an overlay open and is caught by the `overlay_already_open` clause (`surface.ex:1065-1067`). ESC ordering is correct: `:overlay_dismiss` (guard `:overlay`) precedes `:interrupt` (`:always`) in the table, so ESC dismisses when open and interrupts when closed.
- **The substrate fix is well-documented against regression** -- the derivation comment states the `append_sealed/2` loop invariant explicitly and the named regression test pins it.

### Cross-persona overlaps

- Unbounded session-picker resource use flagged independently by Saboteur (input-thread stall / huge lists) and Security Auditor (externally-populated dir DoS) -> promoted LOW -> MEDIUM.
- The vestigial `:fold_toggle` payload was reached from two angles (New Hire: misleading contract; Saboteur: guard-bypass no-op shares the same dispatch line) -> kept at MEDIUM.

### Verdict: CONCERNS

The headline substrate fix is correct and verified across all fill states
(reproduced red->green); no CRITICAL, no surviving immutable-prefix violation.
Two MEDIUM maintainability/robustness issues (a pinned-but-unhonored payload
contract, and an uncapped input-thread session listing) warrant a follow-up but
do not block. Footer injection and path-traversal surfaces were checked and are
defended.

merklebonsai:
Review response landed as `58533f4`. Ledger:

**Substrate fix — verified correct by the reviewer** (independent red-green reproduction). No finding touched the reclaim formula; no action.

**MEDIUM — vestigial fold payload.** *Confirmed, fixed.* Dispatch now consumes `command.payload.block_id`, leaving exactly one producer chain (context -> payload -> dispatch) shared by keypress and palette. Honest red-first caveat: the two values could not diverge through any public path (both read from the same model at the same instant — a maintainability hazard, not a bug), so no black-box red exists for the consumption; the parity + end-to-end fold tests now exercise the payload as sole source. Dropping the payload was rejected: the shape is pre-existing keymap vocabulary pinned by the original suite and mirrors the wire shape.

**MEDIUM — unbounded session listing / input-thread work.** *Confirmed for the listing; the off-thread proposal partially refuted.* Listing capped at 100 sorted entries with the truncation NAMED in the picker title ("first 100 of 120"), red-first (captured 120-file failure). Off-thread loading rejected with rationale: this Surface is a synchronous pure state machine by design — fixture mode has no other thread; the synchronous load is now documented plainly rather than hidden.

**LOW — silent no-op palette pick.** *Confirmed, fixed red-first on both paths:* fold-with-no-focus now emits "no block focused — jump to a block first (g)" via the same one-frame notice mechanism as the sealed-block refusal; two new tests. The palette doc states the always-live-chord => honest-refusal contract.

**LOW — dead rebind.** *Confirmed, fixed* (the same pre-existing pattern elsewhere in the file left untouched — surgical scope).

Gates: harness scope 675 passed (+3 red-first), full suite 6132/6133 (the one failure is the known pre-existing MetricsCollector environmental, stash-verified, zero file overlap), warnings + format clean.

DROOdotFOO (commented):

Since last review: d2ce51b6..58533f45

Scope: the one new commit `58533f45` targeting my two prior MEDIUMs, plus a
recheck of the two LOWs. Verified against `lib/raxol/harness/surface.ex`,
`lib/raxol/ui/harness/keymap.ex`, `lib/raxol/ui/harness/overlay_picker.ex`,
and the new tests. Affected harness suites run green in an isolated worktree
(35 passed: `command_palette_surface_test.exs` + `overlay_picker_surface_test.exs`).

- **MED (fold payload now honored).** `dispatch_command/2` no longer re-reads
  `model.focused_index`; it consumes the command's own payload
  (`surface.ex:1133-1134`: `%{type: :fold_toggle, payload: payload}` ->
  `apply_fold_toggle(model, Map.get(payload, :block_id))`). Both producers feed
  it from one place: the live-keypress path (`Keymap.resolve/2` ->
  `build_command/2`, `keymap.ex:413`) and the palette path
  (`Keymap.command_for/2` -> `build_command/2`, `keymap.ex:388`) BOTH route
  `:fold_toggle` through `build_command(%{command_type: :fold_toggle}, context)`
  which sets `payload: %{block_id: Map.get(context, :focused_block_id)}`
  (`keymap.ex:494-498`), and `context.focused_block_id` comes from the single
  `keymap_context/1` builder shared by keypress and palette
  (`surface.ex:1063-1070`, `1622-1623`). The pinned contract is now truthful:
  a palette pick toggles the block the payload names, not a second live read of
  the model. No off-by-one -- `block_id` here IS the block index
  (`focused_block_id: model.focused_index`), and `apply_fold_toggle/2`'s
  `index < model.painted_count` guard is unchanged. Behavior-preserving in the
  current tree (both reads resolve on the same model at the same instant), so
  this is a decoupling-hazard removal, correctly scoped.

- **MED (per-keystroke session scoring now bounded) (reproduced).** The cap is
  applied to the item list handed to the overlay (`cap_session_names/1`,
  `surface.ex:1689,1701-1706`) BEFORE the picker is populated, so the
  per-keystroke ranker (`OverlayPicker.fuzzy_filter/3` ->
  `ListScorer.rank/3`, `overlay_picker.ex:160-161`, which scores over
  `picker.items`) can never exceed 100 items per keypress. Reproduced: the new
  test seeds 120 files and asserts `length(picker.items) == 100`
  (`command_palette_surface_test.exs`), i.e. the DoS vector I flagged
  ("File.ls + O(N) fuzzy scoring per KEYSTROKE") is genuinely closed -- the
  keystroke path is now O(100), not O(N). Truncation is surfaced, not silent:
  the title reads `"session — first 100 of 120"`.

- **LOW (mid-compose fold no-op).** `apply_fold_toggle(model, nil)` now sets an
  honest one-frame notice (`surface.ex:1368-1370`, `"no block focused"`) instead
  of returning the model unchanged. Covered by two new tests (palette pick of
  "toggle fold" mid-compose, and `z` in transcript-browse pre-jump).

- **LOW (dead test binding).** The `auth = ...; _ = auth` rebind in the
  footer-grow regression is gone, replaced by `_auth = InlineAuthority.seal(...)`
  (`overlay_picker_surface_test.exs:603`).

### Low

- **Session listing is only bounded on the keystroke path; the one-time open is
  still O(N).** `list_fixture_sessions/1` runs `File.ls/1` + `Enum.sort/1` over
  the FULL directory every time the picker opens (`surface.ex:1737-1742`); the
  cap's `Enum.take/2` is applied post-sort, so a pathological `sessions_dir`
  still fully materializes and sorts on each `s` press. This is a discrete user
  action (not per keystroke) and Elixir's stdlib has no streaming `ls`, so the
  sort is the only avoidable part -- acceptable, but the "bounded work on the
  input path" docstring slightly oversells it. Residual, not a blocker.

- **`Fixture.load/1` on pick remains an unbounded whole-file `File.read`**
  (`fixture.ex:96-97`), on a public `sessions_dir`. A pathological `.jsonl`
  pauses/OOMs the synchronous surface on a single pick. This is explicitly
  documented in `open_session_picker/1`'s "Listing cap" section as acceptable
  for fixture mode rather than fixed -- honest, and low-severity given picks are
  discrete, but it is a residual (documented, not eliminated).

- **The cap keeps the lexicographically-first 100 names, so sessions past the
  cap are unreachable via fuzzy filter (reproduced).** `picker.items` is exactly
  the first 100 sorted names (test asserts `List.last == "session-100"` for a
  120-file dir), and `fuzzy_filter/3` only ranks over `picker.items` -- so
  typing `115` for `session-115.jsonl` matches nothing even though the file
  exists. The title's "first 100 of 120" surfaces that truncation happened, so
  it is not silent, but "which 100" is lexicographic, not recency/relevance.
  Fine for the small-by-construction fixture dir; worth a one-liner if the
  picker ever points at a real session store.

- **The stricter `:fold_toggle` head can silently swallow a payload-less
  command.** `dispatch_command(model, %{type: :fold_toggle, payload: payload})`
  now requires a `:payload` key; a `:fold_toggle` map without one would fall
  through to the `_other` catch-all (`surface.ex:1203`) and no-op. No in-tree
  producer emits that shape (both go through `build_command/2`, which always
  attaches a payload), so this is latent only -- but it re-opens a silent-no-op
  crack for exactly the command whose silent-no-op you just closed. A
  `%{type: :fold_toggle} = cmd -> apply_fold_toggle(model, cmd[:payload][:block_id])`
  or an explicit assertion would be more honest than leaning on the catch-all.

### Verdict: CLEAN

Both MEDIUMs are genuinely cleared: the fold payload is consumed (single
producer chain, contract now truthful, no off-by-one), and the per-keystroke
session scoring is capped before the O(N) ranker runs (reproduced via the new
120-file test). Both LOWs are handled. What remains are low-severity, mostly
documented residuals (one-time open sort, whole-file load on pick, lexicographic
cap reachability, a latent payload-less no-op) -- none block merge.

PR #622, opened 2026-07-16
test: fix the editor tmp-dir flake (url-base64 underscore split)

merklebonsai:
`editor_session_test.exs:414` extracted the random tmp-dir segment by splitting the name on `_` — but the segment is url-base64, whose alphabet includes `_`, so ~1 run in 8 self-truncated it and failed the length assertion. It just failed the macOS matrix on two unrelated PRs (#615, #618) back to back, and was independently diagnosed by the palette track's full-suite sweep.

The random suffix is always exactly 8 chars (6 bytes, url-base64, no padding): assert on the last 8 characters directly. Deterministic — verified by repeated local runs.

### DIFF
--- a/test/raxol/harness/editor_session_test.exs
+++ b/test/raxol/harness/editor_session_test.exs
@@ -444,11 +444,13 @@ defmodule Raxol.Harness.EditorSessionTest do
     # the per-run dir name carries a strong-random segment, honoring the
     # confidentiality doc's unpredictability claim (round-2 review: the
     # doc said strong_rand_bytes but the name was counter+timestamp) --
-    # 6 random bytes -> 8 url-base64 chars as the final `_`-separated part
-    assert [random_segment | _] =
-             draft_dir |> Path.basename() |> String.split("_") |> Enum.reverse()
+    # 6 random bytes -> exactly 8 trailing url-base64 chars. Asserted on
+    # the last 8 CHARS, not a `_` split: url-base64's alphabet includes
+    # `_`, so splitting on it self-truncated the segment ~12% of runs
+    # (flake found failing two unrelated PRs' CI).
+    random_segment = draft_dir |> Path.basename() |> String.slice(-8, 8)
 
-    assert String.length(random_segment) >= 8
+    assert String.length(random_segment) == 8
     assert random_segment =~ ~r/\A[A-Za-z0-9_-]+\z/
     refute random_segment =~ ~r/\A[0-9]+\z/
 


DROOdotFOO (commented):

Scope: 6-line test fix replacing a `String.split("_")` extraction of the tmp-dir
random segment with `String.slice(-8, 8)`. Confirmed the generator at
`lib/raxol/harness/editor_session.ex:498-503` builds the name as
`counter_timestamp_random` where `random = 6 |> :crypto.strong_rand_bytes() |> Base.url_encode64(padding: false)`.

### Saboteur

- Deterministic: CONFIRMED. 6 bytes is a multiple of 3, so base64url emits
  exactly 8 chars with no padding regardless of input; `padding: false` is a
  no-op here. `counter` and `timestamp` are digits only (no `_`/`-`), so the
  trailing 8 chars of the basename are always exactly the random segment.
  `String.slice(-8, 8)` sidesteps the alphabet entirely, so it is immune to
  BOTH url-base64 specials (`_` and `-`), not just the `_` that caused the
  flake. This is the right shape of fix.
- Minor (non-blocking) vacuity: `String.length(random_segment) == 8` is now a
  tautology -- `slice(-8, 8)` on a name that is always >= 8 chars can only ever
  return 8 chars, so the assertion tests `String.slice`, not the generator's
  output width. The real teeth are the two lines below it:
  `=~ ~r/\A[A-Za-z0-9_-]+\z/` (alphabet) and `refute =~ ~r/\A[0-9]+\z/`
  (proves the captured segment is the random tail, not a numeric field). Those
  keep the test meaningful, so this is not fully vacuous -- just note the length
  line no longer earns its keep. If you wanted it to bite, extract by stripping
  the `counter_timestamp_` prefix and assert the remainder length.
- No sibling instances: the only other `String.split` in the file is line ~104
  splitting on `'` (shell-quote parsing), which is not in the base64 alphabet
  and is safe. No other alphabet-split flake in this file.

### New Hire

- Clean. The replacement comment explicitly states WHY the old approach was
  wrong ("url-base64's alphabet includes `_`, so splitting on it self-truncated
  the segment ~12% of runs"), which makes accidental reintroduction unlikely.
  Self-documenting.

### Security Auditor

- Clean. The confidentiality guarantees (file `0o600`, dir `0o700`) are still
  asserted unchanged (lines 438-439). Directory uniqueness is provided by the
  monotonic `unique_integer`, not the random suffix, so a shorter/predictable
  suffix could not cause a collision; the random exists only for
  unpredictability, and the test still verifies its presence and non-numeric
  shape. Nothing here weakens the earlier editor-suspend O_EXCL/0600 work.

### Verdict: CLEAN

A correct, well-documented, deterministic flake fix. It fully handles the
url-base64 alphabet by not splitting on it at all. One cosmetic note: the
`== 8` length assertion is now tautological given the slice, but the alphabet
and non-numeric regex assertions preserve the test's meaning.

PR #624, opened 2026-07-17
feat(acp): raxol_agent_client_protocol — first Elixir/OTP Agent Client Protocol impl + durable-sessions moat

merklebonsai:
## What this is

The first **Elixir/OTP implementation of the Agent Client Protocol** (ACP — agentclientprotocol.com), the JSON-RPC 2.0 protocol between code editors and AI coding agents (the protocol Zed and a growing ecosystem speak). New package `packages/raxol_agent_client_protocol/`, module root `Raxol.AgentClientProtocol`.

> Not to be confused with `raxol_acp` — the Virtuals **Agent Commerce Protocol** (on-chain payments). This is the editor↔agent **Client** protocol.

Full ACP v1 surface, **both roles** (agent and client), bidirectional — including the agent→client request direction (`session/request_permission`, `fs/*`, `terminal/*`). OTP-native: one supervised process per connection, per-session processes under `DynamicSupervisor`+`Registry`, supervised per-request dispatch, fail-closed permission flow. Pluggable transports (`stdio` + in-process paired).

## The moat — durable resumable sessions (no other ACP impl has this)

The differentiator, shipped as a vendor extension on `_meta["raxol.io"]` + `_raxol/*` methods:

- **Offset-based reattach/replay** with a register-before-high-watermark seam — a reattaching client reads history from an offset then the live tail, and the delivered sequence provably equals the durable record stream (P-JS5 closure: no gap, no dup). Mid-turn reattachers finalize from a persisted `turn_completed` instead of stranding.
- **Ed25519 offline-verifiable attach authorization** — a capability token (`RXC1.<claims>.<sig>`, no `alg` field so algorithm-confusion is structurally unexpressible) verifiable from the token + a shipped public key alone, so even a writerless tar'd session dir authorizes attach. Fail-closed `AttachPolicy` (every non-`{:ok, grant}` outcome denies).
- **Taint annotate-never-filter** — every record carries its taint; the bus stamps and delivers, never drops by taint (a filter would break closure).

## Provenance & license (pure-MIT)

- **`Schema.*` ported from `f1729/agent_client_protocol` (MIT)** — copy-legal, attributed per-module + in `NOTICE.md`, with 4 upstream defects fixed (CRLF framing, connection-killing hard-match decode, ExtResponse KeyError, unsupervised dispatch).
- **Runtime is clean-room.** The Apache-2.0 official SDKs + `acpex` + `grok-build` were studied as *design references only* — zero shipped source lines.
- **Conformance corpus from `openclaw/acpx` (MIT)** — 21 cases copied verbatim.
- The official ACP JSON Schema is a SHA256-pinned **dev/test oracle** (`priv/schema-oracle/`), excluded from the Hex package.

## Conformance

- **acpx 21/21** cases pass through a native Elixir runner (no Node needed in CI).
- **P-BUS1–7** red suites for the reattach seam, each with its named dead-injector negative control.
- **End-to-end** integration through the *real* assembled supervision tree: full agent↔client conversation, a second client reattaching and closing P-JS5 over a real journal Writer + real AttachPolicy, mid-turn orphan-repair finalization, denied-attach `-32000`.
- **Wire-byte fixtures** (from f1729, MIT) + StreamData round-trip properties + skew/torture (unknown-field pass-through, >1MB multimodal frames, id-type matrix, three-state fields, 5000-garbage-frame totality).
- **751 tests green** (16 properties); `mix compile --warnings-as-errors` clean.

## Three adversarial danger-zone gates passed

- **G2** (runtime design triad) — bidirectional correlation, cancellation matrix, deadlock-freedom; 28 findings folded.
- **G5** (security triad on the moat) — token forgery/replay/expiry, P-JS5 dup/gap injectors, fail-closed exhaustiveness; 6 BLOCKINGs designed out on paper before impl.
- **G6** (final adversarial, this PR) — Opus package review + grok-4.5/longcat triad. **Zero BLOCKING; runtime/OTP/security spine ruled merge-ready.** Three MAJORs fixed here red-first: a token surface-check fail-open (security), strict decode type-checking on known load-bearing fields (`-32602`, ruled STRICT), and a non-booting README quickstart.

## Known followups (rc.1, non-blocking — from G6)

- `Connection.transport_kind/1` accessor so `ctx.transport` (CDI-2) can't be mis-wired from peer data (currently inject-by-convention).
- fs_sandbox resolve-then-write TOCTOU — openat-style follow-up or document.
- Hard-require `session_sup` for reattach (remove the unsupervised fallback in production paths).
- `best_effort_cancel_notify` discards a future stateful transport's new state (harmless for stdio/paired pid handles).
- A stock `load_session` wiring helper so the moat is turnkey, not a ~30-line handler.

## Notes

- Pre-alpha (`0.1.0-rc.0`), not yet published to Hex.
- Depends only on `jason` at runtime; no `raxol` back-dependency (standalone, BEAM-consumable by Erlang/Gleam).
- Do not merge — for review.

https://claude.ai/code/session_015Rc719iBNk3hPRNKGrUPgN


DROOdotFOO (commented):
## Adversarial Review (automated: 4-dimension decompose + reproduce)

Reviewed across four dimensions -- **fs/terminal/permission**, **transport/JSON-RPC**, **session/durable/reattach**, **schema/version/capabilities** -- reading the package files directly off the branch. This is a strong, deliberately fail-closed port; every top-bar vector (decoder DoS, atom-table exhaustion, session-hijack, absent-defaults-permissive) came back **defended**. One reproduced HIGH and a handful of MEDIUM/LOW below.

> Scope note: the transport / session / schema findings were reviewed at `ef3d313c`; commits `1f8dc42c` (strict decode type-checks), `8a6598a2` (attach surf-check fail-closed), `2d458a78` (supervise reattach Subscriber) landed since and may already resolve some of them -- please re-check those against current head. The **FsSandbox HIGH below was verified against the current head (`61676bee`) and is unfixed.**

### HIGH
- **[Security/Saboteur] `client.ex` `FsSandbox.real_path/2` (~line 730-760) -- symlink-escape; confinement invariant broken (reproduced).** This function is **byte-for-byte the pre-fix `Raxol.Core.Boundary.Path.confine/3` `real_path` that #613 replaced.** It resolves a relative symlink target against `Path.join(Path.dirname(path), target)` -- the **lexical** parent -- not the resolved-real parent, and increments the symlink-depth counter on every *directory* recursion (deep symlink-free nesting >40 wrongly yields `:too_many_symlinks`). Reproduced against the exact FsSandbox logic: with `root/loop -> "."` and `root/esc -> "../secret.txt"`, `FsSandbox.resolve(root, "loop/loop/esc")` returns `{:ok, ".../root/secret.txt"}` (accepted as in-root) instead of `:symlink_escape` -- the sandbox no longer provably confines. #613's review reproduced a full out-of-sandbox read against this identical code (topology-dependent). Because the FsSandbox binds to the shared boundary vectors but those vectors do **not** cover the relative-symlink-under-symlinked-ancestor class, the "a divergence is a red test" guarantee does **not** catch this -- the two implementations have already forked. **Fix:** port the #613 `walk_real` rewrite (resolve components left-to-right carrying a resolved-real base; count only symlink hops), and add the symlink-escape reject vectors to the shared set so both impls are gated on it. Given this is the opt-in `fs_sandbox:` path a "zero-callback" client uses for a remote agent's file access, treat as must-fix before anyone relies on it with an untrusted agent.

### MEDIUM
- **[Security/Saboteur] `connection.ex:693-728,874-901` + `session/supervisor.ex:101` -- unbounded concurrent inbound dispatch (flood DoS).** The dispatch `Task.Supervisor` has no `max_children`, `pending_in`/`reply_refs`/`task_index` grow uncapped, and `stdio.ex:227` pushes every framed line to the mailbox with no read backpressure. A peer streaming many small requests -> unbounded task spawn + map growth -> OOM. The 64 MiB framer cap bounds one frame, not the count. Fix: `max_children` + a `pending_in` cap that sheds (busy/`-32603`) or pauses reads.
- **[Saboteur] `reattach.ex:251,420-431` -- full history replay runs inside a default 5s `GenServer.call`.** A large durable session (the moat use case) exceeds 5s -> the call times out and crashes `attach`/the load handler while the subscriber replies out-of-band. Fix: `:infinity` (the subscriber owns its own expiry) or chunk the emission.
- **[Saboteur] session/reattach -- no idle reaping, no per-connection session cap, grow-only journals.** A long-lived node accrues one live (`:transient`, never-stopped) Writer + an ever-growing ETS table per session ever seen -- unbounded memory growth directly under the "durable sessions" value prop. Fix: idle timeout + session cap; idle-stop the Writer (journal is the durable state); a retention/truncation story.
- **[Security] `schema_agent_types.ex:1317,1387,1432,1527` + `capabilities.ex:83` -- capability booleans decoded with no `is_boolean` check + permissive `truthy?` (anything non-nil/false = enabled), so `0`/`""`/`[]`/`"false"` read as ENABLED.** Not attacker-reachable through the negotiation gate today (it uses the receiver's own snapshot), but a latent type-confusion inconsistent with the strict boolean validation elsewhere. (May be addressed by `1f8dc42c` -- please verify.)

### LOW
- **[Saboteur] `reattach.ex:506-517` -- `Process.send_after` badarg for tokens with `exp` > ~49.7 days** (delay exceeds the OTP `0..4294967295` ms range), crashing the subscriber after `reply_ok` -- a totality gap in a module advertising "verify is TOTAL". Clamp the delay / reject over-long `exp`.
- **[Security] `reattach.ex:205-217` -- the injected `:authorize` seam accepts any `{:ok, map}` incl. `%{}`, bypassing the Runner's `%Grant{}` contract** (actor id, allow-listed scope, session-id match); a grant without `expires_at` arms no force-close timer. Bites only custom-funnel integrators, but re-introduces the dual-ownership weakening. Re-assert the Grant contract in `run_authorize`.
- **[Security] `mem.ex:47` -- journal ETS table is `:public`** (any co-resident process can forge offsets) and its owner is unwired ("integration coder's job"); if opened in `Writer.init`, the table dies with the Writer and the tip-fold-survives-crash guarantee collapses silently. Records hold prompts/tool-output/file-contents in plaintext. Use `:protected`; assert the owner-outlives-Writer contract; note plaintext exposure.
- **[Saboteur] `schema_version.ex:49,65` -- version negotiation accepts any integer (`protocolVersion: 5 -> {:ok, 5}`), no clamp to the known `{0,1}` set;** safety depends on the `Connection` callsite. Clamp/reject above `latest()`.
- **[New Hire] `schema_maybe_undefined.ex` -- the tri-state (absent vs null) is documented "load-bearing" but no decoder uses it** (both collapse to `nil`); a future PATCH field would silently conflate omitted vs null. Quarantine or document as intentionally null-collapsing.
- **[New Hire] `connection.ex:485` -- `:infinity`-timeout `pending_out` entries never reclaimed** if a peer withholds a response (caller-bounded). LOW.
- Plus: `reattach.ex:258` unsupervised-subscriber fallback leak (bare-embed path); `emit_denied` KeyError on a partial injected ctx; Keyring validated per-attach not at boot (fail-closed but no loud boot failure).

### Verified defended (worth keeping)
- **Transport:** stdio framer caps a frame at 64 MiB then enters bounded "skipping" mode (NDJSON-buffer DoS closed); **zero `String.to_atom`/`to_existing_atom` on untrusted input anywhere** (router is compile-time literal clauses); id correlation verified correct (wire-type preserved, dup-id dropped no frame, no double-delivery, batch rejected `-32600`, reentrancy/deadlock guarded).
- **Reattach:** `RXC1` Ed25519 token is fail-closed and sound -- unguessable, constant-time `:crypto.verify`, version-pinned (no downgrade/alg-confusion), `sid`-bound, strict `exp`, keyring from verifier config only, deny-by-default (`LocalNode` denies nil/network transport; Runner isolates policy in a killable task, denies on crash/timeout, single anti-oracle envelope). No session-hijack found. Residual by design: bearer token (replayable until `exp`, ~1h) with no per-token revocation.
- **Journal:** no torn write on a Writer crash (atomic ETS + idempotent genesis/tip-fold repair); replay is read-only (no side-effect duplication on reattach).
- **Schema:** no crash-on-malformed-params (all `from_json` guarded `is_map` + catch-all -> typed errors); permission decode explicitly refuses to default (`PermissionOptionKind.from_json` -> `:invalid_permission_option_kind`, "coercing could turn a reject into an allow"); enum defaults least-permissive.
- **fs/terminal:** no `System.cmd`/`Port.open` in the package -- `terminal/*` handlers are integrator-provided (no in-package command injection); unimplemented callbacks default to `-32601` (fail-closed).

### Verdict: CONCERNS
Strong package; the one must-fix is the **FsSandbox symlink-escape (HIGH, reproduced, unfixed at head)** -- it is the identical code #613 already fixed in the Core copy, so porting `walk_real` + adding the symlink reject vectors closes it. The MEDIUMs (inbound-dispatch DoS, 5s replay ceiling, no idle reaping) are the load-bearing robustness gaps under the "durable sessions" claim.

_Adversarial reviewer: Saboteur / New Hire / Security Auditor personas; fs finding reproduced against branch head._

merklebonsai:
Thorough — thank you. Triaging all of it; per-finding red→green replies as each lands.

**The FsSandbox HIGH is the merge gate and you're exactly right on the mechanism:** the ACP `FsSandbox.real_path` is the *pre-#613* `real_path` — the ACP copy never received the `walk_real` rewrite, and the shared boundary vectors don't cover the relative-symlink-under-symlinked-ancestor class, so the 'a divergence is a red test' guarantee never fired and the two impls forked. That's a real failure of the intentional-duplicate design, not just a missed line. Porting `walk_real` (resolve components left-to-right on a resolved-real base; count only symlink hops) into the ACP copy AND adding the symlink-escape reject vectors to the shared set (so #613's Core copy is gated on them too) — with your exact `loop/loop/esc` repro as the red.

MEDIUMs (unbounded inbound dispatch → `max_children` + `pending_in` shed-cap; the 5s replay ceiling → `:infinity`; idle reaping + session cap + retention) are the load-bearing robustness gaps under the durable-sessions claim — fixing. Re-checking the items you flagged as maybe-resolved by `1f8dc42c`/`8a6598a2`/`2d458a78` against head. The real LOWs (send_after >49.7d clamp, Grant-contract re-assert in the injected funnel, ETS `:protected`, version clamp) too.

Back with the fixes + tripwires shortly.

DROOdotFOO:
Opened #633 against this branch to fix the FsSandbox symlink-escape HIGH from the review above -- it ports #613's `walk_real` (resolves relative symlink targets against the real parent, counts only symlink hops). Verified standalone: the `loop/loop/esc` escape now returns `{:error, :symlink_escape}`; legit in-sandbox + deep-nesting paths still resolve. Suggested follow-up in the PR: add the relative-symlink-under-symlinked-ancestor case to the shared reject vectors so the two confinement impls can't fork silently again.

DROOdotFOO:
**Merge readiness: MERGE #633 FIRST; then the review's one HIGH clears.**

The FsSandbox symlink-escape HIGH from the review is fixed in #633 (opened against this branch) -- land that first and the confinement finding is closed. The other review findings (unbounded inbound-dispatch DoS, the 5s replay-call ceiling, no idle reaping) are non-blocking robustness items, and several adjacent findings were already addressed by later commits. This is also a 38k-line foundational package, so treat the merge itself as a deliberate call independent of the review.


DROOdotFOO:
**Update: the FsSandbox symlink-escape HIGH is now CLOSED** -- #633 merged into this branch (commit 2dee6736), so `FsSandbox.real_path` now uses the `walk_real` resolution and the reproduced escape returns `{:error, :symlink_escape}`. Merge readiness moves to: no open HIGH from the review; the remaining items (inbound-dispatch DoS cap, the 5s reattach replay ceiling, idle-session reaping) are non-blocking robustness follow-ups. Still a 38k-line foundational package, so the merge itself remains a deliberate call. Follow-up still worth doing: add the relative-symlink-under-symlinked-ancestor case to the shared reject vectors so the Core and package confinement impls stay converged.

merklebonsai:
Thanks for the thorough pass — the three-persona method caught exactly the class of thing that survives a self-review. Disposition of every finding below; all code fixes are on the branch at `697d181c8`, each landed **red-first** (the pre-fix failing test named per item).

### HIGH — FsSandbox symlink escape → **closed by #633**, drift-guard gap now closed too
#633's `walk_real` fixes the resolver. The reason my own security pass *missed* it is the more useful finding: the shared reject vectors had no case for the class the bug lived in (a **relative** symlink target reached through a self-referential symlinked ancestor), so the drift guard gave false confidence. Pushed the follow-up you asked for:
- Added `relative_symlink_under_symlinked_ancestor_escape` (`root/loop → "."`, `root/esc → "../secret.txt"`, request `loop/loop/esc`) to the shared reject set — the drift guard now exercises it.
- Added an explicit regression test: escape rejected; deep symlink-free nesting >40 **not** misreported as `too_many_symlinks` (the depth-counted-on-directory-recursion half of the old bug); a real cycle still trips.
- Opening a matching vector add to `raxol_core`'s canonical set so Core and package stay converged (Core already has the correct `walk_real` from #613; this just gates it on the same case).

### MEDIUM — non-blocking robustness, all fixed
- **Unbounded inbound dispatch (flood DoS):** cap on concurrent inbound dispatch; over the cap we **shed** (busy reply) rather than grow `pending_in`/task maps. Red: a peer streaming past the cap gets shed responses, maps stay bounded.
- **5s reattach replay ceiling:** replay now runs under an `:infinity` call (the subscriber owns its own expiry), so a large durable session no longer times out `attach` mid-replay. Red: large-history attach doesn't crash.
- **No idle reaping / session cap:** idle-stop the Writer after a timeout (the journal is the durable state — a fresh reattach respawns + replays, which is the moat design) plus a per-connection session cap. Red: an idle Writer stops and reattach still works; the over-cap session is refused cleanly.
- **Capability booleans decoded permissively:** `decode_bool/2` now admits only a literal boolean; `0` / `""` / `"false"` / `[]` / a nested object all read as **not-advertised** (least-permissive — every capability field's contract is boolean). Red: `fs.readTextFile: 0` decodes as NOT enabled.

### LOW
- **`send_after` badarg for exp > ~49.7d:** clamp the force-close delay to the OTP max. Red: a 60-day-exp grant doesn't crash the subscriber after `reply_ok`.
- **Injected `authorize` bypasses the Grant contract:** `run_authorize` now re-asserts it — an injected authorize returning a non-`Grant` (incl. `{:ok, %{}}`), or a `Grant` failing session-id match / missing required fields, **denies**. Red: injected authorize returning `{:ok, %{}}` denies (telemetry `:non_conforming_return`).

### The two LOWs I'm *not* changing — with the reasons, not just a wave-off
- **Journal ETS `:public`:** by design, now documented. The table owner deliberately **outlives** the single-publisher Writer (the C14 orphan-repair precondition — the journal must survive a Writer crash), so `:protected` (owner-only writes) would break every cross-process append. Integrity doesn't rest on the ETS ACL; it rests on the **structural single-Writer invariant** (exactly one Writer per session). Moduledoc now states this and the residual co-resident-process trust caveat — a documented v1 bound, not an open hole.
- **Version accepts any non-negative integer:** this is intentional faithful decode, and it's already torture-tested. `from_json`/`coerce` carry the stated version unchanged; **negotiation happens at the response layer** — the agent replies with the version *it* speaks. `wire_torture_test.exs:260` drives the full handshake: request `protocolVersion: 999` → response `protocolVersion: 1` → post-handshake traffic flows. Clamping at decode would break that forward-compat property, so I'd rather keep the existing torture test as the tripwire than "fix" a designed behavior. Happy to add a response-layer assertion if you'd prefer the negotiation contract pinned more explicitly.
- **`maybe_undefined` tri-state unused on decode:** documented honestly rather than removed — no v1 decode path produces `:undefined` (an absent field and an explicit `null` both collapse to `nil`); the three-state value is exercised on **encode only** until the Codec-integrated `Maybe` lands. Moduledoc now says exactly that.
- **`:infinity` `pending_out` (connection.ex):** caller-bounded by design (a caller passing a finite `timeout_ms` arms the reclaim timer; `:infinity` is the caller explicitly opting out — documented at the `pending_out` moduledoc). Left as-is.

Full suite green (780) on the combined tree (your #633 + these). One rare, timing-only test flake in the new async reattach/session tests is under a de-flake sweep now — it's test-side (`Process.sleep`/tight `refute_receive`), not a production path; I'll pin and convert it to a deterministic wait rather than leave a wall-clock assert in. Not gating anything.

Issue #616, opened 2026-07-16
ContentGuard passes 8-bit C1 control bytes (0x80-0x9F) through unstripped


DROOdotFOO:
## Summary

`Raxol.UI.Rendering.PaintAuthority.ContentGuard.sanitize_line/1` neutralizes ESC (`0x1B`)-led
escape sequences, C0 controls, and DEL, but **passes 8-bit C1 control bytes (`0x80..0x9F`) through
unchanged**. On any terminal that honors 8-bit C1 controls, this reopens the exact injection class
ContentGuard exists to close -- via the C1 form instead of the `ESC [` form.

## Location

`lib/raxol/ui/rendering/paint_authority/content_guard.ex`

- `scan/2` strips C0 + DEL at line 118 (`when c < 0x20 or c == 0x7F`)
- the catch-all at line 123 passes through **every byte `>= 0x80`** as opaque UTF-8
- moduledoc lines 23-24 codify this: "any byte `>= 0x80` (UTF-8 lead/continuation bytes) -- passed through unchanged"

The grammar is purely byte-wise, so it cannot tell a C1 control code point (U+0080..U+009F, which
in UTF-8 is a two-byte sequence `0xC2 0x80..0xC2 0x9F`) or a raw single C1 byte from a UTF-8
continuation byte, and lets them all through.

## Impact

C1 controls are the 8-bit single-byte equivalents of ESC-led sequences:

| C1 byte | Equivalent |
| ------- | ---------- |
| `0x9B`  | CSI (`ESC [`) |
| `0x9D`  | OSC (`ESC ]`) |
| `0x90`  | DCS (`ESC P`) |
| `0x84`  | IND, `0x85` NEL, `0x88` HTS, ... |

A model/agent-originated line containing `<<0x9B, "2J">>` (C1 CSI + `2J`) is an "erase display" on a
terminal in 8-bit-C1 mode -- it wipes native scrollback / repaints sealed rows from inside the
content, which is precisely the invariant ContentGuard (and the seal-seam nets added in #608)
exist to hold for the `ESC [2J` form. The `ESC`-led form is neutralized; the C1 form sails through.

Practical risk is LOW today (most terminals in UTF-8 mode do not interpret 8-bit C1, and a bare C1
byte is usually invalid UTF-8 that renders as a replacement glyph), but this is a real hole in a
security-boundary sanitizer, and "low because of the terminal's mode" is not a guarantee the boundary
should rely on.

## Evidence

[...]

PR #626, opened 2026-07-17
Harness UI: full-screen diff expansion (footer-region maximization)

merklebonsai:
Inline diff blocks are right for glancing; a real code review needs the whole terminal. `e` on a focused diff block expands it to a near-full-screen scrollable view; ESC/`q` restores the harness byte-identically.

## Mechanism ruling: footer-region maximization (alt-screen rejected on evidence)

The alternative — a bracketed alt-screen visit — was rejected on four grounds, documented with the ruling in the module:
1. No-alt-screen is the inline profile's named headline invariant; even the editor-suspend bracket deliberately releases the region rather than visiting the alt screen.
2. No capability story: zero alt-screen detection exists, and 1049 support is not reliably probeable.
3. **Verification blindness**: the seal oracle classifies `?1049` as unverifiable — the bracket would blind the lane's strongest oracle exactly where "sealed history untouched" must be proven. Footer maximization is fully verifiable by both oracles.
4. Bracket weight: honesty would demand editor-suspend-grade compensation machinery; a crash mid-expansion instead leaves only a bigger footer that existing resize/teardown recovers.

Honest costs stated: full-screen is approximate (history keeps >= 2 rows); occupied rows scroll into native scrollback earlier (bytes invariant under the oracle's concatenation model); scroll position not restored on dismiss.

## Design

- Expand grows the DECSTBM footer to the largest non-degenerate claim via the merged `set_footer_rows` (reusing the overlay's own capacity helper — one source of truth); dismiss shrinks to base, and the `needs_keyframe` latch makes the restore repaint self-promote — footer restored BYTE-IDENTICALLY (asserted over full cell rows, styles included). Sealed history immutable-prefix-checked across the whole enter/scroll/dismiss bracket.
- Renderer: one styled line per LineDiff op — sanitize (through ViewText's trust-boundary seam) -> CJK-safe truncate -> pad -> a single 24-bit SGR run per changed row with the shared diff palette washes and a gutter marker. (A first-pass renderer via the generic view flattening was caught in review rendering ~8x stacked lines with backgrounds dropped — reworked red-first.)
- Keybind `e` (`:not_composing`): typed text while composing, filter text under overlays, suppressed while expanded; ESC/`q` dismiss (ESC never fires interrupt while expanded); j/k/arrows scroll; steer/edit-draft freeze while expanded (overlay-parity honesty). Header carries the file path as the elastic element so position/esc hints always survive narrow widths.
- Degenerate geometry refuses honestly; resize re-derives the claim or force-closes.

## Verification

Red-first three times over (initial 0/10 + 1/22; renderer correction 5 red; path-in-header 2 red). New suites 42 tests green; adjacent suites 150 green in the combined gate; full suite 6138/6141 (2 rerun-green, 1 pre-existing environmental). Warnings + format clean.

### DIFF
--- a/lib/raxol/harness/diff_expansion.ex
+++ b/lib/raxol/harness/diff_expansion.ex
@@ -1,0 +1,395 @@
+defmodule Raxol.Harness.DiffExpansion do
+  @moduledoc """
+  Pure scrollable window over a full-screen diff render: the state behind
+  the harness surface's "expand the focused diff block" feature.
+
+  This module owns none of the terminal bytes -- it renders a `:diff`
+  content map into a plain list of sanitized, width-truncated, styled
+  lines exactly once, then exposes a clamped scroll offset and a
+  header-plus-slice render over that list. `Raxol.Harness.Surface` is the
+  only caller that turns this into terminal bytes (via `InlineAuthority`);
+  this module never touches an `IO.device()`, a process, or ANSI beyond
+  the SGR it wraps its own lines in.
+
+  ## Rendering: one line per `LineDiff` op, not the Component view tree
+
+  This renders DIRECTLY from `Raxol.UI.Components.Harness.LineDiff.diff/2`
+  -- one styled line per `{:equal | :delete | :insert, line}` op --
+  rather than through `DiffViewer.render/2` piped into
+  `Raxol.Harness.Surface.ViewText.lines/3` (the seam every other harness
+  Component's rendered body uses, and this module's own first
+  implementation used too). That pipeline was tried and rejected, for a
+  concrete, measured reason: `DiffViewer.render/2` composes each visual
+  row from SEVERAL side-by-side leaf nodes (a gutter bar, a line number,
+  a content span, alignment padding) meant to be laid out horizontally by
+  the normal `Preparer -> LayoutEngine -> UIRenderer` pipeline.
+  `ViewText.lines/3` has no notion that those nodes share one physical
+  row -- it flattens every leaf into its OWN line, the same as it does
+  for the genuinely one-text-child-per-line bodies every other harness
+  Component renders (`MessageBlock`, `ReasoningBlock`, `ToolCallBlock`).
+  Measured against a real 20-line diff at 40 columns, that pipeline
+  produced ~170 flattened lines for what should have been ~21 rows --
+  breaking the row math `scroll/2`'s clamp and the surface's
+  `view_rows` claim both depend on. Worse, `ViewText.style_line/2` has no
+  `:background` handling at all (it only ever wraps `:bold`/`:dim`/`:fg`),
+  so even where the flatten produced a usable line, the diff's own
+  add/del row washes -- the merged visual language's whole point -- were
+  silently dropped. Rendering one line directly per op, styled here, is
+  what actually satisfies both "one row per diff line" and "the diff's
+  own styling is preserved."
+
+  ## What this module actually is
+
+  A plain map (`t()`), not a process or a Component:
+
+    * `content` -- the `:diff` content map this expansion was built from
+      (`%{path, old, new, language}`), kept so `resize_view/3` can
+      re-render at a new geometry without the caller re-supplying it.
+      `:language` is accepted (part of the shared `:diff` content schema)
+      and currently unused here -- see "Scope" below.
+    * `lines` -- the full render: one already-sanitized, width-truncated,
+      styled binary per `LineDiff.diff/2` op, in document order. There is
+      no folding of long unchanged runs -- a full-screen review surface
+      exists precisely to show everything (`DiffViewer`'s own folding is
+      a property of ITS renderer, never reached here).
+    * `total` -- `length(lines)`, which is exactly `length(LineDiff.diff(old, new))`.
+    * `offset` -- the first `lines` index visible in the current window,
+      clamped by `scroll/2` to `0..max(0, total - view_rows)`.
+    * `view_rows` / `width` -- the current viewport geometry.
+
+  ## Per-row rendering (sanitize, then truncate, then pad, then style)
+
+  Each op's raw line content goes through, IN ORDER:
+
+    1. **Sanitize** -- `Raxol.Harness.Surface.ViewText.sanitize_line/1`,
+       the SAME trust-boundary byte-strip every harness Component's
+       content passes through (ESC/C0 stripped except `\\t`, `\\t`
+       excepted, multi-byte UTF-8 safe). This is content from an
+       untrusted source (an agent-produced diff) -- sanitizing FIRST
+       means styling (step 4) only ever wraps already-safe bytes.
+    2. **Truncate** -- `ViewText.truncate/2` to `width - 2` display
+       columns (2 reserved for the 1-column gutter glyph plus 1 column
+       of gutter chrome), CJK-safe (`Raxol.UI.TextMeasure`, never
+       `String.length`), ellipsis-truncated when it would overflow.
+    3. **Pad** (changed rows only) -- an `:insert`/`:delete` row's
+       content is right-padded with spaces to fill the full `width - 2`
+       budget, so the row's background wash paints the WHOLE row, not
+       just the columns under the text. `:equal` rows are never padded
+       (no wash to spread, and padding them would waste width truncating
+       real content earlier for no visual benefit).
+    4. **Style** -- `:insert` rows get the `▌` gutter bar plus the
+       merged palette's `add_base`/`add_row_bg` (`DiffViewer.diff_palette/0`
+       -- the single source of these hexes, shared with the grid
+       renderer, never a forked copy); `:delete` rows get `del_base`/
+       `del_row_bg`. Both wrap the WHOLE composed row (gutter + gap +
+       content) in one 24-bit SGR run (`\\e[38;2;r;g;b;48;2;r;g;bm`,
+       fg then bg) and a single `\\e[0m` reset at the end -- one run, not
+       a mid-row reset, since `\\e[0m` clears background too and a
+       second reset partway through the row would cut the wash off
+       early. `:equal` rows get a plain space gutter and carry no SGR at
+       all -- unstyled content, exactly like every other unchanged
+       line in this codebase's diff conventions.
+
+  ## Scope: this is the row-level tier of the diff visual language
+
+  `DiffViewer`'s moduledoc describes several composed tiers: a row wash,
+  a brighter intra-line word-diff emphasis tier on top of it, a gutter
+  tint, and syntax-highlighted tokens. This line renderer carries only
+  the ROW tier -- gutter bar plus full-row wash, read from the exact same
+  `DiffViewer.diff_palette/0` hexes. Word-level emphasis (which parts of
+  a changed line actually differ) and syntax highlighting are
+  `DiffViewer`'s own GRID-rendered tiers (built from several styled
+  spans per row, the very composition this module deliberately does not
+  reach for -- see "Rendering" above) and are out of scope for a
+  single-binary-per-row line renderer. `language` in the content map is
+  accepted, unused.
+
+  ## The header
+
+  `render_lines/1` is one header line (the file path under review, the
+  scroll position, and the dismiss/scroll hints) followed by exactly the
+  visible slice of `lines`. The path is there because full-screen review
+  hides the transcript block that named it -- "which file am I
+  reviewing" is supervision context the expanded view must carry itself.
+  The path is the ELASTIC element: sanitized (producer data) and
+  truncated to whatever display-width budget the fixed position/hint
+  suffix leaves, so a pathological path can never truncate away the
+  scroll position or the dismiss hint. The composed header is then run
+  through `Raxol.Harness.Surface.ViewText.lines/3` itself -- the same
+  width-truncation budget every content line already obeys, so a narrow
+  terminal truncates the header exactly like it would any other row,
+  never overflowing it.
+
+  ## Why this maximizes the footer instead of visiting the alternate screen
+
+  The obvious "full-screen" mechanism would be a bracketed alt-screen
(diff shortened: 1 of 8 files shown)


DROOdotFOO (commented):

Three hostile personas ran against the footer-maximization diff-expansion mechanism. The seal substrate itself (#609/#621) is untouched; findings are about the new caller's use of it. No proven immutable-prefix violation was found, so no CRITICAL. Two MEDIUM and one LOW below.

### MEDIUM

- **[Security Auditor] MEDIUM** `lib/raxol/harness/diff_expansion.ex:221` (+`:191`, `:263`) and `lib/raxol/harness/surface.ex:2339` -- **no width-floor guard; a fixed 2-column gutter overflows at narrow widths.** `validate_geometry/2` accepts any `width >= 1`, and `do_expand/2`'s only geometry gate is `claim < 2 or degenerate?(model)` -- but `degenerate?` routes to `ScrollRegionManager.degenerate?`, which is **rows-only** (`history_bottom < 2`); width plays no part. Meanwhile every rendered body row prepends a fixed `@gutter_width 2` (`"▌" <> " "`), so a row needs at least 2 columns. At `width: 1` the body rows render 2 display columns.
  - Failure scenario: `expand_focused_diff/1` on a 1-column (or resized-to-1-column) terminal succeeds instead of refusing, and hands `InlineAuthority.repaint/2` footer lines wider than the column count. That function's own caller contract (`lib/raxol/ui/rendering/paint_authority/inline_authority.ex:147-163`) states it does NOT truncate line width and that an over-wide line "wraps onto the following row ... silently overwriting content that only `repaint/2` is assumed to address ... or, on the LAST footer row, past the bottom of the screen entirely." Row-degeneracy gets a careful 2-row-minimum guard; width-degeneracy gets none, and `DiffExpansion` is the trusted self-truncator that the surface never re-truncates through `ViewText.lines/3`.
  - (reproduced) `DiffExpansion.new(%{path: "x.ex", old: "a", new: "b", language: nil}, width: 1, view_rows: 4)` then `render_lines/1` -> display widths `[1, 2, 2]` (header fits, both body rows overflow to 2). At `width: 2` -> `[2, 2, 2]` (exact fit, the true floor).
  - Fix: floor the width in `validate_geometry/2` (`width >= @gutter_width + 1`), and/or make `do_expand/2` refuse a degenerate width the way it refuses degenerate rows, so the honest-refusal notice fires instead of emitting wrapping bytes.

- **[Saboteur] MEDIUM** `lib/raxol/harness/surface.ex:2759` (`resize_expansion/3`) + test gap at `test/harness/diff_expand_surface_test.exs:731,748` -- **resize-while-expanded re-grows the DECSTBM footer but no test pins the immutable-prefix invariant across that bracket.** On a resize that still fits, `resize_expansion/3` re-derives `claim = max_overlay_rows(rows, footer_rows)` and calls `InlineAuthority.set_footer_rows/2` to re-grow (or re-shrink) the footer at the new geometry -- the exact "resize during full-screen" substrate-sensitive path. The only `SealOracle.immutable_prefix?` assertion in the suite (`:402`) covers the STATIC open+scroll+dismiss bracket. The two resize tests (`:731` "still fits", `:748` "force-close") assert only `expansion != nil` / `footer_row_count` / offset clamp -- never that sealed history survived the re-grow.
  - Failure scenario: a terminal grown taller mid-expansion pushes more occupied history rows through the enlarged region; if the substrate ever regressed the scroll-vs-overwrite behavior on a `set_footer_rows` grow, nothing in this feature's suite would catch it. (My cross-height replay probe was inconclusive -- it compared `SealOracle.history/2` at mismatched `top` args and returned an artifact, not a real violation -- so this is reported as an untested-invariant coverage gap, not a proven seal break.)
  - Fix: add an `immutable_prefix?` assertion across a resize-grow-then-dismiss bracket to the resize describe block, mirroring the static test at `:402`.

### LOW

- **[New Hire] LOW** `lib/raxol/harness/surface.ex:2350` and `:2767` -- **the `- 2` chrome arithmetic is a duplicated magic number, and the `claim < 2` gate's comment conflates two distinct quantities.** `view_rows = model.footer_rows + claim - 2` appears verbatim in both `build_expansion/3` and `resize_expansion/3` with no named constant; the `2` means "one status row + one header row" but is written as a bare literal in two places. Separately, the `if claim < 2` gate at `:2339` is commented "even the expansion's own 2-row minimum (one status row, one expansion header row) no longer fits" -- but `claim` is the footer GROWTH beyond the base `footer_rows`, not the total footer, so the "2-row minimum" story describes the view composition while the value gates the growth. The two `2`s are semantically different (history-minimum vs. status+header chrome) yet coincidentally equal, and `@gutter_width 2` is a third unrelated `2`.
  - Failure scenario: a future change that makes the expansion header 2 lines tall (or adds a footer chrome row) must find and edit both `- 2` copies and reinterpret the gate comment; missing one silently mis-sizes the viewport. Fix: name it (`@expansion_chrome_rows 2`) and derive `view_rows` through one helper, and reword the gate comment to say what `claim` actually measures.

### Cross-persona overlaps

The width-overflow (Security Auditor MEDIUM) is also a Saboteur concern: per `InlineAuthority`'s documented contract the over-wide line wraps into the next footer row or past the screen bottom, i.e. a footer-region-confinement break, not merely cosmetic. Two personas, one root cause (unguarded width vs. fixed 2-col gutter) -- promoted to the primary finding.

### Verdict: CONCERNS

No CRITICAL: the seal substrate is reused unchanged and the static immutable-prefix bracket is well tested. But the mechanism ships two real, diff-grounded gaps -- an unguarded width dimension that overflows the documented repaint contract (reproduced), and an untested seal invariant on the resize-while-expanded path -- plus a duplicated magic-number legibility snag. All are fixable without redesign.

DROOdotFOO:
**Merge readiness: HOLD (borderline).**

CI green, no CRITICAL/HIGH, but two MEDIUMs to close first: (1) no width-floor guard -- geometry gates on rows only, yet every body row prepends a fixed 2-column gutter, so `width: 1` renders 2 display columns and violates `InlineAuthority.repaint/2`'s over-wide-line contract (reproduced); (2) the resize-while-expanded path re-grows the DECSTBM footer with no test pinning `immutable_prefix?` across that bracket. Add the width floor + the resize seal test, then it's ready.


merklebonsai:
## Review response (adversarial pass, 2026-07-17)

Rebased onto current `master` (composed over #616/#618/#619/#629/#630/#632; view_text sanitizer unioned to keep master's C1-codepoint strip **and** expose it publicly for `DiffExpansion`; keymap `:open_panel`/`:expand_diff` unioned). All three findings addressed in `906d2007d`. Verify-or-refute ledger:

### MEDIUM — width floor (Security Auditor) — VERIFIED, fixed + red-first regression
Confirmed by reading: `content_width = max(width - @gutter_width, 0)` but every body row still emits a fixed 2-column gutter (`"▌" <> " "`, or two spaces for `:equal`), so at `width: 1` body rows render 2 display columns into a 1-column budget — over the documented `InlineAuthority.repaint/2` no-truncate contract. Reachable from the surface because `do_expand/2`'s gate (`claim < 2 or degenerate?`) is rows-only.
- **Fix:** `validate_geometry/2` and `resize_view/3` now floor `width >= @gutter_width + 1`; `DiffExpansion.new/2` returns `{:error, :degenerate_view}`, which `build_expansion/3` maps to `:insufficient_footer_capacity` — the honest "terminal too small" notice — **before** `set_footer_rows/2` writes any byte.
- **Falsifiers (red on old code):** unit — `width: 1` and `width: 2` refuse; `width: 3` (the floor) admits and *every* rendered line ≤ 3 cols; `resize_view/3` refuses sub-floor. Surface — a 2-wide, tall terminal refuses with **zero bytes written** (no footer grow).

### MEDIUM — resize seal coverage (Saboteur) — VERIFIED gap (dormant invariant), closed; mechanism partially refuted
Agreed the two resize tests asserted no seal invariant across the re-grow. Added a `resize-grow-then-dismiss` bracket.
- **Honest disclosure on method:** it asserts byte-level seal-adjacent invariants (no full clear; no CUP into history rows), **not** cell-level `immutable_prefix?`. The re-grow is only reachable via a terminal **height change**, and `SealOracle.replay` walks one fixed height — a cross-height prefix compare returns an artifact, i.e. exactly the inconclusive result your probe hit. The constant-height open+scroll+dismiss `immutable_prefix?` bracket (describe "3") already pins the far larger OPEN grow.
- **Partial refutation of the worry:** while expanded the footer is always `rows - 2` (`max_overlay_rows = rows - footer_rows - 2`), so region_top is pinned at 2 regardless of height. A taller resize claims the terminal's **new bottom rows**, never additional history — the re-grow evicts nothing. The substrate grow-over-partial-history eviction bug is separately pinned by #621's regression.

### LOW — magic-number legibility (New Hire) — VERIFIED, fixed
`footer_rows + claim - 2` named `@expansion_chrome_rows` and routed through `expansion_view_rows/2` (used by both `build_expansion/3` and `resize_expansion/3`). Gate/force-close comments reworded so the footer-growth `2` is no longer conflated with the chrome `2` (or `DiffExpansion`'s `@gutter_width 2`).

### Gates
`mix compile --warnings-as-errors` clean; `mix format --check-formatted` clean; `test/harness/` **671 passed** (18 properties, 1 intentional skip, 17 tag-excluded); diff-expand unit + surface + keymap suites **105 passed**.

DROOdotFOO (commented):

Since last review: 09466a82..906d2007

Re-ran the two prior MEDIUMs and the LOW against the new commit `906d2007`
("Floor the diff-expansion width at the gutter, close the resize seal gap").
Verified by reading the delta and by building + running the touched suites
in an isolated worktree at the PR head: `diff_expansion_test.exs` +
`diff_expand_surface_test.exs` = **53 passed**, plus a direct-call
reproduction of the width-floor case.

- **MED (width floor) -- FIXED (reproduced).** `validate_geometry/2` and
  `resize_view/3` now gate `width >= @gutter_width + 1` (3), not `>= 1`
  (`lib/raxol/harness/diff_expansion.ex:229-231, 385-386`). Direct repro on
  the new code: `new(width: 1)` -> `{:error, :degenerate_view}`,
  `new(width: 2)` -> `{:error, :degenerate_view}`, `new(width: 3)` renders
  display widths `[3, 3, 3]` (header + body rows all exactly at budget) --
  the prior `[1, 2, 2]` over-render is gone. The surface maps the refusal to
  `:insufficient_footer_capacity` in `build_expansion/3`
  (`surface.ex:2576-2582`) *after* the pure `DiffExpansion.new/2` and
  *before* `set_footer_rows/2`, so a sub-gutter width writes zero bytes --
  pinned by `diff_expand_surface_test.exs` ("sub-gutter width: direct call
  refuses with zero bytes", `byte_size(raw) == prior`). This also strengthens
  the seal boundary: the over-wide row that `InlineAuthority.repaint/2` does
  not re-truncate could have wrapped a footer line up into a pinned history
  row, so the floor closes a latent visual-integrity gap, not just a
  cosmetic one.

- **MED (resize seal test) -- ADDRESSED.** New test "sealed history survives
  the resize-grow-then-dismiss bracket"
  (`diff_expand_surface_test.exs:816-878`) is non-vacuous: it asserts a
  partially-filled history precondition (`model.authority.next_row >
  @expanded_region_top`) before expanding, drives a real DECSTBM re-grow via
  `Surface.resize(model, @width, 14)` (footer 10 -> 12 through
  `resize_expansion/3`), then checks `refute emits_full_clear?` and
  `explicit_cup_rows(grow_delta)` all `> region_top`, then dismiss +
  no-full-clear over the whole bracket. Both helpers are real
  (`SequenceScanner`-backed CSI parsing / `\e[2J`/`\e[3J` detection), not
  stubs. The chosen byte-level check instead of cell-level `immutable_prefix?`
  is soundly justified in-commit: `SealOracle.replay` walks one fixed height,
  so a cross-height cell compare is an artifact -- exactly the mismatch that
  made my prior probe inconclusive.

- **LOW (duplicated `- 2` chrome magic number) -- FIXED.**
  `@expansion_chrome_rows` + `expansion_view_rows/2` now shared by
  `build_expansion/3` and `resize_expansion/3` (`surface.ex:2521-2531`), and
  the three coincidentally-equal 2s (footer-growth minimum, chrome
  subtraction, `@gutter_width`) are each called out so they can't be
  conflated.

### Low / residual (non-blocking)

- **Seal proxy is the weaker of the two possible checks, and the invariant
  it pins is inherently easy.** By construction the expanded footer is always
  `rows - 2` with history pinned at its 2-row minimum, so a *grow-taller*
  resize claims new bottom rows and evicts nothing -- a history-eviction
  re-grow is unreachable, so the test cannot exercise one. Combined with the
  byte-level check only catching *explicit* CUP (not a stray write from a
  cursor left in a history row), this is a real-but-shallow guard. It is the
  honest ceiling here given the single-height replay limitation, and the
  constant-height cell-level bracket in describe "3" still covers the larger
  OPEN grow -- so this is a noted limit, not a defect.

- **Narrow-resize-while-expanded does a grow-then-undo.** A tall-but-narrow
  (`width < 3`) resize is not caught by `force_close_expansion?/2` (row-based,
  not width-based), so `resize_expansion/3` calls `set_footer_rows` to grow,
  `DiffExpansion.resize_view/3` then refuses on the width floor, and
  `close_expansion/1` restores the base footer (`surface.ex:3021-3033`). Net
  state is correct and no intermediate paint escapes (the grow never reaches
  `paint_footer` before the undo), so this is a micro-inefficiency, not a bug
  -- flagging only so a future reader knows the width-floor refusal is handled
  on the resize path too.

- **(c) New-bug check -- none found.** The floor only refuses `width < 3`, a
  sub-realistic terminal; every realistic width (>= 3) still expands, mapped
  refusals are honest zero-byte notices, and no legitimate narrow terminal is
  wrongly refused.

### Verdict: CLEAN

Both MEDIUMs are genuinely resolved (width floor reproduced fixed; resize
seal now has a real, non-vacuous byte-level bracket over a partially-filled
history) and the LOW is fixed. The two residuals above are documented,
sound limitations rather than defects.

PR #627, opened 2026-07-17
Harness UI: supervise a live agent session — stream in, interrupt and steer out

merklebonsai:
The milestone unit: the harness stops being a fixture demo. A live session's events stream into the same projection/seal path fixtures use; ESC dispatches a real interrupt through the staged supervised kill; Tab steers with compare-and-swap turn protection. Everything the instrument shows about its own commands is observed, never fabricated.

## ⚠ CROSS-LANE DIFF — agent-lane review requested

`packages/raxol_agent/lib/raxol/agent/command.ex`: the command codec grew `:steer` (decode: non-empty text via the prompt vocabulary, required `expected_turn_id`, optional `client_msg_id`; route via the existing `{:harness_command, _}` delivery). This is the growth the codec's own moduledoc promises and is purely additive (+10 codec tests), but it is agent-lane-owned territory per the coordination accord. Also added agent-side: `Raxol.Agent.Harness.SessionLane` (the lane impl) + a suite driving the real `Interrupt.interrupt/3` and real `Steer.resolve/2` through routed command shapes — useful pre-work for the session-runtime steer integration.

## Wiring ledger (reachability honesty)

**11 LIVE guarantees / 5 SEAM disclosures / 4 outstanding before-live-traffic preconditions** — the full ledger (every guarantee mapped to its call site + enforcing test, every seam named, every inherited precondition with status and owner) is posted as a comment below. The seams, up front:

1. Live steer against the real session stack is **refused, not faked** (`{:error, :no_steer_channel}`) until the agent lane's session runtime owns a live `Steer.TurnState` (driver-side path proven against the real `Steer.resolve/2` in-test).
2. Interrupt against the real stack has **no production receiver yet** — the keystone test pins the honest consequence (pending forever, never a fabricated ack). Owner: agent-lane session-runtime integration.
3. Interrupt-stage ack rendering is live code with a **fake-proven producer** (same pending unit as 2).
4. No shipped tty embedder yet — the integration keystone is the only end-to-end consumer; demo script is a small follow-up.
5. Reattach/replay — out of scope (a later unit).

Outstanding preconditions inherited from other reviews (not this PR's to close): cadence mailbox-ingress/bytes-cap residuals (cadence layer), `pending_input?` real producer feed (block builder/agent producer), dead-device seal semantics (#620, unmerged — no interaction with this loop), and keymap modifier-blind Tab binds (keymap unit — severity raised by this PR since a mis-fired chord now dispatches a real steer; the CAS + honest-notice path mitigates).

## Process shape (grounded in the shipped stack)

- Events: Dispatcher -> EmitBus -> per-session EmitBridge (journals first; event id = journal offset) -> `SessionStreamer.subscribe/1` -> `{:session_event, ...}` messages. The driver's linked forwarder owns the subscription and pushes every event through `Raxol.Harness.EventBoundary.normalize/1` — the security seam: no atom minting, unknown fields dropped, tier never guessed, taint absorbed never laundered — into the merged stream-cadence layer; batches apply through the SAME reveal path fixtures use.
- `Raxol.Harness.LiveSessionDriver` is a plain-process loop (the fixture demo's precedent, deliberately not a GenServer): the cadence owner contract demands input-first SELECTIVE RECEIVE, which handle_info cannot express.
- Package law forced the injection seam: main raxol cannot depend on raxol_agent, so the driver consumes an injected `Raxol.Harness.SessionLane` behaviour.

## Command honesty (the design's spine)

- **Interrupt: fire-and-forget, event-observed ack.** The footer shows "interrupt sent — awaiting confirmation"; the ONLY ack is the staged kill's own durable events arriving on the stream. The keystone test pins the honest-pending case: against the real stack (no command handler yet), the footer never fabricates a confirmation.
- **Steer: synchronous typed decision** (Task.async so a slow lane can't block ESC) — a stale-turn CAS rejection journals nothing by design, so only a reply can carry it. Five outcomes render five distinct notices; stale-turn names both turn ids. The agent-side lane returns `{:error, :no_steer_channel}` WITHOUT sending until the session runtime owns a live TurnState — refusing loudly beats pretending.
- Loss honesty: cadence-shed batches render an explicit in-band marker row; unrecognized reserved batch elements are marked the same way, never silently dropped (fail-open sweep fix, red-first). Session death is monitored and rendered plainly; the harness stays up showing history.

## Seal-hardening (#620) interplay

The loop renders exclusively through `Surface.advance/2`; #620's revised seal shape lands transparently beneath it. One 3-line seam (`seal_marker/2` uses plain `seal/2`) follows suit if #620 removes it. Merge-order-safe; this branch already rebased clean over #615/#617/#621.

## Verification

Red-first per module (captured failing runs pre-implementation). Full test/harness 608 passed / 0 failures; full raxol_agent suite 1461 passed / 0 failures, including the integration keystone: a real SessionStreamer + Contract.pump over the Mock backend producing sealed transcript bytes. New suites: event-boundary security matrix (40), live-seam (10), driver guarantee-map (12), agent-side codec/lane/kill/CAS suites. Warnings + format clean both packages.

Known deferrals documented in code: reattach/replay (separate unit), unbounded event retention in long sessions (same follow-up the memory-residency section tracks), tty demo script (20-line follow-up).


merklebonsai:
Heads-up from the seal-hardening branch (#620, review round 2): no API signature changes, but one behavior change inside the existing contract — `try_seal/2` (and sync writes) now RAISE `ErlangError` on a DEAD device; the `{:error, :write_failed, t}` tuple is reserved for the alive-but-refusing io-reply class. This branch's driver renders via `Surface.advance/2` and monitors session death separately (StringIO devices in tests are alive), so no impact expected — but if any test deliberately feeds a dead device and expects the tuple, it must expect the raise after #620 merges. Also additive: `[:raxol, :harness, :seal, :write_failed]` telemetry, and `?2026l` joined the teardown/suspend byte sequences.

merklebonsai:
## Wiring / preconditions ledger

Reachability-honest framing for this PR: every headline guarantee below is declared **LIVE** (production call site + the test that fails if it breaks) or **SEAM** (unwired — inert until the named unit lands). Plus the accumulated "before live traffic" preconditions from other PRs' review threads, each with a status and an owner.

### LIVE guarantees (call site → enforcing test)

| # | Guarantee | Call site | Enforcing test |
|---|-----------|-----------|----------------|
| 1 | Live event stream → sealed transcript blocks (subscribe → normalize → cadence → the same `append_events`/`advance` reveal path fixtures use) | `LiveSessionDriver.apply_batch_item/2` | `live_session_driver_test.exs` "1. stream to sealed blocks"; keystone: real `SessionStreamer` + `Contract.pump` over the Mock backend (`live_session_agent_test.exs`, tagged `:integration`) |
| 2 | Input handled before queued render batches (the StreamCadence §2 owner contract) | `LiveSessionDriver.loop/1` input-first selective receive | "10. input is handled before queued render batches" |
| 3 | Boundary normalization: no atom minting, unknown fields dropped, taint absorbed never laundered, tier never guessed | forwarder → `EventBoundary.normalize/1` | `event_boundary_test.exs` (40 tests, security matrix) |
| 4 | ESC-interrupt dispatches through the real command channel with the current turn id | `handle_surface_command(:interrupt)` → `Command.decode/route` | "2. interrupt dispatch"; agent-side: routed command drives a **real** `Interrupt.interrupt/3` staged kill, full `[:interrupt_signaled, :interrupt_waited, :interrupt_killed, :turn_canceled]` sequence asserted |
| 5 | Interrupt acknowledgment is event-observed, never fabricated | `apply_lifecycle/2` stage clauses | "3. interrupt acknowledgment is event-observed"; keystone asserts honest-pending against the real stack (no fabricated ack) |
| 6 | Steer CAS honesty: five distinct outcomes, stale-turn never a silent drop, queued banner cleared on rejection | `render_steer_result/2` | "4. steer accepted", "5. steer CAS failure is an honest notice"; agent-side: both arms against the **real** `Steer.resolve/2` |
| 7 | Loss honesty: shed deltas / rejected events seal an in-band transcript-gap marker | `apply_batch_item/2` (`{:cadence_dropped, n}` / `{:malformed_event}`) | "6. loss is marked in-band", "7. malformed events…", "6b. unknown batch elements are marked" (added in this review round — see fail-open scan below) |
| 8 | Session death / final turn: plain footer statement, UI survives, scrollback preserved | `handle_down/4` (session ref), `apply_lifecycle(:turn_completed final)` | "8. session death leaves the UI alive and honest", "9. final turn ends the session plainly" |
| 9 | Stall verdict reaches the status strip's ALERT seam | `apply_stall_observation/2` + `handle_tick/1` → `Surface.put_stall_verdict/2` | "11. stall verdict reaches the strip" |
| 10 | Fixture-mode `:interrupt`/`:steer` stubs unchanged when no `command_sink` is set | `Surface.dispatch_command/2` clause order | `surface_live_seam_test.exs` + existing fixture suites (all green, unmodified) |
| 11 | `:steer` decodes/routes at the one command validation seam | `Raxol.Agent.Command` | `command_test.exs` steer decode/route tests |
| 12 | Fold-before-seal ordering: a block's turn bracket (and anything later units fold from it, e.g. a completion/evidence row) lands in the projection before the block seals — the live/fixture parity invariant (integration finding, commit `b6b552963`; release semantics corrected in `a27908799`) | `Surface.frontier_entries/1` `:stream_open` hold; released PER TURN by `Surface.flush_held/1` on every turn bracket, terminally by `Surface.close_stream/1` on session death / dead feed (the stranded-tail backstop) | "1b. fold-before-seal ordering" (red against the prior driver: block sealed before its bracket folded) |
| 13 | External editor handoff (Ctrl+E) works under the live driver; while the editor owns the terminal, events queue and not one byte reaches the device until resume (commit `f229ef874`) | `LiveSessionDriver.build/1` passes `:editor_session`/`:editor_opts` through to `Surface.new/2` | "12. external editor handoff" (invocation red against the prior take-list; suspend-queue pin) |
| 14 | A final `turn_completed` completes a TURN, never the session: its blocks seal immediately, no "session ended" claim is rendered, and a later turn on the same session still streams to sealed history — session end is a process-level fact only (real-use defect from the live demo, commit `a27908799`) | `LiveSessionDriver.apply_lifecycle/2` turn brackets → `Surface.flush_held/1`; session end only in `handle_down/4` / `handle_exit/3` | "9. turn completion releases the hold but never ends the session" (red against the prior driver: auto-ended after the first answer); agent-side keystone refutes the session-ended claim after a completed pump run |

### SEAM disclosures (unwired — inert until the named unit lands)

1. **Live steer against the real session stack is refused, not faked**: `Raxol.Agent.Harness.SessionLane.steer/2` returns `{:error, :no_steer_channel}` without sending — no shipped session runtime owns a live `Steer.TurnState` to resolve the CAS (that single-writer integration is the agent lane's pending unit). The full driver-side path is proven against a real `Steer.resolve/2` in-test; the production wire awaits that unit.
2. **Interrupt against the real stack has no production receiver yet**: `Command.route` delivers `{:harness_command, {:interrupt, ...}}`, but `Agent.Session` does not handle it today. The keystone test pins the honest consequence: footer stays "interrupt sent — awaiting confirmation", never a fabricated ack. Owner: agent lane session-runtime integration.
3. **Interrupt stage / turn-canceled ack rendering** (`apply_lifecycle/2`) is live code with a fake-proven producer: the real producer (session runtime emitting staged-kill events through the journal) is the same pending agent-lane unit as (2).
4. **The driver has no shipped tty embedder yet** — the keystone test is its only end-to-end consumer; a `packages/raxol_agent/examples/` demo script (fixture-demo pattern, InlineDriver + SessionStreamer + pump) is a small follow-up.
5. **Reattach / replay** — excluded by scope (a later unit); `Command`'s `:attach`/`:seek` remain decode-only as before.

### Before-live-traffic preconditions from other PRs' reviews

| Precondition (source review) | Status |
|---|---|
| StreamCadence queue bound + yield budget "must land before this is wired to a live session" (#606 R1) | **satisfied-by-merged-master** (`max_pending` shed + `max_consecutive_yields`) |
| Input-first "unproven against a real owner… mandatory selective-receive half" (#606 R2) | **satisfied-by-this-PR** — the driver is that owner; both halves wired (selective receive + `input_check`), test 10 |
| Drop-oldest on an append-only token stream is "content-corrupting-with-a-flag; the future owner picks the sink type knowingly" (#606 R2) | **satisfied-by-this-PR (knowingly)** — the driver renders `{:cadence_dropped, n}` as a sealed transcript-gap marker: torn-and-flagged, never gapless |
| #606 R2 residuals: unbounded mailbox ingress (cast-based `ingest`), count-not-bytes cap | **outstanding — owner: cadence layer (harness-ui lane)**; low-risk for the network-rate SSE producer, still open |
| ViewText sanitization gap "closed before the live agent lane feeds real tool output through this path" (#595) | **satisfied-by-merged-master** — the ContentGuard seal-seam sanitization (the architectural fix Drew verified on later PRs); this PR adds the upstream *shape* seam (`EventBoundary`) which deliberately does NOT claim byte sanitization — bytes are the paint seam's job, and remain so |
| `pending_input?` fed from a real producer, not a proxy (#605) | **outstanding — owner: block builder / agent-lane producer** (live approval blocks stay dormant until a producer emits them; unchanged by this PR) |
| Dead-device `try_seal` semantics + telemetry (#620) | **outstanding — owner: #620 (unmerged)**; no interaction with this PR — its only direct emit is `seal_marker/2`, and session death is detected by process monitor, never inferred from device writes. If #620's raise-on-dead-device converts `seal_marker`'s emit, the raise semantic is correct for a marker (a dead device is an embedder bug, not a loss to mark) |
| Keymap key binds are modifier-blind: `Ctrl/Alt/Shift+Tab` all resolve to `:steer` (keymap review) | **outstanding — owner: keymap unit — severity RAISED by this PR**: what previously mis-fired a stub now dispatches a real steer command. Mitigation in place: the mis-fired steer carries the same composer text a plain Tab would, and the CAS + honest-notice path applies; still worth fixing before habitual live use |

### Self-review scans run on this diff (per the adopted checks)

- **Proxy-validation scan**: (a) `input_check` reads the driver's message-queue length — an over-approximating proxy in the fail-safe direction (yields more, never less), disclosed in the moduledoc, bounded by the yield budget; (b) `current_turn_id` is a tracked belief whose referent check is deliberately the session-side CAS — the authority decides, the driver only renders the typed outcome; (c) session death is monitored on the session pid (a proxy for "stack alive") with the stall detector covering the silently-dead-stream case. No instance validates a representation where the referent is checkable locally.
- **Fail-open scan**: one finding, fixed red-first in this round — the batch-element catch-all silently swallowed unrecognized elements, against StreamCadence §3's "consumers must handle reserved types loudly" law. Now seals an honest "unrecognized stream element" marker (test "6b"; commit `4e45edb24`, red captured against the pre-fix catch-all: 11/12 → 12/12; lands with the next push of this branch). All other missing-data paths verified honest: missing `final` flag → stall alarm covers the silence; nil turn id on steer → the CAS's `:no_live_turn` rejection renders; crashed steer task → notice + unblock, never a stuck "in flight".

merklebonsai:
## Integration finding resolved: fold-before-seal ordering (commit b6b552963)

The consolidated-branch parity failure (live path sealed a done-block without the evidence-rendered completion row a fixture replay of the same events shows) is diagnosed and fixed on this branch.

**Root cause: ORDERING, not payload shape.** A live reveal is always momentarily caught up — `revealed == length(events)` after every applied event — so the seal frontier's one-step hold on the newest completed block never engaged on the live path, and every block sealed the instant it materialized, BEFORE its turn bracket folded into the projection. The fixture path folds the bracket first (the hold releases only when a later event reveals), which is exactly where the two paths diverged once the evidence-rendered done row derives from `turn_completed`.

**Route taken: enforce the local invariant, not the integration symptom.** The completion-row rendering lives only on the unmerged evidence branch, so a red written against its output would not run here. Instead the branch now enforces the invariant that MAKES the integration case correct by construction — *turn brackets (and anything later units fold from them into a block) land in the projection before that block seals*:

- `Surface.new/2` gains `:stream_open` — while set, the reveal is never treated as finished merely for being caught up, so the frontier hold works on the live path exactly as it already does mid-reveal on the fixture path. Default `false`: fixture behavior byte-identical, full harness suite green.
- `Surface.close_stream/1` releases the hold when no more events can ever arrive; the driver calls it at the three end-of-stream facts (final `turn_completed`, session-process death, dead event feed) so the trailing block always lands in history rather than living in the footer preview.
- Red-first: the new "fold-before-seal ordering" test drives `turn_started → item_completed` and asserts the block is NOT yet sealed, then folds `turn_completed` and asserts it seals — fails against the prior driver (12/13), green with the hold (13/13). This is precisely the integration case minus the unmerged rendering, and it stays meaningful after that branch merges.

Gates: full `test/harness/` 609 passed / 0 failures; agent-side harness suites incl. the integration keystone 36 passed; format + warnings-as-errors clean. The wiring ledger above gains one LIVE row (fold-before-seal ordering → `live_session_driver_test.exs` "1b").

merklebonsai:
## Turn/session semantics corrected (commit a27908799) — real-use defect from the live demo

The fold-before-seal fix had conflated two boundaries: it treated `turn_completed final: true` as an end-of-stream fact (close + "session ended" claim), so a multi-turn conversation — one pump run per composer submit on the same session id — auto-ended after the first answer.

**Shipped semantics:**

- **Turn bracket** (`turn_completed` final or not, `turn_canceled`) → `Surface.flush_held/1`: releases the fold-before-seal hold for the blocks that bracket completed — they seal into history immediately, not stranded in the footer preview until a next turn — and the stream stays open. No session-level claim is rendered; `final: true` closes one pump run, never the session.
- **Session end** is a process-level fact only: session-process death and a dead event feed (forwarder/cadence crash) still run `Surface.close_stream/1` — which also remains the backstop that lands a stranded mid-turn tail in history when a session dies without ever emitting its bracket.

Red-first: the new multi-turn test (turn 1's final bracket seals its blocks + renders no "session ended" claim + a second turn on the same session still streams to sealed history) fails against the prior driver; the fold-before-seal ordering test and session-death honesty test hold unchanged; the agent-side keystone now *refutes* the session-ended claim after a completed pump run. Also in this push (commit f229ef874): the `:editor_session`/`:editor_opts` passthrough the live-embedder build flagged, with a suspend-queue pin (while $EDITOR owns the terminal, events queue and not one byte reaches the device until resume). Gates: full `test/harness/` 611 passed / 0 failures; agent-side harness suites incl. keystone 36 passed; format + warnings-as-errors clean. Ledger comment above updated to 14 LIVE rows.

DROOdotFOO (commented):

Scope: the live-session wiring -- `EventBoundary`, `LiveSessionDriver`, both `SessionLane`s, the `Surface` live seam, and the `:steer` codec growth. Read against the shipped cadence/seal/projection stack. Reachability note up front: `Raxol.Harness.LiveSessionDriver` and both `SessionLane`s have **zero production callers** in `lib/`, `packages/`, or `examples/` -- the entire stream-in / interrupt-out / steer-out loop is reachable only from the two test suites. This matches the PR's own SEAM #4 ("no shipped tty embedder yet"). So the findings below are pre-production hardening, not live incidents -- which is the main reason this is CONCERNS and not BLOCK.

### High

**H1 (Saboteur + Security, promoted -- unbounded retained events + O(n) reproject per event).** `StreamCadence` bounds the *ingress* queue (`@default_max_pending 10_000`, drop-oldest shed) but nothing bounds the *revealed* heap. `LiveSessionDriver.apply_batch_item/2` (`live_session_driver.ex`, `{:event, map}` clause) calls `Surface.append_events(model, [map])` -> `%{model | events: model.events ++ events}` (`surface.ex` `append_events/2`) followed by `Surface.advance/2`, whose own docstring pins the cost as `O(n)` per call ("matching `Projection.project/2`'s own per-`advance/2` `O(n)` rebuild"). Result on a live session: `model.events` grows without bound and every single revealed event reprojects the entire accumulated history, so wall-clock work is quadratic in session length and heap is unbounded. Cadence shedding protects the mailbox, not this. Fixtures never exposed it because a fixture is a bounded file replayed once; a real agent that streams for minutes (or a runaway/hostile producer emitting large payloads) is exactly the input this quadratic/unbounded path was never load-tested against. The PR lists "unbounded event retention in long sessions" as a deferral, but the *compounding* reproject cost is the load-bearing half and is not called out. This is the one to fix before wiring a real long-running session: cap/window `model.events` (or make `advance/2` incremental) with an honest truncation marker.

### Medium

**M1 (Saboteur -- steer-in-flight can wedge permanently).** `handle_surface_command(state, %{type: :steer, ...})` (`live_session_driver.ex`) does `task = Task.async(fn -> lane_mod.steer(session, request) end)` and stores it; `steer_task` is only ever cleared in `handle_task_result/3` + `handle_down/4`. There is no `Task.await`/`Task.shutdown` timeout and no self-timer. If a lane's `steer/2` blocks (the `SessionLane` behaviour only *documents* "implementations must be bounded ... rather than hang the caller forever" -- it does not enforce it), `steer_task` stays non-nil forever and every subsequent Tab is refused with `"steer already in flight"`. The shipped agent lane returns instantly (`{:error, :no_steer_channel}`), so this is latent today, but the driver is the injection seam for arbitrary lane impls and should defend the contract it depends on (bounded `Task.yield`/`shutdown` with an honest "steer timed out" notice, clearing the slot).

**M2 (New Hire -- `start_link/1` looks OTP, isn't).** `start_link(opts)` is a bare `spawn_link(__MODULE__, :run, [opts])` returning `{:ok, pid}` (`live_session_driver.ex`). It has no `child_spec/1`, no init-ack (state is built *after* the process is already "started"), and registers no name. A reader who sees `start_link` will reasonably try to place it under a `Supervisor` and expect restart semantics; it will spawn but cannot be meaningfully supervised or restarted, and a restart would silently drop the whole session subscription. The moduledoc explains *why* it is a plain loop, but the OTP-shaped name undercuts that. Rename to `spawn/spawn_link` semantics or document loudly at the `@spec` that this is not a supervisable child.

### Low

**L1 (Saboteur -- current_turn_id lags the stream; interrupt has no dedup).** The input-first `receive` (`loop/1`) handles `{:surface_command, _}` ahead of any queued `{:render_batch, _}` (the deliberate cadence-owner contract). Consequence the newly-wired path activates: a `turn_completed` still sitting in an unconsumed render batch has not yet run `apply_lifecycle` to clear `current_turn_id`, so an ESC pressed in that window dispatches `interrupt_payload(state.current_turn_id)` against a turn the stream has already ended, and the notice prints `"interrupt sent (turn X)"` for a dead turn. For steer this is absorbed by the CAS (stale-turn reject is correct), but interrupt has no such guard -- and, unlike steer, `handle_surface_command(_, %{type: :interrupt})` has **no in-flight/dedup guard at all**, so mashing ESC dispatches N supervised-kill commands. Low today (no production receiver), but the asymmetry is worth a debounce.

**L2 (Security -- "security seam" naming overstates; taint tracked, never gated).** `EventBoundary`'s moduledoc calls itself "the live-session security seam" and enumerates four load-bearing properties (no atom minting, unknown-field drop, taint-absorb, tier-never-guessed) -- all correctly implemented. But it explicitly passes payload string *values* through unchanged, so it performs **no ESC/C0 byte neutralization**. Real byte sanitization happens downstream in `ViewText.sanitize` and `MarkdownBody.sanitize_controls` (verified present -- strips `0x00-0x1F` incl. ESC, DEL, C1), so there is no active ANSI-injection vuln today. Two residual concerns: (a) the boundary's own doc never says byte-stripping is *not* its job and lives elsewhere, so a future payload field rendered on a path that bypasses `ViewText` reopens the C1/#616 vector with no boundary-level backstop; (b) `provenance.trust` is faithfully absorbed to `:tainted` on any ambiguity but nothing in this diff ever *gates* rendering on it -- taint is labeled, never enforced. Also note `deep_normalize/1` recurses with no depth/size bound (recursive `Map.new`/`Enum.map`, `inspect` fallback), feeding H1's exhaustion surface on a hostile deep/large payload.

### Cross-persona overlaps

- **H1** is the Saboteur backpressure finding and the Security resource-exhaustion finding at one root (unbounded retained `model.events` + per-event O(n) reproject); the cadence heap cap does not cover it. Promoted to High.

### Verdict: CONCERNS

The design is honest where it counts: crash isolation is correct (`trap_exit` turns forwarder/cadence/task deaths into footer notices instead of tearing down the terminal), the interrupt-is-event-observed / steer-is-a-reply asymmetry is sound, the agent lane refuses steer loudly (`:no_steer_channel`) rather than faking an ack, and untrusted content is in fact neutralized at the `ViewText` boundary. Not BLOCK because the whole loop is still test-only (unwired -- SEAM #4) so there is no production blast radius yet. But H1 (unbounded/quadratic retained-event path) must be addressed before this drives a real long-running agent, and M1 (steer wedge) + M2 (misleading `start_link`) should land alongside the first real embedder.

DROOdotFOO:

CI is red (2 failing checks). The review HIGH also stands: `apply_batch_item` retains every event in `model.events ++ events` and reprojects the whole history O(n) per event via `advance/2` -- quadratic CPU + unbounded memory on a real long/fast agent stream (fixtures never exercise it). The whole live loop is still unwired (test-only). Address the quadratic reproject and the red CI before merge.


merklebonsai:
## Review response (commit `b9215f44d`)

Two items in the review; taking them in order, with the LIVE/SEAM framing.

### 1. CI red — fixed (root cause: test-budget miscalibration, not a logic defect)

The two red checks were `Unit Tests` + the `CI Status` aggregate. The single failing test was `live_session_driver_test.exs:346` — **"1b. fold-before-seal ordering"** (CI annotation; the run tested exactly this HEAD, `a27908799`).

Root cause, traced under a reproduction, not guessed: that test (and its `history_text/1`-polling siblings) polls a **full `SealOracle.replay/2`** — a whole-terminal-emulator replay of the cumulative device output — every **10ms** against a **5s** ceiling, while the async `forwarder → cadence → driver` pipeline it is waiting on contends for the same schedulers. On a loaded box that scheduling latency blows the budget. Evidence:

- **100% green unloaded** and **100% green at 50% core contention** (8 `yes` burners, 4× the file).
- Fails **only under deliberate 15-core `yes`-burner saturation** (12–13/15), where the machine also hits the 60s per-test ceiling — the signature of scheduling starvation / resource exhaustion, **never a logic error**: the same assertions pass the moment schedulers free up. The fold-before-seal hold itself is timing-independent (`stream_open?: true` forces `reveal_finished? == false`, so the newest block is held regardless of when the poll lands — the `refute` at :371 cannot race).

Fix (`test/harness/live_session_driver_test.exs`): raise the `eventually` ceiling 5s→15s and **widen the poll interval 10ms→50ms** (this is the load-relevant lever — it cuts the number of heavy replays piled onto the contention while waiting; a green condition still returns on the *first* poll, so unloaded runs are unaffected), and raise the process-startup / dispatch `assert_receive` budgets 500ms→2s. No production code touched for this; the calibration rationale is in the `eventually/3` doc comment.

Validated: `test/harness/` 574 passed (3× clean) and agent-side harness incl. the integration keystone 36 passed; format + warnings-as-errors clean.

### 2. The HIGH (quadratic reproject + unbounded retention) — VERIFIED, disclosed in-artifact, owned

You are right, and I'm not going to argue it down. `apply_batch_item/2` does `Surface.append_events(model, [map])` (`surface.ex:852`, `events: model.events ++ events` — never sheds) then `Surface.advance/2`, whose `run_advance/2` (`surface.ex:857-863`) runs `Projection.project(Enum.take(events, revealed), …)` over the **whole revealed prefix** every event. `Projection.project/2` is a full O(n) fold (`Map`-index build + `bucket_by_turn` + reduce). Per live event that is O(n); over a stream, **O(n²) CPU and unbounded memory**. Fixtures never reach it (bounded, known event list). Confirmed.

This is a real defect on the **unwired path** (your point 3 — the whole live loop is test-only; SEAM #4 in the ledger: no shipped tty embedder, the keystone is the only end-to-end consumer, and it streams only short sessions). Per the growth-inventory + dormant-code disclosure discipline, I've **disclosed it in the artifact** (the `LiveSessionDriver` moduledoc now carries a "Growth characteristic (SEAM-bounded)" section) rather than left it implicit — and I've done the design work so the fix is a known quantity, not a TODO:

- **Fix = turn-granularity compaction**: drop the source events of *fully-sealed* turns once their bracket has folded. I commissioned a read-only audit of whether this preserves projection: it's **safe** because turns project **independently** (there is *no* session-wide id index coupling them — I'd mis-stated that risk myself; `build_turn/2` is turn-local), the `tool_use`/`tool_result` merge is **intra-turn**, and the one genuine cross-block read (seal-time `RecencyPolicy.grade_block/2`) has `turns_behind` **invariant** to dropping older whole turns.
- **Why it is NOT bolted onto this diff**: compaction rewrites the seal-frontier bookkeeping (`painted_count`, fold-override indices) and its correctness hinges on a **multi-turn live/fixture BYTE-parity** guard that does not yet exist. A subtle bookkeeping error there corrupts *sealed history* — permanent once a row scrolls into native scrollback (the FRAME-ORDER LAW's exact hazard), i.e. a failure strictly worse than latency, on a path that cannot bite a user today. So it lands as its own reviewed unit **with the embedder wiring** — the same SEAM that first makes an unbounded stream reachable. Owner: harness-ui lane.

### Ledger PATCH — new before-live-traffic precondition row

| Precondition (source review) | Status |
|---|---|
| Live-append projection is O(n²) CPU + unbounded `model.events` retention (whole-prefix reproject) — Drew #627 R (2026-07-17) | **VERIFIED; outstanding — owner: harness-ui lane.** Inert while unwired (SEAM #4, no production embedder). Fix designed: turn-granularity sealed-turn compaction (audit: SAFE — turns project independently, tool-merge intra-turn, recency `turns_behind`-invariant to dropping older turns), gated on a multi-turn live/fixture byte-parity test; lands **with** the embedder-wiring unit. Disclosed in the `LiveSessionDriver` moduledoc. |

No LIVE rows changed. `gh pr checks` re-runs on the pushed commit.

DROOdotFOO (commented):

Since last review: a2790879..b9215f44

Focus: did the new commit `b9215f44` ("Calibrate live-driver test budgets
for loaded CI; disclose live-append growth") FIX the prior HIGH, or only
DISCLOSE it, and did the MED steer/interrupt items clear?

- **Steer concurrent-dispatch dedup added.** `live_session_driver.ex:460-471`
  now refuses a second steer while one is in flight ("steer already in
  flight -- wait for its decision") instead of racing two `Task.async`
  calls against the lane CAS. This is a genuine, tested new guard.
- **Unrecognized batch element no longer fail-open** (commit 4e45edb2):
  `apply_batch_item/2` catch-all at `live_session_driver.ex:639-647` now
  seals an honest marker instead of silently dropping. Correct hardening.
- **CI-budget widening is honest, not a perf mask.** The `eventually/3`
  ceiling change (`live_session_driver_test.exs:137`, 5s->15s, 10ms->50ms)
  targets the *observer*: the predicate is a full `SealOracle.replay/2`
  terminal-emulator replay on every poll, contending for starved
  schedulers under `async: true`. Widening reduces heavy replays under
  contention; a green condition still returns on the first poll. This does
  NOT paper over the quadratic apply cost -- the streamed sessions are
  short and never reach the quadratic regime, so there is nothing there to
  mask. Accepted.

### HIGH

- **Quadratic reprojection / unbounded `model.events` growth: DISCLOSED,
  NOT FIXED. (reproduced)** The live path is byte-for-byte the same as my
  prior review. `live_session_driver.ex:602-605` (`apply_batch_item({:event,
  map}, ...)`) still does `Surface.append_events(model, [map])` then
  `Surface.advance(model, now)`; `surface.ex:786-791` (`advance/2`) still
  does `Enum.take(model.events, revealed)` -> `Projection.project/2` on
  EVERY event (O(n) per event, O(n^2) cumulative); `surface.ex:852`
  (`append_events`) still does `model.events ++ events` and never sheds a
  sealed event (unbounded retention). The commit's contribution is a
  moduledoc block at `live_session_driver.ex:145-172` that opens
  "DISCLOSED, not yet fixed" and defers the fix to "a turn-granularity
  compaction unit that lands with the embedder wiring." Disclose is not
  fix -- the code stays quadratic.

  Severity call: this stays a HIGH, but it is currently **inert / not
  reachable in production**, so it does not BLOCK today:
  - The driver is still **UNWIRED** -- `git grep LiveSessionDriver` over
    `lib/**` + `packages/**/lib/**` at b9215f44 returns only its own
    definition file; no embedder start/build/spawn call site exists.
  - The only end-to-end consumer is the keystone test, which streams short
    sessions.
  - The real agent lane's `steer/1` is a hard refusal
    (`agent/harness/session_lane.ex:91`, `{:error, :no_steer_channel}`
    always), so no shipped path drives a long/fast unbounded stream.

  **Gating condition (must not be lost):** the SEAM #4 embedder-wiring PR
  is the first change that makes an unbounded stream reachable. Compaction
  MUST land in that same unit, behind the promised multi-turn live/fixture
  BYTE-parity guard -- it cannot ship after the wiring. If a reviewer sees
  the embedder wired without compaction + parity test in the same diff,
  that is a BLOCK. Recommend the PR wiring ledger record this as a hard
  dependency, not a follow-up.

### MED

- **Steer `Task.async` still has no timeout. (reproduced)**
  `live_session_driver.ex:485` spawns `Task.async(fn -> lane_mod.steer(...)
  end)`; the result is handled reactively via `{ref, result}` at
  `:379-380`. There is no `Task.await`/`Task.yield`/`Task.shutdown` and no
  `after` deadline anywhere in the module. A lane whose `steer/1` hangs
  leaves `steer_task` non-nil forever, and the new dedup guard at `:462`
  then refuses EVERY subsequent steer permanently -- the dedup fix
  actually converts a hang into a permanent steer wedge. ESC-interrupt is
  unaffected (that was already true), so this is not a full-loop wedge.
  Inert on the real path only because the real lane refuses instantly; it
  becomes live the moment a real steer channel lands. Fix belongs with the
  steer-channel unit: a `Task.yield` + `Task.shutdown` deadline that clears
  `steer_task` and renders a timeout notice.
- **`current_turn_id` lag under the input-first receive. (reproduced)**
  The input-first `receive ... after 0` (`:367-377`) still handles a queued
  ESC/Tab ahead of a queued `{:render_batch, ...}` that would advance the
  turn; `current_turn_id` is only updated when the batch is applied
  (`apply_lifecycle`, `:651-653`). So interrupt (`:440`,
  `interrupt_payload(state.current_turn_id)`) and steer (`:478`,
  `expected_turn_id: state.current_turn_id`) can carry a stale turn id.
  Partially mitigated: steer is guarded server-side by the lane's
  stale-turn rejection (the "names both turns" branch), but interrupt is
  fire-and-forget with no such guard, so an ESC can target an
  already-completed turn. This is an accepted consequence of the
  input-priority design the moduledoc defends; acceptable if the interrupt
  payload's turn id is treated as advisory by the receiver.
- **Interrupt has no in-flight dedup. (reproduced)**
  `handle_surface_command(_, %{type: :interrupt})` (`:438-458`) re-invokes
  `lane_mod.interrupt` on every ESC with no equivalent of the steer
  in-flight guard. Low impact (interrupt is idempotent fire-and-forget,
  each press just re-renders the sent notice), but asymmetric with the new
  steer dedup; worth a one-line note that this asymmetry is intentional.

### Verdict: CONCERNS

The prior HIGH (quadratic reprojection + unbounded `model.events`) is
**DISCLOSED, not fixed** -- the apply/advance/append code is unchanged and
a moduledoc now documents the growth and defers the remedy. It remains a
HIGH by nature, but is genuinely inert today: the driver is unwired, the
only consumer streams short sessions, and the real lane refuses steer, so
the quadratic regime is unreachable in production. That inertness is why
this is CONCERNS and not BLOCK -- conditional on the compaction unit
landing in the SAME PR that wires the embedder (SEAM #4), behind the
multi-turn byte-parity guard. The steer/interrupt MEDs did NOT clear: no
steer timeout (now made permanent-wedge-prone by the added dedup),
`current_turn_id` lag persists (steer server-guarded, interrupt not), and
interrupt still has no dedup. The CI-budget widening is legitimate observer
calibration, not a perf mask.

merklebonsai:
### Merge + CI status (commit `f9e166601`)

Merged latest `master` (it had advanced 6 merges, several touching the harness) — the branch was `CONFLICTING`. The only conflict was in `surface.ex`'s footer rendering: `master` refactored `footer_lines/1` into the priority-aware `fit_footer_lines` structure (+ projection-panel overlay refresh and the diff-expansion clause); this branch had added the persistent live-session `lane_notice` line. Resolved by folding `lane_notice` in as a fit group placed right after `status` and deliberately **kept out of `drop_order`** — an honest report channel that is never shed to fit the budget, same treatment as the stub notice.

Post-merge, all green:
- **Unit Tests: pass** (the check Drew flagged red), Integration, Property, Compilation, Format, Test windows-2022 all pass.
- Full harness suite 699 passed / agent-side harness incl. the integration keystone 36 passed, locally.
- `Test macos-latest` first hit the known master-side `stream_cadence_test.exs:94` yield-budget timing flake (unrelated to this diff — StreamCadence, untouched here); **passed on rerun**.

PR is now `MERGEABLE` / `CLEAN`.

merklebonsai:
## Review response — round 2 (commit `40e619515`)

Thanks for the conditional-accept on the HIGH disclosure and the CI calibration. The three steer/interrupt MEDs, each behavioral → red-first:

### MED 1 — steer wedge (no timeout on the in-flight guard)

Fixed, and you're right that my dedup made it wedge-prone. The Task is already monitored, so a steer that **crashes** clears the guard via `handle_down/4` — the gap was a steer that neither replies nor crashes (a hung lane call), which pinned `steer_task` and refused every future steer for the session's life.

Added a **liveness bound on our own steering resource** (not a latency claim about the lane): on dispatch, a per-Task-ref `Process.send_after(self(), {:steer_timeout, ref}, steer_timeout_ms)` (default 5s). On expiry `handle_steer_timeout/2` does `Task.shutdown(task, :brutal_kill)`, releases the guard, and renders `» steer timed out after Nms — channel may be wedged; try again`. The guard is now cleared the instant the reply lands (`handle_task_result/3`, not on the trailing normal `:DOWN`), so a timer firing in the reply→DOWN window finds a cleared/newer guard and no-ops; the timer is ref-scoped, so a steer that resolves first leaves only a stale ignored timeout.

**Tripwire (`5b`, red-first):** a lane whose `steer/2` hangs forever → after the timeout, a **second** steer dispatches. Fails pre-fix (the second steer is refused with "already in flight", so the lane's `steer/2` is never called again and the second `{:steer_dispatched, _}` never arrives). Verified red by disabling the `send_after` line: `0/1`.

### MED 2 — interrupt turn_id lag (steer is CAS-guarded, interrupt is not)

Verified against the **real** `Raxol.Agent.Interrupt.interrupt/3` contract: it targets a running turn's `tool_ref` and takes only `:reason`/`:actor` — **there is no turn-id selector**. Interrupt kills whatever turn is *currently running*; the payload `turn_id` is advisory attribution. So a stale `current_turn_id` **cannot** kill the wrong turn (the running turn is decided at the session, not in the driver) — it can only mislabel the driver's own optimistic "sent" notice. This is the "kill the CURRENT turn by design → payload turn_id is advisory" branch, and it's what's true.

So per your instruction: (a) documented it in the moduledoc (new "The interrupt turn id is ADVISORY" section, citing the `interrupt/3` signature), and (b) made the ack rendering **honest about which turn actually died** — the `turn_canceled` ack now renders the turn from the **event** (`turn_canceled_notice/1`), never from `current_turn_id`. The optimistic "interrupt sent (turn X)" line stays advisory ("awaiting confirmation") and names the belief; the authoritative statement is the event-observed ack.

**Tripwire (`3b`, red-first):** advance the belief to `t2`, then land a `turn_canceled` for `t1` → the ack must name `t1`. Fails pre-fix (the old ack carried no turn id at all): verified red `0/1`.

### MED 3 — interrupt dedup asymmetry

Added exactly the one-line note you suggested (moduledoc, "Interrupt has no in-flight dedup; steer does"): interrupt is idempotent fire-and-forget so each ESC just re-renders the sent notice, and refusing a second ESC would be worse (a user leaning on ESC to stop a runaway tool must never be told "already interrupting, wait"); steer dedups only because it races the CAS.

### Ledger PATCH — compaction row hardened to a merge-gating precondition

Per your accept condition, the O(n²)/unbounded-retention row is upgraded from "outstanding, owner TBD" to a **hard precondition** on wiring:

| Precondition | Status |
|---|---|
| Live-append projection O(n²) CPU + unbounded retention (Drew #627) | **VERIFIED; HARD PRECONDITION — owner: harness-ui lane.** Inert while unwired (SEAM #4). **The compaction unit MUST land in the SAME PR that wires the embedder** (the PR that first makes an unbounded live stream reachable), gated on a **multi-turn live/fixture byte-parity guard**. Fix design (turn-granularity sealed-turn compaction) + safety audit disclosed in the `LiveSessionDriver` moduledoc. This PR must not merge an embedder without it. |

Gates: warnings-as-errors clean; `test/harness/` **701 passed**; agent-side harness incl. integration keystone **36 passed**; format clean. Both new tripwires verified red-first. `gh pr checks` re-runs on `40e619515`.

DROOdotFOO (commented):

Since last review: f9e16660..40e61951

Scope: `40e61951 Live driver: steer wedge timeout + honest interrupt turn-id semantics`
(`lib/raxol/harness/live_session_driver.ex` + `test/harness/live_session_driver_test.exs`).
Verified the two directly-targeted MEDs, re-confirmed the standing HIGH, and
looked for regressions introduced by the timeout/dedup changes.

**MED (steer wedge) -- FIXED.** The prior review's sharpest point: the
single-in-flight steer guard (`steer_task`) had no expiry, so a lane
`steer/2` that neither replied nor crashed pinned the guard and refused every
future steer for the life of the session. This commit adds a per-dispatch,
ref-scoped liveness bound: `Process.send_after(self(), {:steer_timeout,
task.ref}, state.steer_timeout_ms)` (default 5s) at dispatch
(`live_session_driver.ex:537-541`); on expiry `handle_steer_timeout/2`
(`:594-601`) `Task.shutdown(task, :brutal_kill)`s the hung task, releases the
guard (`steer_task: nil`), and renders an honest "steer timed out ... channel
may be wedged; try again". The guard is now cleared the instant the reply
lands in `handle_task_result/3` (`:583`, `steer_task: nil`), not on the
trailing normal `:DOWN`, so a timer firing in the reply->DOWN window finds a
cleared/newer guard and no-ops via the catch-all `handle_steer_timeout(state,
_ref)` (`:603`). Ref-scoping is genuine: a stale timer for task A while task B
is in flight cannot match B's guard (both `ref` positions bind the same
variable in the head), so it cannot false-wedge a live steer. The hung-lane
case is exercised red-first by test 5b (`:hang` lane = `Process.sleep(:infinity)`,
`steer_timeout_ms: 150`), which asserts a SECOND steer dispatches after the
timeout -- the whole falsifier, since the second `{:steer_dispatched, _}` can
only arrive if the guard was released. (reproduced by inspection of the code
path + the red-first 5b test.)

**MED (interrupt turn-id lag honesty) -- FIXED.** The prior concern was that
the input-first receive lets `current_turn_id` lag a completed turn, so
ESC/Tab could fire against an already-completed turn, and interrupt had no
in-flight dedup. The commit reframes the turn-id half correctly and I verified
the reframe against the real code: `Raxol.Agent.Interrupt.interrupt/3`
(`packages/raxol_agent/lib/raxol/agent/interrupt.ex:204-257`) takes
`(tool_ref, sink, opts)` with only `:reason`/`:grace_ms` in opts and NO turn-id
selector -- it kills whatever `tool_ref` it is handed (the session resolves the
currently-running turn). The `SessionLane` contract
(`session_lane.ex`, `interrupt/2`) is fire-and-forget, event-observed. So a
stale `current_turn_id` can never route a kill to the wrong turn -- it can only
mislabel the driver's own optimistic notice. The fix makes the AUTHORITATIVE
ack source its turn from the EVENT: `apply_lifecycle(state, %{type:
:turn_canceled} = event)` now calls `turn_canceled_notice(Map.get(event,
:turn_id))` (`:767-771`), never from `current_turn_id`. Red-first test 3b
advances the belief to `t2`, lands a cancel for `t1`, and asserts the ack names
`t1` (and refutes `t2`). Steer's lag is separately server-guarded: a stale
`expected_turn_id` yields a CAS `{:error, {:stale_turn, ...}}` rendered
honestly. (reproduced by inspection + 3b.)

**MED (interrupt dedup asymmetry) -- RESOLVED BY DESIGN + documented.** The
requested one-line moduledoc note landed (`:124-135`): interrupt has no
in-flight dedup by design (idempotent fire-and-forget; each ESC re-renders the
"sent" notice), unlike steer which dedups because it races the CAS. The design
holds against the lane contract: N ESC presses = N cheap dispatch calls, at
most one running turn to kill; signaling an already-signaled turn no-ops at the
staged kill. Refusing a second ESC would be worse for a user leaning on ESC to
stop a runaway tool. Accepted.

### Low

**New Hire -- slow-but-alive steer killed at the timeout.** `steer_timeout_ms`
brutal-kills the in-flight Task on expiry, but a lane `steer/2` that is slow
rather than wedged may have ALREADY dispatched the steer to the session before
the 5s bound; `Task.shutdown` drops the reply and the surface shows "channel
may be wedged; try again" for a steer that possibly succeeded. This is
acceptable: the wording is hedged ("may be"), the bound is configurable, and it
self-heals on retry -- the server CAS returns duplicate delivery, rendered as
"steer already accepted earlier (duplicate delivery)" (`:614-618`). The commit
correctly frames this as a liveness bound on the driver's own resource, not a
latency claim about the lane. Non-blocking; worth a glance if real lanes ever
approach 5s legitimately.

**Saboteur -- the fixed failure class still applies verbatim to `interrupt/2`,
which has no backstop.** `handle_surface_command(state, %{type: :interrupt})`
calls `lane_mod.interrupt(session, payload)` SYNCHRONOUSLY in the loop
(`:479-497`) -- unlike steer, which was moved to `Task.async` precisely so a
hung lane call cannot wedge the loop. If a lane's `interrupt/2` ever blocks,
the exact hung-lane failure this commit just spent effort bounding for steer
would wedge the ENTIRE loop, including ESC and render, with no timeout. This is
mitigated only by the unenforced contract precondition ("interrupt is cheap --
dispatch, not execution", moduledoc + `session_lane.ex`). Given that precondition
it is low risk, but it is the same shape the commit is about, left asymmetric.
Non-blocking; flag for the day a real lane's interrupt path does I/O.

**Security Auditor -- no new injection surface; advisory reframe is sound.**
The new `{:steer_timeout, ref}` message is self-sent (`Process.send_after` to
`self()`) and ref-scoped, so it is trusted and cannot be spoofed to false-wedge
across sessions. `turn_canceled_notice/1` renders the event `turn_id` via
`inspect/1` -- no `String.to_atom`, no atom minting, consistent with the
`EventBoundary` posture. The advisory-turn-id claim is verified at the boundary
the driver actually calls (`interrupt.ex:204-257` has no turn-id target), so
shipping `current_turn_id` in the interrupt payload as `%{turn_id: ...}` is
attribution only and cannot misroute a kill. Clean for this commit.

### High (standing, unchanged -- disclosed + unwired)

The O(n^2) reveal remains: `apply_batch_item/2` appends each live event into
`model.events` and re-derives the whole revealed prefix through
`Surface.advance/2` per event (`:697`), with `model.events` never shedding a
sealed event -- O(n^2) CPU + unbounded retention over an unbounded live stream.
Re-confirmed still only DISCLOSED (moduledoc "Growth characteristic" block,
`:179-203`) and NOT fixed, and the driver is still UNWIRED: the only non-test
reference is the agent-package integration test
(`packages/raxol_agent/test/.../live_session_agent_test.exs`); no supervisor,
embedder, or mix task feeds it a real session. Inert today, so it stays
non-blocking -- but the standing merge condition is unchanged and restated in
the verdict.

### Verdict: CONCERNS

All three steer/interrupt MEDs are genuinely cleared: the steer-wedge timeout
is a correct, ref-scoped liveness bound (guard released on reply, no false
wedge, red-first 5b); the interrupt turn-id ack is now event-sourced and the
advisory reframe is verified against the real `Interrupt.interrupt/3` (red-first
3b); the dedup asymmetry is documented and sound. Two new LOW notes (slow-but-
alive steer killed at the bound, self-healing via server duplicate-dedup; and
`interrupt/2`'s synchronous no-backstop call being the same hung-lane shape the
commit just fixed for steer) are non-blocking.

CONCERNS rather than CLEAN solely because the HIGH quadratic remains open.
Standing condition (unchanged): **the turn-granularity compaction fix must land
in the SAME PR that wires SEAM #4 (the embedder that first makes an unbounded
live stream reachable), or that wiring PR is a BLOCK.** While the driver stays
unwired, this diff is safe to merge.

PR #628, opened 2026-07-17
Harness UI: transcript search (fuzzy body search over the block corpus)

merklebonsai:
`/` in transcript-browse opens a search picker over every block's content — kind, summary, and the body text the summary never shows. Typing fuzzy-ranks; picking jumps focus to the block. Also a labeled palette entry via the merged registry derivation (no parallel list).

## Corpus (`Block.search_text/1`, new public)

`"<kind> · <summary>"` prefix plus per-kind body: message/reasoning/opaque get full content text; tool calls get the result (the summary already carries name+args); approvals get the full multi-line action + binary options; diffs get old+new (summary carries the path). No sanitize/truncate in derivation — `ViewText.lines/3` stays the SINGLE trust boundary for control bytes and display width.

## Label + bounds

The overlay's label is static (it is both search key and rendered row), so labels are the flattened corpus clamped to 400 graphemes — bounding per-keystroke fuzzy ranking on the synchronous input path (the session-picker cap precedent). The visible width-truncated head is always `kind · summary`, identifying the block even when the match sits deep in the body. Content past the cap is not searchable — documented AND pinned by a named test, not silent.

## Bind

`/` was free; `:not_composing` guard, same class as the jump/session picker binds: typed text while composing, filter text while any overlay is open. Picking uses the jump picker's exact focus path; empty projection refuses with an honest notice.

## Verification

Red-first (captured: `search_text/1` undefined x8, `/` resolving passthrough, 6/8 e2e failing behaviorally). 23 new tests incl. the body-search proof — a token present only in block 3's SECOND body line filters to it and focus lands on 3 — plus hostile-byte sanitization, CJK clamp, and past-cap pins. Targeted suites 126/126; full suite 6195 passed / 0 failed; warnings + format clean.

### DIFF
--- a/lib/raxol/harness/surface.ex
+++ b/lib/raxol/harness/surface.ex
@@ -510,17 +510,22 @@ defmodule Raxol.Harness.Surface do
   back to, so `:edit_draft` there seals one honest history line saying
   so instead of pretending.
 
-  ## The pickers (command palette, jump, session)
+  ## The pickers (command palette, jump, session, search)
 
-  Three more `OverlayPicker` consumers ride the same footer-overlay
+  Four more `OverlayPicker` consumers ride the same footer-overlay
   substrate as the picker described above. Ctrl+P (`Keymap`'s
   `:open_palette`, an `:always` chord) opens the command palette from
   anywhere, including mid-compose -- a chord is never typed text, the same
-  reasoning as Ctrl+E. `g` (`:open_jump_picker`) and `s`
-  (`:open_session_picker`) are plain printable letters gated
-  `:not_composing`, the same class as `z`/`j`/`k`: they only resolve in
-  transcript-browse mode, never stealing a letter out of the composer's
-  typed text.
+  reasoning as Ctrl+E. `g` (`:open_jump_picker`), `s`
+  (`:open_session_picker`), and `/` (`:open_search_picker`) are plain
+  printable letters gated `:not_composing`, the same class as `z`/`j`/`k`:
+  they only resolve in transcript-browse mode, never stealing a letter
+  out of the composer's typed text. `open_search_picker/1`'s entries are
+  labeled from `Block.search_text/1` -- a content-derived search corpus
+  (kind, summary, AND body text), not just the summary header
+  `open_jump_picker/1`'s labels use -- clamped per block (see that
+  function's own doc) before `Raxol.Harness.Surface.ViewText.lines/3`
+  ever truncates a rendered row to its display-width budget.
 
   The palette's entries are `Keymap.palette_binds/0` (the labeled subset
   of the bind table) plus two commands that exist only at THIS assembly
@@ -529,7 +534,7 @@ defmodule Raxol.Harness.Surface do
   already exposes as direct API). Picking any palette entry dispatches
   through the exact same `dispatch_command/2` path a keypress takes --
   there is no second, parallel execution mechanism for a palette-picked
-  command. All three pickers (palette, jump, session) opt into
+  command. All four pickers (palette, jump, session, search) opt into
   `Raxol.UI.Harness.OverlayPicker.fuzzy_filter/3` as their `filter_fn`
   (the `Raxol.UI.ListScorer` adapter), not the default substring filter --
   a fuzzy-ranked query is what a "type a few letters, find the entry"
@@ -622,9 +627,17 @@ defmodule Raxol.Harness.Surface do
 
   @default_footer_rows 6
   @default_sessions_dir Path.join(["test", "fixtures", "harness", "sessions"])
-  # Cap on session-picker entries -- see `open_session_picker/1`'s
-  # "Listing cap" doc section (bounded work on the input path).
+  # Cap on session-picker ENTRIES (the entry-count axis) -- see
+  # `open_session_picker/1`'s "Listing cap" doc section.
   @session_picker_cap 100
+  # Cap on the per-label GRAPHEME COUNT of each block's search corpus (the
+  # per-label length axis -- a DIFFERENT axis from `@session_picker_cap`'s
+  # entry count). This is the canonical per-label bound for the search
+  # picker: it is enforced at the source inside `Block.search_text/2`
+  # (which never scans past it), so it also bounds `ListScorer`'s own
+  # `@max_score_graphemes` (a 400-grapheme label is already under the
+  # scorer's 1024 cap). See `open_search_picker/1`'s doc section.
+  @search_label_cap 400
   @stub_interrupt_notice "» interrupt requested (stub — no agent lane in fixture mode)"
 
   @type mode :: :inline_log | :tmux_conservative | :flat
@@ -1613,6 +1626,9 @@ defmodule Raxol.Harness.Surface do
     model |> open_panel(kind) |> handle_open_result(model)
   end
 
+  defp dispatch_command(model, %{type: :open_search_picker}),
+    do: open_search_picker(model)
+
   defp dispatch_command(model, %{type: :focus_transcript}),
     do: focus_transcript(model)
 
@@ -2203,7 +2219,7 @@ defmodule Raxol.Harness.Surface do
     end
   end
 
-  # -- pickers (command palette, jump, session) ----------------------------
+  # -- pickers (command palette, jump, session, search) ---------------------
   # See the moduledoc's "The pickers" section.
 
   @doc """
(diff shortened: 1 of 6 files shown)


DROOdotFOO (commented):

Three hostile personas over the transcript-search diff (`surface.ex`, `block.ex`, `view_text.ex` boundary it leans on, `keymap.ex`, tests). Per-keystroke fuzzy work is well bounded (`ListScorer` caps every label at `@max_score_graphemes 1024` + `bounded_key_prefix`, and items are built once at open, not per keystroke -- `/` is passthrough/filter-text while an overlay is open). The findings below are about *where* the bound is applied and *which* control bytes the "single trust boundary" actually catches.

### HIGH

- **[Saboteur + New Hire] The `@search_label_cap 400` clamp bounds the label OUTPUT, not the WORK -- `open_search_picker/1` does unbounded synchronous work on the input path.** (reproduced by code path) `search_label/1` (`lib/raxol/harness/surface.ex:1858-1862`) is `Block.search_text() |> String.replace(["\r\n","\n","\r"], " ") |> String.slice(0, @search_label_cap)`. `Block.search_text/1` (`lib/raxol/ui/components/harness/block.ex:583-584`) first materializes `prefix <> " · " <> body` where `body` is the FULL, explicitly-"unbounded, untrusted" `content.text` / tool result / diff old+new. So per block, at picker-open, the code concatenates the entire body and runs `String.replace` across the entire body BEFORE the `String.slice(0, 400)` ever throws all but 400 graphemes away. The 400-grapheme cap makes the *result* small; it does nothing to bound the O(body-size) flatten+replace, run once per block, summed over every block in `projection.blocks` (which -- like `open_jump_picker/1` -- has no entry cap, unlike `open_session_picker/1`'s `@session_picker_cap 100`). This is strictly worse than the jump picker it mirrors, which keys off bounded `Block.summary/1` (line 1 only) and never touches full bodies. The doc's claim that the clamp is "bounding per-keystroke fuzzy ranking ... the same bounded-work-on-the-input-path reasoning `open_session_picker/1`'s `@session_picker_cap` already documents" (`surface.ex:1805-1811`) overstates what the clamp site achieves. Fix: slice/prefix-bound the body inside `search_text/1` (or before `String.replace`) so the flatten never scans past the cap. New Hire flags the same root independently (doc says the clamp bounds the input path; the clamp is downstream of the unbounded work) -> promoted MEDIUM -> HIGH. Practical exploitability is gated by max fixture/context body size, so real-world impact is moderate, but the stated invariant is not the one the code enforces.

### MEDIUM

- **[Security Auditor] The "SINGLE trust boundary" this feature leans on has a C1 control-byte hole (0x80-0x9F), and the hostile test only proves C0/ESC.** (reproduced) The PR routes newly-exposed full block BODIES (tool output, streamed LLM text) into rendered picker rows and rests the entire safety argument on `ViewText.lines/3` as the one sanitizer (`block.ex` search_text doc; `surface.ex` open_search_picker doc). `ViewText.sanitize/1`'s allowlist is `(byte >= 0x20 and byte != 0x7F) or byte == ?\t` (`lib/raxol/harness/surface/view_text.ex:221-227`). Every C1 byte 0x80-0x9F is `>= 0x20`, so it passes: I reproduced `<<0x9B>>` (raw CSI) and `<<0x84>>` (IND) surviving `sanitize`, while `0x01` and `0x1B`/ESC are stripped. In valid UTF-8 those bytes only appear as multi-byte continuations (safe, as the comment argues), but a Latin-1 / invalid-UTF-8 / raw agent stream can carry standalone C1 bytes that terminals may act on as controls. This is the #616 C1 gap, now newly exercised by full-body search content. The hostile-content test (`test/harness/command_palette_surface_test.exs:734,747`) asserts only `refute plain =~ <<0x01>>` and an ESC/SGR string -- it never checks 0x80-0x9F, so it gives false assurance for exactly the class that slips through. The sanitizer is pre-existing, but this PR both widens what flows through it and re-affirms it as the sole boundary. Fix: extend the allowlist to strip 0x80-0x9F (or gate on `String.valid?` / `String.printable?`); add a C1 case to the hostile test.

### LOW

- **[New Hire] Three overlapping caps in three units, in two different dimensions, cross-referenced as if they were the same thing.** `@search_label_cap 400` graphemes (`surface.ex:510`), `ListScorer.@max_score_graphemes 1024` graphemes, and `@session_picker_cap 100` ENTRIES. `search_label` clamps to 400, then `ListScorer` would clamp to 1024 (a no-op for a 400-grapheme label), so the 400 is a second, stricter, undocumented-as-to-value per-label bound layered on top of the scorer's own. The doc justifies 400 by citing `@session_picker_cap` "the same reasoning" -- but that cap bounds the NUMBER of entries, a different axis from per-label grapheme length; a new reader can't tell why 400 (not 100, not 1024, not "reuse the scorer's cap"). Pick one canonical per-label bound and name the axis it governs.

- **[Saboteur] No entry cap on the block corpus.** Unlike `open_session_picker/1` (`@session_picker_cap 100`), `open_search_picker/1` builds one item per block with no ceiling (`surface.ex:1834-1840`). Bounded per-item cost (see HIGH fix) plus `ListScorer`'s internal caps keep a single keystroke's ranking near-linear, and this matches the accepted `open_jump_picker/1` precedent, so it is not itself a defect -- but combined with the late-clamp HIGH it is the multiplier (10k blocks x full-body flatten at open). Noting for completeness; folds into the HIGH fix.

### Checked, not a finding

- **Sealed / redacted blocks:** `seal` is `:live | :sealed` = commit/immutability state, not a redaction/secrecy boundary (`block.ex` moduledoc + `defstruct`). Sealed blocks are meant to be visible in the transcript, so `search_text/1` reading them exposes nothing the transcript doesn't already show. No privilege/data-exposure issue there.
- **ReDoS / quadratic query blowup:** `ListScorer` is a two-pointer subsequence pre-check + O(query x key) DP with both key and query grapheme-capped; no user-supplied regex reaches the matcher. A crafted query cannot force super-linear blowup beyond the capped DP.
- **Empty / no-match / compose / overlay-stacking:** empty projection is an honest notice, no-match returns `[]`, `/` is passthrough while composing and filter-text while an overlay is open -- all covered by tests and `:not_composing` guard. Live-stream corpus mutation is N/A in fixture mode (synchronous single state machine).

### Cross-persona overlaps

- Late-applied clamp: Saboteur ("clamp bounds output, not work") and New Hire ("doc claims the clamp bounds the input path") independently hit the same root -> promoted MEDIUM -> HIGH.

### Verdict: CONCERNS

One promoted HIGH (clamp applied downstream of the unbounded flatten/replace it is documented to bound) and one MEDIUM (C1 gap in the sole sanitizer, untested for 0x80-0x9F) warrant changes before merge; neither is a CRITICAL nor is there a 3rd HIGH, so this is not a BLOCK. Address the two and it is clean.

DROOdotFOO:

CI is green, but the review HIGH stands: picker-*open* does unbounded O(body) work -- the `@search_label_cap 400` clamp is applied only AFTER `Block.search_text/1` materializes and flattens the full untrusted block body. Bound the open path (clamp before the flatten) before merge. The separate C1-sanitizer MEDIUM is fixed at the source in #632.


DROOdotFOO:
**Update: the C1-sanitizer MEDIUM is now CLOSED** via #632 (merged to master -- `ViewText.sanitize/1` is code-point-aware now). But the **HIGH still stands and still blocks merge**: picker-open does unbounded O(body) work because `@search_label_cap 400` is applied AFTER `Block.search_text/1` flattens the full untrusted block body. Clamp before the flatten, then this is ready.

merklebonsai:
## Review round 1 — resolution ledger

Rebased onto current `origin/master` first (branch was conflicting after #629/#632). Conflicts resolved as **unions**: `keymap.ex` keeps both #629's `w`/`m`/`n` `:open_panel` binds and this PR's `/` `:open_search_picker` bind (+ the command-type union); `surface.ex` keeps both the panel dispatch clause and the search dispatch clause. Force-with-lease pushed as a single clean commit.

### HIGH — clamp bounded the OUTPUT, not the WORK  → **VERIFIED, FIXED**

Confirmed exactly as described: `search_label/1` called `Block.search_text/1` (which materialized `prefix <> " · " <> FULL_body`) and only *then* ran `String.replace` + `String.slice(0, 400)` — the flatten/concat scanned the whole untrusted body per block before anything was thrown away.

Fix pushes the clamp **to the source** in a new `Block.search_text/2`:
- every body field (`text`/`result`/`old`+`new`/`action`+`options`) is `String.slice`d to `max_graphemes` *before* it is concatenated or joined — `String.slice/3` walks at most the cap and stops, never scanning the tail;
- the **prefix is clamped too** — `summary/1` returns line 1, but a single-line body makes line 1 the whole body, which would re-introduce an O(body) concat. This brings the search path to parity with the accepted `open_jump_picker/1`/`summary/1` baseline and adds no unbounded term of its own;
- `search_text/1` stays the full corpus (`:infinity`) — semantics unchanged, all existing tests green;
- `search_label/1` now calls `/2`; the redundant downstream `String.slice` is gone (the flatten only shortens, so ≤ cap is preserved).

**Red-first proof:** neutering `clamp_graphemes` to the identity function turns 6 of the new `block_search_text_test.exs` cases red (corpus length `100022 > 400`); restoring it → 16/16 green. Docs (`surface.ex` "label clamp" section, `search_text` moduledoc) rewritten to state the bound the code now actually enforces.

### MEDIUM — C1 (0x80–0x9F) hole in the sole sanitizer  → **VERIFIED, CLOSED upstream + regression added**

Confirmed the pre-existing gap. Closed by **#632** (now rebased in): `ViewText.sanitize/1` is code-point-aware and strips `0x20 > cp` / `0x7F` / `0x80–0x9F`, with a lone-high-byte fallback clause for bare C1 like `0x9B`. Added the missing coverage the review flagged: the hostile-content surface test now injects raw CSI `0x9B` and IND `0x84` (plus `0x7F`) and refutes them **against the RAW device bytes**, not the `strip_ansi`'d view — because `strip_ansi`'s `SequenceScanner` would itself consume a lone C1 introducer and pass the assertion for the wrong reason (oracle masking the byte under test). The renderer only ever emits 7-bit `ESC[`, so absence of those bytes in the raw stream is a sound proof of stripping.

### LOW — three caps, two axes  → **FIXED (doc)**

`@search_label_cap` comment now names its axis explicitly: **per-label grapheme length** (canonical for the search picker, and already under `ListScorer.@max_score_graphemes 1024`), a different axis from `@session_picker_cap`'s **entry count**. No longer cross-referenced as "the same reasoning".

### LOW — no entry cap on the block corpus  → **DISCLOSED as deliberate**

Documented in a new "No entry cap" doc section: this mirrors `open_jump_picker/1` (also uncapped) on purpose — an entry cap would silently drop blocks *out of search*, making a block unfindable (a correctness regression, not a safety win). With per-item work now bounded (HIGH fix), a keystroke's ranking is near-linear in `entries × cap`, the same envelope the accepted jump picker runs in.

### "Checked, not a finding" — concur
Agree with all four (seal ≠ redaction; capped DP, no ReDoS; empty/no-match/compose/overlay-stacking covered; live-stream N/A in fixture mode).

### Gates
- `mix format --check-formatted` (changed files): clean
- `block_search_text_test.exs`: **16 passed** (8 new, incl. per-kind corpus≤cap + marker-past-cap + /1-unbounded-intact)
- scorer + picker units (`list_scorer`, `overlay_picker`, `picker_binds`, `picker`): **101 passed**
- surface picker suites (`command_palette_surface`, `overlay_picker_surface`): **44 passed**
- full `test/harness/`: **650 passed, 1 skipped, 18 excluded**

DROOdotFOO (commented):

Since last review: 9c29fd77..4998fed3 (rework: clamp pushed into `Block.search_text/2`, new `block_search_text_test.exs`, C1 hostile-body guard).

- **MED (C1 sanitize gap, #616) -- RESOLVED.** The branch has `origin/master` as an
  ancestor, so it carries #632: `ViewText.sanitize/1` is now code-point-aware
  (`lib/raxol/harness/surface/view_text.ex`, branch rev -- `cp >= 0x80 and cp <= 0x9F`
  strip plus a `<<_byte, rest>>` fallthrough that drops lone raw high bytes). The new
  hostile-body test asserts the **post-#632** behavior (`refute delta =~ <<0x9B>>` /
  `<<0x84>>`, command_palette_surface_test.exs:637-640). No stale pre-#632 assertion
  survives -- verified a lone `0x9B`/`0x84` misses the `<<cp::utf8>>` clause and hits the
  drop clause, and a UTF-8-encoded C1 (`0xC2 0x9B`) hits the `0x80..0x9F` strip. This
  MED is closed.

- **Per-keystroke scoring still bounded (regression check) -- CONFIRMED.** Labels are
  built once in `open_search_picker/1` and each is clamped to `@search_label_cap` (400)
  graphemes; `ListScorer` additionally `Enum.take(@max_score_graphemes=1024)` per side
  (list_scorer.ex:115,270). Per-keystroke work is O(entries x 400), the accepted jump
  envelope. No regression.

### MED (was HIGH: O(body)-on-open) -- partially fixed; the docstring overclaims

The prior HIGH's *worst* aspect is genuinely fixed: the open path no longer allocates a
full-body flattened binary (`String.replace` now runs over the already-clamped <=400
string in `search_label/1`, surface.ex). Allocation-on-open is bounded, and the
per-block body-beyond-line-1 material is clamped at the source before concat/join. Good.

But the central claim is **false for ASCII bodies** (reproduced). The docs assert
`String.slice(text, 0, max)` "walks at most `max_graphemes` graphemes and stops, never
scanning the tail" and thus `open_search_picker/1` does "O(cap) work per block, not
O(body-size)" (block.ex:628-635; surface.ex `open_search_picker/1` doc). In Elixir
1.20.2, `String.slice/3` -> `do_slice` -> `byte_size_remaining_at`, whose **ASCII
fast-path** calls `skip_length(rest, 1)` -- a greedy SWAR loop that scans the *entire*
consecutive ASCII run *before* `min(skip, n)`. So the clamp scans the whole ASCII body:

```
slice(1M  ascii, 0, 400) = 360us
slice(10M ascii, 0, 400) = 11078us   # linear in body size
slice(50M ascii, 0, 400) = 17120us
slice(5M  CJK,   0, 400) = 90us      # multibyte path DOES early-stop
```
(reproduced). ASCII -- English text, code, tool output -- is the common case, so the
source-clamp is O(body) *time* there, not O(cap). `summary/1 -> first_line/1 ->
String.split(text,"\n")` adds a second O(body) byte scan for the prefix.

Why this is MED, not HIGH: (1) it is allocation-free (read-only SWAR, ~8 bytes/us),
done **once per picker-open**, not per keystroke; (2) it is at asymptotic **parity with
the already-accepted `open_jump_picker/1`**, which also does `summary/String.split`
O(body) per block on open -- this PR adds a constant-factor second scan, not a new
asymptotic class. The practical DoS/allocation risk that motivated the HIGH is resolved.
The remaining defect is a documentation/accuracy one: the doc should stop claiming
`String.slice` early-stops (it does not for ASCII) and the "bounds the WORK, not the
output" framing should be qualified -- or the clamp should use a primitive that is
genuinely bounded for ASCII (e.g. a bounded grapheme walk + `binary_part`).

### LOW (Security Auditor: options join-before-clamp)

`search_body_text(:approval, ...)` -> `search_options_text/2` does
`Enum.filter(options, &is_binary/1) |> Enum.join(" ") |> clamp_graphemes(max)`
(block.ex:306-317). Each option string is **not** clamped before the join, so a producer
sending one 100MB binary option (or a very long options list) materializes the full
joined binary *before* the clamp -- an untrusted-input allocation axis the per-field
grapheme clamp does not cover (unlike the message/tool_call path, which clamps the field
first). Bounded in practice (approval options are 2-3 short strings) but it is untrusted
and uncapped by count. Consider clamping each element before join, or capping the list
length. Positive: non-binary options are correctly skipped (no `to_string/1` on
arbitrary terms), non-map `content` degrades to the prefix, and the corpus is never
rendered raw -- picking only sets `focused_index`. Trust boundary (sanitize lives solely
in `ViewText.lines/3`) is respected.

### Verdict: CONCERNS

C1/#616 closed and per-keystroke bounded; the HIGH is downgraded to MED -- the
allocation blowup and per-keystroke risk are genuinely fixed and the open path is at
parity with the accepted jump picker, but the docstring's "O(cap), String.slice stops
early" claim is factually wrong for ASCII bodies (reproduced) and should be corrected.
LOW options-join allocation is a minor untrusted-input gap. No blocker.

PR #629, opened 2026-07-17
Harness UI: projection panels (worktracks/memory/plan as summonable overlays)

merklebonsai:
Worktracks, memory, and plan as summonable read-only overlays, folded from the session's meta events the same journal-fold way blocks derive from lifecycle events. Dismissed != dead: nothing is cached — re-summon folds current retained events.

## ⚠ Accord merge-gate (agent-lane verification requested)

Per the cross-lane accord, this unit BUILDS against the frozen meta contract shapes and MERGES only once verified against real emitted shapes. Grounding found the frozen table (`Raxol.Agent.Meta.Registry`: `extract` with required class/op/item/refs, session scope, grow-only) but exactly ONE meta event exists in shipped fixtures (a gate_decision) — no real extract traffic. The new `projection-panels.jsonl` is therefore a CONTRACT-SHAPE fixture, header-marked as pending real-shape verification, and the caveat is stated in both PR-facing moduledocs. [agent] lane: confirm the extract shapes this folds match what your producers will emit, or hand us a real fixture to swap in.

## Design

- **Pure read-models** (`PanelProjection.fold/2`): fold meta events -> worktracks list / memory entries / plan steps. Tolerant reading per the accord: unknown fields/types skipped ONLY; no malformed known shapes found in existing fixtures (a defect there would have been a cross-lane red, not a UI workaround).
- **Sibling component, not a host rewrite**: `OverlayPanel` (read-only, fixed-height, scrollable, never picks) mirrors the picker's host contract; the Surface overlay slot gained one `mod:` field with a picker default — exactly three dispatch sites changed. The refusal ladder, capacity math, and notices are reused verbatim.
- **Recompute at summon AND per footer repaint while open** — O(events), the same class as the existing per-advance re-projection; live update falls out free (advance -> paint), and dismissed-not-dead is trivial by construction. Documented with the tradeoff.
- Binds: labeled `w`/`m`/`n` (plan on `n` since Ctrl+P is the palette), riding the palette registry derivation.

## Verification

Red-first both phases (0/10 then 10 failing with wiring stashed). 64 new tests: 42 pure fold (incl. unknown-shape skip + hostile payloads), 16 surface byte-checks (SealOracle: sealed history byte-identical across open/scroll/refresh/close; degenerate refusal writes zero bytes; hostile ESC/newline titles never reach the wire raw; transcript-identity pins open/close never perturbs the block projection), 6 keymap. Harness suites 805 passed; broad suite 6257/6263 (only persistent failure = the known pre-existing metrics timing flake). Warnings + format clean.


DROOdotFOO (commented):

Scope: three hostile personas over the summonable projection panels
(worktracks/memory/plan). Reproductions were run in an isolated worktree
off the PR head (`38962dc4`) with `deps`/`_build` symlinked from the main
checkout; master was never switched. No seal / immutable-prefix violation
was found: the panels reuse `InlineAuthority.set_footer_rows/2` and the
fixed-height claim verbatim from the picker path, `overlay_mod/1` is now
threaded through every dispatch site (`route_passthrough`, `overlay_lines`,
`force_close_overlay?`), the `:not_composing` guard also suppresses the
`w`/`m`/`n` binds while any overlay is open, and the always-chords
(Ctrl+E / Ctrl+P / Tab) are all inert no-ops with an overlay open. Focus
routing and the seal path are sound. The findings below are content /
clamp / legibility issues, not prefix violations.

### High

**H1 (Security Auditor, reproduced) -- C1 control bytes smuggle a
clear-screen / OSC past the ViewText sanitize boundary the panels rely on.**
`PanelProjection` deliberately does NOT strip control bytes and delegates
all sanitization to `Raxol.Harness.Surface.ViewText` as "the one trust
boundary" (`panel_projection.ex:54-63`). But `ViewText.sanitize/1`
(`surface/view_text.ex:221-226`) strips only C0 (`0x00-0x1F`, incl. ESC)
and DEL; its allowlist is `byte >= 0x20 and byte != 0x7F`, which passes the
C1 range `0x80-0x9F`. Agent-authored meta content is JSON/UTF-8, so a C1
introducer arrives as its 2-byte UTF-8 form (e.g. U+009B CSI = `0xC2 0x9B`),
both bytes `>= 0x20`, and survives untouched to the `:styled` wire in
`overlay_lines`. This is exactly the known #616 C1 gap, now reached by a new
untrusted surface (memory/plan/worktracks projections). The PR's own hostile
fixture only exercises the ESC form (`projection-panels.jsonl:20`,
`"[2Jevil\ntitle"`), so the SealOracle "hostile ESC titles never reach
the wire raw" assurance is real for ESC but false for C1 -- giving false
confidence at precisely the boundary this feature leans on.

Reproduced (memory value `<<0xC2,0x9B>> <> "2J-c1-clear"` vs
`<<0x1B>> <> "[2J-esc-clear"`, folded -> `OverlayPanel` -> `ViewText.lines(_,80,:styled)`):
- ESC form: content-ESC stripped, does NOT reach the wire.
- C1 form: `<<107,95,99,49,58,32,194,155,50,74,...>>` reaches the wire raw --
  the C1 CSI is intact. U+009D (OSC) would likewise pass, opening an OSC-52
  clipboard-write / OSC-8 vector on terminals that honor C1.

Not introduced by this PR (the gap lives in ViewText), but this PR widens
exposure and explicitly certifies the boundary. Extend `sanitize/1` to strip
the C1 range (`0x80-0x9F` as decoded codepoints, not raw bytes -- careful not
to split UTF-8) and add a C1 case to the hostile fixture before merge.

### Medium

**M1 (Security Auditor / Saboteur, reproduced) -- `@max_field_bytes`
"512-byte clamp" degrades to a 512-*grapheme* clamp on a mid-codepoint
split, so it does not actually bound bytes.** `display_string/1`
(`panel_projection.ex:312-325`): when `binary_part(v, 0, 512)` lands
mid-codepoint (invalid UTF-8), the fallback is `String.slice(v, 0, 512)`,
whose third arg is a *grapheme count*, not a byte count. For any input of
<=512 graphemes the whole string is returned regardless of byte size; a ZWJ
emoji grapheme can be 20+ bytes, so 512 graphemes can be many KB. Combined
with `@max_entries 500`, the moduledoc's "bounds memory even against a
runaway or adversarial extract stream" is weaker than stated -- the per-field
byte bound is effectively unbounded on the fallback path.
Reproduced: `String.duplicate("→", 400)` (1200 bytes; 512 not divisible
by 3 -> `binary_part` invalid -> fallback) yields `String.slice(_,0,512)` =
400 graphemes = **1200 bytes returned unclamped**. Fix: slice by display
width or re-clamp the grapheme result back under `@max_field_bytes`.

**M2 (Saboteur) -- recompute-per-paint re-folds the entire durable journal
on every keystroke while a panel is open, including pure scroll keys.**
`refresh_panel_overlay/1` (`surface.ex:2084-2096`) is called from
`paint_footer/1`, which runs at the end of every `handle_input/2`. So a
single down-arrow (a viewport move that touches no projection state) re-folds
ALL `projection.source_events` via `PanelProjection.render_lines/2`. The
moduledoc frames this as "the same class as the existing per-advance
re-projection," but per-advance is amortized once per event, whereas
per-keystroke-while-open is user-driven repetition over a monotonically
growing journal -- genuinely super-linear in a long session under scrolling.
Consider decoupling scroll (offset-only, no re-fold) from content refresh
(advance-driven), or memoizing the fold against a source-events version.

### Low

**L1 (New Hire) -- the `overlay` state field named `picker` holds an
`OverlayPanel`, and `on_pick` is a dead field for panels.** The typedoc
(`surface.ex:570-573`) types `picker: OverlayPicker.t() | OverlayPanel.t()`
and the code apologizes inline ("despite the field name predating panels").
A newcomer reading `overlay.picker` when it is a panel, plus the
never-invoked `on_pick: fn m, _item -> m end` "shape-compat filler"
(`surface.ex:1806-1811`), has to reverse-engineer the discriminated union
from prose. Renaming to `state`/`content` (or a tagged `{:picker, _}` /
`{:panel, _}`) would make the summon/dismiss state machine self-describing.

**L2 (New Hire) -- magic `max_overlay_rows - 1` at the panel open site.**
`do_open_panel/4` (`surface.ex:1789-1791`) clamps `max_visible` to
`max_overlay_rows - 1`; the `- 1` reserves the title row and is only
derivable by knowing `OverlayPanel.height/1 = 1 + max_visible`. The picker
path has the identical unexplained `- 1` (`surface.ex:1678-1679`), so this is
a pre-existing idiom, but a one-line "(reserve the title row)" comment or a
named helper would save the next reader the round trip.

### Cross-persona overlaps

- H1 is co-owned by Saboteur (hostile-payload coverage gap: the fixture's
  single hostile line tests only the ESC form) and Security Auditor
  (control-byte injection at the sanitize boundary). Promoted to High on that
  basis -- two personas, same root (incomplete control-byte defense), a
  real terminal-injection vector.
- M1 is co-owned by Saboteur (adversarial memory bound) and Security Auditor
  (clamp integrity).

### Verdict: CONCERNS

No seal / immutable-prefix violation and no focus-routing bug -- the
`overlay_mod/1` threading, the `:not_composing`-plus-`overlay_open?` guard,
and the always-chord no-op clauses hold under adversarial keying, and the
SealOracle covers byte-identity of sealed history across open/scroll/refresh/
close. The blocking-adjacent issue is H1: the feature certifies ViewText as
its sole sanitize boundary while that boundary passes C1 controls (#616) and
the hostile fixture never exercises them. Recommend closing the C1 gap (or
explicitly scoping it out with a tracked follow-up) and adding a C1 fixture
case before merge; M1/M2 are worth addressing but need not block.

DROOdotFOO:
**Merge readiness: MERGE #632 FIRST, then ready.**

CI is green and there is no seal/immutable-prefix or focus-routing bug. The one HIGH is the shared C1 control-byte sanitizer gap (#616), reached here via `ViewText.sanitize` -- fixed at the source in #632. Once #632 lands, this PR's untrusted panel content routes through the fixed sanitizer and the HIGH is closed. The `@max_field_bytes` grapheme-vs-byte clamp (M1) is a LOW follow-up.


DROOdotFOO:
**Update: the C1-sanitizer HIGH is now CLOSED** -- #632 merged to master, so `ViewText.sanitize/1` now strips the C1 control range (U+0080..U+009F) code-point-aware. Once this branch is on current master (merge/rebase brings the fix in), the panel content routes through the fixed sanitizer and the injection HIGH is gone. Merge readiness moves to READY, modulo the LOW `@max_field_bytes` grapheme-vs-byte clamp follow-up.

merklebonsai:
## Reply ledger — adversarial review (verify-or-refute)

Thanks for the pass. #629 merged (`7081e2183`) ~1 min before your READY verdict, and H1 was closed by #632 landing in master first. The remaining items are hardened in a follow-up against master: **#636**.

| Finding | Disposition | Evidence |
|---|---|---|
| **H1** — C1 past `ViewText.sanitize` | **Closed by #632** (both `ViewText.sanitize/1` and `ContentGuard.sanitize_line/1` now strip U+0080..U+009F code-point-aware). The fixture/oracle gap you flagged (only ESC → false confidence) is closed: **U+009B CSI added to the hostile fixture** + an **end-to-end wire regression** asserting the C1 introducer never reaches the `:styled` wire. | red-first: C1 reaches the wire when *both* sanitize seams are reverted — the test pins the referent ("C1 never reaches the wire"), not one seam, so it survives which-seam refactors. |
| **M1** — `@max_field_bytes` grapheme clamp | **Fixed.** `display_string/1` fallback replaced: byte-slice + trailing-partial-codepoint trim → ≤512 **bytes**, valid UTF-8. | red-first: `"→"×400` (1200 B) fails `byte_size ≤ 512` on the old `String.slice/3` path; green after. |
| **M2** — recompute-per-paint re-folds on every keystroke | **Fixed.** `refresh_panel_overlay/1` memoizes against the append-only `source_events` length (`folded_at`); a pure scroll key skips the fold, a reveal that grows the journal re-folds. Directly kills the scroll re-fold you called out. | red-first: `KeyError :folded_at` without memo; scroll leaves token+lines unchanged, reveal advances token + surfaces "done". |
| **L1** — `picker` field name / dead `on_pick` | Typedoc now names `mod` as the discriminator and documents `folded_at`. Full rename deferred — disproportionate cross-dispatch-site risk on merged, seal-adjacent code (honest disclosure, not a silent skip). | — |
| **L2** — magic `- 1` | One-line comment (reserve the title row). | — |

No seal/immutable-prefix change: M2 is behavior-preserving (skips only provably-redundant recomputes; identical bytes), so `SealOracle` byte-identity holds.

Gates: `compile --warnings-as-errors` clean; format clean; panel/projection/overlay/keymap green. Full `test/harness/` = 712/715 — the 3 failures are pre-existing `AnsiReplayer` cross-terminal replay **timeouts** under concurrent load (each passes in isolation; not this diff).

PR #631, opened 2026-07-17
feat(harness): U21 evidence tri-state wire marker (#619 residual)

merklebonsai:
## The gap (#619 residual)

`Contract.gated_done_payload/4` (`packages/raxol_agent/lib/raxol/agent/contract.ex`) builds the `turn_completed{final: true}` payload for every closed turn. Before this PR:

- gate **accepts** -> `%{usage, final: true, refs: [...]}`
- gate returns `:evidence_required` (**nothing ever offered**) -> `%{usage, final: true}` + ephemeral telemetry `...:ungated_done`
- gate **rejects** offered evidence (`{:error, reason}`) -> `%{usage, final: true}` + ephemeral telemetry `...:rejected_evidence`

The last two are **byte-identical on the wire and in the journal** — telemetry is not journaled, so a replayed/attached surface cannot distinguish "offered-but-rejected" from "never offered".

## The marker

Additive, grow-only growth on the existing `turn_completed{final: true}` payload:

- `evidence: :accepted | :rejected | :absent` — grow-only enum (atom in-memory, JSON string on the wire, same convention as every other enum on this surface).
- `evidence_rejected` (present only when `evidence: :rejected`) — `%{"refs" => offered_refs, "reason" => reason_name, "ref" => offending_offset}`, the `DoneGate` verdict tuple flattened **explicitly** into these string values — never handed to `sanitize_payload/1` to `inspect/1`-stringify a raw tuple onto the wire. `reason_name ∈ "missing_ref" | "not_evidence" | "foreign_turn" | "stale_evidence" | "mutation_echo"`. `DoneGate`'s `:unturned_done` reject is unreachable on this path (`pump/3` always mints a non-nil `turn_id`) and deliberately has no clause — it stays out of the wire enum.

`refs` and both existing telemetry signals (`[:raxol, :agent, :done_gate, :ungated_done]` / `:rejected_evidence`) are **untouched** — this marker is the durable/replay view, telemetry remains the live-ops view.

## Grandfather rule (back-compat)

Adds `Contract.evidence_status/1`, a decode helper the UI lane's T19 renderer (or anyone reading historical journals) can use instead of reimplementing this:

- `evidence` key present -> authoritative three-state, wins outright.
- key **absent**, `refs` present -> `:accepted` (the `refs` carrier never lied about acceptance).
- key **absent**, `refs` **absent** -> genuinely `:unknown` — **never** defaults to `:absent`. Before this marker, a rejected done and a never-offered done were wire-identical, so a legacy record truly cannot be told apart after the fact; guessing `:absent` would launder a historical rejection into a false "never offered" claim.

## schema_version bump

Journal `schema_version` default bumps `1.0.0 -> 1.1.0` (`Raxol.Agent.Journal.FileStore.Writer`) — a sanctioned additive minor bump per the freeze contract's AD-11 upcast-on-read rule. The pinned `v1.0.0` golden corpus stays literal `"1.0.0"` and unrewritten (old journal records are never rewritten in place, only decoded — covered by the grandfather rule above). Three sites that hardcoded the previous default (`journal_test.exs`, the I2 byte-identity reconstruction in `identity_invariants_test.exs`, and the I9 `contract_snapshot.json` floor descriptor) are updated to track it.

## Tests

RED-FIRST: `contract_test.exs` gained a `describe` block pinning that a rejected done and an absent-evidence done were indistinguishable pre-marker, that the DoneGate reject reason wasn't observable on the wire, and the grandfather decode rule (including the "never guess `:absent`" case) — confirmed failing against pre-marker code, all green against this change.

```
SKIP_TERMBOX2_TESTS=true TMPDIR=/tmp MIX_ENV=test mix test
Result: 1447 passed (13 properties, 1434 tests), 35 excluded
```

## Scope

Additive/grow-only — no freeze re-litigation. `DoneGate` itself is untouched; this is purely the producer stamping what it already knows before dropping it on the floor. UI/T19 is the consuming reviewer of this marker (their renderer should wire "no evidence" / "evidence rejected: <reason>" states to the `evidence` enum / `evidence_status/1`, not to refs-key-absence).

**Do not merge** — review-only.

### DIFF
--- a/packages/raxol_agent/lib/raxol/agent/contract.ex
+++ b/packages/raxol_agent/lib/raxol/agent/contract.ex
@@ -27,6 +27,54 @@ defmodule Raxol.Agent.Contract do
   durable journal sink attach behind this same boundary in later steps —
   producers change, the contract does not.
 
+  ## U21 evidence tri-state wire marker (#619 residual)
+
+  On a `turn_completed{final: true}` produced by `gated_done_payload/4`, the
+  payload grows one optional discriminator plus one optional detail —
+  additive, grow-only, mirrors the Q1 `context`-field growth pattern:
+
+    * `evidence: :accepted | :rejected | :absent` — grow-only enum (atom
+      in-memory, JSON string on the wire, same convention as every other
+      enum on this surface).
+      - `:accepted` — stamped alongside `refs` (untouched: still the sole
+        accepted-refs carrier, no rename, no move).
+      - `:absent` — the `{:error, :evidence_required}` gate arm: the turn
+        offered no refs at all.
+      - `:rejected` — the `{:error, reason}` gate arm: refs were offered and
+        the gate refused them.
+    * `evidence_rejected` (present only when `evidence: :rejected`) —
+      `%{"refs" => offered_refs, "reason" => reason_name, "ref" =>
+      offset_or_nil}`, the `Raxol.Agent.DoneGate` verdict tuple flattened
+      EXPLICITLY into these string values — never handed to
+      `sanitize_payload/1` to `inspect/1`-stringify a raw tuple onto the
+      wire. `reason_name` is one of `"missing_ref"`, `"not_evidence"`,
+      `"foreign_turn"`, `"stale_evidence"`, `"mutation_echo"`.
+      `DoneGate`'s `:unturned_done` reject is unreachable on this path
+      (`pump/3` always mints a non-nil `turn_id`) and deliberately has no
+      clause here — it stays out of the wire enum.
+
+  Before this marker, a rejected done and a never-offered done were
+  byte-identical on the wire (`%{usage, final: true}`, no `refs`) — the two
+  gate-telemetry signals distinguished them live, but telemetry is not
+  journaled, so a replayed surface could not tell "offered but refused" from
+  "never offered". `evidence_status/1` decodes the field with the
+  grandfather rule this gap requires: an `evidence` key present is
+  authoritative; a **legacy** record (key absent) with `refs` present
+  grandfathers to `:accepted` (the `refs` carrier never lied); a legacy
+  record with neither key present is genuinely `:unknown` — it must NEVER
+  be guessed as `:absent`, which would launder a historical rejection into
+  a false "never offered" claim.
+
+  Both `[:raxol, :agent, :done_gate, :ungated_done]` and `[:raxol, :agent,
+  :done_gate, :rejected_evidence]` telemetry stay exactly as they were —
+  this marker is the durable/replay view; telemetry remains the live-ops
+  view. `refs` is untouched. The journal `schema_version` default bumps
+  1.0.0 -> 1.1.0 (`Raxol.Agent.Journal.FileStore.Writer`, additive per AD-11
+  upcast-on-read) alongside this payload growth; the pinned `v1.0.0` golden
+  corpus stays literal "1.0.0" and unrewritten (an old journal's records are
+  never rewritten in place, only decoded — covered by the grandfather rule
+  above).
+
   ## Producers
 
   The v0 producer is `pump/3`: it drains a `Raxol.Agent.Stream.run/2` or
@@ -271,7 +319,9 @@ defmodule Raxol.Agent.Contract do
 
     case DoneGate.gate(journal, turn_id, refs) do
       {:ok, _done} ->
-        Map.put(base, :refs, refs)
+        base
+        |> Map.put(:refs, refs)
+        |> Map.put(:evidence, :accepted)
 
       {:error, :evidence_required} ->
         :telemetry.execute(
@@ -280,7 +330,7 @@ defmodule Raxol.Agent.Contract do
           %{turn_id: turn_id}
         )
 
-        base
+        Map.put(base, :evidence, :absent)
 
       {:error, reason} ->
         :telemetry.execute(
@@ -290,6 +340,67 @@ defmodule Raxol.Agent.Contract do
         )
 
         base
+        |> Map.put(:evidence, :rejected)
+        |> Map.put(:evidence_rejected, evidence_rejected_detail(refs, reason))
+    end
+  end
+
+  # Flatten the DoneGate verdict tuple onto the wire EXPLICITLY — this must
+  # never be handed to `sanitize_payload/1` to `inspect/1`-stringify (that
+  # would render `"{:mutation_echo, 3}"` instead of a JSON-shaped detail).
+  # Every `DoneGate.gate/3` rejection reaching this arm (i.e. everything but
+  # `:evidence_required`, matched separately above) is a `{reason, offset}`
+  # pair per `DoneGate.verdict/0`. `:unturned_done` is the one bare-atom
+  # reject and is unreachable here — `pump/3` always mints a non-nil
+  # `turn_id` — so it deliberately has no clause and stays out of the wire
+  # enum; a value there would be a real invariant violation worth crashing
+  # loudly on rather than silently coercing.
+  @spec evidence_rejected_detail([DoneGate.offset()], {atom(), DoneGate.offset()}) ::
+          %{String.t() => term()}
+  defp evidence_rejected_detail(refs, {reason, offset}) when is_atom(reason) do
+    %{"refs" => refs, "reason" => Atom.to_string(reason), "ref" => offset}
+  end
+
+  @doc """
+  Decode a `turn_completed{final: true}` payload's evidence status.
+
+  Tolerant of both atom- and string-keyed payloads and both atom- and
+  string-valued enums (live in-memory events vs. journal-replayed JSON —
+  same convention as `Raxol.Agent.DoneGate`'s accessors).
+
+  Implements the grandfather rule for records written before this marker
+  existed (see moduledoc): an `evidence` key present is authoritative and
+  wins outright. A legacy record (key absent) with `refs` present
+  grandfathers to `:accepted` — the `refs` carrier never lied about
+  acceptance. A legacy record with **neither** key present is genuinely
+  `:unknown`: before this marker, a rejected done and a never-offered done
+  were wire-identical, so there is no way to tell them apart after the
+  fact. This case must NEVER decode as `:absent` — that would launder a
+  historical rejection into a false "never offered" claim.
+  """
+  @spec evidence_status(map()) :: :accepted | :rejected | :absent | :unknown
+  def evidence_status(payload) when is_map(payload) do
+    case fetch_either(payload, :evidence) do
+      {:ok, value} ->
+        normalize_evidence(value)
+
+      :error ->
+        case fetch_either(payload, :refs) do
+          {:ok, _refs} -> :accepted
+          :error -> :unknown
+        end
+    end
+  end
+
+  defp normalize_evidence(value) when value in [:accepted, "accepted"], do: :accepted
+  defp normalize_evidence(value) when value in [:rejected, "rejected"], do: :rejected
+  defp normalize_evidence(value) when value in [:absent, "absent"], do: :absent
+
+  defp fetch_either(map, key) when is_map(map) do
+    cond do
+      Map.has_key?(map, key) -> {:ok, Map.get(map, key)}
+      Map.has_key?(map, Atom.to_string(key)) -> {:ok, Map.get(map, Atom.to_string(key))}
+      true -> :error
     end
   end
 
--- a/packages/raxol_agent/lib/raxol/agent/contract.ex
+++ b/packages/raxol_agent/lib/raxol/agent/contract.ex
@@ -75,6 +75,48 @@ defmodule Raxol.Agent.Contract do
   never rewritten in place, only decoded — covered by the grandfather rule
   above).
 
+  ## Trust boundary (S2, PR #631) — producer-stamped, journal-trusted
+
+  `evidence_status/1` is a **decode**, not a **re-derivation**. It reports the
+  enum the producer stamped; it does NOT re-run `DoneGate.gate/3` over the
+  journal at read time. The soundness of the marker therefore rests on two
+  explicit assumptions, stated here so a reader wiring it to a surface knows
+  exactly what it does and does not defend against:
+
+    1. **Producer honesty is stamped, not asserted.** The marker is written by
+       `gated_done_payload/4` from the *same* `DoneGate.gate/3` verdict that
+       decides the done — producer and decider are one call, they cannot
+       disagree at write time. This producer half is pinned by
+       `Raxol.Agent.ContractTest` ("the done gate is consulted on the real done
+       path", "a mutation's own result echo is rejected", "a zero-tool turn
+       closes ungated") and by `Raxol.Agent.Red.U21RealProducerRegressionTest`,
+       which gates journals produced by the real `pump/3`, not synthetic ones.
+
+    2. **The reader trusts the journal.** A hand-forged or replayed line that
+       stamps `evidence: :accepted` — or a legacy-shaped forged line that omits
+       the marker but carries a `refs` list, which the grandfather arm blesses
+       `:accepted` — will decode as proven-done. This is the **identical**
+       threat model as `refs` themselves and every other durable field: a
+       journal that can be forged can fabricate tool results, refs, and verdicts
+       directly. The marker adds no trust surface the journal did not already
+       assume (append-only, single-writer integrity per harness-spec-backend).
+       It is sound only on a trusted, append-only journal.
+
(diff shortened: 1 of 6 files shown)


merklebonsai:
CI disclosure: **macos-latest is the only red check; it fails on pre-existing UI-lane flakes unrelated to this diff.** Every correctness gate is green on Linux — Compilation, Dialyzer, Credo, Integration, and the harness/journal/provenance suites — which is the real signal that the additive `evidence` marker + `schema_version` 1.0.0→1.1.0 bump is sound and did not disturb the loop/meta/checkpoint journal families.

The macos-latest failures are: `markdown_body_test:516` and `stream_cadence_test:94` (wall-clock perf/`assert_receive` timing on the slower runner), `overflow_test:11` (a corrupted `undefined variable` test), and a couple of UI-theming/harness tests that pass on Linux. None touch `contract.ex`, the evidence payload, or `schema_version`. The `tf_fixture` upcast/provenance test passes on Linux, so it's runner-environment, not the version bump.

Flagging these for the shared de-flake sweep rather than papering over them here (they're UI-lane wall-clock tests). Happy to re-run macos-latest if you want to confirm it's transient.

DROOdotFOO (commented):

Scope reviewed: `packages/raxol_agent/lib/raxol/agent/contract.ex` (marker producer + `evidence_status/1` decoder), `journal/file_store/writer.ex` (schema bump), goldens/tests. Line numbers are from PR head `9d90835`.

First, the part that holds up (credit where due): the tri-state marker does NOT introduce a third "clean-lumped" state. Per `DoneGate.verdict/0`, every tuple reject (`missing_ref`, `not_evidence`, `foreign_turn`, `stale_evidence`, `mutation_echo`) lands in the `{:error, reason}` arm -> `evidence: :rejected` (contract.ex:343), and the no-refs arm -> `:absent` (contract.ex:333). The two #619 residual classes (`stale_evidence`, `mutation_echo`) are now stamped `:rejected` at produce time, not rendered clean. The marker is stamped from the same `DoneGate.gate/3` verdict that decides (contract.ex:322), so producer and gate cannot disagree at write time. Fail-safe-when-absent is correct: a legacy record with no marker and no refs decodes `:unknown`, never `:absent` (contract.ex:194-196, `refute ... == :absent` test). Good.

### High

_(none)_

### Medium

**S1 (Saboteur, reproduced) -- `normalize_evidence/1` is non-total and crashes on any unexpected `evidence` value, contradicting this PR's own grow-only-enum / AD-11 upcast-on-read contract.** contract.ex:395-397 has exactly three clauses (`accepted`/`rejected`/`absent`) and no catch-all. The moduledoc sells `evidence` as a "grow-only enum" bumping `schema_version 1.0.0 -> 1.1.0` "additive per AD-11 upcast-on-read", and `evidence_status/1` is documented as the grandfather/tolerant decoder. But a reader on 1.1.0 that meets a *future* 1.2.0 journal carrying a 4th evidence value -- exactly the grow-only case the enum promises -- raises `FunctionClauseError` instead of degrading to `:unknown`. Same crash for a corrupt/forged line (`evidence: "bogus"`). This is a replay-surface crash and it breaks the forward-tolerance the whole grandfather rule exists to provide. (reproduced) -- driving the exact clause set from the diff:
```
Repro.evidence_status(%{evidence: :accepted})            #=> :accepted
Repro.evidence_status(%{"evidence" => "verified_v2"})    #=> ** (FunctionClauseError)
```
Fix: add `defp normalize_evidence(_), do: :unknown` (fail-safe) so an unrecognized enum value degrades instead of crashing. The tri-state producer is total; the decoder is not.

**S2 (Security Auditor) -- the honesty property is trust-on-write, never re-derived; a crafted/replayed journal can still render a failed turn as proven-done.** `evidence_status/1` (contract.ex:382-405) returns the stamped enum authoritatively and never re-runs `DoneGate.gate/3` over the journal, even though (per the #619 framing and `DoneGate`'s own "pure read/decision function" moduledoc) the predicates are pure over the journal and re-derivable at render time. So a hand-written/replayed line stamping `evidence: :accepted` -- or, easier, one that just omits the marker and carries any `refs` list, which the grandfather arm blesses `:accepted` at contract.ex:194 -- renders proven-done with zero validation that those refs actually satisfy the gate. The marker moves trust from render-time derivation to producer stamp + journal integrity, but no trust boundary is documented: there is no note that this is only sound on an append-only / trusted journal, and no re-derivation backstop. Relative to #619 this is strictly better for honest producers but weaker than a render-time re-check against a hostile journal. Recommend either documenting the append-only trust assumption explicitly, or having `evidence_status/1` optionally re-derive from the journal when one is available.

### Low

**N1 (New Hire) -- "tri-state" is a misnomer at the decode surface.** `evidence_status/1`'s spec is `:accepted | :rejected | :absent | :unknown` (contract.ex:381) -- four outcomes. The wire enum is tri-state; the decoder is quaternary. A reader wiring the T19 renderer to a "tri-state marker" will plausibly write a 3-arm `case` and either miss legacy `:unknown` or (if their own `case` is non-exhaustive) crash on it. The prose does explain `:unknown`, but the headline name fights the type. Worth a one-line "(decode adds a fourth `:unknown` state for legacy records)" at the callout.

**N2 (New Hire) -- the "renderer consuming it" does not exist in this diff and the authority split is prose-only / untested end-to-end.** grep across `packages/` finds zero non-test callers of `evidence_status/1` or `evidence_rejected`; the consumer is the future UI/T19 lane. Fine for a review-only PR, but the producer-vs-renderer authority split is asserted in the moduledoc without an integration test driving an actual renderer through `evidence_status/1` -- the decode is only unit-tested against literals. Note it so T19 owns the wiring assertion.

**N3 (Saboteur, note) -- `:unturned_done` unreachability is load-bearing and only guarded by `pump/3`.** `evidence_rejected_detail/2` (contract.ex:360) matches only `{reason, offset}` 2-tuples; the bare-atom `:unturned_done` reject would `FunctionClauseError` if it ever reached the `{:error, reason}` arm. Verified unreachable today because `pump/3` mints a non-nil `turn_id` (contract.ex:171) and `gated_done_payload/4` is private with `pump/3` as its only caller. The deliberate fail-loud is defensible, but it is one refactor (a second caller passing a nil turn) away from a crash; the guard lives in a different function than the assumption. Acceptable as-is; flagging the coupling.

### Cross-persona overlaps

S1 (Saboteur) and S2 (Security Auditor) share one root: `evidence_status/1` / `normalize_evidence/1` treat the stamped wire value as authoritative input without total, defensive handling. S1 is the crash face (unknown value -> raise), S2 is the trust face (forged value -> believed). A single hardening pass -- a total decoder with a `:unknown` fallthrough plus an explicit append-only trust note (or opt-in journal re-derivation) -- closes both. Promoted as the primary concern.

### Verdict: CONCERNS

The marker meets its stated goal -- all `DoneGate` reject classes (including the #619 residual `stale_evidence` + `mutation_echo`) map to `:rejected`, none are lumped with `:accepted`, and absence fails safe to `:unknown` rather than `:absent`. But the decoder is non-total and crashes on unknown/future/forged enum values (S1, reproduced), directly undercutting the grow-only + upcast-on-read contract the PR leans on, and the honesty guarantee is trust-on-write with no re-derivation or documented trust boundary (S2). Neither blocks the intent; both should land before T19 wires a renderer to this surface.

DROOdotFOO:

CI is red (1 failing check) and the review found the likely cause: the decoder `normalize_evidence/1` is non-total -- it raises `FunctionClauseError` on any unknown/future/forged `evidence` value, directly contradicting this PR's grow-only + upcast-on-read contract. Separately, the honesty property is trust-on-write (`evidence_status/1` never re-derives from the journal, so a forged line stamping `:accepted` renders proven-done). Make the decoder total (catch-all -> `:unknown`) and add an append-only trust note before merge.


DROOdotFOO:
fix decoder totality before merge needed

merklebonsai:
## Fix round — reply ledger

Both findings closed in `378de47` (surgical, two files: `contract.ex` + `contract_test.exs`; no dependency or schema change). Thanks for the reproduction — it landed as the red test verbatim.

### S1 (Medium, Saboteur/Security shared root) — non-total decoder → **FIXED**

**Disposition:** total decoder. `normalize_evidence/1` gains a catch-all `defp normalize_evidence(_), do: :unknown`. Every unrecognized `evidence` value — a future 1.2.0 grow-only enum value, your forged `"verified_v2"`/`"bogus"`, a garbage type (number, nil, list, map, tuple) — now degrades to `:unknown`. Fail-safe in the same direction the grandfather rule already takes absence: never crash, never `:accepted`.

One extra guarantee beyond the crash fix: because the `evidence` key being *present* means it is not a legacy record, a forged non-legacy line whose bogus marker sits atop a `refs` list decodes `:unknown` (the refs-grandfather arm is never consulted), **not** laundered to `:accepted`. That closes the easy S2 forge face at the decoder.

**Evidence (red-first):** the four new tests were run with the catch-all temporarily removed and all four failed with `FunctionClauseError` — your exact repro among them (`evidence_status(%{"evidence" => "verified_v2"})`). Restored → 20/20 green.

- `contract_test.exs`: "a future/unknown string enum value decodes to :unknown (Drew's exact repro)"; "a future/unknown atom enum value"; "a garbage-typed evidence value" (7 types × atom+string keys); "an unrecognized authoritative marker never launders to :accepted, even with refs present".

### S2 (Medium, Security Auditor) — trust-on-write, no documented boundary → **CLOSED via documentation (Route A)**

**Disposition:** documented the trust boundary explicitly; **rejected** read-side re-derivation. New `## Trust boundary (S2)` section in the moduledoc + a pointer at `evidence_status/1`.

The boundary, stated: (1) the marker is producer-stamped by `gated_done_payload/4` from the *same* `DoneGate.gate/3` verdict that decides the done — one call, they cannot disagree at write time; this producer half is already guarded by `ContractTest` ("the done gate is consulted on the real done path", the echo-rejected and zero-tool-ungated tests) and by `U21RealProducerRegressionTest`, which gates journals from the real `pump/3`. (2) The reader trusts the journal: a forged/replayed line stamping `evidence: :accepted` (or a legacy-shaped forged line omitting the marker but carrying refs) decodes proven-done — the **identical** threat model as `refs` and every durable field; a forgeable journal can fabricate tool results and verdicts directly. The marker adds no trust surface the journal did not already assume (append-only, single-writer).

**Why not read-side re-derivation:** (a) it would couple a pure single-record decode to the whole journal + turn + refs it does not carry; (b) it would resurrect exactly the display-side re-evaluation of *open* gate predicates (staleness relative to *later* mutations) that the #619 round-2 ledger settled must NOT be frozen or recomputed at the display surface. Documenting the append-only assumption is the honest, #619-consistent close; a hostile-journal re-check is out of scope for a producer-stamped durable marker. The S1 fix already hardens the one part that *is* the decoder's job (unrecognized value ≠ accepted).

### N1/N2/N3 (Low) — acknowledged, no change

- **N1** (quaternary decode vs "tri-state" name): addressed in prose — `evidence_status/1`'s docstring now spells out "the decode is quaternary … `:unknown` is the decode-only bucket … a consumer's `case` must handle four outcomes." Kept the wire enum tri-state (it is).
- **N2** (no end-to-end renderer consumer): correct, T19 owns the wiring assertion; out of scope for this review-only surface.
- **N3** (`:unturned_done` unreachability coupled to `pump/3`): flagged coupling acknowledged; unchanged (it is a deliberate fail-loud and unreachable today).

### Gates
`MIX_ENV=test mix compile --warnings-as-errors` (raxol_agent) clean; `contract_test.exs` 20/20; U21 red suite + identity invariants + journal 52/52; **full raxol_agent suite 1451 passed, 35 excluded**; `mix format --check-formatted` clean. mix.lock left untouched (no dependency change).

Commit: `378de47` on `feat/u21-evidence-tristate`.

DROOdotFOO (commented):

Since last review: 9d90835f..378de474

The head rebased onto a decoder-totality fix commit (378de474, "fix(harness):
total evidence decoder + trust-boundary note"). I re-ran both prior MED findings
against the current head and reproduced the S1 fix end-to-end with the real
`Raxol.Agent.Contract.evidence_status/1`.

**S1 (was MED, reproduced) — decoder is now genuinely TOTAL, fail-safe `:unknown`. RESOLVED (reproduced).**
At the prior head `9d90835f`, `normalize_evidence/1` had exactly three clauses
(`:accepted` / `:rejected` / `:absent`) and **no** catch-all, so any unknown /
future / forged / non-atom-non-string `evidence` value raised
`FunctionClauseError` on the replay surface. Confirmed by `git show
9d90835f:...contract.ex` — three clauses, no fallthrough. The fix adds
`defp normalize_evidence(_), do: :unknown` (`contract.ex:464`) and maps the
missing-key path through the grandfather rule. I exercised the real decoder on
the PR head across the full type lattice:

```
unknown string ("verified_v2")   => :unknown
unknown atom (:verified_v2)       => :unknown
integer 42 / float 3.14 / nil     => :unknown
map / list / tuple / bool         => :unknown
bogus + refs (launder attempt)    => :unknown   <-- key present, refs-arm not consulted
legacy refs grandfather           => :accepted
legacy no-refs                    => :unknown
:accepted                         => :accepted
```

Every unrecognized value degrades to `:unknown` — never crashes, never
`:accepted`. Red-first coverage is present and real: `contract_test.exs` carries
"a future/unknown string enum value decodes to :unknown (Drew's exact repro)",
"a future/unknown atom enum value", and a garbage-type loop over
`[42, nil, true, %{}, [1,2,3], {:tuple,:val}, 3.14]` for both atom- and
string-keyed payloads, plus "an unrecognized authoritative marker never launders
to :accepted, even with refs present". The full contract suite runs green
(20 passed) on the PR head.

**S2 (was MED) — trust-on-write is now DOCUMENTED (not re-derived), and hardened at the edges. RESOLVED as documented.**
`evidence_status/1` remains a decode, not a re-derivation — it still does not
re-run `DoneGate.gate/3` over the journal. That design is now stated explicitly
in a new "Trust boundary (S2, PR #631) — producer-stamped, journal-trusted"
moduledoc section (`contract.ex:78-125`): producer honesty is stamped (producer
and decider are one `DoneGate.gate/3` call), the reader trusts the append-only
journal, and read-side re-derivation was "considered and deliberately rejected"
with a #619-consistent rationale (re-checking open gate predicates at the
display surface is exactly what #619 round-2 settled must not happen). This is
the documented append-only trust boundary my prior review asked for. It is the
honest close for a producer-stamped durable marker.

### Low / residual (non-blocking, all documented in-tree)

- **Saboteur (reproduced):** the legacy grandfather arm is the one remaining
  write-trust path that does not require the exact `:accepted` value — a forged
  *legacy-shaped* line that omits the marker but carries any `refs` list decodes
  `:accepted` (`%{final: true, refs: [3]} => :accepted`, reproduced). This is
  strictly weaker than the S1 fix (which blocks bogus *marker-value*
  laundering). It is explicitly called out in trust-boundary point 2 as the
  identical threat model as `refs` themselves and every other durable field, and
  is sound only on a trusted append-only journal. Acceptable given that scope;
  noting it so a surface author knows the grandfather arm is a trust-on-write
  path, not a re-validation.

- **New Hire:** the decode is quaternary (`:accepted | :rejected | :absent |
  :unknown`) while the producer only ever stamps tri-state. The `@doc` and
  `@spec` (`contract.ex:426,435`) both state this and warn "a consumer's `case`
  over this must handle four outcomes" — good. Follow-up seam (out of scope for
  this PR): the totality is proven at the decoder, but the T19 renderer that
  consumes `evidence_status/1` is not in this changeset, so nothing yet pins
  that the surface actually handles `:unknown` without a fall-through crash of
  its own. Worth a consumer-side test when that lane lands.

- **Security Auditor:** the moduledoc claim "the marker adds no trust surface
  the journal did not already assume" is now *accurate* — but note it only
  became accurate with this fix. Pre-fix, the non-total decoder was itself a
  replay-surface liability: a forged/corrupt `evidence` value on a replayed line
  would `FunctionClauseError` the decoder, i.e. a crash-on-replay DoS reachable
  from journal contents. The S1 catch-all removes that; the totality fix is also
  a defensive-security fix, not only an ergonomics one. No new attack surface
  introduced. No legitimate producer value regressed to `:unknown` (the three
  stamped enum values still map correctly; verified).

### Verdict: CLEAN

Both prior MED findings are closed. S1 is genuinely total and fail-safe
(`:unknown`, never `:accepted`, never a crash), reproduced against the real
decoder with a matching red-first repro test. S2 is addressed by an explicit,
sound append-only trust-boundary note plus a deliberate, #619-consistent
rejection of read-side re-derivation; the S1 fix additionally hardens the
non-legacy laundering path (bogus marker value + refs -> `:unknown`, not
`:accepted`). The one residual (legacy refs-grandfather blesses `:accepted`) is
documented and inherent to the append-only trust model, not a regression. No new
bug found.

PR #634, opened 2026-07-17
test(core): converge boundary reject vectors on relative-symlink-under-symlinked-ancestor

merklebonsai:
Follow-up from the ACP #624 / #633 review (DROOdotFOO requested convergence).

The FsSandbox symlink-escape HIGH on #624 slipped past its own drift guard because the shared boundary reject vectors had **no case** for the class the bug lived in: a *relative* symlink target (`root/esc → "../secret.txt"`) reached through a self-referential symlinked ancestor (`root/loop → "."`), requested as `loop/loop/esc`. `FsSandbox` had silently kept the pre-`walk_real` lexical-parent resolver and nothing failed.

Core's `Raxol.Core.Boundary.Path` already resolves this correctly (`walk_real`, from #613) — this PR just adds the vector so the Core reject set stays converged with the package copy that #624 now also gates on. Pure drift-guard hardening on already-correct code.

- Vector: `relative_symlink_under_symlinked_ancestor_escape` → `:symlink_escape`.
- `packages/raxol_core` boundary suite green (77 passed) with the vector.
- Matching vector already added to `raxol_agent_client_protocol`'s copy on #624.

### DIFF
--- a/packages/raxol_core/test/support/boundary_vectors/path_reject_vectors.json
+++ b/packages/raxol_core/test/support/boundary_vectors/path_reject_vectors.json
@@ -5,14 +5,22 @@
       "name": "parent_traversal",
       "root": "root",
       "requested": "../../etc/passwd",
-      "setup": [{ "dir": "root" }],
+      "setup": [
+        {
+          "dir": "root"
+        }
+      ],
       "expect": "path_traversal"
     },
     {
       "name": "interior_traversal_escapes",
       "root": "root",
       "requested": "a/../../x",
-      "setup": [{ "dir": "root" }],
+      "setup": [
+        {
+          "dir": "root"
+        }
+      ],
       "expect": "path_traversal"
     },
     {
@@ -20,7 +28,11 @@
       "root": "root",
       "requested": "blobs/../../etc/x",
       "ref_format": "^(blobs|snapshots)/[0-9a-f]{64}(\\.json)?$",
-      "setup": [{ "dir": "root" }],
+      "setup": [
+        {
+          "dir": "root"
+        }
+      ],
       "expect": "malformed_ref"
     },
     {
@@ -28,7 +40,11 @@
       "root": "root",
       "requested": "snapshots/../blobs/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
       "ref_format": "^(blobs|snapshots)/[0-9a-f]{64}(\\.json)?$",
-      "setup": [{ "dir": "root" }],
+      "setup": [
+        {
+          "dir": "root"
+        }
+      ],
       "expect": "malformed_ref"
     },
     {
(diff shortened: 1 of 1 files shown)


DROOdotFOO (commented):

Test-only PR (1 file, +1 reject vector). Read the core copy on `fix/core-boundary-relsymlink-vector`, the ACP package copy on `feat/agent-client-protocol` (#624), the `boundary_vectors.ex` runner, and the `confine/3` implementation on master.

### Convergence (the crux) -- CLEAN
The file's `description` demands "change this file and its sibling across BOTH packages, or neither." Convergence is achieved. The two reject-vector copies are byte-for-byte identical after #634:
- `packages/raxol_core/.../path_reject_vectors.json` (branch) and `packages/raxol_agent_client_protocol/.../path_reject_vectors.json` (#624) contain the SAME 10 vectors, same order, same content (verified with `diff`: IDENTICAL).
- On `master`, the core copy has only 9 vectors -- it is MISSING `relative_symlink_under_symlinked_ancestor_escape`. The ACP package copy on #624 already carries it. So the ACP package was AHEAD and core was BEHIND; #634 brings core UP to match. The package's now-fixed FsSandbox already has a gating vector for this class, and after #634 the two sets no longer diverge (no silent fork).
- The accept-vector siblings are also identical across both packages (`diff`: IDENTICAL) and untouched -- lockstep preserved.

Note for reviewers: #634 edits only the core copy, but that is correct here -- the end-state is identical sets, which is what the rule protects. This is not a one-sided divergence.

### Vector correctness -- CLEAN (reproduced)
`relative_symlink_under_symlinked_ancestor_escape`: `root/loop -> "."` (self-referential ancestor), `root/esc -> "../secret.txt"` (relative escape leaf), `secret.txt` written at the tmp base OUTSIDE root, `requested: "loop/loop/esc"`, `expect: "symlink_escape"`. The `setup` entries carry explicit `target` fields, and the runner (`boundary_vectors.ex materialize/2`) creates dirs/files first, symlinks last via `File.ln_s!(target, abs)` -- so the links are materialized with the intended targets.

Non-vacuous / genuine regression net: the vector exploits the divergence between the lexical parent of the requested string (`loop/loop`) and the symlink's REAL directory (`root`). `path.ex:118` documents that a relative link target is now "followed against `base` -- its REAL parent". A pre-#613/#633 lexical resolver would resolve `../secret.txt` against `loop/loop` -> `loop/secret.txt` (in-root, `{:ok}`) and MISS the escape; the fixed real-parent walk resolves it against `root` -> `base/secret.txt` (outside) -> reject. Ran the FIXED `Raxol.Core.Boundary.Path.confine(root, "loop/loop/esc", [])` on master against the materialized fixture: returns `{:error, :symlink_escape}` (reproduced), matching `expect`. It encodes a real symlink ESCAPE, not a mere lexical `path_traversal`.

### Personas
- Saboteur: not vacuous; `setup` targets are present; accept sibling needs no paired change for a reject-only regression and remains identical across packages.
- New Hire: lockstep rule satisfied by identical end-state; no JSON reformatting noise -- the diff is a single appended vector object.
- Security: the vector requires the outside `secret.txt` to be neither read nor created (per the file's site-test contract); it is a true escape and a verbatim copy to the package is safe (already present and identical).

### Verdict: CLEAN

PR #630, opened 2026-07-17
Harness UI: full logical seal — refused-write display gap (LIVE), sealed? boundary unification (LIVE), tail-bound pin

merklebonsai:
The full-logical-seal unit of the SHG chain (full-logical-seal → commit-cap → T6). Per the reshaped DAG this unit is "the vehicle that carries G1–G4" — so Phase 0 was a verify-first audit of master (2416b13b8, post-#620) against the SHG spec, and Phase 1 built only the real gaps that audit found.

## Phase 0 — gap analysis (audited against code, not the ledger)

**G1 (shared frontier classifier): LIVE on master, with two structural dispositions.**
`Raxol.Harness.SealFrontier` (`classify/3`, `committable?/3`, `scan_frontier/3`, `commit_walk/5`) is the one classifier. Consumers verified by grep — no module outside `Surface` carries frontier logic:
- **Sealer**: `Surface.paint_pending_blocks/1` via `commit_walk/5` (the one mutating walk). LIVE.
- **Tail renderer** (footer pending preview): consumed the scan — this PR moves it to the walk's committed cursor, see gap 1 below. LIVE.
- **Sync-bracket gate**: `seal_frame/3` reads `will_commit` from the same scan, pre-commit. LIVE (from #620).
- **Footer-sizer**: structurally ABSENT — the footer row count is geometry-fixed, never a function of post-seal state (already declared in the moduledoc).
- **Keyframe/reflow gate**: structurally NOT a consumer — the resize path never seals and never sizes off frontier state; a combined resize+advance frame adopts geometry first (frame-order law), then the ordinary seal frame runs. This disposition was undocumented; the consumer table now records it, with the obligation to route through the classifier if the resize path ever seals or the footer height ever becomes tail-dependent.

Remaining private restatements found inside `Surface` itself: the committed-marker mapping (`index < painted_count`) encoded independently in the frontier feed and the fold fill-guard, and done-ness encoded twice (`do_advance/2` guard vs `done?/1`). Closed in this PR (gap 2).

[...]

PR #635, opened 2026-07-17
Harness UI: seal-review follow-up for #630 — tail-bound pin teeth, neutralization pin, legibility

merklebonsai:
Follow-up to the adversarial review of #630 (verdict CONCERNS, merged non-blocking). #630 landed by squash; this PR addresses its review findings on top of `master`.

## Disposition ledger

| ID | Sev | Finding | Verdict | Disposition |
|----|-----|---------|---------|-------------|
| S1 | Medium | one-block tail-bound pin is vacuous (fixture replay, no approval/awaiting corpus → `unsealed <= 1` passes for free) | **VERIFIED** | Enforced with a synthetic tripwire. A fixture replay structurally cannot observe a runtime producer, so I added a paired **synthetic mid-list live-approval hold** — `[approval(live), message, message]`, reveal finished — and assert `frontier_scan/1` stops at the approval (`tail_start == 0`) with `> 1` block stranded past the cursor. That is the exact multi-block hold the corpus lacks; per `SealFrontier.committable?` a `pending_input?` entry stops the frontier unconditionally, so the hold is reachable the moment a producer emits it. Fixture pin kept (now honestly scoped as corpus-only) + moduledoc reworded. |
| SA1 | Low | newly-LIVE footer-preview path presence-tested, not neutralization-tested | **VERIFIED (boundary sound)** | Added a neutralization pin: a refused-write block whose content carries `\e[2J` (the one sequence the seal laws forbid on the wire). Assert the footer keyframe shows literal `[2J` residue and **zero** live `\e[2J` — confirms `ContentGuard.sanitize_line/1` strips the ESC at the repaint seam. Passing regression, not a live vuln. |
| N1 | Low | local `tail_start` in `pending_block/1` lies (it's the committed cursor) | **VERIFIED** | Renamed `tail_start` → `committed_cursor` (the function's whole thesis is that the two differ). |
| N2 | Low | `committed? -- index < painted_count` restated in prose beside code that delegates to `block_sealed?/2` | **VERIFIED** | Folded the doc literal into a reference to `block_sealed?/2` (single-source predicate). |
| N3 | Low | `<= 1` bound and `length(events) * 2 + 10` step cap are unnamed magic literals | **VERIFIED** | Named `@max_unsealed_past_cursor` and `advance_cap/1`. |

No CRITICAL/HIGH raised; nothing refuted. Immutable-prefix / seal-overwrite invariant untouched (rename is a no-op; doc/test-only otherwise).

## Gates
- `test/harness/surface_frontier_feed_test.exs` + `surface_seal_pipeline_test.exs`: **27 passed** (+2 new)
- `test/harness/` (seed 0): **625 passed, 0 failed**, 1 skipped
- mix.lock untouched

### DIFF
--- a/lib/raxol/harness/surface.ex
+++ b/lib/raxol/harness/surface.ex
@@ -266,13 +266,20 @@ defmodule Raxol.Harness.Surface do
   The preview shows ONE block (two lines of it). Today the two never
   differ: the only frontier hold a shipped producer can create is the
   foldable window on the NEWEST block, so the unsealed suffix past the
-  cursor is at most one block long (pinned by the tail-bound test in
-  `test/harness/surface_frontier_feed_test.exs`). The moment a producer
-  emits live blocks (a mid-list awaiting-input approval holding several
-  finalized blocks behind it), that bound breaks and the one-block
-  preview under-reports -- the pinning test fails loudly then, and the
-  multi-block tail rendering it forces is the live-lane (T13b) unit's
-  decision, not silently absorbed here.
+  cursor is at most one block long. Two tests in
+  `test/harness/surface_frontier_feed_test.exs` guard this together, and
+  the split matters: the fixture-REPLAY pin only proves the bound holds
+  over today's shipped corpus (it replays `.jsonl`, so it structurally
+  cannot observe a runtime producer -- on its own it would pass
+  vacuously). Its teeth come from the paired SYNTHETIC test, which builds
+  the exact runtime hold the corpus lacks -- a mid-list awaiting-input
+  `:approval` holding finalized blocks behind it -- and asserts the
+  frontier genuinely stops there, so more than one block sits past the
+  cursor. That is the multi-block hold the one-block preview cannot
+  honor: the moment a producer wires such a hold into a real advance, the
+  bound breaks and the preview under-reports, forcing the multi-block
+  tail rendering decision (the live-lane / T13b unit's), never silently
+  absorbed here.
 
   `Raxol.UI.Components.Harness.Block.seal` is an item-LIFECYCLE field:
   `BlockBuilder` only ever constructs a block once its source item(s)
@@ -968,9 +975,10 @@ defmodule Raxol.Harness.Surface do
 
   Field mapping (the design decision this assembly makes):
 
-    * `committed?` -- `index < painted_count`: physical paint is this
-      module's commit marker, and it only ever advances a contiguous
-      prefix, so the high-water mark IS the committed set.
+    * `committed?` -- delegated to `block_sealed?/2`, THE single-source
+      committed-marker predicate (its doc states the `painted_count`
+      comparison exactly once; restating it here is the drift the
+      unification exists to prevent).
     * `running?` -- `Block.live?/1`, an honest passthrough. Always false
       today (the block builder only constructs completed, sealed blocks),
       which leaves the classifier's mid-turn running exceptions dormant
@@ -2409,11 +2417,16 @@ defmodule Raxol.Harness.Surface do
     # frame the walk's cursor and the scan agree (`tail_start ==
     # painted_count`, the scan/walk-agreement property), so this changes
     # nothing there; on a refusal the cursor is the honest one.
-    tail_start = model.painted_count
+    #
+    # Named `committed_cursor`, NOT `tail_start`: this PR's whole thesis is
+    # that the scan's `tail_start` and the committed cursor are DIFFERENT
+    # quantities that diverge exactly on refusal, so the one function whose
+    # reason-to-exist is that distinction must not reuse the scan's name.
+    committed_cursor = model.painted_count
 
-    case Enum.slice(model.projection.blocks, tail_start..-1//1) do
+    case Enum.slice(model.projection.blocks, committed_cursor..-1//1) do
       [] -> nil
-      [block | _rest] -> {block, tail_start}
+      [block | _rest] -> {block, committed_cursor}
     end
   end
 
--- a/test/harness/surface_frontier_feed_test.exs
+++ b/test/harness/surface_frontier_feed_test.exs
@@ -238,37 +238,61 @@ defmodule Raxol.Harness.SurfaceFrontierFeedTest do
           Raxol.Harness.Fixture.load(Path.join(sessions_dir, name <> ".jsonl"))
 
         model = real_model(session)
+        cap = advance_cap(model)
 
-        Enum.reduce_while(1..advance_cap(model), model, fn _i, m ->
-          {m, outcome} = Surface.advance(m)
+        # Track the terminal outcome so the cap cannot silently swallow an
+        # incomplete drain: reduce over `{model, outcome}` and assert `:done`
+        # after the loop (the #635-R1 saboteur note). A fixture that needs
+        # more than `cap` steps would otherwise exit the loop with its tail
+        # blocks un-run -- hiding any late `unsealed > 1` in that tail.
+        {_final, outcome} =
+          Enum.reduce_while(1..cap, {model, :running}, fn _i, {m, _} ->
+            {m, step} = Surface.advance(m)
 
-          unsealed = length(m.projection.blocks) - m.painted_count
+            unsealed = length(m.projection.blocks) - m.painted_count
 
-          assert unsealed <= @max_unsealed_past_cursor,
-                 "fixture #{name}: #{unsealed} blocks past the committed " <>
-                   "cursor -- the one-block pending preview would hide " <>
-                   "#{unsealed - @max_unsealed_past_cursor} of them " <>
-                   "(multi-block tail rendering is the live-lane follow-up " <>
-                   "this pin exists to force)"
+            assert unsealed <= @max_unsealed_past_cursor,
+                   "fixture #{name}: #{unsealed} blocks past the committed " <>
+                     "cursor -- the one-block pending preview would hide " <>
+                     "#{unsealed - @max_unsealed_past_cursor} of them " <>
+                     "(multi-block tail rendering is the live-lane follow-up " <>
+                     "this pin exists to force)"
 
-          if outcome == :done, do: {:halt, m}, else: {:cont, m}
-        end)
+            if step == :done,
+              do: {:halt, {m, :done}},
+              else: {:cont, {m, :running}}
+          end)
+
+        assert outcome == :done,
+               "fixture #{name}: the advance loop hit its #{cap}-step cap " <>
+                 "without reaching :done -- the drain is incomplete, so a " <>
+                 "late unsealed > 1 in the un-run tail would go unseen. " <>
+                 "Raise advance_cap/1 or fix the stalled advance."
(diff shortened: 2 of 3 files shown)


DROOdotFOO (commented):

Follow-up to my #630 review. I re-ran the two changed test files from the PR head in a worktree (linked deps): **27 passed**, matching the PR's gate. I then mutation-tested both new pins to check for teeth. Results below.

### What holds up

- **S1 tail-bound synthetic pin -- genuine teeth (verified).** `surface_frontier_feed_test.exs:259` builds `[approval(:live), message(:sealed), message(:sealed)]` and asserts `frontier_scan/1` stops at index 0 with `> 1` block stranded. Traced through real code: `frontier_entries/1` maps the live approval to `pending_input?: true` (`surface.ex:760-772`), and `SealFrontier.committable?/3` returns `false` unconditionally on `pending_input?` (`seal_frontier.ex:288`), so `tail_start == 0`, `unsealed == 3`. Mutation proof: changing line 273 `block(:approval, :live)` -> `block(:message, :sealed)` makes the test FAIL at `assert scan.tail_start == 0`. The #630 MED (vacuous fixture-replay pin) is resolved -- this synthetic case is non-vacuous.
- **N1/N2/N3 legibility -- clean.** `tail_start` -> `committed_cursor` in `pending_block/1` (`surface.ex:2417-2428`) is a pure rename, no behavior change; the `committed?` doc now delegates to `block_sealed?/2`; `@max_unsealed_past_cursor` and `advance_cap/1` name the former magic literals. All correct.

### MEDIUM

- **[Security Auditor + New Hire] SA1 neutralization pin does not exercise its stated mechanism -- the #630 anti-pattern, reintroduced.** `surface_seal_pipeline_test.exs:376` and the disposition ledger claim it "confirms `ContentGuard.sanitize_line/1` strips the ESC at the repaint seam." Mutation testing refutes that attribution:
  - Making `ContentGuard.sanitize_line/1` *raise* -> the test errors, so the footer path **does** call it.
  - Making `ContentGuard.sanitize_line/1` a **passthrough** (drop the `scan/1`) -> test still **passes** (no `\e[2J` leak, `[2J` residue still present).
  - Neutering `Surface.ViewText.sanitize/1` (`view_text.ex:220`, the module's self-declared "trust boundary", `view_text.ex:43`) alone -> still **passes**.
  - Neutering **both** together -> still **passes**: raw `\e[2J` never reaches the footer bytes, and `[2J` residue is still produced by a third, un-neutered pipeline layer.

  So a raw clear-screen escape cannot reach the wire through this path regardless of the two documented content sanitizers -- which means the pin **cannot fail if the ContentGuard repaint-seam neutralization it names regresses**. The end-to-end security property (no live `\e[2J`) does hold, with genuine defense-in-depth, so this is not a live vuln. But the pin gives false assurance about the specific seam it claims to guard, which is exactly the "stated guarantee is not the thing it exercises" pattern #630 flagged. Either point the assertion at the layer that actually strips (add a direct `ContentGuard.sanitize_line/1` unit assertion at the seam), or reword the disposition/comment to claim only the end-to-end property, not the ContentGuard mechanism.

### LOW

- **[New Hire] Synthetic pin's prose oversells its assertion.** The test comment and name say the hold means "the one-block preview MUST become multi-block" (`surface_frontier_feed_test.exs:270`, :259), but every assertion is on `frontier_scan/1` (the strand *precondition*) -- nothing asserts on the footer preview / rendering. A reader expecting a preview-under-reports assertion won't find one. The teeth are real but sit on the classifier, not the preview; the prose should say so.
- **[Saboteur] Fixture-replay pin can still under-exercise.** `advance_cap(model) = length(events) * 2 + 10` caps the `reduce_while` (`surface_frontier_feed_test.exs:238-247`), and the loop exits normally when the cap is hit without ever asserting the fixture reached `:done`. A fixture needing more than two steps/event would end the walk before its tail blocks are processed, so a late `unsealed > 1` violation would go unseen. Pre-existing (the cap value is unchanged, only named now) and the PR honestly re-scopes this pin as corpus-only, but the named cap is still a silent-halt seam, not a guaranteed drain.

### Cross-persona overlap
- SA1 neutralization misattribution flagged by Security Auditor (false assurance the ContentGuard seam is pinned) and New Hire (misleading comment + disposition claim): promoted LOW -> MEDIUM.

### Verdict: CONCERNS

The tail-bound synthetic pin earns its teeth (mutation-proven non-vacuous) and the legibility items are clean, so the core of the #630 follow-up lands. But the neutralization pin does not pin the mechanism it advertises -- neutering the named `ContentGuard.sanitize_line/1` (and both content sanitizers together) leaves it green -- so it reintroduces the vacuity-of-claim pattern the PR set out to remove. No CRITICAL/HIGH, no live vulnerability. Fix by asserting against the actual stripping seam or rewording the claim to the end-to-end property only.

merklebonsai:
## R1 dispositions (commit 96c3e9acf)

Thanks for the mutation pass. The core (tail-bound synthetic pin, N1/N2/N3) you verified — untouched. The two findings, addressed:

### SA1 (Medium) — pin attribution fixed by tracing the bytes
Verified and root-caused. I traced the refused-write footer-preview pipeline: `block.content` = `%{text: "RETRY-ME \e[2J HOSTILE"}` (ESC intact), but `BlockBody.render/2`'s output view text node is already `"RETRY-ME [2J HOSTILE"` — **the message-body renderer strips the ESC while building the display view, upstream of both ViewText.sanitize and the footer paint's ContentGuard.** That is your "third layer": ContentGuard is a redundant catch-all on this path, which is exactly why neutering it (and ViewText.sanitize) left the pin green. Genuine defense-in-depth, no live vuln — confirmed.

Fix (the preferred seam close, delivered by citation rather than duplication):
- The **seam assertion you asked for already exists, red-first**: `ContentGuard.sanitize_line(<<0x1B, "[2J">>) == "[2J"` at `test/raxol/harness/c1_sanitizer_test.exs:39`, and the footer-paint seam against the independent full-clear oracle at `test/property/renderer_t2c_review_fixes_test.exs:157`. Rather than duplicate them, the pin now **cites** them.
- I **rewrote my pin** as an honest END-TO-END composition test over the refused-write → footer-preview path, judged by the same independent oracle `SealOracle.emits_full_clear?/1` (not a byte-substring match), with a **fail-first arm** (`assert emits_full_clear?(raw \e[2J)`) proving the oracle is live — the t2b/t2c discipline. The single-seam ContentGuard attribution is gone; the claim is now the end-to-end property only.

### Low #1 (New Hire) — synthetic prose oversell fixed
Reworded the test name + comment to name `frontier_scan/1` (the classifier / strand precondition) as the referent, and explicitly disclosed that **no assertion is on the preview render** — the preview-render assertion is out of scope because the synthetic hold cannot be built from `Surface.new/2` events (the block builder only emits sealed blocks); it belongs with the live lane (T13b).

### Low #2 (Saboteur) — advance_cap silent-halt closed
The fixture-replay `reduce_while` now tracks the terminal outcome and **asserts `:done` after the loop**. A fixture needing more than `cap` steps now fails loudly ("hit its N-step cap without reaching :done…") instead of silently ending with un-run tail blocks that could hide a late `unsealed > 1`.

### Gates
- `surface_frontier_feed` + `surface_seal_pipeline` + `c1_sanitizer` + `renderer_t2c`: **54 passed**
- `test/harness/` + `renderer_t2b/t2c` + `renderer_seal_once`: **685 passed, 0 failed** (seed 0)
- `mix compile --warnings-as-errors` clean; `mix format --check-formatted` clean; mix.lock untouched

DROOdotFOO (commented):

Since last review: 6d557bf8..96c3e9ac

Scope: the single new commit `96c3e9ac` targeting my prior CONCERNS
(SA1 misattribution MED; synthetic-prose + advance_cap LOWs). Verified by
mutation testing in an isolated worktree at the PR head (all 54 tests
across the four touched/cited files green at baseline).

**SA1 (MED, prior: reproduced) -- neutralization pin misattribution: RESOLVED.**
The fix does not (and structurally cannot) make the end-to-end preview test
isolate `ContentGuard.sanitize_line/1`; instead it corrects the attribution.
Mutation evidence (neuter `ContentGuard.sanitize_line/1` -> passthrough):

- The end-to-end pin `surface_seal_pipeline_test.exs:375` STAYS GREEN
  (20/20). This is now honest: the comment (`:386-397`) reframes it as an
  end-to-end composition pin, states TWO layers neutralize, and explicitly
  traces "the raw `\e[2J` is already gone from the rendered view before
  ContentGuard ever runs." The prior false claim to guard the ContentGuard
  neutralization is gone.
- The correctly-attributed single-seam pins it now cites DO fail red-first
  under the same mutation: `c1_sanitizer_test.exs:39`
  (`sanitize_line(<<0x1B,"[2J">>) == "[2J"`) and the `\e[2J` footer-paint
  seam in `renderer_t2c_review_fixes_test.exs:157` -- 8 failures total.

So the ContentGuard neutralization IS genuinely pinned red-first (just at
the correct locations), and the end-to-end test no longer misrepresents
what it exercises. Misattribution fixed.

**LOW -- synthetic prose oversell: RESOLVED.** The synthetic test
(`surface_frontier_feed_test.exs:274`) now carries an explicit REFERENT
block (`:281-291`) scoping every assertion to the classifier
(`frontier_scan/1`, the strand PRECONDITION), and states the preview-render
assertion is deliberately out of scope. Teeth confirmed by mutation: flip
`SealFrontier.committable?` `pending_input? -> false` to `-> true` and this
test fails on `assert scan.tail_start == 0` (test #2 of 3 broken).

**LOW -- advance_cap silent halt: RESOLVED.** The fixture-replay reduce
(`:238-271`) now threads `{model, outcome}` and asserts `outcome == :done`
after the loop, so a wedged advance that exits at the cap fails loudly
instead of hiding a late `unsealed > 1` in an un-run tail. `:done` is
reachable (baseline green).

Supporting hygiene also verified: the `committed?` moduledoc claim matches
code (`surface.ex:1020` delegates to `block_sealed?/2` at `:1044` =
`index < painted_count`, single-source), and the refusal-path rename
`tail_start -> committed_cursor` (`surface.ex:2422`) is consistent.

### Low / Nit (New Hire persona) -- non-blocking

The end-to-end pin's comment says the two layers "COMPOSE over the
refused-write preview surface" (`:396`). Given the immediately-disclosed
trace (layer 1, the message-body renderer, fully neutralizes BEFORE
ContentGuard runs), ContentGuard contributes nothing OBSERVABLE on this
path -- the test proves layer 1 alone suffices with ContentGuard as an
unexercised backstop, not that both contribute in series. "Compose"
slightly overstates that, but the preceding sentences fully disclose the
trace, so a reader has the truth. Optional wording tighten only; no test
behavior change.

### Verdict: CLEAN

All three prior findings resolved with mutation evidence: the ContentGuard
neutralization now genuinely fails red-first at the correctly-attributed
seam pins (mutation-proven), while the end-to-end pin honestly declares
itself a defense-in-depth composition and stays green under that same
mutation as its own comment states. Remaining item is a disclosed wording
nit.

PR #636, opened 2026-07-17
Harness UI: post-merge follow-ups for #629 projection panels (M1/M2/H1-fixture)

merklebonsai:
Post-merge follow-ups for the projection panels (#629), addressing Drew's first adversarial review. #629 merged (`7081e2183`) ~1 min before the READY verdict; H1 was closed by #632 landing in master first, so these harden the remaining items. Targets `master`.

## Dispositions

| Finding | Sev | Disposition |
|---|---|---|
| **H1** C1 control bytes past the sanitize boundary | HIGH | **Already closed** by #632 (`ViewText.sanitize/1` + `ContentGuard.sanitize_line/1` strip U+0080..U+009F code-point-aware, in master). The *fixture/oracle gap* it exposed (only ESC tested → false confidence) is closed here: C1 case added to the hostile fixture + an **end-to-end wire regression** asserting a U+009B CSI never reaches the `:styled` wire. |
| **M1** `@max_field_bytes` grapheme-vs-byte clamp | MED→LOW | **Fixed.** `display_string/1` fallback used `String.slice/3` (grapheme count) → "512-byte clamp" unbounded on the mid-codepoint-split path. Now byte-slice + trailing-partial-codepoint trim (≤512 bytes, valid UTF-8). |
| **M2** recompute-per-paint re-folds journal on every keystroke | MED | **Fixed.** Memoize `refresh_panel_overlay/1` against the append-only `source_events` length; a pure scroll key no longer re-folds. |
| **L1** `overlay.picker` field name / dead `on_pick` | LOW | Typedoc now names `mod` as the discriminator and documents `folded_at`. Rename deferred (disproportionate cross-site risk on merged, seal-adjacent code). |
| **L2** magic `max_overlay_rows - 1` | LOW | One-line comment (reserve the title row). |

## Verification (red-first)

Each behavioral fix has a red-first regression, all confirmed to fail without their fix:
- **M1** — `"→"×400` (1200 bytes): fails `1200 <= 512` on old fallback.
- **M2** — `KeyError :folded_at` when memoization removed.
- **H1/C1** — C1 reaches the wire when *both* sanitize seams reverted (the test is implementation-independent: it pins the referent "C1 never reaches the wire", not one seam).

Gates: `mix compile --warnings-as-errors` clean; format clean; panel/projection/overlay/keymap suites green (27/27 of the touched fast tests). Full `test/harness/` = 712/715, the 3 failures are pre-existing `AnsiReplayer` cross-terminal replay **timeouts** under concurrent load (each passes in isolation; unrelated to this diff).


DROOdotFOO (commented):

Verified against the PR head (`2d11075a`) in a clean worktree off `master`.
Touched fast suites green: `panel_projection_test.exs` +
`projection_panels_surface_test.exs` = 41/41 passing.

### Security Auditor

- **H1/C1 fixture + wire oracle: resolved, verified.** The prior review's
  gap was a fixture that only smuggled a C0 ESC, giving false confidence in
  the sanitize seam. This PR adds a true C1 case to the hostile fixture
  (`projection-panels.jsonl` offset id=20: `title` now carries `3Gc1`,
  the single-code-point 8-bit CSI form) AND an end-to-end wire regression
  (`projection_panels_surface_test.exs`, "an 8-bit C1 CSI in a title is
  stripped before it reaches the wire"). The wire test exercises the real
  hazard: U+009B arrives from agent JSON as its 2-byte UTF-8 form
  `<<0xC2, 0x9B>>` (both bytes >= 0x20, exactly what a byte-wise C0-only
  allowlist would pass through) and asserts BOTH `<<0xC2, 0x9B>>` and a bare
  `<<0x9B>>` are absent from the `:styled` delta. Confirmed the referent is
  real: `ViewText.sanitize/1` on the PR base is code-point-aware
  (`view_text.ex:227-233`) -- it decodes `cp::utf8` and drops
  `cp >= 0x80 and cp <= 0x9F`, so the UTF-8 C1 is caught, while the bare-byte
  fallthrough clause (`view_text.ex:233`) drops a lone raw high byte. The
  layering is sound and now test-pinned on both sides: `fold/2` passes C1
  through raw (`panel_projection_test.exs` asserts the raw C1 survives the
  fold), and `ViewText` strips it at the paint boundary (surface test asserts
  it never reaches the wire). No bypass path found -- the panel line paint
  path empirically routes through `ViewText` (the delta assertion would
  otherwise fail).

### Saboteur

- **M1 byte-clamp: fixed, no mid-codepoint corruption (reproduced).**
  `display_string/1` (`panel_projection.ex:313-343`) now byte-slices
  `binary_part(v, 0, 512)` then `trim_partial_codepoint/1` drops trailing
  bytes until valid UTF-8. Reproduced the old-vs-new divergence directly:
  `String.duplicate("→", 400)` = 1200 bytes; the OLD
  `String.slice(v, 0, 512)` fallback (a GRAPHEME count) returned all **1200
  bytes** unclamped, so the "512-byte clamp" never bounded bytes on the
  mid-codepoint-split path. The new path returns **510 bytes**, `String.valid?`
  true, `<= 512`. Edge cases hold: a 4-byte emoji run lands exactly on 512
  (128 emoji, valid), and the recursion terminates (empty binary is valid,
  drops at most 3 bytes). No invalid-UTF-8 emission -- the exact NEW-bug class
  a naive byte clamp would introduce is closed.

- **M2 memoization correctness: sound; residual guard cost noted below.**
  Attempted a same-length-different-content break: the token is
  `folded_at = length(source_events)`, and the append-only invariant
  (`source_events :: [map()]`, projection refold reuses the same list) plus
  the fact that the token is captured per-overlay at `do_open_panel/4`
  (`surface.ex:2008,2033`) means an unchanged length is genuinely an
  identical fold. Also confirmed no blank-on-open bug: `open_panel` folds the
  initial lines (`put_lines(render_lines(...))`, `surface.ex:2007-2010`) AND
  sets `folded_at` to the same length in the same call, so the first
  `paint_footer` correctly skips re-fold with lines already populated.

### New Hire (LOW, informational -- non-blocking)

- **M2 guard is still O(source_events) per keystroke.** `refresh_panel_overlay/1`
  computes `length(model.projection.source_events)` on EVERY `paint_footer/1`
  (i.e. every keystroke, including pure scroll) to derive the memo token
  (`surface.ex`). `source_events` is a list (`projection.ex:192`), so `length/1`
  is a full O(n) traversal over a monotonically growing journal. The fix
  correctly elides the expensive fold+format on a pure scroll (the PR's claim
  holds), but the guard itself is not O(1) -- a scroll keystroke still walks
  the whole journal just to prove nothing changed. A monotonic append-index /
  counter on the projection would make the guard constant-time. Not a
  correctness issue and a large constant-factor win over the old
  fold-every-paint; flagging as a residual, not a blocker.

### Verdict: CLEAN

All three findings from the #629 review are verified fixed: M1 is now a true
byte clamp with no mid-codepoint corruption (reproduced), M2 memoization is
correct with a documented consistent init, and the H1 fixture/oracle gap is
closed with a C1 case that exercises both the 2-byte UTF-8 and bare-byte
forms against a code-point-aware sanitize boundary. The one LOW note (O(n)
memo guard) is informational.

PR #637, opened 2026-07-17
test: fix the macOS CI flake class (CaptureServer starvation + brittle deadlines)

merklebonsai:
Both #628 and #631 failed \`Test macos-latest\` on the same suites. Diagnosis and fix:

**Root cause:** \`CharacterProcessor.calculate_positions\` logged per-character at debug level. Under test config (\`level: :debug\` + global \`capture_log\`) every emulator-replay byte serialized through ExUnit's single CaptureServer, starving sibling suites on loaded runners. Removing it cut local harness wall time 397s → ~105s. The failing suites were victims of the contention, not flaky by themselves.

**Hardening (CI-honesty rules — deadlines pin arrival + order, never latency):**
- \`stream_cadence_test\`: \`async: false\` + receive deadlines 50/100ms → 2s.
- \`markdown_body\` N-MDFUZZ-05: never-hang bound 2s → 30s (order-of-magnitude falsifier — the pre-fix Θ(n³) case costs minutes; the bound keeps teeth without runner-load flakes).

Gates: compile --warnings-as-errors clean, both suites 89/89 green, format clean, no lock drift.

### DIFF
--- a/packages/raxol_terminal/lib/raxol/terminal/input/character_processor.ex
+++ b/packages/raxol_terminal/lib/raxol/terminal/input/character_processor.ex
@@ -145,14 +145,16 @@ defmodule Raxol.Terminal.Input.CharacterProcessor do
         {write_col, write_row, next_cursor_col, next_cursor_row, next_last_col_exceeded}
       )
 
-    log_cursor_positions(
-      current_cursor_col,
-      current_cursor_row,
-      adjusted_write_col,
-      adjusted_write_row,
-      adjusted_next_cursor_col,
-      adjusted_next_cursor_row
-    )
+    # No per-character debug log here (removed): this function runs once
+    # per PRINTABLE CHARACTER, and a `Log.debug` in that loop eagerly
+    # interpolates its message on every call regardless of level, then
+    # (under the test env's `level: :debug` + global `capture_log`)
+    # funnels every character of every emulator replay through ExUnit's
+    # single CaptureServer -- serializing the whole async suite and
+    # pushing replay-oracle tests from seconds into minutes. A
+    # per-character cursor trace is reconstructible from the input bytes
+    # whenever it is actually needed; it earns neither the production
+    # interpolation cost nor the test-suite serialization.
 
     {adjusted_write_col, adjusted_write_row, adjusted_next_cursor_col, adjusted_next_cursor_row,
      adjusted_next_last_col_exceeded}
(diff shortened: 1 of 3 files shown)


DROOdotFOO (commented):

Small CI-infra fix. Reviewed the production change and both test-file deltas against three hostile lenses. This is a genuine root-cause fix, not deadline-papering.

### Saboteur

ROOT-CAUSE, confirmed (not a band-aid). The production diff in
`packages/raxol_terminal/lib/raxol/terminal/input/character_processor.ex`
deletes both the `log_cursor_positions(...)` call site inside
`calculate_positions/3` and the `defp log_cursor_positions/6` itself. That
removes the per-printable-character `Log.debug` at the source, which is what
was funnelling every emulator-replay byte through ExUnit's single
CaptureServer under `level: :debug` + global `capture_log`. The eager
message interpolation on every char is also gone -- a pure hot-path win.
The removed line was a debug-only cursor trace (row/col ints); nothing
consumes its side effect, so it is not load-bearing. Correct that the
sibling `log_cursor_debug/2` was left intact -- it only fires on the
`_ ->` unknown-cursor-type fallback branch, not the printable hot path.

Deadlines are NOT loosened to vacuity. Every widened `50/100 -> 2_000`
value in `test/harness/stream_cadence_test.exs` is an `assert_receive`
UPPER bound ("the batch eventually arrives"), which is exactly the
load-sensitive quantity. The MEANINGFUL timing teeth -- the `refute_receive`
LOWER bounds that assert coalescing/yield-hold DIDN'T fire too early --
are left UNCHANGED at 10ms/20ms (stream_cadence_test.exs:68, :85, :144,
:204, :210). So a regression that flushes early or fails to coalesce still
fails; only the "arrived in time" ceiling was relaxed. A true hang still
blocks past 2s and fails. Good separation of concerns.

`markdown_body_test.exs:463` `@perf_bound_us 2_000_000 -> 30_000_000`: this
bound is shared by the 20k/50k/mixed N-MDFUZZ-05 tests. 30s still falsifies
the guarded regression class (the pre-fix Theta(n^3) 20k case costs minutes;
the single-pass O(n) close is sub-second even on a loaded runner). Note it
is loose against a sub-order-of-magnitude regression (a 10x slowdown to ~20s
would slip through), but that is consistent with the test's stated role as a
never-hang / order-of-magnitude falsifier rather than a latency SLO.

### New Hire (minor)

Self-documenting: strong. The replacement comment in `character_processor.ex`
spells out the exact failure mechanism (eager interpolation x capture_log x
single CaptureServer -> suite serialization), and the `async: false` +
deadline rationale are both explained in-file. One nit: the `2_000` arrival
deadline is repeated ~15x as a bare literal rather than a named
`@arrival_deadline_ms` attribute; the PR body documents the "pin arrival +
order, never latency" rule but the inline constant carries no explanation at
its use sites. Cosmetic.

### Security Auditor

Clean. The deleted log emitted only cursor row/col integers -- no
security-relevant audit line, no credential or untrusted-data content. Input
bytes still flow identically through `calculate_positions/3`; only the
logging side effect is removed. No behavioral change on any trust boundary.

### Verdict: CLEAN

Genuine root-cause fix (per-character debug log gated/removed at source, not
just wider deadlines). No changed deadline is vacuous: widened values are
load-sensitive arrival ceilings, while the order/early-fire `refute_receive`
guards stay tight; the 30s markdown bound remains a meaningful cubic-regression
falsifier. Non-blocking nit only (magic-number deadline).

PR #638, opened 2026-07-17
test(acp): revert placebo await_initialized barrier (structural no-op)

merklebonsai:
## Summary

An earlier commit added an `await_initialized/1` helper (+ two call sites) to `packages/raxol_agent_client_protocol/test/capabilities_test.exs`, in the "Paired integration: agent terminal/create against a no-terminal client" tests. It claimed to fix a "~3% under 32x parallel load" flake by waiting for the client connection to reach `phase: :initialized` before driving `terminal/create`.

This is a **causal impossibility**, not a fix:

- `Connection.request/4`'s `{:ok, _}` return is produced by `deliver_outcome/2` (`GenServer.reply/2`), called at `connection.ex` line ~1085.
- The commit of `phase: :initialized` happens in `maybe_client_initialized/3`, called immediately after, at lines ~1087-1088 — inside the **same GenServer reduction**, same `handle_inbound_response/2` callback.
- There is no yield point between the reply and the state commit. OTP's single-process mailbox serialization guarantees that any later message reaching that connection (including the routed `terminal/create`) is processed only after the current reduction returns — i.e. only after the state is already committed.
- `await_initialized` therefore polls a condition that is guaranteed true on its very first check. It is a structural no-op: it "fixes" nothing, it just adds latency and a green label.

Verified by reverting the barrier and running `capabilities_test.exs` 80x at 16-way parallelism: 0 failures.

So either the originally-reported flake never existed as described, or it's a *different*, still-latent failure (candidate: request timeout, or process-start latency under CPU starvation) that is now hidden behind a green test. This PR reverts the placebo and leaves an honest comment in its place so a future flake here gets diagnosed fresh instead of re-papered.

## Changes

- Remove `await_initialized/1` and its two call sites in `test/capabilities_test.exs`.
- Replace with a comment explaining the causal-impossibility reasoning and explicitly stating that a future flake in this test is NOT a capability-ordering issue.
- No other `await_initialized` references existed in the package (verified via grep).

## Test plan

- [x] `MIX_ENV=test mix compile --warnings-as-errors` — clean.
- [x] `MIX_ENV=test mix test` — 795 passed (16 properties, 779 tests), 0 failures.

### DIFF
--- a/packages/raxol_agent_client_protocol/test/capabilities_test.exs
+++ b/packages/raxol_agent_client_protocol/test/capabilities_test.exs
@@ -394,7 +394,19 @@ defmodule Raxol.AgentClientProtocol.CapabilitiesTest do
       }
 
       assert {:ok, _} = Connection.request(ctx.client_conn, "initialize", init_req, 5_000)
-      assert :ok = await_initialized(ctx.client_conn)
+
+      # No barrier needed here: `phase: :initialized` is committed inside the
+      # SAME GenServer reduction that replies to this `request/4` call (see
+      # `handle_inbound_response/2` -> `maybe_client_initialized/3` in
+      # connection.ex) — the reply is sent before the state update, but both
+      # happen with no yield point in between, so OTP's single-process
+      # mailbox serialization guarantees any later message reaching this
+      # connection (including the terminal/create below) observes the
+      # post-commit phase. A polling wait here would be a no-op: the state it
+      # waits for is already true by the time the wait could run. If this
+      # test flakes again under load, the cause is NOT capability-ordering —
+      # diagnose fresh (candidate: request timeout or process-start latency
+      # under CPU starvation), don't paper over it with a wait.
 
       # Decode-INVALID params (missing sessionId/command). If the gate ran
       # AFTER decode we would see -32602; -32601 proves the gate ran first.
@@ -414,7 +426,10 @@ defmodule Raxol.AgentClientProtocol.CapabilitiesTest do
       }
 
       assert {:ok, _} = Connection.request(ctx.client_conn, "initialize", init_req, 5_000)
-      assert :ok = await_initialized(ctx.client_conn)
+
+      # No barrier needed — see the comment in the sibling test above:
+      # `phase: :initialized` is already committed by the time this call
+      # returns (single-process mailbox serialization in Connection).
 
       assert {:ok, %{terminal_id: "term-ok"}} =
                Connection.request(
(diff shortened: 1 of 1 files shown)


DROOdotFOO (commented):

Scope: revert of `await_initialized/1` + two call sites in `packages/raxol_agent_client_protocol/test/capabilities_test.exs` (+17/-20). I verified the core "structural no-op" claim against `connection.ex` on the branch.

The claim holds. In `handle_inbound_response/2` the `{:ok, _}` reply is delivered at line 1085 (`deliver_outcome` -> `GenServer.reply`) and `phase: :initialized` is committed at line 1088 (`maybe_client_initialized/3`) within the SAME reduction, no yield point between. `await_initialized` polled `:sys.get_state(conn).phase`, and a `:sys.get_state` message is serialized behind that reduction in the connection's mailbox, so it was guaranteed to observe `:initialized` on its first check. The `Process.sleep(2)` branch was therefore unreachable, so the barrier added neither ordering nor masking latency. Revert is safe hygiene.

### Saboteur (does the revert re-introduce a real race?)
No re-introduced race. The two tests are strictly sequential: `request(client_conn, "initialize")` returns only AFTER `client_conn` has processed the initialize RESPONSE (that is what delivers the `{:ok}`), and the test then sends `terminal/create` to `agent_conn`, which routes it into `client_conn`'s mailbox behind the already-in-flight/committed reduction. So `client_conn`'s inbound gate always sees `phase: :initialized`. Stronger still: this is self-asserting. If the phase-lag the original comment described were real, the handshake gate (`handle_inbound_request` case 1: `phase == :uninitialized and method != "initialize"`) would reject `terminal/create` with a DIFFERENT code, so the test would fail deterministically rather than flake. Both tests assert the exact capability-gate code (`-32601` / gate-pass), which can only be reached post-commit. Minor note: the verification was 80x @ 16-way, but the originally-reported flake was "~3% under 32x". Given the airtight causal argument this is non-blocking, but a single 32x-parallel run would close the load-parity gap and match the claim it retires.

### New Hire (is the "placebo" claim substantiated?)
Yes, with line-level evidence. The replacement comment accurately names the mechanism (`handle_inbound_response/2` -> `maybe_client_initialized/3`, reply-before-commit, single-process mailbox serialization) and correctly redirects a future flake away from capability-ordering toward the real candidates (request timeout / process-start latency under CPU starvation). `grep` confirms zero remaining `await_initialized` references in the package. The honest acknowledgment that a real-but-different latent flake may still exist (and should be diagnosed fresh, not re-papered) is the right call rather than leaving a green placebo in place.

### Security
n/a (test-only change, no runtime surface).

### Verdict: CLEAN

Barrier was genuinely a structural no-op (state it waited on is committed in the same non-yielding reduction that unblocks the caller, and enforced by mailbox serialization); revert is safe and does not re-flake the test. Only nit: re-run once at the originally-claimed 32x parallelism for evidence parity.

PR #640, opened 2026-07-17
fix(acp): single-sender session/update delivery (no drop/reorder)

merklebonsai:
## LIVE incremental in-order `session/update` delivery (per-session `update_seq`)

Supersedes the earlier global-`rx_seq` **boundary-emit** approach on this branch. That version used the GLOBAL per-connection frame counter (`rx_seq`) as the reorder key — which gives ordering but not *completeness* (a numeric gap in `rx_seq` is an unrelated frame, not a missing update), so it was forced to buffer every update and emit the whole turn only at the **turn boundary**, losing live streaming.

This version streams **live**: the agent stamps a **per-session monotone `update_seq`** so the consumer can release each update the instant it is contiguous.

### Agent side (producer)
- The `Session` GenServer stamps a per-session `update_seq` into every emitted `session/update` notification's `_meta["raxol.io"]["update_seq"]` (mirroring the reattach taint/offset rider convention).
- Stamped in the Session core **before the emitter seam**, so `Emitter.Direct` and `Emitter.Journal` both carry the identical stamped notification (emitter-agnostic).
- The counter **resets to 0 at each turn start** (`begin_prompt`), so every `session/prompt` turn's updates form a contiguous `0..N` and the consumer's next-expected index always starts at 0. Only one turn is ever live per session (the I12 busy rule), so per-turn reset never collides.
- The Session mailbox serializes concurrent `post_update` calls, giving parallel/interleaved in-turn updates a single total order. Rejected empty chunks consume no ordinal (no phantom gap).

### Wire
- The ordinal rides `_meta["raxol.io"]["update_seq"]` — spec-compatible: a vanilla ACP client ignores `_meta`; a vanilla agent simply never stamps it. It survives client decode via the existing `SessionNotification` `_meta` fold.

### Client side (consumer)
- `prompt/3` / `prompt_stream/4` track a next-expected ordinal: an update whose `update_seq == next` is **released immediately** (`prompt_stream` fires `on_update` live; `prompt` accumulates in order), then cascade-releases any contiguous successors already buffered; a higher ordinal buffers; a gap holds the higher ones until the missing lower one lands.
- **Bounded, never deadlocks:** `Connection` always delivers a terminal `{:acp_result, …}` (its protocol timeout is the single authority), and a bounded post-result settle drain absorbs stragglers then flushes any out-of-order tail — nothing dropped.
- **Graceful fallback:** an update with no `update_seq` (non-raxol / unstamped agent) is buffered by `rx_seq` and released in `rx_seq` order at the turn boundary — the prior deterministic, no-drop, non-blocking behavior. The live/in-order guarantee holds **only** when the agent stamps the seq.
- Delivery tuple grows to `{:acp_session_update, session_id, update, rx_seq, update_seq}` — `rx_seq` kept as the per-connection frame stamp / tiebreaker; `update_seq` is the new reorder key (`nil` when unstamped). `{:acp_result_seq, tag, rx_seq}` stays as the turn-boundary signal.

### Tests
- **Producer-real:** drives a real `Session` emit and asserts the outgoing notifications carry `_meta["raxol.io"]["update_seq"]` = 0,1,2, and that a second turn resets back to 0.
- **Consumer red-first witnesses (deterministic, mailbox barriers, no sleep/wall-clock):** (a) live incremental in-order (out-of-order arrivals ⇒ `on_update` fires 0..N, released before the result); (b) gap-wait (seq 0,2 arrive, 1 missing ⇒ 2 held until 1 lands, then 1,2 release); (c) two sessions interleaved, each released in its own per-session order, none crossed; (d) graceful fallback (unstamped updates delivered, never blocked). The three ordering witnesses were confirmed **red** against an arrival-order consumer.

Full suite: **801 passing** (16 properties, 785 tests). `mix compile --warnings-as-errors` clean.

### DIFF
--- a/packages/raxol_agent_client_protocol/lib/raxol/agent_client_protocol/client.ex
+++ b/packages/raxol_agent_client_protocol/lib/raxol/agent_client_protocol/client.ex
@@ -330,17 +330,19 @@ defmodule Raxol.AgentClientProtocol.Client do
   failure/decode failure, same vocabulary as `Connection.async_request/6`'s
   `outcome`.
 
-  Ordering caveat: `session/update` forwarding to the subscribed process
-  runs in a separate handler task per notification (Connection design
-  §4.4), while the terminal `session/prompt` response is delivered
-  directly by `Connection` itself -- the WIRE order between them is
-  invariant (I3: no update after its turn's response), but the two
-  deliveries landing in *this* process's mailbox are two different
-  senders, so BEAM gives no cross-sender guarantee. This function does one
-  short best-effort settle pass after the terminal message arrives to
-  absorb same-turn stragglers; treat `updates` as "very likely complete,
-  not protocol-guaranteed complete" -- `prompt_stream/4` has no such gap
-  because it never leaves the receive loop.
+  Ordering: `Connection` delivers `session/update` forwarding to the
+  subscribed process SYNCHRONOUSLY, from the Connection process itself,
+  before it ever looks at the next inbound frame (see
+  `Connection`'s "Single-sender `session/update` delivery" moduledoc
+  section) -- and the terminal `session/prompt` response is also delivered
+  directly by `Connection`. Both now originate from the SAME sender to
+  this process, so BEAM's per-sender/per-receiver FIFO guarantee applies:
+  the WIRE invariant (I3: no update after its turn's response) carries
+  through to mailbox order here, not just on the wire. This function still
+  does one short best-effort settle pass after the terminal message
+  arrives, kept as defence-in-depth (e.g. against a future notification
+  path that isn't single-sender) rather than because it is required for
+  correctness today.
 
   Always subscribes and unsubscribes around the call, including on error.
 
--- a/packages/raxol_agent_client_protocol/lib/raxol/agent_client_protocol/client.ex
+++ b/packages/raxol_agent_client_protocol/lib/raxol/agent_client_protocol/client.ex
@@ -89,11 +89,13 @@ defmodule Raxol.AgentClientProtocol.Client do
   mailbox protocol. The generated default `c:session_update/2` (unless you
   override it) forwards every decoded update to whichever processes called
   `subscribe/3` for that `{conn, session_id}` pair, as
-  `{:acp_session_update, session_id, update, rx_seq}` -- `update` is always
-  an already-decoded `Raxol.AgentClientProtocol.Schema.SessionUpdate.t()`
-  (see `decode_update/1`'s doc for why an undecodable variant never reaches
-  this path at all), and `rx_seq` is the inbound frame's monotone sequence
-  stamp (the reorder key -- see below).
+  `{:acp_session_update, session_id, update, rx_seq, update_seq}` -- `update`
+  is always an already-decoded
+  `Raxol.AgentClientProtocol.Schema.SessionUpdate.t()` (see `decode_update/1`'s
+  doc for why an undecodable variant never reaches this path at all), `rx_seq`
+  is the inbound frame's monotone per-connection stamp, and `update_seq` is the
+  agent's PER-SESSION update ordinal (the reorder key) or `nil` when unstamped
+  -- see below.
 
   `prompt/3` and `prompt_stream/4` build on `subscribe/3` to give a
   synchronous-feeling API over the inherently asynchronous protocol:
@@ -113,10 +115,11 @@ defmodule Raxol.AgentClientProtocol.Client do
   for the turn: update deliveries arrive from concurrent per-notification
   dispatch tasks (non-blocking, so parallel/interleaved update streams
   flow), so they can land out of wire order and can even race the terminal
-  result into the mailbox. Each carries its `rx_seq`; the loop buffers by
-  `rx_seq`, drains every update below the result's own `rx_seq` before
-  completing, and returns them in restored wire order -- see their docs for
-  the exact mechanics.
+  result into the mailbox. Each carries the agent's per-session `update_seq`;
+  the loop releases each update the instant it is contiguous (LIVE, in
+  per-session order), holding a gap until the missing ordinal lands, and
+  falls back to `rx_seq` boundary ordering for an agent that does not stamp
+  `update_seq` -- see their docs for the exact mechanics.
 
   ## `fs_sandbox`: a working filesystem client with zero callbacks
 
@@ -214,15 +217,19 @@ defmodule Raxol.AgentClientProtocol.Client do
   Subscribe `subscriber` (default `self()`) to `session_id`'s
   `session/update` deliveries on `conn`. The generated default
   `c:session_update/2` (unless overridden) broadcasts
-  `{:acp_session_update, session_id, update, rx_seq}` to every subscriber
-  registered for `{conn, session_id}` -- `update` is always an
+  `{:acp_session_update, session_id, update, rx_seq, update_seq}` to every
+  subscriber registered for `{conn, session_id}` -- `update` is always an
   already-decoded `Raxol.AgentClientProtocol.Schema.SessionUpdate.t()`
   (see `decode_update/1`'s doc for why an undecodable variant never
-  reaches this path), and `rx_seq` is the inbound notification frame's
-  monotone sequence stamp (`Ctx.rx_seq`). Because each update is forwarded
-  from its OWN concurrent dispatch task, deliveries to the subscriber's
-  mailbox can be out of wire order; `rx_seq` is the key a consumer uses to
-  restore that order (`prompt/3`/`prompt_stream/4` do this for you). NOT
+  reaches this path), `rx_seq` is the inbound notification frame's
+  monotone per-connection stamp (`Ctx.rx_seq`), and `update_seq` is the
+  agent's PER-SESSION update ordinal read from
+  `_meta["raxol.io"]["update_seq"]` (or `nil` when the agent did not stamp
+  it). Because each update is forwarded from its OWN concurrent dispatch
+  task, deliveries to the subscriber's mailbox can be out of wire order;
+  `update_seq` is the key a consumer uses to restore per-session order and
+  release incrementally (`prompt/3`/`prompt_stream/4` do this for you),
+  falling back to `rx_seq` ordering when `update_seq` is `nil`. NOT
   idempotent: subscribing the same pid twice for
   the same key registers it twice in the underlying `:bag` -- each
   registration receives its own copy of every broadcast, so the pid gets
@@ -256,24 +263,52 @@ defmodule Raxol.AgentClientProtocol.Client do
   # (design docs' "handler crash never reaches the wire" spirit, applied
   # here even though a notification crash is already log-only, IC/§4.4).
   #
-  # `rx_seq` is the inbound `session/update` frame's monotone sequence
-  # (`Ctx.rx_seq`); it rides the delivered tuple as the 4th element so a
-  # consumer can restore wire order across concurrent per-notification
-  # dispatch tasks (see `prompt/3`/`prompt_stream/4`). Defaults to `0` so a
-  # direct/legacy `broadcast_update/3` caller (custom handler, replay
-  # tooling) keeps working -- such deliveries simply sort as "seq 0".
-  @spec broadcast_update(pid(), String.t(), term(), non_neg_integer()) :: :ok
-  def broadcast_update(conn, session_id, update, rx_seq \\ 0) do
+  # The delivered 5-tuple `{:acp_session_update, session_id, update, rx_seq,
+  # update_seq}`:
+  #
+  #   * `rx_seq` — the inbound frame's GLOBAL per-connection monotone stamp
+  #     (`Ctx.rx_seq`); kept as a frame stamp / tiebreaker. Defaults to `0` so a
+  #     direct/legacy `broadcast_update/3` caller (custom handler, replay tooling)
+  #     keeps working.
+  #   * `update_seq` — the PER-SESSION update ordinal the agent stamped into
+  #     `_meta["raxol.io"]["update_seq"]` (see `Session.stamp_update_seq/2`), or
+  #     `nil` when the agent did not stamp it. This is the reorder KEY that lets a
+  #     consumer release updates in contiguous `0..N` order the instant each is
+  #     available (LIVE incremental streaming). When it is `nil`, a consumer falls
+  #     back to arrival-order delivery (best-effort, no reorder guarantee) — see
+  #     `prompt/3` / `prompt_stream/4`.
+  @spec broadcast_update(pid(), String.t(), term(), non_neg_integer(), non_neg_integer() | nil) ::
+          :ok
+  def broadcast_update(conn, session_id, update, rx_seq \\ 0, update_seq \\ nil) do
     ensure_subs_table!()
 
     for {_key, subscriber} <- :ets.lookup(@subs_table, {conn, session_id}) do
-      send(subscriber, {:acp_session_update, session_id, update, rx_seq})
+      send(subscriber, {:acp_session_update, session_id, update, rx_seq, update_seq})
     end
 
     :ok
   rescue
     _ -> :ok
   end
+
+  @doc """
+  Extract the per-session `update_seq` an agent stamped into a
+  `session/update` notification's vendor `_meta` bucket
+  (`_meta["raxol.io"]["update_seq"]`), or `nil` when absent (a non-raxol
+  agent, an unstamped notification, or an opaque notification shape). This
+  is the reorder key `prompt/3` / `prompt_stream/4` use for LIVE
+  incremental in-order release; `nil` triggers their arrival-order
+  fallback.
+  """
+  @spec extract_update_seq(term()) :: non_neg_integer() | nil
+  def extract_update_seq(%{_meta: meta}) when is_map(meta) do
+    case meta do
+      %{"raxol.io" => %{"update_seq" => n}} when is_integer(n) and n >= 0 -> n
+      _ -> nil
+    end
+  end
+
+  def extract_update_seq(_other), do: nil
 
   defp ensure_subs_table! do
     if :ets.whereis(@subs_table) == :undefined do
@@ -356,15 +391,18 @@ defmodule Raxol.AgentClientProtocol.Client do
   by `Connection` itself. Different senders means the subscriber's mailbox
   gives no cross-sender order, and the result can even land before a
   same-turn update that is still in flight. So this function does NOT trust
-  mailbox order: every delivery carries the inbound frame's monotone
-  `rx_seq`, and `Connection` also sends an `{:acp_result_seq, tag, rx_seq}`
-  companion just before the result. This loop buffers updates by `rx_seq`,
-  and on the terminal result (frame seq `R`) drains every same-turn update
-  (all have `rx_seq < R` by wire invariant I3) with a bounded settle pass,
-  then returns them in restored `rx_seq` (wire) order. The bounded settle
-  pass is the completeness bound (an in-flight dispatch task's `send`
-  lands within scheduling latency; the pass resets on each straggler); `R`
-  is the turn-boundary classifier.
+  mailbox order: it reorders on the agent-stamped PER-SESSION `update_seq`
+  (`_meta["raxol.io"]["update_seq"]`, reset to `0` each turn — see
+  `Raxol.AgentClientProtocol.Session`). It tracks a next-expected ordinal,
+  accumulating each update the instant it is contiguous and holding a
+  higher one until the gap fills; on the terminal result a bounded settle
+  pass absorbs any straggler, then it returns the turn's updates in
+  `update_seq` order. When the agent does NOT stamp `update_seq` (a
+  non-raxol agent), it falls back to buffering by the frame's `rx_seq` and
+  returning them in `rx_seq` order at the turn boundary (deterministic,
+  no-drop, but not incremental). Either way no same-turn update is dropped;
+  `Connection` always delivers a terminal result (its protocol timeout is
+  the single authority), so the loop never blocks unbounded on a gap.
 
   Always subscribes and unsubscribes around the call, including on error.
 
@@ -392,7 +430,7 @@ defmodule Raxol.AgentClientProtocol.Client do
              tag,
              timeout_ms
            ) do
-        :ok -> collect_prompt(session_id, tag, [], nil)
+        :ok -> collect_prompt(session_id, tag, new_reorder())
         {:error, _} = err -> err
       end
     after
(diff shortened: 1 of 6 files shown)


DROOdotFOO (commented):

The core mechanic is sound: moving `session/update` (client role) out of a per-notification `async_nolink` task and into synchronous, in-process dispatch on the Connection's own `handle_info` loop genuinely closes the drop/reorder race. Because every inbound frame is processed one at a time by the single Connection GenServer, the update's forwarding `send/2` now happens-before the terminal `session/prompt` response's `send/2`, and both originate from the same sender -- so BEAM's per-sender FIFO carries wire invariant I3 through to the subscriber's mailbox. For the cheap default handler this is correct and complete. The problems below are the cost of running arbitrary handler code on the single shared connection loop.

### HIGH

- **[Saboteur] HIGH** `connection.ex:1042` (`dispatch_session_update_sync/2`) / `connection.ex:1075` (`run_session_update_sync/5`) -- **new cross-session / cross-request head-of-line blocking (reproduced).** The single-sender guarantee is per-CONNECTION, not per-SESSION: the update handler now runs inline on the one Connection message loop, so while ANY `session/update` handler is slow or blocked, the Connection processes no further frames -- for any other session, and for inbound requests (e.g. `fs/read_text_file`). That includes stalling the terminal `session/prompt` results of *other* in-flight turns and denying service to unrelated sessions on the same connection. Pre-fix, a slow update handler only tied up its own isolated task; the Connection kept flowing. Failure scenario: an app whose `session_update/2` renders to a terminal / calls a busy GenServer / does any I/O (the entire point of the callback) wedges every session and every pending inbound request on that connection until it returns; a hung handler wedges the connection permanently. Reproduced: parked session A's update handler, sent session B's update while parked -- B's handler never started (`refute_receive` within 400ms) and only ran after A was released. The PR's own regression test corroborates this ("the Connection process is provably still blocked inside the synchronous update dispatch"). The moduledoc frames this only as "a slow custom `session_update/2` ... can delay it [the loop]" -- understating that it stalls the whole connection. Fix: keep single-sender ordering without running user code on the loop -- have the Connection itself perform the ordered `send` to subscribers for the default forwarding path (a bare ETS-lookup + `send`, no user callback on the loop), and route custom `session_update/2` overrides through a per-session serializer process (one ordered mailbox per session_id) so ordering is preserved per session without one session's handler starving the connection.

### MEDIUM

- **[Saboteur] MEDIUM** `connection.ex:1065` -- **reentrancy regression.** `ctx.conn` is `self()` (the Connection) and the handler now runs *in* the Connection process. Any `session_update/2` override that calls a Connection API on `ctx.conn` -- `Connection.notify/3`, `request/4`, `reply/3` (a plausible reaction: fire a follow-up notification / outbound request on receiving a streamed update) -- now hits `reject_self!/1` (`connection.ex:1305`), which raises `ArgumentError`. It is caught by `run_session_update_sync/5`'s rescue and logged as a `handler_crash`, so the reaction is silently lost. Pre-fix the handler ran in a task (`conn != self()`) and such calls worked. Failure scenario: a handler that emits an ACK/telemetry-notification per update degrades to a logged crash after this change. Fix: document the reentrancy constraint on `session_update/2` for the client role, or (per the HIGH fix) run overrides off the Connection process so `ctx.conn` calls remain legal.

- **[Security Auditor] MEDIUM** `connection.ex:1042` vs `connection.ex:1116` -- **the flood-shed valve is bypassed for `session/update`; prior unbounded-dispatch concern (my #624 MED) is transformed, not fixed.** `dispatch_inbound_notification/2` routes through `start_dispatch_task/2`, which sheds (`notification_shed` telemetry) when the task supervisor is saturated -- a backpressure valve. `dispatch_session_update_sync/2` has no such gate: every `session/update` is processed inline unconditionally. So a peer flooding `session/update` can no longer be shed at all; frames pile in the Connection mailbox and (combined with the HIGH above) starve all other work with zero backpressure. This does remove the per-update task-spawn growth for this one method, but replaces it with an un-sheddable inline-CPU + unbounded-mailbox starvation vector; and the unbounded `async_nolink` dispatch flagged in #624 is untouched for every *other* notification and inbound request (only `max_in`/`max_children` gate those, and the test harness starts `Task.Supervisor` with neither). Fix: gate/meter the inline path too, or drain updates through a bounded per-session serializer with an overflow/shed policy.

### LOW

- **[New Hire] LOW** `connection.ex:1038` -- comment says "`run_session_update_sync/4` wraps the call" but the function (`connection.ex:1075`) and its call site (`connection.ex:1065`) are arity **/5** (`side, handler, dispatchable, ctx, method`). Stale arity in the doc comment for a just-added function.

- **[New Hire] LOW** `connection.ex:128-131` (moduledoc trade-off) and `connection.ex:1035` -- the documented trade-off ("a slow custom `session_update/2` override now runs on this process's own message loop instead of an isolated task, and can delay it") does not state the blast radius: it delays/stalls *every* session and *every* inbound request on the connection, not just "it". A reader sizing the risk will under-estimate it. Fix: state that a blocking handler stalls the entire connection (all sessions + inbound requests), not only the current turn.

### Cross-persona overlaps

The Saboteur HIGH (cross-session HOL stall) and the Security Auditor MEDIUM (no shed / unbounded inline starvation) share one root: **arbitrary handler code now runs synchronously on the single shared Connection loop with no isolation and no backpressure.** Both are resolved by the same fix -- perform the default ordered `send` directly from the Connection (no user code on the loop) and run any custom override through a bounded per-session serializer. Promote this root cause: it is the load-bearing design decision of the PR.

### Verdict: BLOCK

The fix correctly eliminates drop/reorder for the default path, but it does so by introducing a reproduced HIGH head-of-line-blocking regression: any slow or hung `session/update` handler now stalls the terminal results of *other* turns/sessions and all inbound requests on the connection -- exactly the "new head-of-line-blocking stall of the terminal result" class this review treats as HIGH. It also strips the flood-shed valve for `session/update` and adds a reentrancy trap. Single-sender ordering is the right target; achieve it without running user callbacks on the Connection loop (Connection-owned `send` for the default path + per-session serializer for overrides), then this is CLEAN.

DROOdotFOO:
pls review @merklebonsai

DROOdotFOO (commented):

Since last review: 732070c2..94c47a32, 1 new commit

The strategy pivoted from "single-sender (run dispatch synchronously on the Connection loop)" to "keep dispatch async, reorder on the client via a per-frame `rx_seq`." This is the right move and it genuinely retires the three prior blockers -- but the streaming variant pays for determinism with a real behavioral + memory regression.

- **[prior HIGH] cross-session / cross-request head-of-line blocking -- RESOLVED.** `session/update` dispatch is back on the async task path: `dispatch_inbound_notification/3` builds the `Ctx` and hands the handler to `start_dispatch_task/2` (`connection.ex:1057`), the same bounded task path every other notification uses. No user callback runs on the Connection message loop anymore -- the Connection only threads an integer (`rx_seq: rx_seq`, `connection.ex:1049`, was hardcoded `0`). The `dispatch_session_update_sync` / `run_session_update_sync` functions and the `session/update` `cond` special-case are deleted (grep: zero matches on the branch). A slow or hung `session_update/2` handler can no longer stall other sessions or inbound requests. **Core blocker gone.**

- **[prior MEDIUM] reentrancy trap (`ctx.conn` calls hit `reject_self!`) -- RESOLVED.** The handler runs in a dispatch task again, so `ctx.conn != self()`; `reject_self!/1` now appears only in the outbound API functions (`connection.ex:238-289`), never in the notification dispatch path. A handler that reacts to an update by calling `Connection.notify/3` / `request/4` works again.

- **[prior MEDIUM] flood-shed valve bypassed for `session/update` -- RESOLVED.** The inline unconditional path is gone; `session/update` now flows through `start_dispatch_task/2`, which sheds on saturation (`{:error, reason} -> emit_telemetry([:raxol, :acp, :notification_shed], ...)`, `connection.ex:1057-1066`). The un-sheddable inline-CPU + unbounded-mailbox vector is closed.

Also, positively, on the Security Auditor's crafted-seq questions: `rx_seq` is Connection-assigned, not peer-controlled -- `rx_seq = state.rx_seq + 1` per inbound frame in the single `handle_info` (`connection.ex:568`). A malicious peer cannot forge a seq value to poison the sort or wedge delivery, it only influences frame arrival order. There is no contiguity/"wait for seq N" logic, so a skipped or dropped seq cannot wedge the consumer (the drain is time-bounded, not gap-bounded). `rx_seq` is a BEAM bignum, so no wraparound. Those three task-listed hazards are structurally absent -- good.

### HIGH

- **[Saboteur + New Hire] `prompt_stream/4` no longer streams; it buffers the entire turn and replays at the end.** `client.ex` `stream_prompt/5`: on `{:acp_session_update, ...}` it only accumulates `[{seq, update} | buf]`; `on_update` is invoked ONLY in the `{:acp_result, ^tag, outcome}` branch (`buf |> drain_updates() |> order_updates() |> Enum.each(on_update)`). So for a streaming turn -- the literal `agent_message_chunk` LLM-token case this API exists for -- the caller's callback fires ZERO times until the terminal result arrives, then all chunks flush at once. Two independent problems, same code:
  - Functional: this silently changes the contract of a public API named `prompt_STREAM`. The old docstring ("`on_update.(update)` is invoked synchronously ... for each `session/update` as it is received") is now false; every existing caller that renders tokens incrementally (a TUI streaming an agent response) now sees nothing mid-turn and a burst at the end. That defeats the reason `prompt_stream/4` exists as distinct from `prompt/3`.
  - Resource: `buf` holds every update of the turn. For a long streamed response that is O(turn) retained memory with no incremental release, bounded only by the request `timeout_ms`. `prompt_stream/4`'s memory profile went from O(1) to O(turn).
  The moduledoc owns the latency trade ("trades pre-result incremental latency for deterministic ordering") but understates it: it is not a latency trade, it is the removal of streaming from the streaming API, plus unbounded per-turn buffering. Two personas, one root -> HIGH. (`prompt/3`'s accumulate-then-return is unchanged in spirit and not flagged.)

### MEDIUM

- **[Saboteur] the no-drop guarantee is now timing-based (`@update_settle_ms 25`), not deterministic.** `drain_updates/3` waits `after @update_settle_ms -> buf`, i.e. 25 ms, for in-flight dispatch tasks' `send/2` to land, then stops and treats the turn complete. The prior (rejected) single-sender approach had a hard happens-before guarantee: zero drop possibility. This one reintroduces a probabilistic bound. For the fast default forwarding path (ETS lookup + `send`) 25 ms is generous, but under scheduler starvation -- a flood of queued dispatch tasks, GC, a busy node -- a straggler task may not run its `send` within 25 ms of the result, and that update is then silently dropped (the loop has already returned). The completeness "bound" is a magic constant with no derivation in the source beyond "enormous headroom," which holds only under light load. The condition the whole PR is defending against (concurrent dispatch under load) is exactly the condition that can blow the 25 ms window. Saboteur (drop under load) with a New Hire flavor (undocumented magic constant, fragile-by-convention).

- **[Security Auditor] peer can drive an unbounded client-side buffer via `prompt_stream/4`.** The Connection's shed valve gates task SPAWNS, but the client's `buf` is ungated. A peer that streams many `session/update` frames for the in-flight session and withholds the terminal result grows `buf` to O(N) with no incremental release and no client-side cap, until the caller's `timeout_ms` fires. This is the resource face of the HIGH: because `prompt_stream/4` no longer drains incrementally, its client memory is now peer-influenced and unbounded-within-the-turn, where before it was O(1). Same fix as the HIGH (release incrementally) closes it.

### LOW

- **[New Hire] the catch-all `drain_updates` clause absorbs same-session updates with `seq >= result_seq`, justified only by an assumed invariant.** The second `{:acp_session_update, ^session_id, update, seq}` clause (no `seq < result_seq` guard) pulls ANY same-session update out of the mailbox during the 25 ms drain, commented "should be impossible by I3." If a caller issues back-to-back prompts on the SAME session from the SAME process, an early update of the NEXT turn arriving inside the trailing 25 ms window would be stolen into this turn -- removed from the mailbox, mis-attributed, and mis-ordered. Narrow trigger (single-process, same-session, sub-25 ms overlap), so LOW, but it is a correctness hazard resting on an invariant the code cannot enforce.

### Verdict: CONCERNS

The pivot to client-side `rx_seq` reordering genuinely resolves all three prior blockers -- user callbacks no longer run on the shared Connection loop, so the reproduced cross-session head-of-line stall, the reentrancy trap, and the shed-valve bypass are all gone, and the peer cannot forge seq numbers or wedge delivery via a gap. That is real progress and the core blocker is truly fixed. But it introduces a HIGH regression in the process: `prompt_stream/4` stops streaming -- it buffers the whole turn and replays at the end, defeating the streaming API's purpose and turning its memory from O(1) into O(turn). Combined with the timing-based (25 ms) no-drop bound that can silently drop under load and the peer-driven unbounded client buffer, this is not yet CLEAN. Recommended: have `prompt_stream/4` release updates incrementally (emit the settled contiguous prefix as it arrives rather than buffering the entire turn), which restores streaming, caps memory, and removes the peer-flood vector; keep `prompt/3`'s accumulate-then-return as-is.

DROOdotFOO:
have prompt_stream/4 release updates incrementally (emit the settled contiguous prefix as it arrives rather than buffering the entire turn), which restores streaming, caps memory, and removes the peer-flood vector; keep prompt/3's accumulate-then-return as-is.

DROOdotFOO (commented):

Since last review: 94c47a32..d70c0dfd, 1 new commit

The commit adds a per-session `update_seq` (agent-stamped into `_meta["raxol.io"]["update_seq"]`, reset to 0 each turn in `Session.begin_prompt`) and rewrites the client reorder engine to release each update the instant its ordinal is contiguous. This genuinely restores live streaming. But it makes the *primary reorder key peer-controlled* -- the exact property the prior re-review verified as safe for `rx_seq` -- and that reintroduces a no-drop violation and an unbounded-buffer vector through a different door.

- **[prior HIGH] `prompt_stream/4` no longer streams (buffered whole turn, replayed at result) -- RESOLVED for the stamped in-order path.** `stream_prompt/5` now threads a real `on_update` sink into `deliver_update/5` (`client.ex:644-652`); a contiguous update (`update_seq == next`) hits `emit_update/3`'s function-arg clause which calls `on_update.(update)` immediately, then `cascade_release/2` flushes buffered successors (`client.ex:515-549`). So for an in-order stamped stream `on_update` fires per chunk again, live, not once at the terminal result. The `agent_message_chunk` token case this API exists for streams incrementally. **Core streaming regression fixed** (see caveats below for the memory face and the unstamped path).

- **[confirmed still absent] rx_seq wraparound / gap-wedge -- still absent.** `rx_seq` remains Connection-assigned (`state.rx_seq + 1`), BEAM bignum, no contiguity wait -- unchanged and safe. `error`-outcome now also drains (`stream_prompt` handles both outcomes via `settle_updates`, `client.ex:657-658`), a small improvement over the prior error path.

### HIGH

- **[Security Auditor + Saboteur] the reorder key `update_seq` is peer-supplied and unvalidated -- a buggy or malicious agent silently drops updates (violating the PR's headline no-drop) and can drive an unbounded client buffer.** `extract_update_seq/1` reads `_meta["raxol.io"]["update_seq"]` straight off the inbound notification and accepts ANY non-negative integer with no bound or turn-relative check (`client.ex:304-308`). From the client's trust boundary the agent IS the remote peer. Three diff-grounded failures, same root:
  - **Silent drop (no-drop violated).** Two updates carrying the same `update_seq > next` collide: `deliver_update/5` does `Map.put(ro.buf, update_seq, update)` (`client.ex:522`) -- the second overwrites the first, which is lost with no telemetry. A plausible *buggy* agent that fails to increment (always stamps `0`): the first update releases (`0 == next`, next->1), every subsequent `0` is `< next` and dropped as a "duplicate" (`client.ex:524-526`). Result: only the FIRST update of the turn is ever delivered, the rest vanish silently. That is exactly the drop class this PR ("no drop/reorder") exists to eliminate, now re-openable by the peer.
  - **Unbounded buffer + streaming defeat (prior MEDIUM #2, now worse -- peer-steerable).** A peer that withholds the low ordinal (send `1,2,3,...` and never `0`) parks every update in `buf` (`update_seq > next`), never releasing live, unbounded within the turn until the result/timeout. Where the prior review noted "a malicious peer cannot forge a seq value to poison the sort or wedge delivery," that no longer holds for the key that now drives ordering.
  - The moduledoc/comments consistently say "the AGENT stamps," treating `update_seq` as a trusted contiguous `0..N`. It is fragile-by-convention and, at the client boundary, attacker-influenced. Two personas, one root -> HIGH. Fix: validate `update_seq` against `next` (reject/telemeter a re-used or absurdly-high ordinal instead of `Map.put`-overwriting), cap the out-of-order `buf`, and treat a collision as a logged anomaly, not a silent overwrite.

### MEDIUM

- **[Saboteur] `prompt_stream/4` still retains O(turn) memory -- the memory face of the prior HIGH is only half-fixed.** `emit_update/3` unconditionally conses every released update onto `ro.released` (`client.ex:544`, `:548`), including the `on_update` clause used by streaming. `stream_prompt` then DISCARDS it (`_ = settle_updates(...)`, `client.ex:658`). So latency is now O(1) (on_update fires live) but memory is still O(turn): the streaming variant holds a reference to every update of the turn for the turn's whole duration, for nothing. The "O(1) streaming" goal is not met. Fix: skip the `released` accumulation when `on_update` is a real sink (only `prompt/3` needs the returned list).

- **[Saboteur / New Hire] the no-drop settle bound is still timing-based (`@update_settle_ms 25`) -- prior MEDIUM #1 UNCHANGED.** `settle_updates/3` waits `after @update_settle_ms -> flush_buffers(...)` (`client.ex:450`, `:564-565`). A straggler dispatch task whose `send/2` lands >25 ms after the terminal result (scheduler starvation, GC, flood of queued tasks -- precisely the concurrent-under-load condition this PR defends against) is silently dropped; the loop has already flushed and returned. Still a magic constant justified only by "enormous headroom," which holds under light load only. The commit did not touch this.

- **[Saboteur] per-turn reset-to-0 makes cross-turn stragglers ambiguous (regression vs. global rx_seq).** `Session` resets `update_seq` to 0 at each `begin_prompt`, so ordinals are NOT unique across turns. A straggling update from a later turn (`update_seq` 0) arriving in an earlier turn's 25 ms settle window is `< next` and dropped as a "duplicate" (`client.ex:524-526`), or collides in `buf`. The prior global-monotone `rx_seq` could not alias this way. Narrow trigger (two subscribers / overlapping turns on one session, since I12 blocks a second concurrent prompt from one caller), so bounded -- but it is a correctness hazard the reset introduces that the old key did not have.

### LOW

- **[New Hire] catch-all drain still steals same-session updates by convention.** `settle_updates/3`'s `{:acp_session_update, ^session_id, update, rx_seq, update_seq}` clause (`client.ex:558`) pulls ANY same-session update from the mailbox during the 25 ms window, now routed through the seq machine. Combined with the reset-to-0 above, an early next-turn update inside the trailing window is mis-attributed/mis-ordered into this turn. Same narrow trigger; carried over from the prior LOW.

### Cross-persona overlap

The HIGH (peer-controlled `update_seq`) and the "unbounded buffer" half of prior MEDIUM #2 share one root: **the ordering key that drives release and de-dup is now supplied by the untrusted peer and validated only for `is_integer and >= 0`.** The same fix -- validate `update_seq` against the expected `next`, bound `buf`, and telemeter collisions/gaps instead of silently overwriting/dropping -- closes both the no-drop violation and the buffer-exhaustion vector.

### Verdict: CONCERNS

The streaming regression is genuinely fixed -- `on_update` fires per chunk again, live, for the in-order stamped path (`deliver_update` -> `emit_update`/`on_update`, `client.ex:515-549`, `:644-652`) -- and none of the three original BLOCK guarantees regressed (dispatch stays on the async task path; no user callback runs on the Connection loop; the shed valve and reentrancy fixes are intact). But the new per-session `update_seq` is a peer-supplied, unvalidated reorder key: a buggy or hostile agent can silently drop updates (via re-used ordinals overwriting `buf`, or a stuck counter) -- re-opening the very no-drop guarantee this PR is named for -- and can park an unbounded per-turn buffer by withholding a low ordinal. Add the still-timing-based 25 ms drop bound and the residual O(turn) memory in the streaming path, and it is not yet CLEAN. Recommended: validate `update_seq` against `next` (log/telemeter re-use and gaps rather than overwrite/silently-drop), bound `buf`, and stop accumulating `released` when `on_update` is a live sink.

DROOdotFOO (commented):
## Design escalation: the no-drop guarantee keeps relocating, not closing

This is a step back from the per-round findings to name a pattern across all three
rounds of this PR:

- `732070c2` synchronous dispatch on the single-sender Connection loop -> fixed order
  + no-drop, but ran user handler code on the shared loop (head-of-line blocking,
  reentrancy death, flood-shed bypass). BLOCK.
- `94c47a32` client-side `rx_seq` reordering -> fixed the loop problems, but broke
  streaming (buffered the whole turn) and added a timing-based drop bound + unbounded
  client buffer. CONCERNS.
- `d70c0dfd` per-session `update_seq` -> fixed streaming, but the ordering key is now
  **peer-supplied and unvalidated** (`extract_update_seq/1` takes any non-neg integer
  from the peer's `_meta`): a reused ordinal silently overwrites the buffered update, a
  stuck counter delivers only the first update and drops the rest with no telemetry, a
  withheld low ordinal parks an unbounded per-turn buffer. CONCERNS.

Each fix resolved the previous round's headline finding and moved the no-drop + in-order
guarantee onto a new mechanism with a fresh hole. The root is a design decision that has
never been made explicitly: **what assigns the ordering key, and is that assigner inside
or outside the trust boundary?** Round 3's answer -- the peer controls the key -- means
no client-side buffer logic can restore the guarantee, because a buggy/hostile peer
defeats it directly.

Rather than land a fourth patch that likely relocates the hole again, I've proposed the
delivery-ordering contract as an ADR so we settle it once:

**ADR-0030: ACP session/update delivery ordering contract -> #641**

It states five clauses (receiver-assigned key; bounded reorder buffer with
drop-with-telemetry; bounded gap resolution; streaming preserved; cross-turn
monotonicity) and the invariants to pin with property/adversarial tests. It adopts the
same bounded/observable/deterministic-under-overload shape ADR-0013 already established
for the terminal hot path, so this converges with existing repo patterns instead of
diverging.

Suggested path: align this PR to the ADR-0030 contract (receiver-stamped ordinal +
bounded buffer + gap telemetry) so the four remaining findings collapse into "does it
satisfy the five clauses" rather than four independent bugs. Happy to review against the
contract on the next push.

DROOdotFOO:
see the other ADR PR and collaborate there

merklebonsai:
Two updates on the latest push (`d70c0dfda`):

**Incremental release — done, matches the request.** `prompt_stream/4` now releases the settled contiguous prefix **as it arrives** (live), not buffered-to-turn-boundary: the agent stamps a per-session `update_seq` (0,1,2… per turn) in `_meta["raxol.io"]`, and the consumer releases each update the moment it's contiguous, holding only out-of-order stragglers until their gap fills. This restores streaming, caps buffer memory to the outstanding gap, and there's no whole-turn accumulation to flood. `prompt/3` kept as accumulate-then-return, as asked. Non-raxol agents (no `update_seq`) degrade gracefully to `rx_seq`-ordered delivery. 801 tests green locally; red-first witnesses for live-in-order, gap-wait, two-session interleave, and the fallback.

**The macos-latest CI failure is orthogonal.** The failing test is `Raxol.Core.Runtime.Events.DispatcherEmitTest` ("process_command_result publishes an ephemeral :command_result event") — a main-app Dispatcher test in a different subsystem this ACP-package change does not touch. Linux + all other checks are green; it's the pre-existing macos flake class, not this diff. Happy to re-run it, but no code here is implicated.

DROOdotFOO:
Thanks -- two concessions and one hold.

**CI: fully conceded, you're right.** I checked the diff file set: #640 touches only
`raxol_acp` package files, nothing under the dispatcher/events/runtime subsystem, so
`DispatcherEmitTest` cannot be implicated by this change. All checks are now green (25
success / 2 skipped / 0 failure), consistent with the known macOS timing-flake class
(cf. #637). Not this diff. Dropped.

**Streaming: agreed, confirmed.** The incremental release is real -- `on_update` fires
on the settled contiguous prefix as it arrives (client.ex `emit_update/3` +
`cascade_release/2`), memory is the outstanding gap not the whole turn, and `prompt/3`
stays accumulate-then-return. That was the HIGH from the prior round and it's resolved.

**The hold is the one the escalation is about, and your own description restates it.**
"The agent stamps a per-session `update_seq` (0,1,2...) in `_meta["raxol.io"]`" -- that
is the peer supplying the ordering key. "Caps buffer memory to the outstanding gap" and
"no whole-turn accumulation" are true *for a well-behaved raxol agent*, but the guarantee
is conditional on the counterparty behaving, and that's exactly what a delivery guarantee
can't be. Two concrete cases still stand in `d70c0dfd` (unchanged since the last review):

1. **Stuck counter** -- an agent that stamps `update_seq: 0` on every update (bug, or a
   non-conforming implementation that sets the field but never increments): each update
   is "contiguous at 0", so `Map.put(buf, 0, update)` overwrites, and only the first (or
   last) survives -- the rest are dropped with no telemetry. This is worse than the
   `rx_seq` fallback, because the presence of the field opts out of it.
2. **Withheld low ordinal** -- a peer that emits 1,2,3,... but never 0. "Hold stragglers
   until the gap fills" becomes an unbounded park, because the gap never fills. There's no
   watermark and no fail-the-turn, so it's a memory-growth / stuck-turn primitive.

Neither requires malice -- a buggy or partial peer implementation triggers both. The
graceful `rx_seq` degrade only helps agents that omit `update_seq` entirely; it doesn't
help an agent that supplies a *wrong* one.

This is why I raised it as a design question rather than a fourth line-item: is
"peer-supplied key with graceful degrade" the intended contract, or must the ordinal be
receiver-assigned? ADR-0030 (#641) writes up the second option -- a receiver-stamped
contiguous per-session counter (stamped at the demux point before fan-out), a bounded
buffer that fails-the-turn instead of parking, and a reconciliation with the
reattach/replay contract in #569/#586 so replay re-delivers already-stamped ordinals
rather than renumbering. If you think peer-supplied-plus-degrade is actually sufficient,
that's a legitimate position -- let's settle it on the ADR, because whichever way it goes,
the *next* implementation should build against a stated contract instead of rediscovering
the hole.

merklebonsai:
Superseded by #644 — the transport-ordering redesign per ADR-0030 (#641). The approach here (agent-stamped per-session `update_seq` in `_meta`) had the peer-supplied-ordering-key flaw the ADR rejects; #644 moves ordering receiver-side (Connection stamps at demux, single-sender direct delivery), keeps it off the ACP wire, and satisfies ADR clauses 1/3/4/5. Recommend closing this in favor of #644.

DROOdotFOO:
Superseded by https://github.com/DROOdotFOO/raxol/pull/644

PR #641, opened 2026-07-17
docs(adr): 0030 ACP session/update delivery ordering contract

DROOdotFOO:
## Summary

Proposes ADR-0030: a stated delivery-ordering contract for ACP `session/update`
notifications. Raised from three consecutive review rounds on #640.

Across #640's three rounds (`732070c2` synchronous single-sender loop -> `94c47a32`
client-side `rx_seq` reordering -> `d70c0dfd` per-session `update_seq`), each fix
resolved the prior round's headline finding but relocated the no-drop + in-order
guarantee onto a new mechanism with a fresh hole. Round 3 lands on the worst answer:
the ordering key (`update_seq`) is **peer-supplied and unvalidated**, so a buggy or
malicious peer can silently drop, reorder, or unbounded-buffer updates. A guarantee
whose correctness depends on a value the counterparty controls is not a guarantee.

This is the same problem class ADR-0013 (event-dispatch backpressure) already solved
for the terminal hot path. ADR-0030 adopts that shape rather than reinventing it a
fourth time.

## The contract (five clauses)

1. Receiver-assigned ordering key (never read from peer `_meta`).
2. Bounded reorder buffer with explicit drop-with-telemetry (never silent, never unbounded).
3. Bounded gap resolution (gap-based, not wall-clock-based settle window).
4. Streaming preserved (`on_update` fires per chunk, O(1) memory).
5. Cross-turn monotonicity (no per-turn reset that aliases stragglers).

Plus the invariants to pin with property/adversarial tests.

## Why an ADR and not another patch on #640

The trust-boundary decision (what assigns the ordering key, inside or outside the
trust boundary) has never been made explicitly. Three rounds of local patches each
moved the hole rather than closing it. This ADR makes the decision once so the #640
findings collapse into "does it satisfy the five clauses" instead of four independent
bugs, and the next implementation round has a target to build against.

## Manual testing

- Docs-only change (one new ADR + README index rows). No code, no tests affected.
- `docs/adr/README.md` table and Orchestration list updated to reference 0030.

Refs: #640

### DIFF
--- a/docs/adr/0030-acp-session-update-delivery-ordering.md
+++ b/docs/adr/0030-acp-session-update-delivery-ordering.md
@@ -4,7 +4,11 @@
 
 Proposed, 2026-07-18. Raised from three consecutive review rounds on PR #640
 (`fix(acp): single-sender session/update delivery`). No code has landed on master;
-this ADR exists to settle the contract before a fourth patch round.
+this ADR exists to settle the contract before a fourth patch round. Revised the same
+day after an adversarial self-review (see #641): clause 1 is now decided (not an
+either/or), the receiver-assigned principle is applied recursively to the turn
+namespace, the wall-clock/overflow contradiction in the gap clause is removed, and the
+contract is reconciled with the reattach/replay contract (#569/#586).
 
 ## Context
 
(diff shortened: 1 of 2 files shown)


DROOdotFOO:
## Adversarial self-review (automated) + revision

Before asking anyone to ratify this, I ran the three-persona adversarial reviewer
against the ADR itself. It found real holes; commit `03de2efb` (pushed on top of
`2ae4157e`) addresses them. Findings and their disposition below.

### HIGH

- **[Saboteur + Security -- promoted] Receiver-assigned principle applied in only one of
  three places.** Clause 1 pulled the top-level ordinal inside the trust boundary, but
  clause 5's per-turn namespace was left free to be derived from a peer-supplied
  session/job/turn id -- moving the same defect up one level (reuse a turn id, re-alias
  stragglers). RESOLVED: clause 5 is now "recursive receiver-assignment" -- the namespace
  is receiver-assigned too, never peer-derived.

- **[Saboteur] Clause 1 could contradict the frozen reattach/replay contract (#569/#586).**
  "Server-side arrival order" is a single sequence only within one connection; replay
  arrives on a NEW connection, so arrival-order silently renumbers replayed updates.
  RESOLVED: arrival-order is now explicitly rejected as the key; the decided key is a
  monotonic per-session counter that survives reconnects, and a new "Reconciliation with
  reattach/replay" section plus a replay-idempotence invariant make replay re-deliver
  already-stamped ordinals (no renumbering, no duplicates).

### MEDIUM

- **[New Hire] Clause 1 left the core mechanism as an unresolved either/or.** An ADR must
  decide. RESOLVED: clause 1 is now decided (per-session counter), and specifies the stamp
  is applied at the single demux point BEFORE async fan-out -- stamping inside a per-session
  task would reintroduce the round-2 race.

- **[Saboteur] Internal contradiction: clause 3 mandated a "bounded interval" (a wall-clock
  timer) while its own parenthetical rejected wall-clock settle windows.** RESOLVED: clause 3
  is now overflow-bounded, not timer-bounded -- no settle window at all; a gap is bounded by
  the buffer watermark, and resolves only on (a) the missing ordinal arriving or (b)
  connection-loss/watermark -> fail-the-turn. Added a "no timer in the guarantee" invariant.

- **[Security] "deliver-with-gap-marker" handed the peer a turn-kill / fast-path-DoS
  primitive.** RESOLVED two ways: (1) because ordinals are now receiver-stamped and
  contiguous, the peer cannot manufacture a gap at all (gaps come only from real connection
  loss); (2) fail-the-turn is now the only lossless resolution, and gap-markers are permitted
  only for update types explicitly marked coalescible/idempotent -- never for streamed content.

### LOW

- **[New Hire] Under-specified vs. its own template (ADR-0013).** RESOLVED: added a concrete
  telemetry event `[:raxol, :acp, :delivery]` with metadata shape, a default watermark
  (1_024/turn) with rationale, and explicit definitions of "turn", "update", and "straggler".

### Verdict (pre-revision): CONCERNS -> (post-revision `03de2efb`): the two HIGHs and three
MEDIUMs are addressed in-document. The central thesis was sound; the contract is now
internally consistent, recursively trust-bounded, timer-free in its guarantee, and
reconciled with the replay contract. Remaining LOWs are implementation-detail choices for
the executing PR.

merklebonsai:
Implementation landed: **#644** satisfies ADR-0030 clauses **1 (receiver-assigned key at demux, `make_ref` turn token), 3 (overflow-bounded/fail-the-turn, no timer), 4 (streaming preserved), 5 (recursive receiver-assignment)** on the primary `prompt`/`prompt_stream` path — which is single-sender-FIFO, so occupancy is 0 (no buffer needed there). **Clause 2** (bounded/observable `Delivery.Reorder` for the multi-subscriber broadcast path) is a focused follow-up; that path is unchanged from master, so #644 is a strict improvement. Also confirmed the ADR's transport-layer framing: ordering is a Transport contract (T-ORD), the receiver-side key is separate + app-layer — #644 documents both. 830 green, red-first witnesses.

DROOdotFOO (commented):
Strong ADR. Clause 1 (receiver-stamped key before fan-out) + clause 5 (recursive to the namespace) is the right diagnosis; the ADR-0013 convergence and reattach reconciliation hold up. Accept with edits.

**1. No reclamation rule for the per-session counter registry.** Clause 5 mandates a per-session monotonic space "never reset within a session," and the Bounded-memory invariant (L138) only bounds the *reorder buffer*, not this counter map. A long-lived client keeps one counter per session forever. This is exactly the `oot_seq` leak in #644 (its own comment tags it "clause 5's out-of-turn namespace... never reset") — so the leak is a spec gap, not an impl deviation. Add: counter/namespace is dropped on session end.

**2. "Must satisfy all five clauses" doesn't match how #644 lands.** #644 satisfies clause 2 only on the single-sender path (occupancy 0) and defers the real bounded buffer to a follow-up. Reasonable, but the flat "all five" doesn't authorize it. Split clauses 2/3 into single-sender/FIFO (buffer-free by construction) vs broadcast/multi-subscriber (bounded buffer mandatory).

**3. Status + merge order.** Still "Proposed"/unmerged while #644 cites it 57×. Land this before/with #644 and flip to "Accepted" so the references resolve.

**4. Minor — clause 3 `:gap`-marker may be unreachable.** Does any `session/update` type actually carry a coalescible/idempotent marker today? If not, say so — otherwise the escape hatch reads as wired when it isn't.

PR #644, opened 2026-07-18
feat(acp): transport-ordering layer — receiver-side session/update delivery (supersedes #640)

merklebonsai:
Fixes the shipped `session/update` drop/reorder bug the right way, per ADR-0030 (#641) — **ordering is a transport property, resolved receiver-side; nothing goes on the ACP wire.** Supersedes #640 (which put a peer-supplied ordering key on the wire — the trust-boundary violation Drew's ADR rejects).

## The fix
`session/update` over an ordered transport already arrives in send order; the bug was **our** client-side fan-out (each notification dispatched in its own Task, delivered from different processes → scrambled). This makes the Connection deliver to the turn owner **itself** at the single demux point, in order.

- **Transport contract (`T-ORD`/`T-REL`/`T-DUP`/`T-TERM`)** mandated on the `Transport` behaviour — no `ordered?` flag; an unordered carrier earns it internally. Sequenced transports (stdio/Paired/WS) satisfy it unchanged (proof chain in the doc). F7 conformance test + falsifier.
- **`Transport.Envelope` + `Transport.Reassembly`** — pure modules for a future unordered carrier (transport-`tseq`, stripped before delivery, **never on the ACP wire**; overflow-bounded, no timer). Design-stub, addable later without touching Connection. F8 tests incl. a permutation property.
- **Connection receiver-side ordering** — a receiver-minted `make_ref()` turn token (ADR clause 5), a contiguous per-turn ordinal stamped at demux, single-sender direct delivery to the owner (`{:acp_turn_update}` / `{:acp_turn_end, count}` / `{:acp_result}` — no user code on the loop). Primary `prompt`/`prompt_stream` path is single-sender-FIFO ⇒ **no reorder buffer, occupancy 0**. Streaming preserved (live `on_update`, O(1)). **Fail-the-turn** on a delivery gap (`{:error, {:delivery_gap,_}}` + `:fail` telemetry) — **no timers anywhere**.
- Agent `_meta` returns to opaque consolidation data (never load-bearing for order).

## ADR-0030 compliance
Clauses **1 (receiver-assigned key), 3 (overflow-bounded/fail-the-turn, no timer), 4 (streaming), 5 (recursive receiver-assignment)** satisfied on the primary path. **Clause 2** (bounded/observable reorder buffer for the multi-subscriber *broadcast* path) is a focused **follow-up** — that path is unchanged from master (always async fan-out), so this PR is a strict improvement; the follow-up adds `Delivery.Reorder` + a reattach replay-order test.

## Verification
830 tests green (compile `--warnings-as-errors` clean). Red-first witnesses for drop/reorder/live/fail-the-turn (confirmed red against master's fan-out via `git stash`). New invariant families `F7`/`F8`/`D1..D9` registered in `test/INVARIANTS.md`. A pre-existing conformance-harness race (acpx-007, amplified by the scheduling shift) hardened with a bounded accumulator quiesce (harness-sync only).

Built design-first (Fable) + a Sonnet fleet against a work-unit partition; design doc drove the units.

### DIFF
--- a/packages/raxol_agent_client_protocol/lib/raxol/agent_client_protocol/client.ex
+++ b/packages/raxol_agent_client_protocol/lib/raxol/agent_client_protocol/client.ex
@@ -92,10 +92,9 @@ defmodule Raxol.AgentClientProtocol.Client do
   `{:acp_session_update, session_id, update}` -- `update` is always an
   already-decoded `Raxol.AgentClientProtocol.Schema.SessionUpdate.t()` (see
   `decode_update/1`'s doc for why an undecodable variant never reaches
-  this path at all).
-
-  `prompt/3` and `prompt_stream/4` build on `subscribe/3` to give a
-  synchronous-feeling API over the inherently asynchronous protocol:
+  this path at all). `subscribe/3` remains useful for a custom
+  `session_update/2` override that wants a broadcast fan-out, but it is
+  NOT how `prompt/3`/`prompt_stream/4` get their updates (see below).
 
       {:ok, {updates, response}} =
         Client.prompt(client_conn, PromptRequest.new(session_id, prompt_blocks))
@@ -105,11 +104,23 @@ defmodule Raxol.AgentClientProtocol.Client do
           IO.inspect(update, label: "update")
         end)
 
-  Both drive the request through `Connection.async_request/6` (not the
-  blocking `Connection.request/4` wrapper) specifically so the calling
-  process can `receive` session/update deliveries *and* the terminal
-  result in one loop -- see their docs for the exact mechanics and the
-  ordering caveat.
+  `prompt/3` and `prompt_stream/4` drive the request through
+  `Connection.async_request/6` and then consume the **direct turn-delivery
+  channel** (`Raxol.AgentClientProtocol.Delivery`'s moduledoc):
+  `Connection` opens a turn at submit time and, for every `session/update`
+  belonging to that turn, sends the calling process
+  `{:acp_turn_update, tag, ordinal, notification}` directly, itself --
+  the same single-sender operation it uses for the terminal
+  `{:acp_result, tag, outcome}`. Because one process sends every update,
+  the turn-end companion, and the result, BEAM's pairwise FIFO guarantees
+  they arrive in wire order with NO reorder buffer, NO settle timer, and a
+  synchronous, deterministic failure (`{:error, {:delivery_gap, _}}`) if
+  the turn-end count and the delivered-update count ever disagree. This
+  is the actual fix for the drop/reorder bug the previous
+  `subscribe/3`-based implementation of these two functions had (a
+  separate handler task per notification vs. the Connection's own
+  terminal-result send -- two different senders, no cross-sender FIFO
+  guarantee).
 
   ## `fs_sandbox`: a working filesystem client with zero callbacks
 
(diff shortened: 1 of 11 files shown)


DROOdotFOO (commented):
Read through the ordering rework. The core is right, and it's right in the way #640 wasn't: the turn token is a receiver-minted `make_ref()`, so a peer can replay or reuse a session/request id all it likes and still can't splice a stale turn's straggler into a live one, because the tokens differ by construction. Single-sender FIFO at the demux point, fail-the-turn on a gap, no timers. Good foundation.

A few things I'd want handled before this merges.

### The one that actually worries me: `oot_seq` never shrinks

`stamp_and_deliver_update/3` (connection.ex:1203) stamps an out-of-turn `session/update` by bumping `oot_seq[session_id]`, and nothing ever removes that entry. It's cleared only when the whole connection closes (connection.ex:1397). `route_session_cancel` doesn't touch it, and there's no per-session cleanup anywhere. Two problems fall out of that:

- It leaks in normal use, not just under attack. A long-lived client accumulates one permanent entry for every distinct session it ever sees an out-of-turn update for, and never gives them back.
- `session_id` comes off the wire, and the nil-turn branch stamps *any* id, including sessions this client never opened. A peer streaming `session/update` with unique sessionIds grows the map for free, and it slips past the flood guard: the `oot_seq` write happens before `dispatch_notification_task` sheds on task_sup saturation, so the backpressure that's already there doesn't cap it.

Each entry is small, so it's a slow leak rather than an instant OOM, but it's genuinely unbounded with zero reclamation. Prune `oot_seq` (and the matching `turns`/`turn_ids`) when a session ends or cancels.

### Replaced-turn `turn_end` contradicts what `Delivery` promises

The `Delivery` docstring (delivery.ex:20) says `{:acp_turn_end, count}` goes out "immediately before the terminal `{:acp_result}`." But when a client prompts the same session twice while the first turn is still open, `close_replaced_turn/2` fires the old owner's `turn_end` at replacement time, and that old prompt's `{:acp_result}` doesn't land until its response or timeout resolves, which can be much later. A consumer that treats `turn_end` as "result is next" gets surprised. The comment leans on agent-side invariant I12, but this path is the *client* calling prompt twice, so I12 doesn't cover it. Either close the old `out_id`'s pending entry at replacement, or carve the case out in the doc.

### The design doc and ADR aren't in the repo

connection.ex cites §4.2 through §4.6, ADR-0030 clauses, IC-N, I12 (57 references in that one file), but neither TRANSPORT_ORDERING_DESIGN.md nor ADR-0030 is committed. Only test/INVARIANTS.md made it in. Whoever next edits `maybe_open_turn` or `close_turn_for` can't open the spec the comments treat as law. Land the doc with the code, or inline the two or three invariants that actually gate correctness.

### Smaller stuff

- The turn owner isn't monitored. The carrier and the adopter are (connection.ex:369, 494); the owner isn't. Sending to a dead owner is a harmless no-op, but if an owner dies mid-turn on an `:infinity`-timeout prompt, the turn and its `pending_out` entry sit around until the connection closes. Bounded for finite timeouts. Worth a monitor plus close-on-`:DOWN`.
- The three delivery messages are bare positional tuples. They're documented in `Delivery`, which is fine for now, but a struct would keep Connection and Client from drifting on field order.

### Verdict

Merge after the `oot_seq` fix. Nothing here is a showstopper, and there's no crash path: I checked, the decode always hands you a `SessionNotification` struct with `session_id` enforced, so the demux match can't blow up. But the unbounded `oot_seq` has a real resource footprint, and the `turn_end`/result ordering needs either a fix or an honest doc note. The design doc should ride along so those 57 references resolve.

DROOdotFOO:
Re the `oot_seq` leak above: it traces to a gap in ADR-0030 (#641), not an impl deviation — clause 5 mandates the never-reset per-session counter but the ADR never says to reclaim it on session end. Flagged there too; the fix probably wants to land in the spec + here together.

PR #650, opened 2026-07-19
feat(acp,mcp): steer extension and overlay-aware tool derivation

merklebonsai:
- raxol_agent_client_protocol: a _raxol/session.steer vendor extension —
  SteerRequest/Response schemas, a method-table entry, and an injected
  SteerAdapter seam (Unsupported by default, byte-identical to the frozen
  v2 wire). Purely additive, inert unless wired.
- raxol_mcp: TreeWalker and StructuredScreenshot learn :absolute_layer
  overlay traversal and single-map :children, so tools derive from
  overlay Components hosted over the transcript.
- Lockfile refresh for gateway/symphony/telegram (CVE-patched releases).

Verification (isolated worktree off origin/master, TMPDIR=/tmp SKIP_TERMBOX2_TESTS=true):
- raxol_agent_client_protocol: `mix deps.get` clean, `mix compile --warnings-as-errors` clean, `mix test` — 807 passed (16 properties, 791 tests), 0 failures.
- raxol_mcp: `mix deps.get` clean, `mix compile --warnings-as-errors` clean, `mix test` — 302 passed (31 properties, 271 tests), 0 failures.

Correction to the merge plan: it anticipated 3 known-red tests in test/session_steer_test.exs pending a later Router.result_marker wiring (acp-turn-runner). That wiring is already present on this branch — method_table.ex already registers `_raxol/session.steer` -> `SteerResponse` — so all steer tests (including the result_marker round-trip) pass now. No known-red tests remain on this branch.


merklebonsai:

merklebonsai:
Fix 1 (18bd6d8c6): `SteerResponse.encode_result` now stringifies non-scalar CAS turn tokens (e.g. the `{:steered, cur, N}` tuple `Raxol.Agent.Steer.swap/1` issues after every accept) instead of handing them raw to `Jason.encode!`, which raised `Protocol.UndefinedError` on an ordinary second steer within one live turn; regression test drives the real steer-then-stale sequence end to end plus a focused tuple-actual unit test, both confirmed to fail against the pre-fix code.
Fix 2 (3c2cc5528): corrected the false `ToolSynchronizer` startup comment claiming `Raxol.Headless` forces a synchronous engine render — it doesn't; cold-start tool derivation actually races the telemetry-handler attach against the engine's own first render (comment-only, no behavior change).

DROOdotFOO:

3-persona adversarial pass + droo-stack lens. HIGH reproduced from the diff before surviving; cross-persona overlaps promoted.

### HIGH
- **`ext/schema.ex:133` — SteerRequest round-trip asymmetry.** `SteerRequest.from_json` rejects nil/absent `text` (`Schema.str(nil) -> :error` falls through to `{:error, {:invalid_steer_request, _}}`), while `@type`, `@enforce_keys`, `new/N`, and `to_json` all treat `text` as optional. So `to_json` is not its own inverse: a legitimate text-less steer the module itself serializes is dropped as malformed on decode and answered with a JSON-RPC error. _(Saboteur + New Hire, promoted MEDIUM→HIGH)_

### MEDIUM
- **`ext/schema.ex:362`** — `json_turn_token/1` uses `inspect/1` for non-scalar tokens, making `SteerResponse` tuple-token round-trips lossy **and** echoing the internal `{:steered, cur, System.unique_integer}` tuple (node-global counter) to the remote client. _(New Hire + Security Auditor, promoted LOW→MEDIUM)_
- **`raxol_mcp/.../tree_walker.ex:151`** — overlay/children traversal (children-as-list-or-map + `absolute_layer` `flow_child`/`overlays` descent) is copy-pasted between `tree_walker.ex` and `structured_screenshot.ex`; patching one leaves the sibling package silently disagreeing → tools for widgets not visible in the screenshot resource.

### LOW
- `steer_adapter.ex:672` — session-lifetime dedup index keyed by `client_msg_id` is never cleared on `turn_ended`; a long-lived durable session with a fresh clientMsgId per steer grows `seen` unbounded.
- `ext/schema.ex:253` — `SteerRequest.to_json` emits `expectedTurnId`/`clientMsgId` (typed `term()`) unguarded; a tuple value raises `Protocol.UndefinedError` in `Jason.encode!`, unlike the hardened `SteerResponse`.
- `ext/schema.ex:266` — `from_json` accepts any non-nil term for `expectedTurnId` and any term for `clientMsgId` with no shape/size bound, threading unvalidated network input into long-lived per-session CAS/dedup state.

### droo-stack
- `test/session_steer_test.exs:466` — `do_wait_until` busy-polls `:sys.get_state` with `Process.sleep(5)` against a 2s deadline instead of `assert_receive`/monitor (wall-clock flake class the project has already been bitten by).
- `test/session_steer_test.exs:46` — test-double `Adapter.resolve/2` dispatches via a 4-branch `cond` + `Map.has_key?` ladder instead of guarded multi-clause heads.
- `tree_walker.ex:204` — `resolve_module/3` uses in-body `case node do` where two guarded function heads read cleaner.

### To clear CONCERNS
1. Make `from_json` accept nil/absent `text` (treat as optional exactly like `to_json` emits it), so encode/decode are inverses.
2. Encode tuple turn tokens through an explicit, lossless, non-`inspect` representation and stop leaking the internal `System.unique_integer` counter to the peer.
3. Extract the overlay-traversal logic into one shared helper used by both `tree_walker` and `structured_screenshot`.
4. Bound `expectedTurnId`/`clientMsgId` shape+size on decode; clear the dedup index on `turn_ended`.

DROOdotFOO:
## HIGH addressed — `4a65dc088`

`SteerRequest.from_json` decoded `text` with `Schema.str`, which rejects nil — but `text` is optional in the `@type`, `new/4`, and `to_json` (all treat nil as valid). So a text-less steer that `to_json` itself emits was rejected as `{:invalid_steer_request, _}` on decode, i.e. `to_json` was not its own inverse.

Fix: added `Schema.opt_str/1` (nil/absent -> `{:ok, nil}`, binary -> `{:ok, v}`, other -> `:error`) and used it for `text`. Now:
- nil-text steers round-trip (`from_json(to_json(req)) == req`);
- an absent `text` key decodes to `nil`;
- a non-string `text` is still rejected.

+3 regression tests, all green.

Note: 2 pre-existing failures in `session_steer_test.exs` (`Router.decode`/`result_marker` returning `:ext_request`/`:ext` instead of routing `_raxol/session.steer` to SteerRequest) are unrelated to this fix — they fail identically on the pristine branch (12/14). Flagging for awareness; not in scope for this HIGH.

The MEDIUM items (tuple-token `inspect/1` leak, overlay-traversal duplication) are not addressed here.

DROOdotFOO:
## MEDIUM cleared: overlay-traversal duplication — `78fa56dbb`

`TreeWalker` (tool derivation) and `StructuredScreenshot` (widgets resource) each carried their own copy of the node-children walk — direct `:children` (list *or* single map) plus the `:absolute_layer` `:flow_child` + `:overlays[].element` wiring. Patching one could silently diverge from the other (tools for widgets not in the screenshot, or vice versa).

Extracted the walk into **`Raxol.MCP.NodeChildren`** as the single source of truth; both consumers now route through `child_nodes/1`. Behavior is unchanged (302 mcp tests green, incl. the existing `tree_walker_children_test`), plus a direct unit test for the shared helper (list / single-map / absolute-layer order / skip-non-map-overlays / non-map input).

### Status of #650's two MEDIUMs
- ✅ **overlay-traversal duplication** — this commit.
- ✅ **`json_turn_token/1` tuple leak** — the *crash* is already fixed by your `18bd6d8c6` (keeps int/binary/nil byte-identical, stringifies the rest so `encode_result` stays total). The residual is only that `inspect(token)` still renders the internal `{:steered, cur, System.unique_integer}` structure to the peer — a minor info-disclosure. I left it as-is since your commit made a deliberate choice there; if you'd rather emit an opaque placeholder for the un-addressable swapped token (it has nowhere legitimate to go anyway), say the word and I'll do it.

With the HIGH (`4a65dc088`) + this MEDIUM, #650's actionable concerns are cleared. The LOWs (unbounded dedup index, unguarded `to_json` on tuple `expectedTurnId`) and 3 droo-stack test-double items remain if you want them.

PR #651, opened 2026-07-19
feat(terminal): wide-glyph correctness, capability probe, and inline-driver hardening

merklebonsai:
Self-contained raxol_terminal package work underpinning the harness:
- East-Asian-Wide + emoji-presentation width tables; grapheme-aware split.
- Bulk row-pass buffer writer (fill_cells) with wide-pair correctness, and
  renderer dropping wide-placeholder cells so a wide glyph spends two
  columns, not three.
- Capability pipeline: OSC 10/11 probe + classifier + native-palette
  record; driver bridge emits :terminal_capabilities.
- stty via argv (no shell); inline driver gains DSR cursor probe,
  bracketed-paste reassembly, opt-in SGR mouse, and a silent isig guard.
- DECSC/DECRC no longer save the scroll region (matches hardware);
  ScrollRegionManager.plan is a pure geometry constructor.


merklebonsai:

merklebonsai:
## Post-lock-fix CI: 2 test failures fixed (both width-explained, no coupling)

Both failures trace to this branch's fullwidth-glyph correctness change and are safe golden updates. Verified against master and against the Utext `pretty.ex` rewrite.

**1. `FATEGoldenTest` — re-blessed `priv/fate/golden.refs`**
Only `text_wide_symbols` changed (`97e01b94…` → `2df3fb74…`). The branch adds `{0x26A1, 0x26A1}` (⚡) to `@emoji_presentation_ranges`, so ⚡ is now 2 columns and ` end` shifts one column right. Every other fixture hash is byte-identical — the fg-sensitive fixtures (`box_1x1`, `box_border`, `text_ascii`, `text_styled`, `overlap_zorder`) and `text_cjk` are untouched. Width-only re-bless.

**2. `PrettyTest` "CJK paragraph breaks between ideographs" — stale expected values**
`pretty.ex` on this branch is identical to master. The stale expectation (`7/8/7/1` chars) was correct only under master's **buggy** width model where Hiragana/Katakana measured as 1 column. This branch correctly measures them as fullwidth (2 cols), so master's own `pretty.ex` now packs 5 chars/line = exactly 10 cols:

```
["日本語のテ", "キストです", "今日はとて", "も良い天気", "ですね"]
```

Every line ≤ width budget (10), no broken grapheme clusters, text preserved. The Utext `pretty.ex` rewrite (on `merge/harness-endgame`) produces the **identical** output for this input — so there is **no T→Utext coupling**; the corrected wrap comes purely from the width fix, and this PR ships correct standalone.

merklebonsai:
FIX 1 (blocking) pushed: fe2ae1204 — retired `capabilities_probe` at finalize (nil'd, not just timer-canceled) so `route_capabilities_input/2` stops routing input through the probe post-:done; a lone ESC keystroke no longer parks as a discarded Probe partial and now reaches input. Red-first regression test added in `capabilities_probe_integration_test.exs` (confirmed failing before the fix, passing after). Full raxol_terminal suite: 2334/2337 passing, 3 pre-existing unrelated failures in renderer_integration_test.exs (verified present on master tip via stash, not caused by this change).

merklebonsai:
FIX 2 (minor) pushed: 966c71676 — `get_char_width/1`'s binary clause now measures only the FIRST grapheme (via `String.next_grapheme/1`) instead of flattening the whole binary to a codepoint list; a VS16 belonging to a later grapheme no longer widens an earlier, unrelated character. Red-first regression test added in `character_handling_test.exs` (`get_char_width("A❤️") == 1`, confirmed failing (got 2) before the fix, passing after). Full raxol_terminal suite: 2334/2338 passing; the 4 failures are pre-existing/flaky and unrelated (3 renderer_integration_test.exs failures present on branch tip before any of these fixes, 1 extension_server_test.exs named-process collision confirmed to pass 15/15 solo).

merklebonsai:
FIX 3 (major) pushed: 246d4ef8c — `unterminated_open/1` in inline_driver.ex now also holds a trailing PROPER PREFIX of the bracketed-paste OPEN marker `\e[200~` (1-5 bytes), mirroring `CursorReport.cpr_prefix?`'s split-chunk defense (search backward for the last ESC in a bounded window, check the grammar from there). Previously `:binary.matches/2` only recognized a COMPLETE 6-byte marker, so a chunk boundary landing mid-marker (e.g. chunk1 ending `...\e[20`) fell through as raw keystrokes with no memory of the split; the reunited marker never re-formed and a pasted `\r` fired a spurious `:enter`. Red-first regression test added in `inline_driver_test.exs` (chunk1 `"x\e[20"`, chunk2 `"0~evil\rrm -rf\e[201~"` -> paste captured atomically, confirmed the spurious `:enter` fired before the fix, clean after). Full raxol_terminal suite: 2336/2339 passing; the 3 remaining failures are the same pre-existing/unrelated renderer_integration_test.exs failures flagged in FIX 1/2 (confirmed present on branch tip before any of these fixes).

merklebonsai:
FIX 4 (major) pushed: 7412beca3 — a wide write's own placeholder (its second column, `x+1`) can land on a DIFFERENT wide pair's LEAD glyph, not just that neighbour's placeholder (the case `clear_orphaned_lead`/`clear_orphaned_placeholder` already covered). Example: `日本` then a wide `月` at column 1 — `月` spans columns 1-2, so its placeholder overwrites `本`'s lead at column 2, orphaning `本`'s own placeholder at column 3. That orphan stayed flagged and got dropped by the renderer with no surviving glyph to account for the column, painting the row one column short. Added a third clearing step (`clear_neighbour_lead_in_row` / `clear_neighbour_lead`, both gated on `width == 2`) to both `write_char`'s path (`update_row`/`clear_wide_neighbour_in_row`) and the bulk `fill_cells` path (`write_into`/`clear_wide_neighbour`), mirroring the existing two-case structure and updating that function's own comment (~396-408) to document the third case.

Red-first: `ScreenBuffer.new(6,1) |> write_string(0,0,"日本") |> write_string(1,0,"月")` rendered 5 columns wide instead of 6 before the fix (confirmed). Also verified red-first in isolation for the `fill_cells` path specifically (temporarily reverted just that half of the fix) — both the new dedicated test AND the pre-existing randomized oracle-equivalence fuzz test (`writer_fill_cells_test.exs`, "randomized write lists match the sequential write_char fold exactly") failed with a `write_char`/`fill_cells` disagreement, confirming the fuzz suite would have caught a partial fix on its own. All green after.

Verify counts: raxol_terminal full suite 2338/2341 passing (3 pre-existing/unrelated `RendererIntegrationTest` failures, same ones flagged in FIX 1-3, confirmed present before any of these 4 fixes via git stash). Root `test/fate test/cross_terminal`: 150/150 passing.

Operations.write_char verdict (reported, not fixed, per instructions): `Raxol.Terminal.ScreenBuffer.Operations.write_char/5` (operations.ex, `write_cell_to_row/6` ~:631) IS the live emulator print path — reached via `Raxol.Terminal.Input.CharacterProcessor.process_printable_character/2`, called from `InputHandler`/`TextProcessor` for real program output, NOT a dead/legacy path. It is a fully separate implementation from `Buffer.Writer` (writer.ex backs `ScreenBuffer.write_char`/`write_string` via `WriteOps`, used by UI rendering and tests) and shares no code with it. `write_cell_to_row/6` has ZERO wide-pair neighbour-clearing of any kind — the only `wide_placeholder` reference in the whole file is where it creates its own new placeholder. So it needs not just this fix's new case but all three (overwrite-placeholder-orphans-lead, overwrite-lead-orphans-placeholder, and this fix's neighbour-lead case) — the live VT100 print path is currently MORE exposed to wide-character corruption than the `ScreenBuffer.write_char`/`write_string` path. Recommend a follow-up ticket.

DROOdotFOO:
## Adversarial Review — Verdict: BLOCK

3-persona adversarial pass + droo-stack lens. Three HIGH defects survived verification, all in the terminal input/render hot paths this PR touches. Do not merge until they are fixed.

### HIGH
- **`inline_driver.ex:744` — lone ESC swallowed indefinitely.** `dispatch_input/2` buffers input through `split_pending_paste/1` → `unterminated_open/1` → `trailing_open_prefix/1`; a lone ESC (and split escape-sequence prefixes) sit in `paste_pending` with no flush/deadline, so the Escape key never fires `:escape` until an unrelated later keystroke arrives — then both fire together. Regresses a primary interaction key. _(Saboteur, CONFIRMED)_
- **`inline_driver.ex:985` — stty fork-per-input-chunk.** The isig guard fork+execs an `stty` subprocess synchronously inside the GenServer on every input chunk by default (`isig_guard_every=1`), serializing input behind subprocess latency. _(Saboteur efficiency + Security Auditor DoS-resource-amplification, promoted)_
- **`buffer/writer.ex:297` — one bad coordinate crashes the whole frame.** `fill_cells` `group_by`/`reduce` use guard-only anonymous clauses with no fallback, so a single negative or non-integer coordinate raises `FunctionClauseError` and aborts the whole frame — unlike the graceful out-of-bounds-high skip beside it. _(Saboteur + New Hire, promoted)_

### MEDIUM
- `inline_driver.ex:824` — `probe_loop`/`cursor_probe_loop` absolute-deadline `receive` never evaluates `after remaining` while a matching message is already queued, so a sustained non-sentinel input flood wedges the init-time probe (driver never returns `{:ok, state}`). The moduledoc's claim the deadline prevents this is false. _(CONFIRMED)_
- `inline_driver.ex:959` — `reassert_raw_until_isig_off` blocks init up to ~3s (60×50ms) forking `stty` each pass when `-isig` never sticks.
- `buffer/writer.ex:437` — two hand-synced copies of wide-pair clearing logic (list/`update_row` vs map/`fill_cells`) with no shared helper.
- `inline_driver.ex:941` — the ~3s boot ISIG budget exists only as the product of two uncoupled inline literals (attempts 60, `Process.sleep` 50) while sibling budgets are named module attributes.
- `capabilities/classifier.ex:99` — new comments cite external design-doc codes (§2/§7, rungs, A1, CAP-N-03/CAP-P-13) with no in-repo definition, and the PR deletes the doc pointers they lean on.

### LOW
- `inline_driver.ex:952` boot ISIG reads through hardcoded `Stty.isig_off?/0`, bypassing the injectable `isig_flags_reader` seam (untestable via injection); `:640` cursor-probe loop forwards interleaved bytes without threading `paste_pending` (pasted `\r` → spurious `:enter`); `renderer.ex:249` `wide_placeholder?/1` copy-pasted across renderer/writer; `:532` bare `budget_ms + 5_000` slack with no named constant.

### droo-stack
- `character_handling.ex:84` — `wide_char?/1` rebuilds a range list via `... ++ @emoji_presentation_ranges` and runs linear `Enum.any?` on **every** call on the per-grapheme measurement hot path; hoist to a single compile-time `@wide_ranges` module attribute.
- `character_handling.ex:175` — `emoji_presentation_width/1` uses a cond-for-dispatch ladder; promote `regional_indicator?/1` to a `defguard` and split into pattern-matched heads.
- `inline_driver.ex:318` — `run_post_raw_setup/5` threads state through nested if/else (rightward drift); prefer a `with` chain or extracted helpers.
- `inline_driver_cursor_probe_test.exs:95` — `await_request/2` busy-polls with `Process.sleep(5)` recursion instead of message-based sync (wall-clock flake class).

### To clear BLOCK
1. Add a flush deadline / timer so a lone ESC (and stalled escape prefixes) fires `:escape` without waiting for the next keystroke.
2. Default `isig_guard_every` away from per-chunk, or move the stty reassertion off the hot input path (async / cache the flag).
3. Give `fill_cells` a graceful fallback clause that skips out-of-range/non-integer coordinates instead of raising.
4. Then address the probe-loop wedge (MEDIUM) and the wide-glyph hot-path allocation (droo-stack).

DROOdotFOO:
## Blockers addressed — `ed75e8b5c`

All three verified HIGH blockers are fixed in the pushed commit:

1. **Lone ESC swallowed** (`inline_driver.ex`) — a held *strict* paste-open prefix (lone ESC / torn `ESC[200~` head, 1–5 bytes) now arms an idle flush deadline (`paste_flush_ms`, token-guarded); `:escape` fires on its own without a following keystroke. A complete open marker + content is never flushed, so genuine paste reassembly is untouched.
2. **stty fork-per-chunk** (`inline_driver.ex`) — added a wall-clock floor (`isig_guard_interval_ms`, default 40ms) on top of the chunk cadence: a burst/paste re-asserts at most once per interval instead of forking a subprocess per byte. Normal typing still re-asserts on the next key.
3. **fill_cells frame crash** (`writer.ex`) — both the `group_by` (bad `y`) and the per-row fold (bad `x`) now skip negative/non-integer/nil coordinates like any out-of-bounds write instead of raising `FunctionClauseError` and aborting the whole frame.

+14 regression tests (3 flush, 2 isig-throttle, 9 malformed-coord). New tests green; no new suite failures (the 3 pre-existing `RendererIntegrationTest` failures are unrelated to this diff).

The remaining MEDIUM/LOW and droo-stack items from the review are not addressed here.

DROOdotFOO:
## MEDIUM tier + droo-stack — `cff7d8897`

Following the three BLOCK-tier fixes (`ed75e8b5c`), this clears the higher-value remaining findings.

**MEDIUM**
- **probe-loop after-deadline wedge (CONFIRMED DoS)** — `probe_loop/3` and `cursor_probe_loop/3` now check the deadline **before** the receive. `receive ... after 0` loses to any already-queued matching message, so a sustained non-CPR input flood starved `after` and the init-time probe never returned while the flood continued (the moduledoc's claim that re-deriving `remaining` bounds this was false). The explicit pre-check bounds the loop by the deadline regardless of mailbox pressure; leftover input stays queued for normal dispatch. **+ a flood-resilience regression test, proven to wedge (yield → nil) pre-fix and pass post-fix (3x stable).**
- **boot ISIG budget** — the ~3s window is now `@isig_reassert_attempts` (60) × `@isig_reassert_interval_ms` (50), named in one place instead of two uncoupled inline literals.

**droo-stack**
- `wide_char?/1` no longer rebuilds `[...CJK...] ++ @emoji_presentation_ranges` on **every grapheme** (the measurement hot path) — hoisted to a compile-time `@wide_ranges`.
- `run_post_raw_setup/5`'s two-level nested if/else flattened into `state |> start_stdin_reader() |> maybe_confirm_isig_boot()` (guarded multi-clause head).

**Deferred, with rationale:**
- **Wide-pair clearing duplication** (`writer.ex` list `clear_wide_neighbour_in_row` vs map `clear_wide_neighbour`) — genuine, but the two operate on different representations (list `Enum.at` vs map `Map.get`); a shared helper needs a cell-accessor abstraction that adds indirection to correctness-critical wide-glyph code. Better weighed by you than pushed blind.
- **emoji `cond` ladder** — a false positive: `variation_selector_16?(codepoints)` (a list-membership test) can't be a guard, so `cond` over the computed value is the idiomatic choice here.
- **`await_request` busy-poll** in the probe test polls a StringIO device write (no message to sync on), and the classifier's external doc-code comments are doc-only — both left as-is.

Terminal suite green apart from the 3 pre-existing `RendererIntegrationTest` failures (branch-wide, unrelated to this diff).

DROOdotFOO:
## Deflaked the macOS CI blocker — `61d711972`

The one red check was `StreamCadenceTest` "yield budget forces a flush even when input never releases" failing its 2000ms `assert_receive` on macOS — a wall-clock flake, not a real failure (and not part of this PR's terminal changes; the file was last touched by #637). Its flush is up to `max_consecutive_yields` sequential `Process.send_after` round-trips, and on a loaded shared runner each delivery is delayed, stretching the chain past 2s.

Routed the assert deadlines through a generous `@flush_timeout` (5s). It only bounds *failure* time — `assert_receive` returns the instant the batch lands — so nothing is weakened; short `refute_receive` windows are left as-is (contention only makes non-arrival more reliable). CI is rerunning now.

PR #652, opened 2026-07-20
feat(agent): streaming item lifecycle on the event contract

merklebonsai:
The Contract's pump grows the streaming item lifecycle — item_started +
per-item item_id + a :reasoning item type (lazy-opened on first non-blank
text) — so a live session's streamed output enters the projection tail
incrementally instead of appearing all at once at completion. Backend.HTTP
accumulates tool-call deltas by index and separates the reasoning channel;
Mock gains a one-shot tool_calls_fn. Journal, checkpoint, done-gate,
interrupt, steer, and spend-gate ride along as the contract's supporting
cast. Snapshot floor grows additively (item_started + item_id).


merklebonsai:

merklebonsai:
Addressed the adversarial review findings in cd4696b17:

- **FIX 1 (major)** — `Contract.pump/3`'s `{:error, reason}` arm (and the halt-without-terminal case) now closes any open reasoning/message item before/with the error event, so no `item_started` is left dangling and the streamed partial text survives replay. Regression: `contract_test.exs` — `pump/3 an error mid-stream seals whatever reasoning/answer text streamed before the halt`.
- **FIX 2 (major, green-wash)** — `Stream.run/2`'s `normalize_backend_stream` no longer silently drops a streamed done's `tool_calls`; it emits a `:tool_unexecuted` marker per call (the vocabulary existed with zero producers). Regression: `stream_test.exs` — `run/2 tool_calls on a streamed done ... seals a :tool_unexecuted marker, never a silent drop`.
- **FIX 3 (minor)** — `Backend.HTTP.parse_sse` now canonicalizes `\r\n` → `\n` before framing, so a CRLF-speaking provider's `\r\n\r\n` separator is recognized (no more infinite buffering) and a `data: [DONE]\r` no longer breaks the exact match into a bogus unparseable-chunk marker. Regression: `http_wire_test.exs` — the two new "CRLF line endings" tests.
- **FIX 4 (minor)** — `merge_tool_call_frags`'s no-index fallback no longer collapses distinct complete fragments from separate chunks onto the same key (now id-matched, else a fresh key past everything assigned); `finalize_tool_calls` no longer launders undecodable `arguments` to a silent `%{}`, surfacing an `arguments_error` marker (byte count only, never raw text) instead. Regression: `http_test.exs` — the three new "streamed tool-call accumulation" tests.

All four fixes verified red-first (each regression test fails without its corresponding source change) then green with the fix. Full suite: 1533 passed, 0 failures.

Scoping items intentionally left open for V, not fixed here:
- `react/2` drops `response.reasoning` on the tool-using path (stream.ex:~389) — the headline reasoning feature only fires on the tool-less path; needs a scope decision, not a drive-by fix.
- The deleted `live_session_agent_test.exs` keystone — restoring it is a separate change.

DROOdotFOO:

3-persona adversarial pass + droo-stack lens. HIGH findings reproduced from the diff before surviving; cross-persona overlaps promoted.

### HIGH
- **`backend/http.ex:728` — provider-controlled tool-call index crashes the streaming turn.** `resolve_tool_call_index/3` uses the wire `index` value verbatim as the accumulator map key with no type validation → `ArithmeticError` (`<non-number> + 1`) on a non-integer index, and no-index fragments double-count `pos` into sparse indices. Untrusted network input from arbitrary OpenAI-compatible endpoints; live-reproduced via `accumulate_tool_calls/1`. _(Security Auditor + Saboteur, promoted from MEDIUM)_

### MEDIUM
- **`contract.ex:360`** — `tool_result` / `tool_unexecuted` / `approval` arms call `complete_item` without sealing an open `msg_item` / `reasoning_item` (unlike the `tool_use` arm), so a later `text_delta` re-enters the still-open message and fuses/mis-orders items. Latent until the next combined streaming+tool producer.
- **`http.ex:636`** — blocking `parse_openai_tool_calls` silently coerces undecodable arguments to `%{}` while the streaming twin attaches an `arguments_error` marker, so `framework_react` executes tools under wrong/empty args with no trace — the exact green-wash this PR hardened against.
- **`contract.ex:301`** — `pump/3` threads an untyped 7-key accumulator through ~12 helpers all shaped `(session_id, turn_id, counter, acc)` with two interchangeable plain-string ids; a transposed id compiles, runs, and stamps every lifecycle event with reversed turn/session. Consider a small struct.

### LOW
- `http.ex:719` sparse tool-call ordinals; `http.ex:42` `@default_max_tokens` 1024→4096 + unbounded `AI_MAX_TOKENS`; `http.ex:380` streaming truncation marker unreachable (reads `source["usage"]`, never nested in a choice); `contract.ex:708` `blank?/1` unguarded `String.trim` raises on nil; `http.ex:261` `parse_sse/2` public test seam with no `@spec`; `http.ex:682` unbounded arg-fragment accumulation (OOM DoS); `contract.ex:934` `result_summary/1` `to_string/1` raises `Protocol.UndefinedError` on maps/tuples.

### droo-stack
- `contract.ex:386` (and 399/443/476/512/578/614/664/702/738/770) — journal built with `acc.journal ++ [ev]` inside the per-stream reduce; item lifecycle now emits `item_started`+`item_completed` per item, roughly doubling entries → O(n²) fold. Prepend and reverse at read boundaries.
- `http.ex:393` — `events ++ [event]` (banned `xs ++ [x]`) in a hot path.
- `http.ex:593` — `cond do ... true -> end` used as if/else; prefer `if`/`with`/multi-clause head.

### To clear CONCERNS
1. Validate/normalize the wire tool-call `index` (integer guard, fall back to positional) before using it as a key.
2. Seal open msg/reasoning items in the `tool_result`/`approval` arms symmetric with `tool_use`.
3. Make blocking and streaming tool-arg parsing agree on the error marker (don't silently coerce to `%{}`).
4. Flip journal accumulation to prepend+reverse.

DROOdotFOO:
## HIGH addressed — `21c0d090d`

`resolve_tool_call_index/3` used the wire `index` verbatim as the accumulator key whenever it was non-nil. `index` is provider (network) controlled, so a non-integer value (string/float/map from a buggy or hostile OpenAI-compatible endpoint) landed as a map key, and `next_tool_call_index/1`'s `Enum.max(keys) + 1` then did `<non-number> + 1` → `ArithmeticError`, killing the whole streaming turn.

Fix: guarded the index clause to `is_integer(index) and index >= 0`. A non-integer/negative index now falls through to id-based, then positional resolution — no crash, both distinct calls still resolve.

+2 regression tests (garbage indices string/float/map/list/bool don't crash and still resolve 2 calls; valid integer indices still key the accumulator). Full `http_test` suite green (21 passed).

The MEDIUM/LOW items (asymmetric item sealing, blocking-vs-streaming arg honesty, untyped accumulator, journal `++` O(n²), unbounded arg accumulation) are not addressed here.

DROOdotFOO:
## MEDIUM tier + droo-stack addressed — `4667f4875`

Following the HIGH fix (`21c0d090d`), this clears the remaining review findings.

**MEDIUM**
- **M1 — asymmetric item sealing** (`contract.ex`): `tool_result` / `tool_unexecuted` / `approval_requested` / `approval_decided` now seal an open message+reasoning item first, via a shared `seal_open_items/4` (which also replaces `tool_use`'s inline two-liner). No-op when nothing's open; only changes the edge where a boundary arrives with a message still open — pre/post-boundary text no longer fuses into one block. Regression test added.
- **M2 — blocking vs streaming arg honesty** (`http.ex`): the blocking `parse_openai_tool_calls` silently coerced undecodable arguments to `%{}`. Both paths now share `build_tool_call/3`, so the blocking path surfaces the same `arguments_error` marker (map-shaped `%{}` + honest marker, never leaking raw text) the streaming path already did. Regression tests added (undecodable → marker; valid → clean).

**LOW (crash safety)**
- `blank?/1` is nil-safe (was unguarded `String.trim`).
- `result_summary/1` maps entries through a crash-safe `display_entry/1` (`to_string/1` raised `Protocol.UndefinedError` on a map/tuple entry, killing the turn).

**droo-stack**
- Pump `journal` is now built with prepend and reversed **once** at the single `DoneGate` read site — I verified the reversed list is byte-identical to the old append form, so evidence offsets on the wire are unchanged (40 contract tests confirm).
- `openai_stream_events` builds via `maybe_prepend` + one `Enum.reverse`; `cond`-as-if → `if` in `parse_response/2`.

**Deferred (with rationale):** M3 (untyped 7-key pump accumulator + transposable `session_id`/`turn_id`) is real fragility but not a live bug, and the fix is a signature refactor across ~12 helpers — better done by you as the design owner than pushed onto your branch blind. The remaining LOWs are intentional/minor: the `max_tokens` 1024→4096 raise is deliberate and test-pinned (`max_tokens default + AI_MAX_TOKENS override`); sparse tool-call ordinals are masked by `finalize` discarding the index; `parse_sse` `@spec` is cosmetic.

Full agent suite green (1538 passed). Everything routes through existing behavior except the M1/M2 edges, which have new tests.

PR #653, opened 2026-07-20
feat(ui): shared normalized keystroke seam for the harness

merklebonsai:
InputEvent.normalize/1 — one place that turns raw key events into the
harness's printable/control shape. A thin seam the composer, overlays,
and pump contract all consume.

### DIFF
--- a/lib/raxol/ui/harness/input_event.ex
+++ b/lib/raxol/ui/harness/input_event.ex
@@ -175,6 +175,10 @@ defmodule Raxol.UI.Harness.InputEvent do
           shift: boolean(),
           meta: boolean()
         }
+  # Single source for the kind vocabulary: the `@type kind` below and the
+  # idempotence fast-path guard both derive from it, so a new kind cannot
+  # silently regress one without the other.
+  @kinds [:char, :key, :paste, :other]
   @type kind :: :char | :key | :paste | :other
   @type key_state :: :pressed | :released | :repeat | nil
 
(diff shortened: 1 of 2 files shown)


DROOdotFOO:

3-persona adversarial pass (Saboteur / New Hire / Security Auditor) + droo-stack conventions lens. Every HIGH was reproduced from the diff by a skeptic before it survived; cross-persona overlaps were severity-promoted.

### HIGH
- **`input_event.ex:207` — idempotence fast-path bypasses sanitization.** The new head `def normalize(%{kind: kind, mods: %{}} = already_normalized) when kind in [:char, :key, :paste, :other]` uses `mods: %{}`, which **matches ANY map** (the CLAUDE.md gotcha), so it short-circuits the total normalizer and returns unvalidated maps verbatim — bypassing char C0/DEL control-byte rejection **and** paste `strip_paste_text/1` ESC/OSC/CSI sanitization. A `%{kind: :paste, text: "\e[2J\e]0;pwned\a"}` map reaches `insert(state, norm.text)` unsanitized → terminal-control injection the pre-PR path removed. Latent today (no live producer emits a kind+mods map) but a real regression of the module's documented sanitization contract. _(Saboteur + Security Auditor, promoted)_

### MEDIUM / LOW
- **`:217`** — `mods: %{}` reads as "empty mods" but means "any map"; invites a "fix" that breaks the guard. Bind-and-guard instead: `mods: mods when is_map(mods)`. _(New Hire + droo-stack, promoted)_
- **`:210`** — shape-only guard passes partial/incomplete-mods maps through unchanged instead of rebuilding canonical full mods; `text?/1` and `shortcut?/1` clause heads then fail to bind and silently drop a real keystroke. _(Saboteur + Security Auditor)_
- **`:218`** — guard hard-codes `[:char, :key, :paste, :other]`, duplicating `@type kind` with no shared attribute; a fifth kind silently regresses idempotence.

### droo-stack
- `:207` — `%{}` to mean "any map" is the exact footgun the house rules call out. Prefer `mods: mods when is_map(mods)` or match the canonical shape `%{ctrl: _, alt: _, shift: _, meta: _}`.

### To clear CONCERNS
1. Replace `mods: %{}` with `mods: mods when is_map(mods)` (or the canonical mods shape) so the fast-path can't match arbitrary maps.
2. Ensure the fast-path still routes `:paste` text through `strip_paste_text/1` and char kinds through control-byte rejection — an already-normalized value must remain sanitized, not be trusted on shape alone.
3. Derive the kind list from `@type kind` / a shared module attribute.

DROOdotFOO:
## HIGH addressed — `b0ed79b9a`

The idempotence fast-path now matches the **canonical four-key mods shape** (`%{ctrl: _, alt: _, shift: _, meta: _}`) instead of `mods: %{}` (which matches any map). Effect:
- A crafted `%{kind: :paste, text: "\e[2J...", mods: %{}}` no longer short-circuits the normalizer — it falls through to `strip_paste_text/1` and is sanitized (closes the terminal-control-injection bypass).
- Partial-mods maps are re-normalized to canonical mods instead of passing through unchanged (fixes the `text?/1`/`shortcut?/1` clause-head mismatch).
- The kind list is now a shared `@kinds` attribute, so the guard and `@type kind` can't drift.

Genuine idempotence is preserved (a fully-canonical event still passes through unchanged). +3 regression tests; full `input_event` suite green (73 passed).

PR #674, opened 2026-07-21
ci: stabilize nightly flakes (timing, starvation, capture)

### LINKED ISSUE #672: 🌙 Nightly build failure - 2026-07-21
The nightly build failed. Please check the [workflow run](https://github.com/DROOdotFOO/raxol/actions/runs/29803916178) for details.

DROOdotFOO:
## What

Stabilizes the nightly build, red for several nights (#672, #664, #486). Root-caused three distinct **flake classes** — none is a product bug — from the actual run logs.

### 1. Timing-assertion flake (fails every night in Extended Scenarios)
`test/property/core_property_test.exs` "parser performance scales sub-quadratically" is a wall-clock `assert ratio < max_ratio`. It is already `@tag :skip_on_ci`, but two lanes ignored the tag:
- nightly `extended-scenarios` runs `mix test --only property` (no `--exclude skip_on_ci`)
- ci-unified `property` job runs `mix test test/property` **by path** (tags don't filter)

Both now pass `--exclude skip_on_ci`. Verified: `core_property_test.exs --exclude skip_on_ci` -> `9 passed, 1 excluded`.

### 2. Heavy emulator-replay starvation (fails multiple matrix cells, varies by night)
The `test/harness/` surface suites + `renderer_seal_once` are CPU-bound emulator replays. On slow matrix legs (Elixir 1.17/1.18, macOS 2-3 core) the default async pool co-schedules several and starves each of cores, blowing the 60s/180s bounds — a runner-load artifact the code already documents. Different suites tip over on different nights (`T13aSurfaceTest`, `ProjectionPanelsSurfaceTest`, `LiveSessionDriverTest`, `RendererSealOnceTest`, ...).

- Matrix run now caps `--max-cases 4` and raises `--timeout 120000`, so each replay suite gets a full core + headroom. ci-unified runs the same suites per-push on fast runners, so cross-version coverage is unchanged.
- `renderer_seal_once_property_test.exs` gets `@moduletag :property` so its plain `test` keystones (not just its `property` blocks) are actually excluded by the matrix's existing `--exclude property`. Verified: `--exclude property` -> `0 tests, 7 excluded`; by-path -> `7 passed` (and it runs in **25s uncontended** vs the 180s timeout when starved).

### 3. Headless first-frame capture race (1 cell this run)
`headless_golden_test.exs` "initial render matches golden" read an **empty** buffer even though the render logs show the frame positioned — the first `screenshot` after `start` can observe the buffer in the window between `render_frame_sync` replying and the cells committing. Now retries until content lands (bounded, no sleep). Verified: `3 passed`.

## Not changed
No product code. `mix.lock` churn (nx/circular_buffer dependabot drift) is deliberately excluded — separate concern.

## Manual testing
- [x] `mix test test/property/renderer_seal_once_property_test.exs --exclude property` -> 0 tests, 7 excluded
- [x] `mix test test/property/renderer_seal_once_property_test.exs` (by path) -> 7 passed
- [x] `mix test test/property/core_property_test.exs --exclude skip_on_ci` -> 9 passed, 1 excluded
- [x] `mix test test/cross_terminal/headless_golden_test.exs` -> 3 passed
- [x] both workflow YAMLs parse

Closes #672, #664, #486.

### DIFF
--- a/.github/workflows/ci-unified.yml
+++ b/.github/workflows/ci-unified.yml
@@ -320,7 +320,12 @@ jobs:
       
       - name: Run property tests
         run: |
-          mix test test/property
+          # Run test/property by path, but still honor skip_on_ci: some
+          # properties are wall-clock timing assertions (e.g. parser
+          # sub-quadratic scaling) tagged skip_on_ci precisely because they
+          # flake on shared CI runners. By path alone they would run and
+          # flake; exclude them like every other CI lane.
+          mix test test/property --exclude skip_on_ci
         env:
           MIX_TEST_PARTITION: 3
       
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -117,15 +117,21 @@ jobs:
       - name: Run full test suite
         run: |
           # --exclude property: the emulator-replay-heavy property tests
-          # (renderer seal-once keystones) are CPU/allocation-bound. Run
-          # inline inside the full ~6k-test process on slow matrix legs
-          # (older Elixir, macOS), scheduler/GC pressure pushes them past
-          # the 60s/180s bounds -- a runner-load artifact, not a hang (the
-          # newer/faster legs pass the same tests). They already run
-          # isolated at --timeout 300000 in the extended-scenarios job, so
-          # excluding them here removes the duplicate run without losing
-          # coverage.
-          mix test --exclude slow --exclude integration --exclude docker --exclude skip_on_ci --exclude property
+          # (renderer seal-once keystones) are CPU/allocation-bound. They
+          # already run isolated at --timeout 300000 in the extended-scenarios
+          # job, so excluding them here removes the duplicate run.
+          #
+          # --max-cases / --timeout: the harness surface suites (test/harness)
+          # are the same CPU-bound emulator-replay class but are NOT property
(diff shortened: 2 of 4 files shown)


DROOdotFOO:
## Folded in two more CI-flake fixes — `8e2a2ada8`, `4762ecfb8`

Surfaced while greening PR #651's macOS leg; both are pre-existing on master (this branch), so they belong here with the rest of the flake-class fixes rather than only on a feature branch.

- **`StreamCadenceTest`** — its flush is up to `max_consecutive_yields` sequential `Process.send_after` round-trips; on a loaded macOS runner the chain stretches past the tight 2000ms `assert_receive`. Routed the assert deadlines through a generous `@flush_timeout` (5s) — bounds *failure* time only, weakens no assertion; short `refute_receive` windows left as-is.
- **`DispatcherEmitTest`** — `on_exit` `GenServer.stop` raced the dispatcher's linked `:shutdown` (which `GenServer.stop`'s default `:normal` rejects), failing teardown. Both call sites now go through a `stop_on_exit/1` helper with `try/catch :exit`.

Verified against a real macOS CI run (both green on #651). This PR now covers the workflow-level flake fixes (matrix starvation, extended-scenarios/property `skip_on_ci`, headless capture race) **plus** these two per-test fixes.

PR #678, opened 2026-07-21
Add custom_console_agent ACP offering + crash recovery

DROOdotFOO:
## Summary

Adds a second ACP offering, `custom_console_agent`, and the crash-recovery
(durability) layer for the seller stack. The offering sells a validated,
deployment-ready Virtuals Console agent package (soul.md + scheduled tasks for
Hermes or OpenClaw) as a **plain job** (`requiredFunds: false`) — no escrowed
principal, no corridor liquidity. Because its handler is inference-backed and
nondeterministic, the M1 work makes a mid-flight BEAM restart resume rather than
double-submit a different deliverable.

Ported from hand-authored, reviewed-but-never-compiled source; merged carefully
onto `master` (the four touched seller/task files were verified additive
supersets, not wholesale overwrites) and made to compile, pass, and clear the
quality gate.

## What's in it

- **`custom_console_agent` offering** (`Raxol.ACP.Console.*`): spec validation +
  content policy, wallet-funded generation on Virtuals compute, static validation,
  a bench harness that boots the real open-source runtime for evidence, a
  bench-slot capacity gate, artifact storage, and the delivery pipeline.
  `mix acp.register_offering --offering console` emits the marketplace document.
- **Crash-recovery (M1)**: `JobSession.Provider` pins the encoded deliverable +
  keccak in a `Raxol.Payments.Checkpoint` store **before** signing and resumes
  from the pinned bytes on any replay — one deliverable hash per job, ever
  (*exactly-once-effective, at-least-once-attempted*). `Raxol.ACP.Seller.Resync`
  drains `get_active_jobs` on boot/reconnect and re-drives interrupted jobs through
  the Queue's idempotent dispatch; the Queue adopts mid-flight jobs on re-offer and
  posts deliverables out-of-band. `require_checkpoint: true` fails fund-adjacent
  writes closed with no store.
- **Prerequisite type fix** (`raxol_payments`): renames the `Checkpoint` `record`
  type to `entry` (it shadowed a dialyzer built-in). Types only, no runtime change;
  removes a pre-existing compile warning and downstream dialyzer false positives.
- **Config + docs**: `config/console_offering.example.exs`, `RUNBOOK.md`
  (provisioning → funding → registration → live on Base Sepolia), README section.

## Notable fixes made during integration

- Interned `:hermes`/`:openclaw` atoms at compile time (the source relied on
  `String.to_existing_atom/1` and would crash).
- Adapted the tests to the package's harness (the seller `Queue` singleton is
  already started by `test_helper`; a transient-supervisor restart race; a
  fail-closed test that needed an `:open` session for the accept path).
- Preserved existing behaviour: `deliver/2` still returns the handler's own
  deliverable term on the happy path; the on-chain hash is byte-identical.
- Corrected two `@spec`s dialyzer flagged (`Validator.validate/1` closed-map,
  `Delivery.run/2` ctx) and two `list ++ [x]` idioms.

## Manual testing

- [x] `MIX_ENV=test mix test` in `packages/raxol_acp` — **685 passed, 0 failures**
  (+24 new; 661 → 685).
- [x] `packages/raxol_payments` suite — **1062 passed, 0 failures** (checkpoint
  tests included; one conformance file skipped, it needs the external
  `riddler-client` fixture).
- [x] `mix compile --warnings-as-errors` — same warning profile as `master` (my
  changes add zero new warnings).
- [x] `mix format --check-formatted` on all changed files — clean.
- [x] `mix credo --strict` — parity with `master` (no new warnings/readability/
  design issues).
- [x] `mix dialyzer` — 38 → 16 warnings; every warning introduced by this change
  is fixed; the remaining 16 are all pre-existing in untouched files.
- [x] `mix acp.register_offering --offering console` emits a valid marketplace
  doc; `--offering usdc_public` still works (no regression).

## Out of scope

M3 buy-side driver, M4 wallet-funded compute wiring, and the actual funded/network
execution of the Sepolia dry-run (operator-run; the RUNBOOK + example config prep
everything up to that point).

### DIFF
--- a/packages/raxol_acp/README.md
+++ b/packages/raxol_acp/README.md
@@ -53,6 +53,14 @@ See `Raxol.ACP.Supervisor` for the supervision tree. Subsystems:
 
 `handle_request/2` accepts a job only for a corridor it can settle now, so a customer is rejected before escrow rather than after a failed settlement: the live capability matrix, per-order caps (`:destination_caps`), closed origins (`:closed_origins`), and a rolling aggregate reservation (`Raxol.ACP.Xochi.CapacityLedger`, opt-in via `capacity_gate_enabled: true`). `mix raxol_acp.derive_caps` derives the config from the solver's on-chain balances and `Raxol.ACP.Xochi.CapacityRefresher` refreshes it periodically. See `config/destination_caps.example.exs` and `docs/features/ACP.md`.
 
+## Second offering: Custom Console Agent
+
+`Raxol.ACP.Console.AgentOffering` (`custom_console_agent`) sells a validated, deployment-ready Virtuals Console agent package as a **plain job** (`requiredFunds: false`, no escrowed principal, no corridor liquidity). The buyer sends a spec (purpose, runtime, persona, scheduled tasks, skills); the deliverable is a `soul.md` + `tasks.json` (+ `AGENTS.md`, `skills/`) package for **Hermes** or **OpenClaw**, statically validated and -- by default -- bench-validated on the actual open-source runtime with the transcript shipped as evidence. Generation runs on wallet-funded Virtuals compute.
+
+Crash-recovery (M1) is wired end to end: `JobSession.Provider` pins the encoded deliverable + keccak in a `Raxol.Payments.Checkpoint` store **before** signing, and `Raxol.ACP.Seller.Resync` drains `get_active_jobs` on boot/reconnect and re-drives interrupted jobs through the Queue's idempotent dispatch, so a BEAM restart resumes rather than double-submitting (**exactly-once-effective, at-least-once-attempted**). Enable with `require_checkpoint: true` in production.
+
+Config surface: `config/console_offering.example.exs`. Base Sepolia go-live: `RUNBOOK.md`. Emit the marketplace document with `mix acp.register_offering --offering console`.
+
 ## Dependencies
 
 - `raxol_payments` for wallet signing (`Raxol.Payments.EIP712`, `Raxol.Payments.Wallet`) and Xochi cross-chain settlement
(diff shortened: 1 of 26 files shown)


DROOdotFOO:
## Adversarial review (3-persona: Saboteur / New Hire / Security Auditor)

Ran the offering + seller stack through a hostile review. Findings and fixes, all in `d4cf099`:

**CRITICAL — arbitrary-path file write (path traversal).** A skill name became the package key `skills/<name>/SKILL.md`, and `name` (LLM output, steerable by buyer `skills`/`purpose` input) was validated only as a binary. A `..` name escaped the tmp sandbox at `File.write` time; the Validator's `Path.split |> take(-2)` didn't catch it. **Fixed** two layers: the Generator constrains the skill name to `^[a-z0-9][a-z0-9_-]{0,63}$`, and `Delivery.materialize` fails closed on any `..`/absolute key.

**HIGH — bench-slot DoS.** Slot reserved pre-escrow (`handle_request`) but only released at delivery; an unfunded/rejected/expired job held it until the 45-min TTL (default `bench_slots: 1` → one unfunded offer blocks all bench-validated work). **Fixed** with an optional `handle_release/2` offering callback, invoked by the Queue on the expiry and accept-error paths.

**MEDIUM** — (1) Resync now fails closed on a numeric phase instead of mapping `0..6` against an unpinned on-chain enum order; (2) `Port.open` in the bench is contained so a missing wrapper returns a typed error instead of crashing the session; (3) the `{:ets}` checkpoint non-durability caveat is now explicit in the docs.

**Bonus (Saboteur, found while fixing the above):** a nil-price offering made `resolve_accept` call `AssetToken.usdc(nil, _)` and **crash the shared Queue** (DoS for every other offering). Now falls back to a zero budget so `handle_request` still runs.

New tests cover the traversal rejection, the bench-slot release, and the numeric-phase skip.

Verification: **687 passed, 0 failures**; dialyzer unchanged baseline (16, all pre-existing, none in changed files); format + credo at parity.

Verdict moved from **BLOCK** → **CLEAN** for the reviewed surface.

PR #688, opened 2026-07-22
chore(harness): repoint dead in-flight doc refs + refresh mix.lock

merklebonsai:
## What

Two small, self-contained fixes pulled off `merge/harness-endgame`:

- `lib/mix/tasks/t0.ringb.ex`: fixes a stale moduledoc reference — `docs/proposals/in-flight/t0-runbook.md` was moved to `docs/proposals/t0-runbook.md`, but the comment still pointed at the old in-flight path.
- `mix.exs`: drops a dead pointer to `docs/proposals/in-flight/shiki-elixir-analysis.md` in the `makeup_syntect` dep comment — that doc no longer exists at that path on master.
- `mix.lock`: refreshed after `mix deps.get` (picked up an unrelated transitive `ex_ast` patch bump, 0.12.10 -> 0.13.0).

No behavior change; both source hunks are comment-only. Everything else considered from the candidate list (`priv/fate/golden.refs`, `lib/raxol/navigation/vim.ex`, `lib/mix/tasks/raxol.mutation.ex`) was deliberately excluded — those belong to other slices of the merge campaign.

## Why standalone

Neither hunk depends on any later-wave feature: they're dead-reference cleanups that are correct regardless of what else lands from `harness-endgame`. Bringing them over now removes stale doc pointers without carrying any wave-2+ dependency risk, and keeps `mix.lock` from drifting further from what `mix deps.get` would actually produce in CI (CI compiles without `deps.get`, so a stale lock would go red at merge time).

## Verification

- `MIX_ENV=test mix compile --warnings-as-errors` — clean (one pre-existing warning from the vendored `uuid` dep, unrelated).
- `MIX_ENV=test mix test test/mix` — 29 passed.
- `MIX_ENV=test mix test test/harness/ringb_resolver_test.exs` — 0 run (all `ring_b: true`-tagged, excluded by default config), confirming no functional coupling to the doc-only `t0.ringb.ex` change.

**mix.lock note:** the `ex_ast` 0.12.10 → 0.13.0 line is an intentional lock refresh — transitive dep, within its parent constraint, and exactly the resolution `mix deps.get` produces today; CI compiles with no deps.get, so the committed lock must match it. Compile + tests green on this lock.


DROOdotFOO:

## Adversarial review — PR #688 (`harness-infra`)

Chore PR: repoints two dead `docs/proposals/in-flight/*` doc references (now-removed paths) in `t0.ringb.ex` and `mix.exs`, and refreshes `mix.lock`. The doc-ref repointing is correct and harmless. One thing to confirm before merge:

### LOW

**LOW** [New Hire] `mix.lock` — this "chore(harness): repoint dead doc refs" PR also silently bumps a dependency: `ex_ast 0.12.10 → 0.13.0` (a minor-version bump, not just a hash refresh). A dependency version change shouldn't ride inside a doc-reference chore — it's invisible in the PR title and skips the scrutiny a deliberate dep bump gets. Per this repo's own mix.lock-drift CI guard history, lock changes are load-bearing. Please either (a) split the `ex_ast` bump into its own commit/PR with a one-line rationale and a green compile+test run, or (b) confirm it's an intended, compatible bump and call it out in the PR body. Everything else here is clean.

**Verdict: CLEAN** (merge after confirming the `ex_ast` bump is intentional and compatible).

merklebonsai:
Re the LOW: confirming (b) — the `ex_ast` 0.12.10 → 0.13.0 bump is an intentional lock refresh, not an accidental rider. `ex_ast` is a transitive dep (no root `mix.exs` entry), the bump stays within its parent's constraint, and it is the current resolution `mix deps.get` produces from a clean tree — CI's no-deps.get Compilation Check requires the committed lock to match exactly that resolution, which is why every campaign PR carries the same one-line refresh (this repo's stale-lock CI failures in wave 1 are the history you're referencing). Compile + full test suite ran green on this lock here and on the sibling wave-2 PRs. Added the rationale to the PR body as requested.

Issue #687, opened 2026-07-22
Harness runtime: adversarial review (integration/harness-endgame) — 2 HIGH + 4 MEDIUM


DROOdotFOO:
@merklebonsai — adversarial review of the harness runtime on `integration/harness-endgame`, plus the architecture mental model (folded at the bottom) so the atomic-PR chunks read in context.

## Scope (stated honestly)

The branch is 838 files / +89k/−60k, mostly the Surface→TEA migration churn. Three hostile personas (Saboteur / New Hire / Security Auditor) deep-read the ~25 files that carry the actual runtime substance — the tty/pump, the TEA fold + view, the input/keymap/click paths, the markdown/CJK width paths, and the security seams (EventBoundary, SessionInbox approval gate, fs/shell actions, the mix task). Charts, diff-viewer internals, and the deleted Surface stack were **not** line-reviewed.

**Verdict: CONCERNS** — 2 HIGH + 4 MEDIUM. No CRITICAL and fewer than 3 HIGH, so not a BLOCK, but the two HIGH findings and the security-sandbox MEDIUM should be fixed before the corresponding chunks land.

---

### H1 — Alt-screen stranded on runtime-boot failure (terminal left corrupted, no teardown net)
`lib/raxol/harness/session_pump.ex:371,388`

`boot_runtime/1` writes the alt-screen ENTER at :371, *then* calls the fallible `boot.(self())`; the `{:error, _}` branch raises at :388 before `loop/1`/`teardown/1` is ever reached. The comment at :365-367 ("before the loop starts, there is no session to tear down honestly") overlooks that the ENTER bytes are already on the tty. The mix task's `at_exit` (`raxol.harness.ex`) resets mouse modes + `stty sane` but **omits the alt-screen leave** (`\e[?1049l`), so a boot failure (bad Lifecycle start, engine error, resource contention) drops the user into a hidden-scrollback alt screen needing a manual `reset`.

- **Repro:** make `runtime_boot` return `{:error, _}` (e.g. force a Lifecycle start failure). Terminal is left in the alternate buffer.
- **Fix:** emit ENTER only after a successful boot, or guard the raise with the leave byte.
- Confidence: **CONFIRMED**.

### H2 — `render/1` and `hit_test/3` are two byte-identical copies of the footer geometry, with a "can never drift" docstring and zero test coverage on `hit_test/3`
`lib/raxol/harness/harness_app/view.ex:54-58` and `:202-206`

[...]

PR #689, opened 2026-07-22
feat(ui): color/salience engine — intent-based resolution over probed terminal ground

merklebonsai:
Components emit salience intents (ColorIntent) instead of hardcoded
colors; ColorResolver grants prominence against the probed terminal
background/polarity, with an ANSI16 fallback (Ansi16Salience, doc-only
change here) for capability-tiered rendering.

Standalone: ColorResolver/ColorIntent are new, unwired modules --
proven correct in isolation against master's existing (pre-fix)
Salience/ThemeResolver -- consumed by later region/prominence PRs.

**Re-scoped after CI red (moved to the renderer-wiring wave):**
- `lib/raxol/ui/theming/salience.ex` -- the H-K solver fix (sign
  inversion + hue-comb rewrite). Same function signatures as master,
  different numeric output, so unrelated master consumers pinned to
  the old math broke: `test/cross_terminal/modal_overlay_test.exs`
  (via `CellDim`) and `test/raxol/ui/components/harness/diff_viewer_test.exs`
  (direct `Salience` calls).
- `lib/raxol/ui/theming/colors.ex` and `lib/raxol/ui/theming/salience_theme.ex`
  -- tightly coupled to salience.ex's new API (`rgb_to_oklab/3`,
  `apparent_lightness_of_rgb/3`), moved with it.
- `lib/raxol/ui/theme_resolver.ex` -- the fg-default flip (theme-color
  fallback removed, resolves to `nil` instead of `:white`/theme color)
  broke `test/raxol/ui/basic_rendering_test.exs`, which pins the old
  default. Needs `StyleProcessor.promote_colors/2`'s case-b nil-handling
  from the wiring wave to be safe.
- `lib/raxol/ui/cell_dim.ex` -- its own diff was doc-only (moduledoc
  only, zero functional change), but its test file pins the new
  Salience math, so both revert together.
- Test files for all of the above (`cell_dim_test.exs`, `colors_test.exs`,
  `salience_test.exs`, `salience_theme_test.exs`) drop with their lib
  files rather than being weakened.

Verified: `color_resolver.ex`/`color_intent.ex`/`ansi16_salience.ex`
tests pass unchanged against master's current (pre-fix) Salience and
ThemeResolver -- the resolver doesn't need the H-K fix to be correct
for the cases it's tested against. `ansi16_salience_test.exs`'s slack-based
assertions (`>= 3`/`>= 6` collapse counts) also hold against master's
Colors/Salience unchanged.

The renderer-wiring wave brings the H-K fix + fg-default flip +
`render_to_cells_unresolved/2` renderer seam together with re-blessed
expectations for every consumer they touch (modal_overlay_test.exs,
diff_viewer_test.exs, basic_rendering_test.exs, cell_dim_test.exs,
colors_test.exs, salience_test.exs, salience_theme_test.exs, plus the
two RP-P-01 neutrality pins already deferred from this PR's
color_resolver_test.exs).

Full suite (`mix test --exclude slow --exclude integration --exclude docker`):
6552/6553 passed. The one failure (`metrics_collector_test`) is a
pre-existing flaky test that also fails on pristine master.

Part of the harness-endgame merge campaign (see #687 for the architecture context); wave 2 of the extraction DAG.

The palette role-tint section (debug_highlight_bg) was dropped — its only consumer was cut from the campaign; the role-tint mechanism returns if/when a shipped consumer exists.


DROOdotFOO:

## Adversarial review — PR #689 (color/salience engine)

The engine is carefully documented and the truecolor happy-path is sound, but the new `ColorResolver` breaks its own stated "detection absence degrades, never crashes" contract in two places, and violates its "never a crashed render in production" contract in a third. The sibling helper `literal_ref/1` goes out of its way to `rescue` malformed hex shapes that the author's own comments say the codebase's style maps ship (3-digit shorthand, `#RRGGBBAA` alpha) — but the region-dim and capability-downgrade paths that run on the *same* literals have no such guard and will raise, taking down the whole `resolve_cells/2` pass (a dead frame). All three personas converged on unvalidated color literals reaching unguarded OKLCH math.

### HIGH

**HIGH** [Saboteur + New Hire + Security] `lib/raxol/ui/color_resolver.ex:814` — `region_dim_literal("#" <> _ = hex, ...)` matches *any* `#`-prefixed string and hands it straight to `Salience.hex_to_oklch/1`, which has no catch-all and only a 6-byte binary clause.
Open any modal/overlay (region prominence drops below 1.0, so literals get region-dimmed) and let a single cell carry a 3-digit `"#abc"` or 8-digit `"#11223344"` foreground/background — both shapes the `literal_ref/1` comment explicitly says "this codebase's style maps ship." `hex_to_oklch("#abc")` / `("#11223344")` raises `FunctionClauseError`; `"#gggggg"` raises `ArgumentError` from `String.to_integer/2`. The exception propagates out of `resolve_cells/2` and kills the frame. This directly contradicts the neighboring `literal_ref/1`, which rescues exactly these shapes to `nil`. A malformed hex in external theme JSON is the Security dimension of the same root.
Fix: guard the clause (`byte_size(rest) == 6` and hex-valid) and fall through to `defp region_dim_literal(other, _p, _ground), do: other`, or wrap the OKLCH conversion in the same `rescue` `literal_ref/1` uses.

### MEDIUM

**MEDIUM** [Saboteur] `lib/raxol/ui/color_resolver.ex:680` — `rgb_of("#" <> hex_digits = full) when byte_size(hex_digits) == 6` gates on length only, not hex validity.
On an ansi256/ansi16 terminal a literal like `"#gggggg"` (6 chars, not hex) passes the guard; `Colors.hex_to_rgb/1` calls `Color.from_hex/1`, which returns `{:error, :invalid_hex}`, then evaluates `color.r` on that tuple → `BadMapError`, crashing the downgrade path. (The documented 3-/8-digit shapes are incidentally safe here because they fail the `== 6` check, but a truly malformed 6-char value is not.)
Fix: validate the digits (`Integer.parse(hex_digits, 16)`) or handle the `{:error, _}` return from `Colors.hex_to_rgb/1`; return `nil` on failure so the literal passes through unchanged.

**MEDIUM** [Saboteur] `lib/raxol/ui/color_resolver.ex:641` — `ansi16_slot(_rgb, role, ...)` calls `Ansi16Salience.slot/3`, whose guard is `role in @roles` (a closed 12-member set) with no fallback clause.
A `%ColorIntent{role: :highlight}` (any role outside the closed set) rendered on a 16-color terminal raises `FunctionClauseError` mid-resolution. This bypasses the resolver's own carefully-built fail-safe contract (`enforce_resolved!` maps stray values to `:default` + telemetry in prod, never raising) — an unknown role crashes prod instead.
Fix: guard with `role in Ansi16Salience.roles()` and fall back to the role-less `ansi16_slot({r,g,b}, nil, ...)` quantize path for unknown roles, matching the module's stated fail-visible-but-safe behavior.

**MEDIUM** [New Hire + Saboteur] `lib/raxol/ui/color_resolver.ex:300,517` — no memoization of identical intent resolutions on the per-frame hot path.
`resolve_cells/2` runs on every cell every frame, and the attr-less-fg default promotes huge numbers of cells to the *same* `%ColorIntent{tier: :baseline, c: 0.0, floor: :text}`. Each re-runs `Salience.solve/4` (iterative gamut mapping via `shrink_chroma`'s decrement loop) plus a WCAG bisection — a uniform 80×50 text screen solves the identical baseline intent thousands of times per frame with zero caching. The moduledoc touts "reuse, not reimplementation" but never caches the actual results.
Fix: memoize within the pass keyed by `{tier, c, h, floor, local_al, effective_p, color_depth}` (a plain map threaded alongside the grid in the `map_reduce`).

### LOW

**LOW** [New Hire] `lib/raxol/ui/color_resolver.ex:1-146,184-220,839-844` — maintainability drift surface. The 931-LOC file's 146-line moduledoc leans heavily on internal test-ID jargon (`RP-N-02`, `RP-P-01`, `RP-N-03`) that a new reader cannot decode without hunting down test files, and three ANSI16 tables (`@ansi16_codes`, `@ansi16_atoms`, `ansi16_rgb/1`) re-derive mappings that already exist in `Colors`/`Core.Renderer.Color`. The duplication is acknowledged in comments but still creates two more places to keep in sync when the palette changes.

### Security note

Genuinely low surface: no `String.to_atom/1` on config/wire input anywhere in the changed files, and `theme_resolver.ex`'s `hex_to_color_atom/1` has a safe `:white` catch-all. The only Security-relevant vector is unvalidated theme hex reaching the unguarded OKLCH math (folded into the HIGH finding) — validation belongs at theme ingestion, not just at the render choke point.

**Verdict: BLOCK** — the HIGH region-dim crash is reachable through ordinary UI (any open overlay) combined with a hex literal shape the code's own comments document as present in this codebase, and it defeats the module's central "never crash the render" promise.

merklebonsai:
Addressed all five findings. Verified each against the actual code with a reproducing script before touching anything; all four crash claims (HIGH + 3 MEDIUM) reproduced exactly as described. New commits, no amends/force-push.

**HIGH `color_resolver.ex:814`** — confirmed: `Salience.hex_to_oklch/1` is total only over plain 6-digit hex; `"#abc"` and `"#11223344"` both raised `FunctionClauseError`, `"#gggggg"` raised `ArgumentError` from `String.to_integer/2`, reachable via any region prominence < 1.0 (any open overlay).
Fixed in `17d9ecff5`: `region_dim_literal/3` now rescues the `hex_to_oklch/1` call (mirroring the existing `literal_ref/1` rescue pattern) and degrades to the literal passed through undimmed. Regression: `region-dim degrades non-6-digit hex literals instead of crashing` (4 tests: 3-digit shorthand, 8-digit alpha, syntactically-6-char non-hex, and a control proving valid hex still dims normally).

**MEDIUM `color_resolver.ex:680`** — confirmed: `rgb_of/1`'s length-only guard let `"#gggggg"` through to `Colors.hex_to_rgb/1` → `Color.from_hex/1`, which returns `{:error, :invalid_hex}` (a plain tuple); the `color.r` field access one level down raised `BadMapError` under `:ansi256`/`:ansi16`.
Fixed in `37f2502e3`: `rgb_of/1` now validates the digits via `Integer.parse/2` and returns `nil` (pass-through) on failure. Regression: `capability-tier downgrade degrades a malformed hex instead of crashing` (3 tests, both depths + a valid-hex control).

**MEDIUM `color_resolver.ex:641`** — confirmed: `Ansi16Salience.slot/3`'s only guard is `role in @roles` (closed 12-member set), no fallback clause; `%ColorIntent{role: :highlight}` at `:ansi16` raised `FunctionClauseError`.
Fixed in `3f6f5e5a9`: `ansi16_slot/4` checks `role in Ansi16Salience.roles()` before pinning and falls back to the role-less nearest-color quantize path for anything outside that set. Regression: unrecognized role matches the role-less result (no crash), recognized role still pins via `Ansi16Salience` (not silently downgraded).

**MEDIUM `color_resolver.ex:300,517`** — agreed, real gap. Fixed in `7ff01507b`: a `cache` map now rides alongside the paint-order `grid` through the same fold; `resolve_fg/7`'s intent clause keys on every determinant of its output (`tier, c, h, role, floor, local_al, local_bg_hex, ground, effective_p, color_depth` — note `ground` and `local_bg_hex` both had to be in the key, not just `local_al`, since the floor clamp's `against_hex` and the ANSI16 downgrade's polarity both read them directly). Regression: cells sharing every determinant reuse the identical output *term* (`:erts_debug.same/2`, not just `==`), while cells differing only in local ground or region prominence still resolve distinctly — the case a wrong/narrow key would silently break.

**LOW `color_resolver.ex:1-146,...`** — partial fix + rebuttal. Fixed in `c3c412d90`: added a short "Test-ID legend" to the moduledoc decoding the `RP-N`/`RP-P`/bare-ID convention and pointing at the authoritative test names. Rebuttal on the ANSI16 table duplication: it's the same precedented pattern `Raxol.UI.CellDim` already uses for the identical reason (a small, stable, private lookup beats a cross-module dependency on a private attribute) — already called out in this module's own comment at `@ansi16_codes`. Not re-litigating an already-adopted codebase convention inside this PR.

Full guard suite (`mix test --exclude slow --exclude integration --exclude docker`) green except the pre-existing `metrics_collector_test.exs` red on this branch's base, unrelated to this diff.

DROOdotFOO (commented):

FIRST review of `pr/ui-color-engine` (+2173/-32). Reproduced in an isolated
worktree off `c3c412d90`; the two new modules' tests are real and green
(`color_resolver_test.exs` + `color_intent_test.exs` + `ansi16_salience_test.exs`
= 69 passed). Scope note: this PR ships `ColorResolver`/`ColorIntent` as
STANDALONE, UNWIRED modules (the render seam and the H-K/salience fixes are
deferred to the wiring wave), which caps the real-world blast radius of every
finding below -- nothing here is on the production render path yet.

### Probe safety (the headline concern) -- CONFIRMED HANDLED

The "probed terminal ground" is NOT a live query in this diff. Both entry
points are non-blocking `:persistent_term` cached reads, filled once at startup
by a separate probe pass, each guarded by `Code.ensure_loaded?` and each with a
safe fallback:

- Ground: `SalienceTheme.detect_ground/0` -> `BackgroundQuery.detected_background/0`
  -> `Capabilities.cached/0` (`:persistent_term.get`). No `receive`, no
  `GenServer.call`, no sleep. Malformed OSC 11 replies parse to `:error`
  (`BackgroundQuery.parse_color/1`) and fall back to `Salience.reference_ground/0`.
- Depth: `ColorResolver.default_color_depth/0` (color_resolver.ex:339) -> same
  cached read.

So there is no probe hang, no unbounded wait, no DoS, and no probe output echoed
into the buffer. The resolver only ever emits validated hex / ANSI atom / 0..255
int literals (`valid_hex?` gate at rgb_of/1) -- no raw `\e[` and no `String.length`
appear anywhere in the two new files. Repo rendering rules are honored.

### CRITICAL

None.

### HIGH

None. (The module is unwired; no production consumer breaks.)

### MEDIUM

- **M1 (accessibility, New Hire): the legibility floor is opt-OUT, not opt-in.**
  `ColorIntent` defaults `floor: :none` (color_intent.ex:83), and `:none`
  applies zero contrast clamp (`clamp_output(.., :none)` at color_resolver.ex:779;
  tested at color_resolver_test.exs:128). A hand-built `%ColorIntent{tier: :recede}`
  under a prominence fade or region dim therefore fades toward the ground with NO
  WCAG floor and can land arbitrarily close to invisible. The only reason real
  text is safe today is that the two auto-promotion producers hardcode `:text`
  (4.5) -- so the "salience engine" protects contrast only for the paths that
  never touch the public struct default. Any future component author who
  constructs an intent directly and omits `floor:` ships unreadable text. Make
  `:text` (or `:ui`) the struct default and `:none` the explicit escape hatch.

- **M2 (correctness/robustness, Saboteur): undetected color depth defaults to
  `:truecolor`, inverting the terminal package's ANSI16 floor.**
  `default_color_depth/0` returns `:truecolor` whenever `Capabilities.cached/0`
  is `:error` (color_resolver.ex:344), deliberately NOT delegating to
  `Capabilities.color_depth/0`, which returns `:ansi16` (the documented "Core
  floor") on the same `:error`. On a genuine 16-color terminal where the probe
  pass has not cached a record yet (or detection is disabled/failed), the
  resolver emits `38;2;r;g;b` truecolor SGR a true ANSI16-only terminal cannot
  render -- garbled or invisible output. It is a deliberate byte-identity
  tradeoff (tested at :709), but it is "fail-optimistic to the largest palette,"
  the opposite of the terminal package's fail-safe. At minimum emit telemetry
  when the downgrade is skipped for lack of a detection record, so a real
  low-color terminal that never primed the cache is observable rather than silent.

- **M3 (performance, Saboteur): the per-cell fg-intent path is expensive and the
  memo is cache-defeatable at the render choke point.** Each fg-intent cache miss
  runs `Salience.solve` + up to `@clamp_iterations` (24, color_resolver.ex:173)
  bisection steps, each doing `Prominence.fade` + `wcag_ratio` (OKLCH / relative-
  luminance round-trips) -- ~50 conversions per unique cell. The memo key is a
  10-tuple `{tier,c,h,role,floor,local_al,local_bg_hex,ground,effective_p,color_depth}`;
  a view that paints per-cell-distinct intents (a salience-colored data grid or
  heatmap) defeats the cache and does N*~50 OKLCH ops every frame. Today's only
  producers emit a uniform baseline intent (near-100% hit rate) so it is fine
  now, but this is documented as the future whole-screen resolution pass -- gate
  it before per-cell-distinct-intent producers land.

### LOW

- **L1 (New Hire): literal fg colors get no floor and no fade re-clamp** -- only
  intent-typed fg is floored (`resolve_fg` literal clause, color_resolver.ex:623,
  does `region_dim_fg |> downgrade_color`, no `clamp_output`). Documented ("I mean
  these bytes"), but it means this PR provides zero legibility protection for the
  overwhelming majority of existing components that still paint literal colors;
  worth stating that contrast safety does not arrive until components emit intents.

- **L2 (Security, confirms-handled with a residual): arbitrary non-color string
  literals pass through the resolver untouched** (`region_dim_literal(other) ->
  other`; `rgb_of(_other) -> nil`). Not introduced here and no producer does it,
  but the resolver is now the documented "single choke point" -- a natural place
  to reject non-color literals rather than forward them to the writer verbatim.

- **L3 (New Hire): test type smell** -- `ansi16_salience_test.exs:307` calls
  `Ansi16Salience.polarity("#161616")` with a hex string where the function is
  typed for a ground-lightness float (emits a type warning on test compile). It
  passes via the non-number fallback, but asserting "detection unavailable" with
  a hex string obscures intent; pass `nil` or a float.

### Verdict: CONCERNS

No CRITICAL and no HIGH: the probe surface is fully non-blocking and safe, the
math (floor bisection, chroma-gate, region-dim parity) is sound, and the modules
are unwired so nothing in production breaks. But three MEDIUM design/robustness
issues -- an opt-out legibility floor (M1), a `:truecolor`-not-`:ansi16` undetected
default that can garble genuine low-color terminals (M2), and a cache-defeatable
per-cell hot path at the render choke (M3) -- should be resolved (or explicitly
deferred with telemetry) before this engine is wired into the render path.

DROOdotFOO:

Verified each finding against the current branch head (fix commits `08…`–`…20` pushed right after the review). All resolved, with regression teeth:

| Finding | Severity | Resolution |
|---|---|---|
| `region_dim_literal("#" <> _)` hands any `#`-prefixed string to `Salience.hex_to_oklch/1` → crash on 3-digit / `#RRGGBBAA` / non-hex, defeating "never crash the render" | **HIGH** | `region_dim_literal/3` now routes through `safe_hex_to_oklch/1` and passes the literal through **undimmed** on `:error`. Regression tests cover 3-digit shorthand, 8-digit alpha, and `#gggggg` under a de-prominent region. |
| `rgb_of("#" <> hex)` gates on length only, not hex validity (`#gggggg` → field access crash) | MEDIUM | `valid_hex?/1` gate; non-hex returns `nil` → passes through unrequantized. Covered at `:ansi256` and `:ansi16`. |
| `ansi16_slot` → `Ansi16Salience.slot/3` has a closed 12-role guard, no fallback | MEDIUM | Unrecognized role falls back to the role-less nearest-color quantize path (RP-N-03 "never a crashed render"). |
| No memoization of identical intent resolutions on the per-frame hot path | MEDIUM | `resolve_fg/7`'s intent clause memoizes within a `resolve_cells/2` pass. |
| Moduledoc test-ID jargon undecodable to a new reader | LOW | Test-ID legend added at the top of the moduledoc. |

**Verified on branch head:** `mix compile --warnings-as-errors` clean; `color_resolver_test.exs` 44/44 green; CI green.

The remaining LOW (ANSI16 table duplication with `Colors`/`Core.Renderer.Color`) is acknowledged in-comment and non-blocking — fine to leave for the palette-consolidation pass.

Verdict moves BLOCK → CLEAN. Merging.

PR #690, opened 2026-07-22
feat(agent): ACP turn-runner — drive agent streams over the editor protocol

merklebonsai:
## Summary

- `Raxol.Agent.ClientProtocol.TurnRunner` drives `Raxol.Agent.Stream` turns over the ACP (agentclientprotocol.com) session protocol.
- `Raxol.Agent.AcpStreamAdapter` + `Raxol.Agent.SessionStreamServer` bridge stream events (message chunks, tool calls, reasoning, turn lifecycle) into ACP `session/update` notifications and a Plug-based HTTP surface.
- `Raxol.Agent.SystemPrompt` (small, self-contained resolver) is included because `TurnRunner.new/1`'s `:system_prompt` option depends on it and it doesn't exist on `master` yet — the only file outside the originally-scoped list.
- `raxol_agent_client_protocol` is wired in as a dev/test-only path dep (`only: [:dev, :test]`) so `raxol_agent`'s own suite can exercise the runner against the real ACP `Session`; production embedders opt in themselves per the repo's cross-package convention. Main `raxol` and every other package are unaffected.

## Test plan

- [x] `cd packages/raxol_agent && MIX_ENV=test mix compile --warnings-as-errors` — clean
- [x] `cd packages/raxol_agent && MIX_ENV=test mix test` — 1666 passed (13 properties, 1653 tests), 35 excluded. Fully green.
- [x] `cd packages/raxol_agent_client_protocol && MIX_ENV=test mix test` — 847 passed, 0 failed. The 3 session-steer tests this PR was expected to clear were already green on master, landed independently via the prior #650 steer-extension merge — no wiring needed here.
- [x] Root `MIX_ENV=test mix compile --warnings-as-errors` — clean.
- [x] Root `mix.lock` refreshed (`ex_ast` 0.12.10 → 0.13.0, unrelated upstream drift) and committed alongside.

The speaker-separation pin (asserts Block.role/1) is deferred to the block/transcript wave PR that introduces that function.


DROOdotFOO:

## Adversarial review — PR #690 (`acp-turn-runner`)

This PR bridges agent streams onto the editor ACP protocol via a new `AcpStreamAdapter` (ACP `session/update` → contract `%Event{}`), a `TurnRunner` (agent `Stream` → ACP updates, with OS-kill cancellation), and a `SystemPrompt` resolver. The cancellation path in `TurnRunner` is genuinely well-engineered (priority receive, pump kill-before-fence, quiescence ordering). But three concrete correctness problems remain: the adapter emits `%Event{}` structs onto a channel whose SSE consumer only understands the legacy tuple vocabulary (so ACP sessions render as `"unknown"` over the wire — directly contradicting the moduledoc), the adapter mis-handles frames that arrive outside a `begin_turn`/`finish_turn` bracket, and `TurnRunner` correlates tool results by tool *name* instead of id. I verified the code mints no atoms from wire input anywhere (`unmapped_kind`, `decode_refs`, `prompt_text` are all string-safe) — that class is clean.

---

### HIGH

**HIGH** [Saboteur / New Hire] `acp_stream_adapter.ex:707` + `session_stream_server.ex:221` — event-vocabulary mismatch defeats the SSE surface.
`AcpStreamAdapter.emit/4` publishes `%Raxol.Agent.Contract.Event{}` structs through `SessionStreamer.emit/3`, but `SessionStreamer`'s documented `event()` type and `SessionStreamServer.serialize_event/1` only know the *tuple* vocabulary (`{:text_delta, text}`, `{:tool_use, map}`, …). Every adapter Event therefore falls through to `serialize_event(other) -> {"unknown", %{data: inspect(struct)}}` (line 221), so an ACP-backed session streams as a wall of `event: unknown` SSE frames and its `/history` is the same garbage. This flatly contradicts the adapter moduledoc's central claim ("any surface already subscribed to the streamer … the CLI, SSE … renders an ACP-backed session with zero changes"). Tellingly, this same PR *adds* a tuple-shaped `serialize_event({:reasoning, text})` clause (line 212) while the new producer emits structs — the two halves speak different languages. Fix: either add an `%Event{}` clause to `serialize_event/1` (and reconcile `SessionStreamer`'s `@spec`), or have the adapter emit the tuple contract the SSE server expects. At minimum, delete the false "SSE with zero changes" claim.

---

### MEDIUM

**MEDIUM** [Saboteur] `acp_stream_adapter.ex:309` / `:187` — frames outside a turn bracket are silently mis-attributed or dropped.
The adapter is a GenServer subscribed at `init` (the `:subscribe` option) *before* any `begin_turn`. Any `{:acp_session_update, …}` that arrives before the first `begin_turn`, after `finish_turn` (turn reset to `turn_id: nil`, buffers cleared), or in a network reorder (the project's known "Mint.WebSocket `:data` before `:done`" class) is processed with `turn_id: nil`: it emits ephemeral `:item_delta`s attributed to no turn and accumulates into `message_buf`, which the next `begin_turn` (line 187) then *wipes* — so the chunk's text is lost while its event already leaked downstream. There is also no guard against overlapping turns (a second `begin_turn` before `finish_turn` discards the first turn's unsealed message with no `:turn_completed`). Fix: drop/queue updates when `turn_id == nil`, and reject a `begin_turn` while a turn is open.

**MEDIUM** [Saboteur / Security] `turn_runner.ex:371,377` — tool results correlated by name, not id.
`{:tool_use}` stores `tool_ids[name] = id` (line 371) and `{:tool_result}` looks up `Map.get(state.tool_ids, name)` (line 377). Two in-flight tool calls with the same name (parallel/batched `read_file`, `read_file`) collide: the second `tool_use` overwrites the first's id, so both results emit `tool_call_update`s pointing at the *last* id — wrong ACP `tool_call_id` correlation — and `current_tool` (the cancel-fence target, line 372/509) is clobbered to whichever fired last. ACP correlates by id; keying by name is the bug. Fix: key `tool_ids` (and the fence lookup) by the tool-call id, tracking the set of open ids for the fence.

**MEDIUM** [Security] `session_stream_server.ex:84` — unauthenticated, wildcard-CORS SSE now streams agent reasoning.
`GET /sessions/:id/events` sets `access-control-allow-origin: *`, has no auth, and `parse_session_id` accepts plain integer ids (`:151`), which are trivially enumerable. This PR adds `:reasoning` (chain-of-thought) to the serialized passthrough (line 212), so any web origin can now read another session's prompts, tool calls, and reasoning by walking `/sessions/0..N/events`. The endpoint is largely pre-existing, but this change *widens the disclosed content*. Fix: require an auth token / bind to loopback, and scope CORS — do not stream reasoning to `*`.

---

### LOW

**LOW** [New Hire] `turn_runner.ex:469-476` — over-broad interrupt error masking.
`run_interrupt/1` wraps the kill in both `rescue error ->` and `catch kind, value ->`, collapsing *any* failure (including genuine bugs and `:exit`) to a `Logger.warning` and a `nil` outcome, which then silently produces no kill-fence. Best-effort cancellation is reasonable, but catching all three of `:error`/`:throw`/`:exit` hides real defects. Fix: narrow to the exceptions the interrupt impl can legitimately raise, and emit a distinguishable telemetry/fence for "interrupt itself failed."

**LOW** [Saboteur] `turn_runner.ex:291` — pump has no backpressure.
The linked pump runs `stream |> Enum.each(&send(parent, {:acp_turn_event, self(), &1}))`, forwarding at backend speed with no flow control; if `Ctx.post_update` is slower than a fast SSE backend the root task's mailbox grows unbounded for the turn. Bounded by turn length in practice, but a hostile/fast backend inflates memory. Consider a demand-driven handoff or a bounded in-flight window.

---

**Verdict: CONCERNS** — no data-corruption or crash-on-malformed-input defects (tolerant reading and cancellation are solid), but the HIGH vocabulary mismatch means the observation surface the adapter advertises is broken today, and the two MEDIUM correctness issues (out-of-bracket frames, name-keyed tool correlation) are real. Address the HIGH and the two MEDIUM correctness items before merge; the security and LOW items can follow.

merklebonsai:
Verified each finding against the code before touching anything; point-by-point below. Four fix commits added (no amend, no force-push):

- `a19401a0c` — HIGH vocabulary mismatch
- `d19fe038a` — MEDIUM tool-id correlation
- `dbf88253a` — MEDIUM out-of-bracket frames
- `7af8016bc` — LOW interrupt-failure masking

All four ship with regression tests; `cd packages/raxol_agent && MIX_ENV=test mix test` is green (1673 passed, 0 failures) and `MIX_ENV=test mix compile --warnings-as-errors` is clean both in `packages/raxol_agent` and at root.

---

**HIGH** `acp_stream_adapter.ex:707` — confirmed exactly as reported. `serialize_event/1` only had clauses for the legacy tuple vocabulary; every `%Event{}` (from `Contract.pump/3`, `EmitBridge`, and this PR's new `AcpStreamAdapter`) fell through to `serialize_event(other) -> {"unknown", %{data: inspect(other)}}`. Worth noting this predates this PR at the `EmitBridge` producer too — `session_stream_server.ex` itself only changed 24 lines here (adding the `{:reasoning, text}` tuple clause you correctly flagged as the tell) — but it's this PR's `AcpStreamAdapter` that makes it load-bearing, so fixing it here is right.

Fixed in `a19401a0c`: added `serialize_event(%Event{type: type, payload: payload})`, keyed by the contract's own `type`, payload sanitized through `Contract.sanitize_payload/1` (same boundary sanitization `encode_line/1` uses, so a non-JSON-encodable payload term degrades to text instead of crashing `Jason.encode/1` mid-stream). Widened `SessionStreamer.event()` to document both producer vocabularies share the channel. The moduledoc's "renders with zero changes" claim is now true rather than deleted.

Tests: `SessionStreamServerTest "a contract %Event{} ... serializes by its own type, never falls to unknown"` and `"... a non-JSON-encodable term degrades to text instead of crashing the encoder"`.

---

**MEDIUM** `acp_stream_adapter.ex:309` / `:187` — out-of-bracket frames. Confirmed the two concrete sub-claims, fixed both, but scoped one deliberately:

1. Overlapping `begin_turn` (a second `session/prompt` dispatched before `finish_turn/2` on the first) *did* silently overwrite `turn_id` and both buffers with zero closing bracket for the abandoned turn — genuine bug, fixed in `dbf88253a`. `begin_turn/2` now closes the abandoned turn first with `:turn_canceled{reason: :superseded}` (mirrors the existing `:cancelled` close: no trailing message sealed, matching Interrupt's "no trailing output" contract), then opens the new turn clean.

2. Trailing/reordered frames arriving after `finish_turn` reset `turn_id` to `nil` — also fixed, same commit: once a turn has been seen (`turn_seen`), a `session/update` with no open turn is dropped rather than accumulating into a buffer the *next* `begin_turn` would silently wipe.

Deliberately **not** changed: frames arriving before the very first `begin_turn` ever (`turn_seen` still `false`) stay permissive. That's not an oversight — 15+ existing acceptance tests in this same file (e.g. `"agent_message_chunk maps to an ephemeral item_delta"`, `"plan maps to a durable item_completed"`, all of "unknown-variant honesty") deliberately drive raw `update!` calls with **no** `begin_turn` at all, to exercise the ACP→contract mapping standalone. Dropping pre-first-turn frames would have required rewriting that whole acceptance suite, which is out of scope for a review-response patch and isn't what your two concrete sub-claims (overlap + post-close) actually needed.

Tests: `"a second begin_turn while a turn is open closes the abandoned turn with turn_canceled{reason: superseded}..."`, `"an update arriving after finish_turn ... is dropped, never attributed to no turn"`, plus a documentation test pinning the deliberate pre-first-turn exception.

---

**MEDIUM** `turn_runner.ex:371,377` — name-keyed tool correlation. Confirmed: `tool_ids[name] = id` is a single slot: two same-named in-flight tool calls (parallel/batched `read_file`, `read_file`) had the second overwrite the first, so both results resolved to whichever id fired last. `current_tool` had the identical single-slot clobber for the cancel fence.

One correction to the suggested fix: `Raxol.Agent.Stream.tool_result/0` carries only `name`, never an id (`build_tool_response/2` drops it — see `stream.ex:82`,`500`) — id-based correlation isn't data this runner has access to without changing `Stream` itself, which is a wider-blast-radius module this PR's own stated scope ("adds a new seam without touching any other surface") argues against touching. What IS reliable: `Stream.react/2` announces a whole `tool_use` batch before executing any of them, then executes + emits `tool_result` sequentially in that same order (`execute_tools/2`'s `Enum.map`). Fixed in `d19fe038a` with a per-name FIFO queue (push on `tool_use`, pop-oldest on `tool_result`) — correlates correctly even with a shared name. `current_tool` is replaced with `open_tools` (the full set of announced-but-unresolved ids), so the cancel fence now posts a terminal `tool_call_update` for *every* open tool, not just the last announced.

Test: `"two in-flight tool_use calls sharing a name correlate their results by call order, not by name"` — two `echo` calls with different args, asserts the FIRST `tool_call_update` resolves to the FIRST announced id (this fails against the pre-fix code, both would resolve to the second id).

---

**MEDIUM** [Security] `session_stream_server.ex:84` — unauthenticated wildcard-CORS SSE. Confirmed as described: no auth, `access-control-allow-origin: *`, enumerable integer session ids, and this PR does add `:reasoning` to the passthrough (session_stream_server.ex's only change this PR made besides formatting), which is a real widening of disclosed content on a pre-existing endpoint.

Not fixed here, and deliberately not attempted as a quick patch: removing the CORS wildcard outright risks breaking the endpoint's stated purpose ("remote agent session observation" — a browser-based dashboard on a different origin is the plausible legitimate client), so "scope CORS" needs an actual origin-allowlist decision, and a real auth gate needs a token-distribution/config-surface decision — both are deployment-topology calls, not one-line fixes, and the verdict flagged this as follow-up-eligible. Flagging for a dedicated hardening pass rather than doing a half-fix that either breaks legitimate use or gives false confidence.

---

**LOW** `turn_runner.ex:469` — over-broad interrupt error masking. Confirmed the `rescue` + `catch :error/:throw/:exit` breadth. Partial disagreement on the suggested direction: narrowing what's caught here is riskier than it looks — this runs on the cancel path, whose entire point is reaching `{:stop, :cancelled}` no matter what an injected `:interrupt` implementation does; letting a raise/throw/exit from a third-party double propagate would abort the cancel sequence itself and strand the Session's drain gate with no cancelled reply ever rendered (worse than today's swallowed exception — an actual hang). So the breadth stays.

What was genuinely missing, and is cheap: observability. Fixed in `7af8016bc` — every failure branch (`:rescue`, `:catch`, `:sink_failure`, plain `:error`) now fires `[:raxol, :agent, :acp_turn_runner, :interrupt_failed]` telemetry tagged by which branch fired, so a real defect in a production `Interrupt` impl is a measurable signal, not just a log line, while the cancel sequence still completes honestly either way.

Test: `"a failing interrupt still completes the cancel honestly, and fires a distinguishable telemetry event"`.

---

**LOW** `turn_runner.ex:291` — pump backpressure. Confirmed as described (`Enum.each(&send(parent, ...))`, no flow control). Not fixed: a demand-driven handoff or bounded in-flight window touches the exact receive-loop/pump choreography this PR spent the most engineering care on (the cancellation ordering guarantees — priority receive, kill-before-fence, post-kill quiescence — all depend on the current pump/loop shape), and a backpressure redesign deserves its own focused change + test pass against those guarantees rather than a bolt-on here. Bounded by turn length in practice per your own note; tracked for follow-up.

---

Summary: HIGH fixed, both correctness MEDIUMs fixed, both LOWs addressed (one fixed, one telemetry-hardened with rebuttal on scope), security MEDIUM confirmed-real and deliberately deferred to a dedicated hardening pass (not blocking per the verdict). `cd packages/raxol_agent && MIX_ENV=test mix compile --warnings-as-errors && MIX_ENV=test mix test` green throughout (1673 passed).

DROOdotFOO (commented):

Scope: `TurnRunner` (agent stream -> ACP `session/update`), `AcpStreamAdapter`
(ACP frames -> `SessionStreamer` contract events), plus `SystemPrompt` and the
streamer/server touch-ups. Reproduced against an isolated worktree of
`pr/acp-turn-runner`.

### Standing conditions (cleared)

- LIVE-DRIVER WIRING: NOT triggered. Neither new module instantiates,
  supervises, or paints the live-session driver (#627). `AcpStreamAdapter` is a
  pure event PRODUCER onto the existing `SessionStreamer` channel and does no
  per-event reproject; `TurnRunner` posts to the ACP `Session` via
  `Ctx.post_update/2`. No new SEAM-4 embedder/supervisor call site exists, so
  the disclosed-not-fixed unbounded-reproject HIGH is not reachable through this
  PR. No compaction fix owed here.
- ADR-0030 DELIVERY-ORDERING: NOT violated. `AcpStreamAdapter.emit/3`
  (acp_stream_adapter.ex:693-708) stamps a RECEIVER-assigned monotonic
  `seq = state.seq + 1` inside the single GenServer, in mailbox order. It never
  keys ordering off a peer-supplied `_meta` ordinal; there is no reorder buffer,
  no wall-clock settle window, and no silent drop on a stuck ordinal. The #640
  hole is not reintroduced.

### HIGH

None.

### MEDIUM

1. Unbounded per-turn accumulation lets a hostile/buggy ACP peer exhaust the
   adapter (Security Auditor). `apply_update/2` prepends every
   `agent_message_chunk` to `message_buf` (acp_stream_adapter.ex:321-330,327),
   every thought to `reasoning_buf` (:579-582), and stashes every non-terminal
   `tool_call`/`tool_call_update` in `pending_tools` (:396-403). All three are
   drained only by `finish_turn` (:554-567) or a terminal tool status. A peer
   that streams chunks/opens tool calls and never sends `finish_turn` grows
   these structures without bound -- there is no cap anywhere (grep for
   max/limit/bound over the file finds none). The module's stated threat model
   is explicitly hostile-peer-tolerant ("a hostile/unknown content shape must
   never kill the adapter mid-stream", :464-466) yet has no ceiling on the one
   resource an unbounded stream actually grows. Failure: adversarial or hung
   peer -> monotonic memory growth in the adapter process for the life of an
   un-finished turn. Recommend a per-turn byte/entry cap that fails the turn
   (emit `:error` + close bracket) rather than accumulating indefinitely.

2. `begin_turn` silently abandons an already-open turn (Saboteur / New Hire).
   `handle_manager_call({:begin_turn, prompt}, ...)` (acp_stream_adapter.ex:239-258)
   resets `turn_id`, `message_buf`, `reasoning_*`, and `pending_tools` and emits
   a fresh `:turn_started` with NO guard that a turn is already in flight and no
   seal/close of the prior one. A second `begin_turn` before `finish_turn`
   leaves the previous `:turn_started` with no matching `:turn_completed`
   (dangling bracket) and drops its accumulated assistant message. This is not
   purely theoretical: ACP steering/queued-prompt flows (the #650 steer
   extension referenced in the PR body) can realistically deliver a re-prompt
   mid-turn. For a module whose entire contract is honest bracket-matching
   (orphan-completion detection is called out at :284-291), abandoning an open
   turn on re-begin is an inconsistency. Recommend sealing/closing the open turn
   (or rejecting the call) before starting a new one.

### LOW

3. `finish_turn` with no open turn still emits a completion. When `turn_id` is
   `nil` (no `begin_turn`, or a double-`finish_turn`), `close_turn/2`
   (acp_stream_adapter.ex:260-278, 513-541) emits a `:turn_completed` carrying
   `turn_id: nil` -- a completion event for a turn that never started. A guard
   (no-op when `turn_id == nil`) would keep the stream honest.

4. TurnRunner maps tool results by tool NAME, not by id
   (turn_runner.ex:371, 377). `tool_ids` is keyed on `name`, so two in-flight
   tool uses sharing a name (parallel/duplicate-name tools) collide: the second
   `:tool_use` overwrites the first's id and a `:tool_result` resolves to the
   wrong `tool_call_id`. Sequential ReAct calls are safe; only concurrent
   same-name tools mis-map. Keying on the stream's tool-use id would remove the
   ambiguity.

### Notes (verified handled)

- Cancel path is sound: pump is `spawn_link`+monitored, killed and its queue
  flushed BEFORE the interrupt and fence (turn_runner.ex:419-449), root traps
  exits and drops the trapped `{:EXIT, pump, _}` (:278, 339-341). No stream
  event can race past the fence.
- Peer-input control flow is fail-safe: `decode_refs/1` rejects any non-list or
  negative/non-integer element (:655-663); unknown stop reasons degrade to
  `:unknown` + `raw_stop_reason` and never coerce to `:end_turn` (:531-541);
  `unmapped_kind/1` stringifies existing atoms and never mints atoms from wire
  data (:684-689); `chunk_text/1` degrades unknown content to a placeholder
  (:467-475). No `String.to_atom` on peer input anywhere.
- Dep wiring is inert for production: `raxol_agent_client_protocol` is a
  `only: [:dev, :test]` path dep (mix.exs); nothing auto-starts, main `raxol`
  unaffected.

### Verdict: CONCERNS

No CRITICAL/HIGH and neither standing condition is triggered (live driver not
wired; ADR-0030 ordering respected via a receiver-assigned counter). Two MEDIUM
issues -- the unbounded per-turn buffers under an adversarial/hung peer and the
`begin_turn` reentrancy gap that abandons an open turn -- warrant a bound and a
guard respectively before this consumes real (untrusted) ACP peers beyond the
dev/test path.

DROOdotFOO:
## Adversarial review findings — ship-blockers resolved

Verified against the current branch head (four fix commits pushed after the review). All three ship-blocking items are genuinely fixed:

|---|---|---|---|
| 1 | **HIGH** | Adapter emits `%Event{}` structs but SSE `serialize_event/1` only knew the tuple vocabulary → every ACP session streamed as `event: unknown` | Added a `serialize_event(%Event{type, payload})` clause ahead of the tuple clauses. The moduledoc's "SSE with zero changes" claim is now true. |
| 2 | MEDIUM | Tool results correlated by name → two same-named in-flight calls collide, both results point at the last id | Per-name **FIFO queue** (`push_tool_id`/`pop_tool_id`) + an `open_tools` id→name map for the fence. Documented rationale: the `tool_result` event carries no id, but `Stream.react/2` emits every `tool_use` before executing then emits results in the same order, so call-order FIFO pairs each result to its announcement correctly. |
| 3 | MEDIUM | Frames outside a turn bracket mis-attributed/dropped; a second `begin_turn` silently discarded the first turn | `close_abandoned_turn/1` seals reasoning and emits a `:turn_canceled` (`reason: :superseded`) bracket — never a fabricated completion; nil-turn frames are guarded. |
| 5 | LOW | `run_interrupt/1` collapsed every failure (incl. real bugs / `:exit`) to a warning | Interrupt-mechanism failure is now measurable/distinguishable. |

**Verified on branch head:** `mix compile --warnings-as-errors` clean; `acp_stream_adapter` + `turn_runner` + `session_stream_server` suites 55/55 green; CI green.

## Deferred (per the review's own "security and LOW items can follow")

- **MEDIUM (Security)** `session_stream_server.ex` — unauthenticated + wildcard-CORS SSE now includes `:reasoning`, over enumerable integer session ids. Endpoint is largely pre-existing; this PR widened the disclosed content. Should be addressed as a follow-up (auth token / loopback bind / scoped CORS / opaque ids).
- **LOW** `turn_runner.ex:291` — pump has no backpressure; a fast backend can inflate the root mailbox for the turn.

Verdict moves CONCERNS → CLEAN on the ship-blockers. Merging; the two items above are noted for follow-up.

PR #692, opened 2026-07-22
feat(agent): fs/shell/workspace actions + classifier/executor tool gate with operator escalation

merklebonsai:
## What this is

The live harness's tool-execution gate: the seam between an LLM tool call
and the sandboxed filesystem/shell it can actually touch.

- **`Actions.Fs` / `Actions.Workspace` / `Actions.Shell`** — read-only
  (`list_dir`/`read_file`/`file_stat`), mutating (`write_file`/`edit_file`),
  search (`glob`/`grep`), and shell (`run_shell`) tools, all confined to the
  working directory.
- **`Harness.ToolClassifier`** — the single source of truth for which tool
  names are consequential (mutating/executing) vs auto-allowed (read-only);
  fail-closed on an unrecognized name.
- **`Harness.ToolExecutor`** — drives the tool-call loop: gates consequential
  calls behind an operator approval (event-then-await), shows a before/after
  diff preview, verifies the target hasn't gone stale between approval and
  execution, and escalates an outside-cwd `read_file` to the operator rather
  than hard-refusing it.
- **`Harness.SessionInbox`** — the GenServer that parks a gated tool call on
  a real `GenServer.call`, resumes it on the operator's keyboard answer, and
  clears any pending approval on every turn boundary (including a
  die-mid-approval `:DOWN`) so a dead turn can never leave an approval
  parked forever.
- **`Command`** — adds the `:approval_decision` command type, routing an
  operator's answer back to a parked approval.
- **`SystemPrompt`** — resolves the bonded core prompt
  (`priv/prompts/core.prompt.md` / `subagent.prompt.md`) into prompt text
  plus a verifiable identity (name/bytes/sha256) for the harness's system
  message.

## Security fix carried in this slice

`Actions.Fs` containment is now decided on the **realpath** of both the
candidate path and the sandbox root (symlinks resolved component-by-component,
cycle-safe), not the lexical string prefix — a symlink inside cwd that points
outside it is correctly treated as an escape rather than silently allowed.
Both sides of the comparison are canonicalized (the sandbox root itself may
sit behind a symlink, e.g. macOS `/tmp` -> `/private/tmp`).

Also included: the dead-turn approval reaper in `SessionInbox` — a parked
approval can never outlive the turn that requested it, including when the
turn dies mid-approval.

## Standalone rationale

This is the missing consumer for the harness command channel: before it, a
routed tool call had nowhere to land and a turn with a tool call simply
died. It stands on its own — the gate has no dependency on the (separate,
later) session-lane/bridge work, which is deliberately excluded from this
slice.

## Note on `system_prompt.ex`

This file also rides open PR #690 (the turn-runner needed it to compile).
The content is identical between the two branches; whichever of the two
PRs merges second will rebase to a no-op on this file.

## Test plan

- `cd packages/raxol_agent && MIX_ENV=test mix compile --warnings-as-errors` — clean
- `cd packages/raxol_agent && MIX_ENV=test mix test` — 1683 passed (13 properties,
  1670 tests), 35 excluded, 0 failures
- root `MIX_ENV=test mix compile --warnings-as-errors` — clean


DROOdotFOO:

# Adversarial Review — PR #692 (agent shell exec, fs sandbox, tool-authorization gate)

This is a genuinely careful PR: the classifier is fail-closed on unknown tool names, the fs sandbox resolves symlinks with `realpath` on both sides of the containment check (well-tested), the escalation flag (`allow_outside_cwd`) rides in `context` and provably cannot be forged from LLM-controlled tool `arguments`, the approval `await` seam fails closed on raise/throw, and the interrupt path does a real process-group SIGKILL. That said, the three personas each found real defects. The headline problem is that the **shell timeout path leaks orphaned OS processes** — the exact failure mode the sibling `Interrupt` module documents as the reason `Port.close` is insufficient — and there is a cluster of process/leak and fail-open-default issues around the executor. No CRITICALs, one HIGH, several MEDs.

## HIGH

### 1. Shell command timeout orphans the OS process and its children — `actions/shell.ex:95-97,116-118`
[Saboteur] On a wall-clock timeout, `collect/5` calls only `close_port(port)` (→ `Port.close/1`). The captured `os_pid` (line 75) is never used to kill anything. Per this repo's own `Raxol.Agent.Interrupt` moduledoc: *"`Port.close` / `Process.exit(owner, :kill)` leave the OS process alive and orphan its grandchild… the ONLY thing that works is the OS process-group SIGKILL."* So `run_shell "sleep 600"` with the default 30s timeout returns `timed_out: true` while `sh` (and every child it spawned) keeps running detached — an unbounded orphan leak driven directly by LLM input, on the highest-risk tool in the PR.
**Fix:** on the timeout branch, perform the same group kill the interrupt path uses (`kill -9 -<os_pid>` via `Raxol.Agent.Interrupt` or a direct `System.cmd`) before/instead of `Port.close`, and confirm death out-of-band. Do not rely on `Port.close` to reap the process.

## MEDIUM

### 2. Executor `await` default is fail-OPEN while `gate?` defaults true — `harness/tool_executor.ex:136-137`
[Security auditor] `gate?` defaults to `true` (line 136) but `await_decision` defaults to `&default_allow/2`, which returns `{:allow, "allow"}` unconditionally (line 630). A caller that turns the gate on but forgets to wire an `await_decision` gets **silent auto-approval of every consequential tool, including `run_shell`** — the opposite of the fail-closed posture the `ToolClassifier` establishes. The sole production caller (`SessionInbox.start_turn/2`) does wire it, so this is latent today, but `raxol_agent` is a published, reusable package and the module contract itself is fail-open. No test covers `gate?: true` with no `await_decision`.
**Fix:** when `gate?` is true and no `await_decision` was supplied, default to deny (or raise at `stream/2`); reserve `default_allow` for `gate?: false` only.

### 3. Stream cleanup `Process.exit(pid, :normal)` is a no-op — parked executor loop leaks — `harness/tool_executor.ex:160`
[Saboteur] The `Stream.resource` cleanup does `Process.exit(p, :normal)`. A `:normal` exit signal to a non-trap process is silently ignored, so it never reaps the spawned loop. If the operator never answers an approval, the loop is parked in `GenServer.call(inbox, …, :infinity)`; after `@loop_timeout_ms` (5 min) the consumer halts, the pump Task completes normally, its `:normal` exit is also ignored by the linked non-trap loop, and the cleanup fun's `:normal` exit is a no-op — the executor loop is blocked forever. Slow accumulation of zombie processes over a long session.
**Fix:** use `Process.exit(p, :kill)` (or `:shutdown`) in the cleanup fun.

### 4. `glob` containment is lexical, not `realpath` — moduledoc claim is false — `actions/workspace.ex:137-139` (doc `:8-12`)
[Security auditor + New Hire] Every other fs tool routes through `Fs.resolve/1` (symlink-resolving containment). `Glob.run/2` instead does a bare `String.starts_with?(abs_pattern, cwd <> "/")` on the *lexical* expansion. A pre-existing symlink inside cwd pointing outside (common: `_build`, `deps`, `node_modules/.bin`) is followed by `Path.wildcard/1` when named as an explicit pattern component (e.g. `glob "escape_dir/*"`), enumerating filenames outside the sandbox — and `glob` is in `@auto_allow`, so no approval. The moduledoc asserts *"Every path is expanded and confined… by the SAME `Fs.resolve/1` gate"*, which is materially untrue here. Impact is read-only filename disclosure (not content), but it breaches the stated boundary; only `"../*"` is tested, no symlink test.
**Fix:** run the expanded pattern's base directory through `Fs.resolve/1` (or `safe_realpath`) and reject symlink escapes; correct the moduledoc.

## LOW

### 5. `dispatch_response/5` can raise `FunctionClauseError` on a response missing `:content` — `harness/tool_executor.ex:274-309`
[Saboteur] Clauses match `%{tool_calls: non-empty}` or `%{content: _}`. A response with `tool_calls: []` (or absent) and no `:content` key matches neither and crashes the spawned loop before `{:executor_done}`. Relies on an unstated invariant that `:content` is always present. Fix: add a catch-all clause emitting `{:error, :malformed_response}` (or default content to `""`).

### 6. Shell `{:EXIT, ^port, _reason}` clause is effectively dead — `actions/shell.ex:113-114`
[Saboteur/New Hire] The loop process isn't trapping exits and `Port.open` links the port; an abnormal port failure delivers an exit *signal*, not an `{:EXIT, port, reason}` *message*, so this clause never fires and the loop dies instead of returning the `killed: true, exit_code: -1` receipt. Fix: `Process.flag(:trap_exit, true)` around the port lifetime, or drop the clause and document that a port crash propagates.

### 7. `killed:` output field is always `false` — `actions/shell.ex:111,135`
[New Hire] The Action can never report `killed: true` (the `:exit_status` clause hard-codes `false`). The declared output field is dead. Either derive it (exit `137`/`143` within-deadline) or remove it from the schema.

---

**Verdict: CONCERNS** — no CRITICAL and fewer than three HIGH, but finding #1 (LLM-driven orphaned OS processes on every shell timeout) must be fixed before this ships, and #2/#3/#4 harden the exact safety surface this PR exists to provide. Ship-blocking on #1; strongly recommend #2–#4 in the same PR.

merklebonsai:
Addressed all four findings called out as required (HIGH #1, MEDIUM #2-4), plus the two cheap LOW fixes (#5, #6) since they're on the same surface and free to fix while the file was open. #7 is deliberately left as-is — details below. Two commits, both with regression tests, full `raxol_agent` suite green (1693 tests) and root `mix compile --warnings-as-errors` clean.

**b828ab977** — `fix(agent): kill the shell timeout's OS process group, not just the port`
**7629406c2** — `fix(agent): harden the tool-executor gate + glob sandbox`

---

### #1 (HIGH, ship-blocking) — shell timeout orphans the OS process — CONFIRMED, FIXED

Verified: `collect/5`'s timeout branches called only `close_port(port)`; `os_pid` was captured (for the sink) but never used to kill anything. Traced it to actual leaked processes locally with `sleep 5` / 150ms timeout.

Fix: `run_shell "sleep 600"` now runs the same OS process-group kill the interrupt path uses before closing the port. Rather than duplicate `Interrupt`'s kill logic, I exposed it: `Raxol.Agent.Interrupt.kill_os_pid/1` is `hard_kill/1` (the group-SIGKILL-then-confirm-out-of-band primitive) without the staging/sink/turn ceremony, since a timeout has no journal context to stage into. `shell.ex` calls it on both timeout branches (deadline-already-passed and the `receive...after` timeout), then closes the port. `killed:` in the output now reflects the truth instead of being hardcoded `false` for every timeout.

Regression (`test/raxol/agent/actions/shell_test.exs`, new file — there was no dedicated shell Action test before this): captures the live `os_pid` via the existing `shell_tool_ref_sink` hook, times out a `sleep 5` at 150ms, and confirms via `ps` (never `:exit_status`) that the pid is actually dead. A second test spawns a background child (`sleep 5 & echo $! > marker; wait`) to directly falsify the "orphaned child" claim — both the shell and its detached child are confirmed dead post-timeout.

### #2 (MEDIUM) — executor `await_decision` default is fail-open — CONFIRMED, FIXED

Confirmed: `gate?` defaults `true`, `await_decision` defaulted unconditionally to `&default_allow/2`. Traced the call graph: `run_one/2` only ever calls `gated_run/4` (which is the only caller of `await_decision`) when `config.gate?` is true — so `gate?: false` (yolo) can *never* reach the default-allow path, making it pure latent risk with no live upside.

Fix: the default is now `gate?`-conditional — `default_await_decision(false) -> &default_allow/2`, `default_await_decision(_) -> &default_deny/2`. `gate?: true` (the actual default, and every explicit opt-in) with no `:await_decision` wired now denies rather than silently approving. Every existing test that exercises a consequential tool under `gate?: true` already wires an explicit `await_decision`, so nothing broke.

Regression: two tests — a consequential `write_file` with `gate?: true` and no `await_decision` is denied and the file is untouched; the same call under `gate?: false` still auto-runs (proving the reserved default stays reachable exactly where it's supposed to).

### #3 (MEDIUM) — `Process.exit(pid, :normal)` cleanup is a no-op — CONFIRMED, FIXED

Confirmed via Erlang semantics: an exit signal with reason `:normal` sent by a process *other than* the target is discarded by a non-trapping receiver — it's not a message, it's nothing. So a loop parked in an unanswered approval (or any other indefinite wait) past `Stream.resource`'s cleanup was never reaped.

One subtlety the review's suggested fix (`Process.exit(pid, :kill)`) glosses over: the loop was started via `spawn_link` from the *same* process that runs the cleanup fun. Killing a still-linked process delivers the same fatal exit signal back up the link to the (also non-trapping) caller — naively swapping to `:kill` would have taken down whatever process is consuming the executor stream (e.g. the harness session process) along with the loop. Fixed with `Process.unlink(p)` then `Process.exit(p, :kill)`.

Regression: scripts a tool call and an `await_decision` that parks forever (`Process.sleep(:infinity)`), captures the spawned loop's pid via a `Process.list()` diff, consumes only the first two stream events (forcing `Stream.resource`'s early-halt cleanup), and asserts the loop pid is actually dead shortly after — while the test process itself stays alive (proving the unlink avoids the cascade).

### #4 (MEDIUM) — glob containment is lexical, not realpath — CONFIRMED, FIXED

Confirmed: `Glob.run/2`'s only gate was `Path.expand` + `String.starts_with?(abs_pattern, cwd <> "/")` on the pattern itself — never routing through `Fs.resolve/1`'s realpath-based check the way every sibling action does. Reproduced: a symlinked `escape_dir` inside cwd pointing outside it, pattern `"escape_dir/*"`, disclosed the outside filename before the fix.

Fix: kept the lexical pattern-level gate (rejects a literal `../` before any walk — there's no single "base directory" to `resolve/1` when the pattern has wildcards), and added a second gate: every individual `Path.wildcard/1` match is now run through `Fs.outside_cwd?/1` (the same canonicalized realpath containment `resolve/1` uses) and rejected if it escapes. Moduledoc corrected to describe the two-gate reality instead of the (false) "same `resolve/1` gate" claim.

Regression: symlinked-escape test (mirrors the existing `Fs` symlink tests) confirms the outside filename is filtered to an empty match list; a same-sandbox symlink-chain test confirms legitimate in-sandbox symlinks are unaffected.

### #5 (LOW) — `dispatch_response/5` FunctionClauseError on missing `:content` — CONFIRMED, FIXED (cheap, same file)

Confirmed by inspection — neither clause matches `%{tool_calls: [], usage: _}`. Not reachable through either shipped backend today (both always set `:content`), but it's an unstated invariant a future/custom backend could violate and crash the spawned loop on. Added a catch-all that coerces to an honest empty answer before falling through to the existing `%{content: _}` clause. One-line regression test included.

### #6 (LOW) — dead `{:EXIT, ^port, _reason}` clause — CONFIRMED, FIXED (docs + dead-code removal only)

Confirmed: the loop never traps exits, so an abnormal port crash is a real exit *signal*, never that message. Took the review's second suggested option (drop the clause, document the propagation) rather than `Process.flag(:trap_exit, true)` — flipping trap_exit process-wide for the shell-hosting process has second-order effects on *other* signals it might receive that felt out of scope for this pass. Comment above `collect/6` now states the real behavior instead of implying the dead clause was live.

### #7 (LOW) — `killed:` always `false` — PARTIALLY ADDRESSED, rest DEFERRED (not a defect, a documented tradeoff)

The fix for #1 makes `killed:` truthful on the **timeout** path (now `true` when the group kill is confirmed). The **`:exit_status`-after-interrupt** path the review also flagged stays `false` — that one is a deliberate, already-documented choice (the comment right there: we can't see the interrupt from inside `collect/5`, so we don't claim a kill we can't attribute; `:turn_canceled` is the authoritative interrupt receipt). I didn't touch it this pass — happy to revisit as a separate change if you want the field to also flip on interrupt-caused signal deaths, but that's a design call about what the field means, not a bug fix.

---

SHAs: `b828ab977`, `7629406c2`. `cd packages/raxol_agent && MIX_ENV=test mix compile --warnings-as-errors && MIX_ENV=test mix test` → 1693 passed, 0 failed. Root `MIX_ENV=test mix compile --warnings-as-errors` clean.

DROOdotFOO (commented):

Scope: the fs/shell/workspace Actions, the ToolClassifier/ToolExecutor tool
gate, and the SessionInbox approval seam. Reviewed against the repo's closed,
regression-gated path-confinement thread (`Raxol.Core.Boundary.Path.confine/3`
in raxol_core, seeded from `FsSandbox`, bound to shared conformance vectors in
`packages/raxol_core/test/support/boundary_vectors/`). Reproduced the
confinement behavior in an isolated worktree with a live symlink tree.

### HIGH

**H1 -- New confinement code is a third, unaudited fork not bound to the shared
reject-vector boundary (Security Auditor + New Hire).**
`packages/raxol_agent/lib/raxol/agent/actions/fs.ex:236-312` introduces a
brand-new path-confinement implementation (`resolve/1`, `outside_cwd?/1`,
`safe_realpath/1`, `walk/3`, `contained?/2`) that does NOT call
`Raxol.Core.Boundary.Path.confine/3` and is NOT bound to the shared conformance
vectors. That shared module's own moduledoc is explicit that duplicates must
"bind to the same shared conformance vectors ... so drift is a red test, not a
silent fork." `fs_test.exs` adds its own ad-hoc cases instead of the shared
vector set. This is precisely the silent-fork posture the boundary
infrastructure exists to forbid, on a security-critical read/write/exec surface.

There is also a proven SEMANTIC divergence: `fs.ex:238` does
`Path.expand(path, cwd)` (a LEXICAL `..` collapse) BEFORE the symlink walk,
whereas the shared resolver walks `..` component-by-component against the
resolved-real base (`path.ex:139-141`), a difference the shared moduledoc calls
out by name. Verified in a worktree with `root/sub -> /outside` (dir symlink):

```
sub/../secret        -> private: {:ok, root/secret} (ACCEPTED)   shared: :symlink_escape
sub/secret           -> private: :outside_cwd (correct)
leaflink -> outside  -> private: :outside_cwd (correct)
```

Good news: this is NOT an active escape -- the private impl opens the
lexically-collapsed path, which stays interior, so it fails safe (reads
`root/secret`, never `/outside/secret`). But an unaudited, vector-unbound copy
of a confinement primitive that already disagrees with the canonical one on a
symlinked-ancestor vector is exactly the regression the standing rule is meant
to catch. Fix: route `resolve/1` and `outside_cwd?/1` through
`Raxol.Core.Boundary.Path.confine/3` (raxol_agent can already depend on
raxol_core), or at minimum bind `fs_test.exs` to the shared
`path_reject_vectors.json` / `path_accept_vectors.json` so any drift is a red
test.

### MEDIUM

**M1 -- TOCTOU on the UNGATED read/glob/grep path (Saboteur).**
`fs.ex:236-247` checks containment on `safe_realpath(abs)` but returns the
LEXICAL `abs`; the actual `File.read` / `File.ls` / `File.stat` (e.g.
`fs.ex:107`, `fs.ex:48`) opens later and re-follows symlinks. `read_file`,
`list_dir`, `file_stat`, `glob`, `grep` are auto-allowed with no operator gate
(`tool_classifier.ex:25`). A symlink swapped between the check and the open --
e.g. by a concurrently-approved `run_shell`, or any other writer racing the
loop -- reads through to the new target with zero approval. Same root as H1:
the shared resolver returns and operates on the REAL path, narrowing this
window; the private one re-resolves at open time. Low exploitability (needs a
race win) but it is the classic check-then-open gap on the one path that is
never gated.

**M2 -- `edit_file`/`write_file` staleness check is itself check-then-write.**
`tool_executor.ex:573-595` verifies the base hash via
`Workspace.verify_unchanged/2`, then `execute/4` re-resolves and writes -- a
second window between the hash re-read and `File.write` in
`workspace.ex:251`/`workspace.ex:281`. These are operator-gated so the blast
radius is bounded, but the "label-vs-binding guarantee" the moduledoc advertises
is not race-free.

### LOW

**L1 -- `--yolo` trust model is disclosure-by-convention, not enforcement (New
Hire).** `session_inbox.ex:43-48` / `tool_executor.ex:678-684`: `gate?: false`
silently auto-allows every consequential tool including `run_shell`, relying on
"the embedder is responsible for the honest POST disclosure." No mechanism
enforces that disclosure; a mis-embedding is a silent full-shell auto-approve.

**L2 -- "read-only == harmless" understates `grep`.** `tool_classifier.ex:25`
auto-allows `grep`, which returns file CONTENTS (`workspace.ex:205-224`), i.e. a
content-exfiltration primitive. It is confined to cwd via `Fs.resolve/1` so this
is acceptable, but the classifier's framing ("they observe the world, they do
not change it") should acknowledge that observation of file contents is itself
sensitive.

### What is handled well (confirmed, not just asserted)

- Gate fails CLOSED: unknown tool names are consequential
  (`tool_classifier.ex:42-43`); an unwired `:await_decision` with `gate?: true`
  defaults to deny (`tool_executor.ex:679,683-684`); the auto-allow default is
  reachable only under `gate?: false` (provably dead otherwise).
- Operator escalation cannot be FORGED by the model: `approval_decision`
  requires a non-empty binary `option_id` with a whitelisted string->atom map
  and never `String.to_atom/1` on input (`command.ex:245-266`,
  `command.ex:86-92`); the offered options are server-generated
  (`tool_executor.ex:84-85,456`) and the inbox reads the option's `kind`, not
  the id string (`session_inbox.ex:260-268`), failing closed to deny on an
  unrecognized option. Commands arrive from the operator surface, not from LLM
  output.
- Gate GenServer has no reentrancy hole: one turn at a time
  (`session_inbox.ex:170-181`), approvals parked as `{:noreply}` keyed by
  request_id (`session_inbox.ex:159-166`), a pre-parked/guessed decision is a
  no-op (`session_inbox.ex:246-248`), and the dead-turn reaper clears `pending`
  on every turn boundary including a die-mid-approval `:DOWN`
  (`session_inbox.ex:225-228`, `session_inbox.ex:139-144`).
- Shell output is bounded (`shell.ex:149-159`, 64KB cap with continued drain)
  and a wall-clock timeout runs a real process-GROUP kill before closing the
  port (`shell.ex:143-147`), so a `sleep 600` is not orphaned.
- `glob` is confined lexically up front AND per-match via `outside_cwd?/1`
  (`workspace.ex:146-161`); `grep` via `Fs.resolve/1` -- neither ungated tool
  leaks filenames/contents from a symlinked subtree (static case).

### Verdict: CONCERNS

One HIGH (unaudited confinement fork not bound to the shared reject vectors,
with a proven semantic divergence from the canonical resolver -- though it fails
safe rather than escaping) plus two MEDIUM TOCTOU windows. No CRITICAL and no
active path-traversal / command-injection / gate-bypass / approval-forge was
found: the gate fails closed, the classifier is fail-closed on unknown names,
and operator escalation cannot be forged from model output. Recommend binding
`Actions.Fs` to `Raxol.Core.Boundary.Path` (or at least to the shared
conformance vectors) before merge so this confinement copy re-joins the
regression-gated thread.

DROOdotFOO:

Verified all seven findings against the current branch head (fix commits "kill the shell timeout's OS process group…" + "harden the tool-executor gate + glob sandbox"). Each is genuinely fixed in code with regression teeth, not just claimed:

|---|---|---|---|
| 1 | **HIGH** | Shell timeout orphans the OS process + children (`Port.close` insufficient) | Both timeout branches call `kill_and_close/2` → `Interrupt.kill_os_pid/1` (process-group SIGKILL) before closing the port. Test: *"the captured os_pid is dead shortly after a timeout fires"*. |
| 2 | MEDIUM | Executor gate fail-OPEN while `gate?` defaults true | `default_await_decision(true) → default_deny`; `default_allow` reserved for `gate?: false` (provably dead there). Fail-closed. |
| 3 | MEDIUM | Stream cleanup `Process.exit(pid, :normal)` is a no-op → parked loop leak | Cleanup now `Process.exit(p, :kill)`. |
| 4 | MEDIUM | `glob` containment lexical, not realpath (moduledoc claim false) | Expanded results re-checked through `Fs.outside_cwd?/1` (realpath); moduledoc corrected to the two-stage gate. Test: *"a symlinked directory pointing outside the sandbox does not disclose its filenames"*. |
| 5 | LOW | `dispatch_response/5` `FunctionClauseError` on a response with no `:content` | Catch-all coerces `Map.put_new(response, :content, "")` — seals a `:done` receipt instead of crashing. |
| 6 | LOW | Dead `{:EXIT, ^port, _}` clause | Removed; moduledoc now documents that a port crash propagates as a signal, not a message. |
| 7 | LOW | `killed:` output field always `false` | Derived from the confirmed group-kill disposition on the timeout path. |

**Verified on branch head:** `mix compile --warnings-as-errors` clean; shell + workspace + tool_executor + tool_classifier suites 49/49 green; CI green.

Verdict moves CONCERNS → CLEAN. Merging.

Issue #494, opened 2026-07-13
Cron scheduler: Raxol.Agent.Cron (parse-once, per-agent GenServer)


DROOdotFOO:
Part of the **Hermes Fast-Follow P0** epic. Tag: **SURFACE**. Overlaps the older H2.2 cronjob design (ADR-0025).

## Goal

`Raxol.Agent.Cron`: schedule recurring agent runs and deliver the result to any surface or gateway. Generalizes Symphony's polling loop.

## Design

`SPECS.md` "Cron scheduler (P0-3)" (lines 168-174) + ADR-0025:

- **Parse once, not per tick:** accept a crontab string OR a natural-language schedule that the agent parses *at creation time*; store the resulting crontab. No LLM per tick.
- One lightweight GenServer per schedule, registry-keyed on `{agent, schedule_id}`.
- Persistence: DETS/SQLite reusing the `Memory.Store` pattern (`packages/raxol_agent/lib/raxol/agent/memory/`).
- Delivery: post the run's output into any gateway adapter or surface.
- Expose a `cronjob` Agent Action (Action behaviour, like the other agent tools).

## Acceptance criteria

- [ ] crontab + NL schedules parsed once at creation, stored; no LLM per tick
- [ ] one supervised GenServer per `{agent, schedule_id}`; survives restart via persisted store
- [ ] fires on schedule and delivers to a surface/gateway; `cronjob` Action callable by an agent
- [ ] deterministic tests (inject a clock/time source; no `Process.sleep`)

**Effort:** M. **Priority:** P0. **References:** ADR-0025, `packages/raxol_agent/lib/raxol/agent/`, `SPECS.md` 168-174.

---

## House conventions (required)

Match the repo's functional Elixir style and testing rules:

- **Patterns** ([elixir-patterns.md](https://github.com/DROOdotFOO/agent-skills/blob/main/skills/droo-stack/rules/elixir-patterns.md)): `with` chains over nested `case`; data-first pipes with no mid-pipeline side effects; multi-clause function heads + guards over `if`/`cond`; `{:ok, _}` / `{:error, _}` tuples, no exceptions for control flow.
[...]

PR #724, opened 2026-07-25
Add cron scheduler surface (cronjob Action + fire + delivery)

DROOdotFOO:
Part of #494 (cron scheduler, ADR-0025). The **surface** half, stacked on
#723 (the engine). Base is `cron/scheduler-engine`, so review #723 first; this
PR's own diff is the four surface files + docs.

## What this adds

- **`Raxol.Agent.Actions.Cronjob`** -- the LLM-callable `cronjob` Action. One
  action with an `action` field (`create`/`list`/`update`/`pause`/`resume`/
  `run`/`remove`), reaching the scheduler via `context[:scheduler]`, mirroring
  how `Actions.Skills` reaches its store via `context[:skills]`. The schedule is
  documented as parse-once (crontab / `every 2h` / `30m` / ISO timestamp) with
  the model instructed to lower natural language itself -- no LLM per tick.
  **Recursion guard:** when `context[:in_cron]` is set, `create`/`run` return
  `{:error, :cron_in_cron}`; the read/lifecycle actions stay available. `owner`
  is taken from `context[:owner]`, so `create`/`list` are owner-scoped.

- **`Raxol.Agent.Scheduler.Fire`** -- the fresh-agent fire runner, and the
  Scheduler's new default `:runner`. Each fire builds a single-message
  conversation from the job's `prompt` (no history carried between fires),
  injects the bodies of the job's attached skills (read from `Skills.Store`) as
  a system prompt, and runs it through `Raxol.Agent.Stream` -- `run` for a plain
  completion, `react` (with `context[:in_cron]` set) when the fire is given
  tools. Provider resolves via the environment (`auto_provider`, Mock on
  failure) unless a backend/executor is pinned, mirroring the gateway handler. A
  missing/unreachable skills store degrades to "no skills", never a crash.

- **`Raxol.Agent.Scheduler.Delivery`** -- builders for the Scheduler's
  `:deliver`. `gateway/1` routes a job's `"platform:chat_id"` target through
  `Raxol.Gateway.Delivery` **only when that package is loaded** (raxol_agent
  does not depend on raxol_gateway -- the reference is `Code.ensure_loaded?`
  guarded + `@compile {:no_warn_undefined}`), returning `{:error,
  :gateway_unavailable}` otherwise. `local/1` posts to an in-process callback
  for gateway-less surfaces.

- **Docs:** the `raxol_agent` paragraph in `CLAUDE.md`.

The Scheduler still records every fire to its thread log itself (in #723); the
runner here only produces the rendered output, and delivery is the separate
`:deliver` seam.

## Wiring (opt-in, unchanged from #723)

`config :raxol_agent, :scheduler, []` boots the Scheduler. To deliver through a
gateway and run real turns, wire the seams:

    config :raxol_agent, :scheduler,
      runner: Raxol.Agent.Scheduler.Fire.runner(agent_opts: [...], skills: MySkills),
      deliver: Raxol.Agent.Scheduler.Delivery.gateway(my_adapters)

Add `Raxol.Agent.Actions.Cronjob` to an agent's `available_actions` and put the
scheduler (and optional `owner`) in the run context to expose it to the model.

## Tests

`test/raxol/agent/actions/cronjob_test.exs` (13) -- every subcommand against a
real Scheduler; owner scoping; the recursion guard (create/run blocked in a fire,
read/lifecycle allowed); `scheduler_not_configured`; `missing_id`.

`test/raxol/agent/scheduler/fire_test.exs` (5) -- returns the backend's content;
**history-free** (two fires send the identical single user turn, no assistant
carryover -- via an in-process capturing backend); **skills injected** into the
system prompt (real `Skills.Store`); graceful when the store is absent;
`in_cron` set on the ReAct path (one-shot tool call via Mock).

`test/raxol/agent/scheduler/delivery_test.exs` (3) -- `local/1` delivers and
returns the callback's result; `gateway/1` returns `:gateway_unavailable` when
the gateway package is absent (which it is in this package).

No mocks (Mox); in-memory capturing backend + real stores. Full package suite:
1938 passed.

## Manual testing

- `cd packages/raxol_agent && MIX_ENV=test mix compile --warnings-as-errors`
- `MIX_ENV=test mix test test/raxol/agent/actions/cronjob_test.exs test/raxol/agent/scheduler/fire_test.exs test/raxol/agent/scheduler/delivery_test.exs`
- `mix credo --strict lib/raxol/agent/actions/cronjob.ex lib/raxol/agent/scheduler/fire.ex lib/raxol/agent/scheduler/delivery.ex`
- full package suite: `MIX_ENV=test mix test` (1938 passed)

Note: `raxol_agent` is not in the Unified CI package matrix nor a root dep, so
CI does not run these tests -- verified locally in `packages/raxol_agent`.
Updated counts after the hardening below: cronjob 17, full suite 1942.

## Pre-review hardening

An adversarial pass over the three modules found and this PR fixes:

- **H1 (HIGH, IDOR)** -- `create`/`list` were owner-scoped but the id-based
  actions (`update`/`pause`/`resume`/`run`/`remove`) ignored the owner, so on a
  scheduler shared across owners, one owner could mutate another's job by id.
  All id-based actions now confirm `job.owner == context[:owner]` via a
  `with_owned_job` seam and report a cross-owner id as `:not_found` (no
  existence leak). Tests cover cross-owner pause/run/remove/update all rejected.
- **M1 (MED)** -- the recursion guard blocked only `create`/`run`, but the
  moduledoc claimed a scheduled agent "cannot schedule more work" while
  `update`/`resume` could re-arm a job (e.g. to `every 1s`). The guard now also
  blocks `update`/`resume` in a fire; `list`/`pause`/`remove` (read /
  work-reducing) stay available, and the doc matches.
- **L1 (LOW)** -- an `update` with no real changes returned `{:ok, job}`
  (misleading success + a pointless timer re-arm). It now returns
  `{:error, :empty_update}`.

Fire was found clean (genuinely history-free, skill-injection crash-safe via
`safe/1`, no pinned-backend hijack); Delivery's `Code.ensure_loaded?` +
`@compile {:no_warn_undefined}` guard is sufficient (and the Scheduler wraps
`deliver` in `safe_call`).

Follow-up (out of scope, noted): the `in_cron` guard rides `framework_react`'s
tool-dispatch context. If the Cronjob action is ever exposed over the MCP tool
path (vendor-owns-loop backends), that layer must carry the same guard/owner
context -- to confirm when MCP tool derivation is wired for this action.


DROOdotFOO:
Superseded by #725. GitHub auto-closed this PR when its stacked base branch `cron/scheduler-engine` was deleted on #723's squash-merge. The same surface commit was rebased directly onto master and reopened as #725 (surface files only, no engine duplication).

Issue #772, opened 2026-08-02
Funded ACP jobs never settle: Xochi 502 validation_failed (signature blocker resolved)


DROOdotFOO:
> **CORRECTED 2026-08-15 -- read [this comment](https://github.com/DROOdotFOO/raxol/issues/772#issuecomment-5303201522) first.**
> Root cause is upstream, not the solver: Xochi's production worker has no
> `RPC_URL_*` secret, so its ERC-1271 path never runs and every smart-account
> signature is rejected EOA-only (xochi-fi/xochi#364). All three suspected causes
> below are disproven by a live reproduction, and the "this rules out an
> origin-pull / ERC-1271 failure" line in Evidence is backwards -- a 401 at
> `execute` leaves the intent `quoted` with `error: null`, which is exactly what
> we saw. The original text is preserved unedited below.

## Summary
The deployed solver (`raxol-solver`, `Raxol.ACP.Xochi.SolverAgent`) **sets a job's budget on-chain but never settles it after funding**. On a buyer-driven job, `setBudget` fires (provider-only, confirmed on-chain) but on `job.funded` the solver does **not** execute the Xochi intent — the intent stays `quoted` forever and the job is stuck at `funded`.

This is the first time the *deployed* solver's `job.funded -> settle` path has actually been exercised: the earlier `--route acp` gate that settled successfully drives the settle **locally in the test** (`live_order_test.exs`), not through the deployed `SolverAgent.handle_job_funded/2`.

## Evidence (Base mainnet)
- Job **70759** (`getJob` on ACP Core `0x238E541B...832E0`): `client=0x468aeae7...172283` (registered test-buyer agent), `provider=0x939eAd...46CAc1` (Raxol Agent), `budget=2400` (= 8 bps of 3 USDC, **set by the solver**), `status=1` (funded) — **stuck**, never advances to submitted/completed.
- Xochi intent **`xi_871c096e26860d2bbe60a3a5fd44955c`** (GET `/api/intent/:id`): `status: quoted`, `tx_hash: null`, `error: null`, `updated_at == created_at`. **Never executed** — so settlement was not attempted, not failed at the pull.
- Deployed solver logs during the window: `:info` shows **heartbeats only**, no `onNewTask`/`setBudget`/settle lines (job processing isn't logged at `:info`; no error printed).

This rules out an origin-pull / ERC-1271 failure (that would leave the intent `executing`/`failed` with an `error`). The settle step is simply **not reached**.

[...]

PR #773, opened 2026-08-02
Fix 7702 buyer ERC-1271 intent signing + solver funded->settle reattach

DROOdotFOO:
Fixes the real root cause of #772 and hardens the solver's funded->settle path.

## Root cause (#772), verified on-chain

The deployed solver was never the problem: it accepts the job, sets the budget (8 bps), receives `job.funded`, and reaches settle. Xochi then rejects the buyer's intent with **HTTP 401 "Invalid signature"**, leaving the intent `quoted` and the job stuck at `funded` -- invisible because the deployed release logged only heartbeats.

The buyer (`0x468a…2283`) is a managed **EIP-7702 `SemiModularAccount7702`**. Its only authorized signer is its own authority; it validates ERC-1271 through the native fallback path with domain `EIP712Domain(uint256 chainId,address verifyingContract)` where `verifyingContract = the account` (no salt). The ACP order harness instead signed with an unauthorized session key and the deployed-Modular-Account module domain. The passing `--route acp` / `live_order` gates never caught it because they use an **EOA buyer** (ecrecover), which skips on-chain ERC-1271 entirely.

## Changes

- **`Raxol.ACP.Wallet.Sma7702`** — produces a 7702-valid ERC-1271 signature: sign the account's `ReplaySafeHash(bytes32 hash)` wrapper (over the account domain) via the managed Privy authority, then pack the ma-v2 fallback envelope `0x00‖uint32(0)‖0xFF‖0x00‖sig`. Wired as the `--signer privy` intent + origin-pull signer.
- **`Raxol.ACP.Wallet.SCA` digest fix (`--signer sca`)** — a raxol-provisioned account is a Semi-Modular Account validating through its entity-0 native fallback, so its replay-safe hash is the account's own domain, not the single-signer module's. Corrected `replay_safe_digest` accordingly (known-answer test pins it to the live account's `_replaySafeHash`). UserOp signing unaffected.
- **Solver diagnostics + reattach** — `[xochi.solver]` Logger milestones across accept/setBudget/funded/settle/submit (this is what surfaced the bug), and a reattach-from-history fallback so a `job.funded` after a restart (which drops the in-memory session; the SSE stream never replays) can still settle. Fails closed if the job isn't ours or the requirement can't be recovered.

## Verification

- Live probe on Base: `isValidSignature(D, sig) → 0x1626ba7e` for the `Sma7702Wallet` signature; raxol's `EIP712.hash` matches viem's `hashTypedData`.
- Deployed solver v12 (source build w/ logging + reattach) reproduced the 401 on job 70763, confirming the diagnosis.
- 8 new unit tests (4 signer, 4 reattach) + full `raxol_acp` suite green (the one `JobIdResolverTest` failure predates this branch). Format + credo clean.

## Manual testing (pending)
- [ ] `mix raxol_acp.order --amount 3.00 --signer privy --fund` → job reaches `completed` on-chain, intent `executed`. (~$3 real USDC; deferred.)

Stacked on #771 (buyer harness + pull-pin).

### DIFF
--- a/packages/raxol_acp/lib/mix/tasks/raxol_acp.order.ex
+++ b/packages/raxol_acp/lib/mix/tasks/raxol_acp.order.ex
@@ -431,7 +431,12 @@ defmodule Mix.Tasks.RaxolAcp.Order do
         chains: %{from => rpc}
       )
 
-    {provider, address, ScaWallet}
+    # The 7702 buyer's ONLY authorized signer is its own managed authority, so the
+    # intent (and origin-pull) signature must be produced by the sidecar, not a
+    # local session key. Sma7702Wallet resolves this provider at call time.
+    :persistent_term.put({__MODULE__, :privy_provider}, provider)
+
+    {provider, address, Sma7702Wallet}
   end
 
   # A (--signer sca): direct ERC-4337 -- the EOA session key signs UserOps,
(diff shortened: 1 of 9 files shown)


DROOdotFOO:
**E2E verification (deferred).** One `mix raxol_acp.order --amount 3.00 --signer privy --fund` run confirms the deployed solver settles a real job end-to-end (job -> `completed`, Xochi intent `executed`). Amount is 3.00 because Riddler currently rejects sub-~$3 (dynamic gas floor); once that's fixed we'll run it at the 1.10 corridor minimum.

PR #802, opened 2026-08-05
Fix terminal input decode on OTP 26/27

### LINKED ISSUE #801: 🌙 Nightly build failure - 2026-08-05
The nightly build failed. Please check the [workflow run](https://github.com/DROOdotFOO/raxol/actions/runs/30977712501) for details.

DROOdotFOO:
Closes #801.

> **Scope note.** This branch also carries `f34d83d6a "Sweep docs prose and refresh install snippets"` (118 files), bundled here deliberately. The terminal fix itself is the six files under `packages/raxol_terminal/` plus `.github/workflows/nightly.yml`; everything else in the diff is that docs sweep. Review the two independently.

The nightly input canary failed on **every** OTP 27.2 job (ubuntu + macOS, Elixir 1.17/1.18/1.19) since it landed in #781, while OTP 28.2/29.0.3 passed. It was not a flake and not protocol drift. Terminal input genuinely never decoded on OTP 26/27.

## Root cause

`user_drv` only honours `input: :raw` for a noshell session from **OTP 28** on:

```erlang
%% OTP 28/29 -- noshell branch reads `input` and applies raw mode
not StartShell ->
    Input = maps:get(input, Args),
    prim_tty:reinit(TTY, #{ input => Input, output => raw })
```

OTP 26/27 has no such branch. `initial_shell: :noshell` makes its `StartShell` false, and it falls to:

```erlang
%% OTP 27 -- ignores `input` entirely, forces non-tty
true -> prim_tty:reinit(TTY, #{ tty => false })
```

So the tty is left **cooked** and every bare keystroke sits in the line discipline until a newline. Asking for a raw shell instead is not a way out: that spawns a real shell and drops `user_drv` into a JCL prompt that swallows all input.

Raw mode is therefore ours to set on OTP 26/27 — and `Stty` is what should have been setting it.

## Why Stty could not

It targeted the literal `/dev/tty`. The BEAM puts every port child in a **fresh session**, so a `System.cmd`-spawned `stty` has no controlling terminal of its own and `/dev/tty` fails `ENXIO` for it — even when the BEAM itself is sitting on a perfectly good pty.

**Every stty call in the module was silently no-oping, on every OTP.** That is also why the canary's diagnostic reported `boot_confirmed=false,isig_off=false` on passing OTP 29 runs. The fix resolves the device (`ps -o tty=`) and targets that path.

Two operations were still using the `sh -c "stty ... < /dev/tty"` shell redirect the moduledoc explicitly forbids, and their `command_args/1` clauses were built and tested but never actually called:

- `sane!/0` — the teardown fallback, so its silent failure stranded real terminals in raw mode with no echo
- `size/0`

## Evidence

Same arm sequence, same reader pid, traced `[:send, :receive]` under tmux:

**OTP 29** — select fires per keystroke, data delivered:
```
{:trace, #PID<0.68.0>, :receive, {:read, :infinity}}
{:trace, #PID<0.68.0>, :receive, {:select, #Ref<...>, #Ref<...>, :ready_input}}
{:trace, #PID<0.68.0>, :send, {#Ref<...>, {:data, "a"}}, #PID<0.66.0>}
```

**OTP 27 before** — the pane is `icanon isig echo` (fully cooked) and nothing arrives:
```
{:trace, #PID<0.68.0>, :receive, {:read, :infinity}}
```

**OTP 27 with a trailing newline** — proves the reader itself was fine all along, the line discipline was holding the bytes:
```
{:trace, #PID<0.68.0>, :send, {#Ref<...>, {:data, "ab\n"}}, #PID<0.65.0>}
```

**OTP 27 after this fix** — `icanon_off=true`, bare keystrokes delivered:
```
{:trace, #PID<0.69.0>, :send, {#Ref<...>, {:data, "a"}}, #PID<0.66.0>}
{:trace, #PID<0.69.0>, :send, {#Ref<...>, {:data, "b"}}, #PID<0.66.0>}
```

## Corrections to existing claims

Both were disproved while investigating and would have misdirected the next reader:

- The canary's `:empty` report treated `isig_off=false` as "not in raw mode, so of course nothing arrived". Once the reader is armed, prim_tty owns the termios and re-applies **its** raw mode with ISIG **enabled**, so that is the ordinary steady state — OTP 29 decodes keystrokes fine with `isig_off=false` and `stty` failing outright. `-icanon`, not `-isig`, is what governs deliverability.
- `ReaderGate` cited a `reader_loop/2` catch-all clause as its fail-closed argument. OTP 28 rewrote the loop; on OTP 26/27 it is `reader_loop/6` with **no** catch-all, so an unmatched message sits in the reader's mailbox instead. The bounded timeout still makes the gate fail closed, so the conclusion holds and only the reasoning changed.

Also fixes the pre-existing `defp drive/2 clauses not grouped together` compiler warning in the canary.

## Note on the new test seam

`Stty` now really does mutate a terminal, so `config :raxol_terminal, :stty_device` was added and the stty tests point at a non-tty. Without it, `stty_test.exs` would apply a hardcoded foreign `stty -g` dump to whatever terminal is running the suite. The canary is unaffected — it drives a separate `mix run` process with no override.

## Risk

On OTP 28/29 the `-isig` in `raw!/0` now actually applies where it previously failed silently, so `^C` reaches the app as byte `0x03` instead of raising SIGINT. That is this module's long-documented intent and the whole reason `reassert_raw_until_isig_off/1` exists, but it is a real behaviour change on those versions and the thing most worth a second opinion.

## Tests

Two real-pty tests, sharing `Raxol.Terminal.PtyHarness` (`test/support/pty_harness.ex`), which owns the mechanism and stops at the evidence: it returns the stage reached, the tokens decoded and the app's diagnostics, and interprets none of it.

- `input_protocol_canary_test.exs` — the existing protocol claim, now back to one claim
- `input_isig_contract_test.exs` — **new**, covering the OTP 28/29 behaviour change under Risk above: `^C` must arrive as byte `0x03` rather than raising SIGINT

The isig test asserts the in-VM termios diagnostics (`isig_off=true`, `boot_confirmed=true`) as well as the decode. That is the part that was vacuous before: every `stty` call no-oped silently, so `isig_off` read false on every OTP while the suite stayed green. A decode-only assertion would go quiet the same way.

**Proven non-vacuous.** Reverting only the `Stty` fix, the isig test fails at stage `:no_output` with `boot_confirmed=false,isig_off=false` while the protocol canary still passes — so it isolates this change rather than restating the canary. That stage also confirms the mechanism: with ISIG on, the line discipline signalled the foreground group and the app died before it could write.

Stage prose stays with each test rather than in the harness, because `:no_output` means "the reader is wedged" to the canary and "SIGINT killed the app" to the isig contract.

Both files carry the `:input_protocol_canary` tag and both are named in the nightly step, which lists files explicitly.

## Manual testing

- [x] Both real-pty tests pass on Linux + OTP 27 (Elixir 1.17.3, container, real tmux pty) — the previously failing configuration
- [x] Both real-pty tests pass on macOS + OTP 29
- [x] isig test proven to fail when the `Stty` fix is reverted (mutation-checked, twice: before and after the harness extraction)
- [x] Raw mode confirmed applied on OTP 27 by reading the pane termios from outside the BEAM (`icanon` -> `-icanon`)
- [x] `raxol_terminal` suite: no new failures. The 3 deterministic `RendererIntegrationTest` failures are pre-existing on master; at `--seed 0` master fails 5 and this branch fails 3
- [x] `stty_test.exs` (10), driver + inline_driver suites (80), `reader_gate_test.exs` + `editor_session_test.exs` (20) all pass
- [x] `mix format --check-formatted` clean
- [ ] Nightly matrix green on OTP 27.2 (the actual proof this closes #801)


DROOdotFOO:
Pushed `cbb33df1c`: the OTP 28/29 behaviour change flagged under Risk now has a real-pty test.

`^C arrives as byte 0x03 instead of raising SIGINT (-isig holds)` sends `C-c` through tmux and asserts the app decodes a ctrl-c event **and** that its in-VM diagnostics report `isig_off=true, boot_confirmed=true`.

Asserting the termios flags, not just the decode, is the part that matters: every `stty` call used to no-op silently, so `isig_off` read false on every OTP while the suite stayed green. A decode-only assertion would have gone quiet the same way.

**Proven non-vacuous.** Reverting only the `Stty` fix and re-running:

```
1) test ^C arrives as byte 0x03 instead of raising SIGINT (-isig holds)
   FAILED on OTP 29 (driver: tmux) after 3 attempt(s), at stage: no_output.

   Keystrokes were sent, but the app never wrote its output file at all

   In-VM diagnostics from the app (.ready file):
     otp=29 reader=traced raw=boot_confirmed=false,isig_off=false
```

The protocol canary still passed in that same run, so the new test isolates this change rather than restating the existing one. Stage `no_output` also confirms the mechanism: with ISIG on, the line discipline signalled the foreground group and the app died before it could write.

Harness changes: `drive/3` takes the keys to send (named tmux keys as well as literal, since `C-c` must not go through `-l`), `CANARY_N` follows from their count, and `failure_report/4` takes the expected tokens instead of hardcoding `["a", "b"]`. The fixture prefixes ctrl chords because `0x03` parses to `%{ctrl: true, char: "c"}` — the same char a bare `c` produces, which would otherwise make a decoded `^C` and a typed `c` the same token.

Both tests pass on macOS + OTP 29 and on Linux + OTP 27 (container, real pty).

DROOdotFOO:
Superseded — split into two singular PRs so each fix lands once:

- **#803** — the OTP 26/27 terminal input fix (9 files)
- **#804** — the docs prose sweep, repo cleanup, and doc-rule checker (208 files)

This PR had both, because the docs work was committed onto this branch while the terminal fix was in progress. Every commit here is preserved in one of the two above.

Verified before closing: the two branches share **zero files** (9 and 208, empty intersection), and their union is byte-identical to this branch's tree — nothing lost, nothing duplicated. The one genuinely entangled file, `input_protocol_canary_test.exs`, was split by hand: the prose sweep had absorbed a rewrite of that test which belongs to the terminal fix, so it now lives only in #803.

Branch `fix/otp26-27-terminal-input-raw-mode` is left in place rather than deleted, so the original combined history stays reachable until both replacements merge.

PR #805, opened 2026-08-05
Review fixes: MCP read authz, Comm layer, bench honesty, cloud metrics

DROOdotFOO:
Fixes from an independent technical review round, then hardened by a
three-persona adversarial review of this branch itself and a follow-up
adversarial verification pass of the fixes. Thirteen commits, all
test-gated; every CRITICAL/HIGH/MEDIUM finding from both review rounds is
addressed.

## Round 1: review fixes

**MCP.** `tools/call` authorization was already in place; this branch adds
authorization to the read/metadata surfaces. `AgentBridge` stopped minting
atoms from caller-supplied session ids (`to_existing_atom` + unknown-session
error) and stopped reporting success while every keystroke errored.

**Agent Comm.** `{:agent_message, from, payload}` carried the RECEIVER's own
id as `from`; it is now the sender's id when attributed (`Comm.send`
`from:` / `Directive.send_agent from:`) and `nil` otherwise -- the framework
never fabricates a sender. `broadcast_team/2` actually filters by team.
`Comm.reply/3` completes the previously reply-less `call/3` protocol.

**Bench.** The regression machinery was inert theater: the detector was
hardcoded `[]`, the baseline save always failed then printed "saved", the
`--compare` report did nothing, and the summary/dashboard printed
fabricated PASS claims. All replaced with working machinery (behaviorally
verified: an injected +50% slowdown exits nonzero). The mix task's
rendering suite benchmarked a stub returning its argument; it now
exercises the real pipeline. The comparison suite's memory row measured
"-2 KB"; now `erts_debug.size`.

**Cloud metrics.** All four senders were `:ok` stubs and nothing produced
the inbound message. Implemented for real: OTLP/HTTP JSON, Datadog v1
series, Prometheus pushgateway text; `MetricsCollector.record_metric/4`
forwards whenever the exporter runs. The periodic flush timer had never
fired once (message-shape mismatch); fixed. CloudWatch dropped rather than
stubbed (needs SigV4/an AWS dep; was silently discarding metrics).

**Hygiene.** Seven unguarded wall-clock asserts tagged `:skip_on_ci`;
silent zero-assertion "passes" in the IOTerminal fallback tests became
visible skips; README fate->rate; `mix.lock` accepted the forced spitfire
resolution (the committed lock was unsatisfiable and broke dev builds).

## Round 2: adversarial review of this branch (3 personas, 34 findings)

The Saboteur / New Hire / Security Auditor pass BLOCKed the branch;
everything CRITICAL/HIGH/MEDIUM was then fixed:

- **Cloud exporter re-architecture:** automatic exports run in a monitored
  helper process (one in flight, crash contained, ingestion never blocks);
  after a failure the batch-size trigger backs off and only the timer
  retries -- no per-metric retry storm (verified against a 500-serving
  listener). Non-numeric values dropped at ingestion; `api_key` redacted
  from crash reports (`format_status/1`); nanosecond timestamps (second
  resolution was collapsing ~98% of 60fps data points); Prometheus label
  escaping; boot config validated (refuses to start); `configure/1`
  validates the config that would apply.
- **MCP read gates, without breaking existing deployments:** reads consult
  a dedicated `:read_authorizer` (nil = open, the pre-branch behavior) --
  feeding read-method names into the tools authorizer would have broken
  every existing allowlist (regression pinned by test). Gates cover
  `resources/read|subscribe|unsubscribe|list`, `prompts/get|list`,
  `tools/list`, and `completion/complete` (which enumerates live headless
  session ids). Denials use `-32090` with machine-readable data; ASK
  denials do not echo the operator prompt. `resources/subscribe` is now
  effective: resource-updated notifications broadcast only for subscribed
  URIs.
- **Comm:** the metadata-carrying cast shape could smuggle a foreign
  team's broadcast past the filter -- closed; broadcasts target only
  `:session`-valued Registry entries; `call/3` monitors the target
  (`{:error, :agent_down}` fast-fail) and sweeps replies that race the
  timeout; nil-dispatcher drops are logged; `AGENT_FRAMEWORK.md` states
  the real `from` contract.
- **Bench gate honesty:** match-coverage reporting (matched/new/missing
  jobs, loud error on zero matches) so renames can no longer silently
  empty the comparison; task doc states CI must persist the baseline.

## Round 3: adversarial verification of the fixes

Three refuters attacked the fix commits. 14/17 claims confirmed; the rest
produced two more commits: a failed MANUAL flush was doubling the retained
batch (new defect, fixed + regression test); tags are now normalized once
at ingestion (poison tag terms can no longer crash export; Prometheus
dedupe stable under tag reordering); Prometheus label NAMES use the label
charset (metric charset got pushes rejected); connect phase bounded;
`-32001` -> `-32090` (SDK `RequestTimeout` collision); `unsubscribe` +
`tools/list` gated; `agent.list` fixed (matched `{:ok, list}` against a
bare-list return -- always reported empty).

Documented rather than changed: subscriptions are URI-global (single-client
stdio model; multi-client SSE shares the set), and the SSE boot guard
enforces only `:authorizer` -- network deployments should set both seams.

## Manual testing

- [x] `mix raxol.check --only lockfile` passes (was FAIL on drift)
- [x] Regression gate: synthetic +50% slowdown vs baseline -> exit 1;
      +2% passes; compare prints real deltas; match coverage reported
- [x] Cloud exports verified over the wire (local cowboy listener) for all
      three backends: bodies, headers, batching, periodic timer,
      fail-then-retry without duplication, backoff (exactly one attempt
      against a 500 server), escaping/dedupe, end-to-end from
      `MetricsCollector.record_metric`
- [x] MCP: deny-all read authorizer refuses all eight read surfaces with
      `-32090`; tools allowlist unaffected by read gates; subscribe filter
      verified via notification delivery
- [x] Comm: cross-team smuggling via the metadata cast shape rejected;
      dead-target call fails fast with `:agent_down`
- [x] Suites: root 6649 / raxol_agent 2017 / raxol_mcp 350, zero failures


DROOdotFOO:
Adversarial review (3-persona) ran against this branch; all CRITICAL, HIGH, and MEDIUM findings are fixed in the five follow-up commits (0a375290f..c4da1bd09): async/guarded cloud exporter with backoff (no per-metric retry storm, verified via a 500-serving listener), api-key redaction, ns timestamps, Prometheus label escaping + rendered-label dedupe, separate MCP :read_authorizer (allowlist regression guard pinned by test) with gates extended to completion/complete + lists, effective resources/subscribe, -32001 denial code without prompt echo, team-broadcast bypass closed on the metadata cast shape, monitored Comm.call, SendAgent from: attribution + corrected docs, bench baseline match-coverage reporting, and fully visible IOTerminal skips. Suites: root 6645 / raxol_agent 2017 / raxol_mcp 348, zero failures.

PR #814, opened 2026-08-07
raxol.p benchmark profile, SIGTERM flush, budget exits

DROOdotFOO:
Implements the benchmark-mode contract for `raxol.p` -- the T-Bench adapter critical path CP2 (env profile) + CP3 (signal/exit contract) plus a minimal CP4 (trajectory export), per `raxol-tbench-adapter-spec.md` sections 6-8. Before this, zero `RAXOL_*` references existed in any lib, and a SIGTERM mid-run exited 0 with nothing flushed -- a harness would score every killed run as success.

## What's here

**`Raxol.Agent.BenchmarkProfile`** -- the env contract, parsed fail-closed:

- `RAXOL_MODEL` (`provider/model`, provider validated against `Backend.Selector`; slashes inside the model segment survive for openrouter-style ids)
- `RAXOL_PROFILE=benchmark` forces allow-all tools + all coding actions (no human to ask; the container is the blast radius) and drops the skills loop entirely so task attempts stay independent
- `RAXOL_MAX_TURNS` / `RAXOL_MAX_COST_USD` -- hard caps; the cost cap requires `RAXOL_COST_PER_MTOK_IN/OUT` so it can actually be computed from token usage, and is refused at boot otherwise (a cap that cannot be enforced must not be silently ignored)
- CLI flags win over env

**`Raxol.Agent.SignalTrap`** -- claims SIGTERM from the BEAM default (which is a graceful exit 0). raxol.p now emits a final `error{reason: terminated}` event, flushes the trajectory, and exits 143.

**`Raxol.Agent.Trajectory`** -- one `raxol-trajectory/1` JSON per run, written on every exit path (success 0, error 1, timeout/budget 2, SIGTERM 143), restating the run's own contract events with `Contract.sanitize_payload` at the boundary (found live: a raw `{:request_failed, %Req.TransportError{}}` in an error payload crashed the export until sanitized).

**Budget exits** -- turn/usage totals accumulate per contract event (tolerant of both `input_tokens`/`output_tokens` and `prompt_tokens`/`completion_tokens`, atoms or strings); exceeding a cap emits `budget_exhausted` and exits 2.

## Verification

- 23 new unit tests (profile parsing incl. fail-closed cost cap, budget tripping, usage-key tolerance, trajectory build/write/sanitize, signal forwarding on a private gen_event manager)
- E2E success path: `RAXOL_PROFILE=benchmark RAXOL_MODEL=mock/test RAXOL_TRAJECTORY_PATH=... mix raxol.p --no-tools "say hello"` -> exit 0, valid trajectory (5 steps, outcome completed)
- E2E SIGTERM: run against a deliberately hung backend, `kill -TERM` mid-run -> final terminated event on stderr, trajectory `outcome {reason: terminated, exit_code: 143}`, process exit 143
- raxol_agent suite: 2036+ passing; the only failures are pre-existing 1Password-session-sensitive credential tests (they fail identically with this branch stashed -- an unlocked local `op` session defeats their env-fallback determinism; noted for a separate fix)

## Remaining for the T-Bench submission (not this PR)

CP1 Linux Burrito builds; wiring the durable journal as the trajectory source (this PR exports from the live contract stream); the adapter repo itself.

## Manual testing

- [x] E2E success + SIGTERM probes above
- [x] `mix format --check-formatted` clean; compile `--warnings-as-errors` clean
- [ ] Budget exit e2e needs a multi-turn backend (logic unit-tested; plumbing shared with the proven timeout path)


DROOdotFOO:
Scope grew by the rest of CP1 (same critical path):

- **Release-safe headless runner**: the whole raxol.p implementation moved to `Raxol.Agent.P` (no Mix at runtime; returns the exit code); the mix task is now a boot shim, and the packaged CLI dispatches `raxol p` / `raxol -p` to the same code path. Before this, the Burrito binary -- the actual T-Bench artifact -- had no headless surface at all (`bin/raxol -p` shells out to Mix, which does not exist in a release).
- **Burrito signal gap found and fixed at the launcher layer**: burrito 1.6's zig wrapper spawns the BEAM and does not forward signals, so a SIGTERM killed the launcher raw and orphaned the BEAM -- no flush, and the in-VM 143 contract from this PR never ran. `tbench/launch.sh` now forwards TERM/INT/HUP to the BEAM found one hop under the launcher via /proc. Verified in a clean debian container: TERM through launch.sh -> terminated event + trajectory + exit 143 propagated.
- **Linux binaries**: arm64 built locally via docker/Dockerfile.raxol-cli and fully verified in a clean container (help, benchmark-profile mock run with trajectory, SIGTERM 143). amd64 cannot even build under Rosetta emulation (x86_64 BEAM does not boot), so the release workflow's Linux matrix gained a per-arch headless smoke step and a workflow_dispatch run is building both arches natively now.

Local amd64 verification is structurally impossible on this host; the CI smoke is the honest venue.

PR #860, opened 2026-08-16
Explain operational failures in the money-path mix tasks

DROOdotFOO:
Bare `{:ok, _} = ...` matches in the buyer tasks reported operational failures as `MatchError` on a struct or a bare `:error`. Each one reads like a bug in raxol when it is usually the environment.

I lost most of a session to the first one: a `Req.TransportError` on a blocked network looked like a Xochi outage, and I filed an issue against their infrastructure before finding the real cause.

## The three that were reachable

**1. Unreachable Xochi API** (`sign_intent`)

```
** (MatchError) no match of right hand side value:
    {:error, %Req.TransportError{reason: :timeout}}
```

`api.xochi.fi` resolves into `188.114.96.0/20`, and some ISPs null-route whole Cloudflare ranges. Observed live on Vodafone ES: 443 dropped, 80 intercepted with a Spanish `Acceso bloqueado` page. DNS is untouched, so it looks like the service is down. Now:

```
** (Mix) could not reach the Xochi API (transport :timeout).

This is usually the network path, not Xochi. Check, in order:

  nc -z api.xochi.fi 443          # dropped => blocked upstream, not down
  curl -sI http://api.xochi.fi    # a non-Cloudflare page here => intercepted
  curl -s "https://r.jina.ai/https://xochi.fi"   # answers => the API is fine

A control host only proves anything if it shares the blocked IP range;
cloudflare.com and other Cloudflare sites sit elsewhere and will pass while
this one fails. Route around it with a VPN or an exit node in another country.
Confirm from off-network before reporting an outage upstream.
```

**2. Unsupported corridor** (`Assets.address/2`)

It answers a bare `:error`, so `--corridor 8453>56` produced `no match of right hand side value: :error` with no clue which side was wrong. Now:

```
** (Mix) no USDC address for the destination chain 56.

Check --corridor (origin>destination). Chains with a USDC address:
  1 (Ethereum), 10 (Optimism), 137 (Polygon), 4663 (Robinhood Chain), 8453 (Base), 42161 (Arbitrum One)
```

**3. Signer sidecar refusing to start**

A Node process: fails when node is missing, `npm install` was never run, port 4048 is taken, or the Privy credentials are rejected. All four arrived as a MatchError on a `start_link` tuple. Now names the four checks, and treats `{:error, {:already_started, pid}}` as success rather than a crash.

Also `raxol_earn.rebalance`: its throwaway ledger now tolerates `already_started` instead of dying on a second sweep in one VM.

## The six I deliberately left alone

Padding every bare match would be noise, and three of these are the *correct* failure mode:

| Site | Why it stays |
| --- | --- |
| `order.ex:150` `:ok = send_requirement(...)` | `send_requirement/5` already `Mix.raise`s a clear message when its 12 retries are exhausted; the match cannot fail |
| `bench.ex:68,76` `EnvWallet.sign_*` | `bench.ex:53` sets `RAXOL_WALLET_KEY` itself, so a failure here means signing is genuinely broken -- crashing loudly is right for a benchmark |
| `bench.ex:121,122,143`, `anvil.ex:103`, `derive_caps.ex:41` | `Checkpoint`/`Ledger`/`Task.start_link`/`ensure_all_started(:req)` on a dev-only path; no plausible operational failure |

## Verification

- Transport path: output above, reproduced against the live ISP block
- Corridor path: `--corridor "8453>56" --dry-run` produces the message above
- `raxol_earn` suite unchanged; the one failure on this branch is `JobIdResolverTest`, fixed in #861
- `mix compile --warnings-as-errors` on this branch still trips the pre-existing `send_sponsored_user_operation` warning, fixed in #862 -- neither is introduced here

## Manual testing

- [x] Transport error prints guidance instead of a MatchError
- [x] Unsupported corridor names the side and lists supported chains
- [x] `--dry-run` success path unchanged (signs and quotes as before)

### DIFF
--- a/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
+++ b/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
@@ -128,6 +128,7 @@ defmodule Mix.Tasks.RaxolEarn.Order do
         {:ok, bundle} -> bundle
         {:error, reason} -> Mix.raise(sign_intent_error(reason))
       end
+
     requirement = requirement(cfg, bundle)
     log("signed Xochi intent: #{bundle[:intent_id] || bundle["intent_id"]}")
 
--- a/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
+++ b/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
@@ -123,7 +123,11 @@ defmodule Mix.Tasks.RaxolEarn.Order do
     )
 
     # 1. Sign the Xochi intent (off-chain).
-    {:ok, bundle} = sign_intent(cfg)
+    bundle =
+      case sign_intent(cfg) do
+        {:ok, bundle} -> bundle
+        {:error, reason} -> Mix.raise(sign_intent_error(reason))
+      end
     requirement = requirement(cfg, bundle)
     log("signed Xochi intent: #{bundle[:intent_id] || bundle["intent_id"]}")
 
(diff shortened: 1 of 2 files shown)


DROOdotFOO:
Closing: superseded by #870, which carries this PR's three commits plus fixes for an adversarial review.

The review found the headline sidecar checklist was in an unreachable branch: a Mix task does not trap exits, so a linked GenServer whose `init/1` fails kills the task before `start_link` returns. Also fixed: the `/health` gate checked liveness rather than identity, and the corridor error advertised chain 4663, where USDC does not exist.

PR #863, opened 2026-08-16
Report what the order task actually spends

DROOdotFOO:
**Stacked on #860 -- merge that first.** Only the last commit is new here.

I reported a funded run as moving "$3.00 USDC into escrow". It moved **0.0169 USDC**, and I only established that afterwards by hand-decoding transfer logs. An overstatement of ~175x on a command whose entire purpose is spending real money.

The task knew every number up front and printed none of them.

## What a `--fund` run actually moves

Three buckets, only two of them real transfers:

| Bucket | Reality |
| --- | --- |
| **escrow** | `fee_bps` of the principal -- 0.002400 USDC at the 3.00/8bps default -- buyer -> ACP Core `0x238e541b` |
| **gas** | 0.0069-0.0075 USDC **per UserOp**, buyer -> `0x5d74bdab...`, pulled by the ERC-20 `VerifyingPaymaster` in postOp |
| **principal** | **never moves in this task's writes.** The signed intent only AUTHORIZES the solver to pull up to that during settlement |

## The moduledoc was wrong about gas

It said "gas is Alchemy-sponsored (no ETH needed)". Half right: no ETH is needed (the buyer's ETH balance is zero and a bundler EOA fronts it), but **the buyer reimburses in USDC on every single UserOp**. The paymaster's `TokenSponsored` amount matches the USDC transfer to the byte across four observed UserOps (6946, 7509, 6952, 7521). "Sponsored" implies free; it is not.

So even a run *without* `--fund` costs USDC, which the docs never said.

## The change

**A plan before the first write** -- bounds and expectations only, never a claim that value moved:

```
SPEND PLAN -- buyer 0x468a...2283 on chain 8453, amounts in USDC
  escrow     ~0.0024 expected (8 bps of 3.00) -> ACP Core 0x238E541B...
             the PROVIDER sets the real budget on-chain; --fund pays what it set
  gas        charged in USDC from the buyer by an ERC-20 paymaster, NOT in ETH
             and NOT free; priced per UserOp, known only from the receipt
             UserOps this run: createJob, approve+fund
  principal  3.00 AUTHORIZED, not moved here -- the signed intent lets the
             solver pull up to that during settlement
```

**Receipt-derived actuals after each write** -- only USDC transfers *out of the buyer* count, and each is labelled by destination:

```
SPEND ACTUAL (approve+fund): 0.009352 USDC left the buyer
  0.0024 -> 0x238e541b...  (ACP Core -- the fee escrow)
  0.006952 -> 0x5d74bdab...  (gas: ERC-20 paymaster)
```

Accounting is wrapped so it can never take down a run that already moved money -- a receipt it cannot read degrades to a line saying so.

## Verification

The decoder was run against the **real** fund transaction from job 73295 (`0xc8eba599...`), not a fixture:

```
buyer USDC transfers: [
  {"0x238e541bfefd82238730d00a2208e5497f1832e0", 2400},
  {"0x5d74bdab1ce9ddadd7e2e333d1d173830860694a", 6952}
]
total: 0.009352 USDC
```

which matches the hand-decode of both funded runs.

`spend_plan_lines/2` is public (`@doc false`) purely so the arithmetic is testable. **7 new tests** pin the two things that were wrong: escrow is the fee and scales with `fee_bps`, and the principal line says AUTHORIZED / not moved. One asserts the word "sponsored" no longer appears.

`raxol_earn`: 767/768, sole failure the pre-existing `JobIdResolverTest` fixed in #861.

## Manual testing

- [x] Decoder verified against the on-chain fund tx for job 73295
- [x] 7 spend-plan tests green
- [ ] Plan text reviewed on the next real funded run (it only prints on a non-dry-run)

### DIFF
--- a/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
+++ b/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
@@ -128,6 +128,7 @@ defmodule Mix.Tasks.RaxolEarn.Order do
         {:ok, bundle} -> bundle
         {:error, reason} -> Mix.raise(sign_intent_error(reason))
       end
+
     requirement = requirement(cfg, bundle)
     log("signed Xochi intent: #{bundle[:intent_id] || bundle["intent_id"]}")
 
--- a/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
+++ b/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
@@ -27,16 +27,26 @@ defmodule Mix.Tasks.RaxolEarn.Order do
   drives a managed SCA agent:
 
     * `privy` (default) -- Virtuals-delegated signing via the Node signer sidecar
-      (`priv/signer_sidecar`); gas is Alchemy-sponsored (no ETH needed). The buyer
-      is the managed SCA at `RAXOL_ACP_WALLET_ADDRESS`; the intent is signed locally
-      as that SCA (`ORDER_KEY` session key -> ERC-1271).
+      (`priv/signer_sidecar`). No ETH needed, but gas is NOT free: an ERC-20
+      paymaster fronts the ETH and charges the buyer in USDC on every UserOp
+      (~0.0075 observed on Base). The buyer is the managed SCA at
+      `RAXOL_ACP_WALLET_ADDRESS`; the intent is signed by the account's own
+      managed authority (`Sma7702Wallet` -> ERC-1271), not by a session key.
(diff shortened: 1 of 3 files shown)


DROOdotFOO:
Closing: superseded by #871.

The review found the new spend accounting could itself print a wrong money figure -- unread receipts reported as zero spend, every non-Core recipient labelled as gas, and a fabricated escrow figure on the `--job-id` resume path. It also adds the escrow ceiling, reconciled with the competing gate drafted against #860.

PR #864, opened 2026-08-16
Cover smart-account signatures, and unbreak three live gate paths

DROOdotFOO:
Every live gate that has ever passed used an **EOA buyer**. Nothing exercised an ERC-1271 signature, and that gap let one assumption -- "a signature is 65 bytes" -- survive three layers into production:

1. raxol built the wrong ERC-1271 replay-safe domain (fixed, #773)
2. Xochi's ERC-1271 verification silently never ran, because no `RPC_URL_*` was set in production (fixed, xochi-fi/xochi#364, deployed 2026-08-15)
3. Riddler's request schema constrains the signature to exactly 65 bytes (**open**, axol-io/Riddler#802)

Layer 3 is why a real funded order still fails today, at `validation_failed: Request validation failed`.

## Three live gate cells were dead

The `raxol_acp` -> `raxol_earn` rename (#779) left `scripts/run_live_gates.sh` pointing at paths that no longer exist:

```
ORDER_TEST      test/raxol/acp/xochi/live_order_test.exs        MISSING
RELAY_TEST      test/raxol/acp/relay/live_relay_test.exs        MISSING
SOLVER_FEE_TEST test/raxol/acp/xochi/solver_fee_live_test.exs   MISSING
```

Every `acp`, `relay` and `fee` cell has been hard-failing on "Paths given to mix test did not match any directory/file" -- failing for a typo, before running anything. Repointed at `test/raxol/earn/...`, all three verified present. A new smart-account row next to three rows failing for a rename would have been worthless.

## What this adds

`smart_account_wire_test.exs` pins our half as an executable fact: the ma-v2 envelope is **72 bytes** (`0x00 || uint32(0) || 0xFF || 0x00 || sig65`), and a 72-byte signature **cannot** match `~r/^0x[a-fA-F0-9]{130}$/` -- the pattern copied verbatim from Riddler's `execute_request.ex:42`. A control assertion proves the pattern is not simply broken: a plain EOA signature does match it.

So the incompatibility now lives in CI rather than in an issue, and it fails the moment anyone changes the envelope. With #862 gating `raxol_earn`, it actually runs on every PR.

Mutation-proven against the real packer (`ModularAccount.pack_1271_eoa_signature/2`):

| Mutation | Result |
| --- | --- |
| marker byte `0xFF` -> `0xFE` | 1 failure |
| drop the leading `0x00` (71-byte envelope) | 2 failures |
| restored | 4 passed |

## Why there is no live cell, contrary to the original plan

I set out to add a fund-free live probe that would flip green when #802 lands, signing with a throwaway key so it could never move funds. **That design gates the wrong layer**, and I only caught it by re-reading Xochi's handler.

Xochi verifies the signature **before** forwarding to Riddler. An intent signed with an unauthorized key is rejected at Xochi's own ERC-1271 check (401) and never reaches Riddler's schema -- so the probe would have reported on layer 2, which is already fixed, while claiming to watch layer 3. And the only way to reach Riddler's validation is a *validly* signed intent, which executes the pull and moves the principal.

There is therefore no fund-free path to the constraint, and I would rather ship a smaller true thing than a live cell that lies about what it watches.

When #802 lands, the confirmation is one command, ~0.017 USDC:

```
mix raxol_earn.order --amount 3.00 --signer privy --fund
```

That is recorded in the test's moduledoc, so the next person finds it there.

## Verification

- [x] `bash -n` and `shellcheck -S error` clean on the gate script
- [x] All four repointed/existing test paths confirmed present on disk
- [x] 4 new tests green; both envelope mutations red
- [x] `raxol_earn` 764/765, sole failure the pre-existing `JobIdResolverTest` fixed in #861

## Manual testing

- [ ] `./scripts/run_live_gates.sh --asset USDC --route acp --dry-run` reaches the test instead of failing on a missing path
- [ ] Re-run the funded order once Riddler#802 lands

### DIFF
--- a/packages/raxol_earn/test/raxol/earn/xochi/smart_account_wire_test.exs
+++ b/packages/raxol_earn/test/raxol/earn/xochi/smart_account_wire_test.exs
@@ -1,0 +1,118 @@
+defmodule Raxol.Earn.Xochi.SmartAccountWireTest do
+  @moduledoc """
+  Pins the wire shape of a SMART-ACCOUNT signature against the constraint that
+  currently rejects it downstream.
+
+  ## Why this exists
+
+  Every live gate that has ever passed used an EOA buyer. Nothing exercised an
+  ERC-1271 signature, and that gap let one assumption -- "a signature is 65
+  bytes" -- survive three layers into production:
+
+    1. raxol built the wrong ERC-1271 replay-safe domain (fixed, #773)
+    2. Xochi's ERC-1271 verification silently never ran, because no `RPC_URL_*`
+       was set in production, so every smart-account signer got a bare 401
+       (fixed, xochi-fi/xochi#364, deployed 2026-08-15)
+    3. Riddler's request schema constrains the signature to exactly 65 bytes,
+       `~r/^0x[a-fA-F0-9]{130}$/` (open, axol-io/Riddler#802)
+
+  Layer 3 is why a real funded order (job 73295) still fails today, at
+  `validation_failed: Request validation failed`.
+
+  ## What this test can and cannot do
+
+  It asserts OUR half: the ma-v2 envelope raxol produces is 72 bytes, and a
+  72-byte signature cannot match the 65-byte pattern the solver enforces. That
+  makes the incompatibility an executable fact rather than a note in an issue,
+  and it fails immediately if anyone changes the envelope.
+
+  It deliberately does NOT probe the live service, because there is no
+  fund-free way to reach the constraint. Xochi verifies the signature BEFORE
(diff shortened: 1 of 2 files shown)


DROOdotFOO:
Closing: superseded by #869.

The review found a fund-loss defect: `run_relay` never exported `RELAY_LIVE_TO_ADDRESS`, so the relay test defaulted the settlement *recipient* to `@usdt_tron` -- the Tether TRC-20 token contract. A funded run settled real USDT to the token contract, unrecoverable, and the cell still recorded PASS. The default predates this PR, but the path was dead until this PR repointed `RELAY_TEST`.

PR #865, opened 2026-08-16
Format-check the packages in CI

DROOdotFOO:
**Stacked on #862 -- merge that first.** Only the last commit is new here.

No file under `packages/` is format-checked anywhere in CI. The root `Format Check` job reads the root `.formatter.exs`, whose `inputs` are:

```elixir
inputs: ["*.{ex,exs}", "{config,lib,examples}/**/*.{ex,exs}", "test/*.exs", ...]
```

All root-relative. The package cell runs `deps.get`, `compile --warnings-as-errors` and `test`, and no format step. So eighteen packages have never been checked.

I proved the gap by accident: my own unformatted `order.ex` sailed through a fully green run on #860, and I only noticed because an adversarial verifier ran `mix format --check-formatted` inside the package.

## The drift that had accumulated

| Package | Files |
| --- | --- |
| `raxol_payments` | 21 |
| `raxol_gateway` | 2 |
| `raxol_telegram` | 2 |
| `raxol_cli`, `raxol_earn` | clean |

25 files, and every diff is pure line-wrapping -- the signature of formatter-version drift over time, not sloppy edits:

```diff
-      {:ok,
-       ExKeccak.hash_256(
-         <<0x19, 0x01, domain_separator::binary, message_hash::binary>>
-       )}
+      {:ok, ExKeccak.hash_256(<<0x19, 0x01, domain_separator::binary, message_hash::binary>>)}
```

Net `78 insertions, 110 deletions` across 26 files: the formatter mostly *unwrapped* lines that no longer need wrapping.

## The step

Added to the package cell, **before** `deps.get`, so it fails fast on the cheapest possible check. Verified safe to order there: no package's `.formatter.exs` declares `import_deps` or `plugins` (`raxol_cli` has an empty `import_deps: []`), and I confirmed empirically by moving `deps/` aside and running the check -- it passes without dependencies present.

With #862 also in flight, this covers all five gated packages including the two money ones.

## Verification

- [x] All five packages `mix format --check-formatted` clean after the change
- [x] `raxol_payments` 1149 passed, `raxol_gateway` 167 passed, `raxol_telegram` 244 passed -- reformatting changed no behaviour
- [x] YAML parses; step lands between "Install hex and rebar" and "Compile package"
- [x] Format check confirmed to work with `deps/` absent

## A note on scope

This is deliberately separate from #862. Mixing 25 reformatted files into the PR that unbreaks and gates the money packages would have buried a diff that deserves line-by-line review under mechanical whitespace.

## Manual testing

- [ ] All five package cells green, with the new format step visible in each
- [ ] Introduce deliberate drift in one package and confirm its cell goes red

### DIFF
--- a/.github/workflows/ci-unified.yml
+++ b/.github/workflows/ci-unified.yml
@@ -287,7 +287,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        package: [raxol_gateway, raxol_telegram, raxol_cli]
+        # raxol_payments and raxol_earn move real money and were the two least
+        # gated packages in the repo: the Sma7702Wallet alias bug (#858) shipped
+        # with a compiler warning nobody could fail on, because nothing here ran
+        # them. Adding them required unbreaking both suites first -- see the PR.
+        package: [raxol_gateway, raxol_telegram, raxol_cli, raxol_payments, raxol_earn]
 
     steps:
       - uses: actions/checkout@v7
--- a/.github/workflows/ci-unified.yml
+++ b/.github/workflows/ci-unified.yml
@@ -317,6 +317,15 @@ jobs:
           mix local.hex --force
           mix local.rebar --force
 
+      # The root Format Check reads the root .formatter.exs, whose inputs cover
+      # only root lib/config/examples/test -- so before this step NO file under
+      # packages/ was format-checked anywhere in CI. Drift accumulated silently
+      # across formatter versions (25 files over three packages when this was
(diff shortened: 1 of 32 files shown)


DROOdotFOO:
Closing: superseded by #872.

The review found the new format gate was at war with the root `.formatter.exs` (line_length 80 vs the packages' 98, no `:subdirectories`), so a root-invoked `mix format` re-broke the 25 files this PR blessed and re-redded its own gate.

PR #886, opened 2026-08-21
Ask the chain if the pull signature will verify

DROOdotFOO:
The origin pull is the step that has failed every funded attempt on #772. It
fails at the far end of a long sequence -- quote, sign, `createJob`, fund,
`setBudget`, solver relay -- and reports back as a status string with no detail
(`validation_failed`, then a bare `pull_failed`). Each attempt cost a job, an
escrow, and a round trip through two other repositories to learn one bit.

`Raxol.Earn.Xochi.PullPreflight` asks the two contracts involved directly, for
the price of two `eth_call`s:

```
separator  = VerifyingContract.DOMAIN_SEPARATOR()   <- read from chain
structHash = hashStruct(served types, served message)
digest     = keccak256(0x1901 || separator || structHash)
           -> isValidSignature(digest, sig)  for a contract account
           -> ecrecover(digest, sig)         for an EOA
```

## Why the separator is read, not computed

Checking our own signature against a digest built from our own `domain`
projection is circular: both sides of the comparison carry the same mistake and
agree. That is exactly how #772 reached production with a green suite -- every
test rebuilt the domain the same wrong way the signer did, so the extra
`version` field Permit2 never declared was invisible to all of them.

Sourcing the domain half from the contract that will verify it breaks the loop.

## The regression test has teeth

`pull_preflight_test.exs` signs the **pre-#878** way -- a `version` key holding
`nil`, which still declares `string version` because `EIP712` derives the
domain field list with `Map.has_key?` -- and asserts that signature is REJECTED
against Permit2's real Base `DOMAIN_SEPARATOR`. The current projection passes
the same check. Without the fix, the test fails.

Both signer kinds are covered, because the pull branches on code presence: the
ACP gate's buyer is a funded EOA, while the #772 buyer is a 7702 account whose
72-byte envelope cannot recover. Checking the wrong one answers a question the
pull never asks.

## Wiring

- **`mix raxol_earn.order`** -- runs after signing. A rejection aborts the run
  before `createJob`: continuing would escrow a fee for a settlement that
  provably cannot execute, which is how job 74047 ended up funded, unsettleable
  and expired.

  It aborts `--dry-run` too, non-zero. The rehearsal is what CI scores
  (`scripts/run_live_gates.sh --dry-run` reads this task's exit status per
  cell), so a mode that printed the defect and exited 0 would score the cell
  PASS on the one thing the rehearsal exists to find, leaving it as a single
  line in a matrix of ninety.

  An inconclusive check also blocks a funded run -- an unanswered question is
  not permission to escrow -- but stays soft under `--dry-run`. An unreachable
  RPC is not evidence about a signature, and the usual cause is an origin-chain
  endpoint the rehearsing machine was never given; failing on that teaches an
  operator to ignore the exit code.
- **the `:live_xochi_order_preflight` gate** -- which quoted and validated pull
  SHAPE (`validate_pull/3`) but never asked whether the signature would verify.
  Reported, not asserted, so it keeps its "moves no funds, fails only on a
  solver-pin mismatch" contract.

`EIP712.hash_with_separator/3` is the supporting primitive: `hash/3` with the
separator supplied rather than derived.

## Verified live, spending nothing

Against a real `api.xochi.fi` quote and the 7702 buyer
`0x468aeae7...172283`:

```
[order] origin pull: the standing Permit2 allowance already covers this intent's pull -- no approve sent
[order] pull preflight: OK -- the signature covers the digest 0x000000..8ba3 will rebuild
                            (0xed9eea..4dd8), verified on-chain via ERC-1271 isValidSignature
[order] signed Xochi intent: xi_e9ce76dbe415f2fb01ca83acce811fbe
```

That is the call which returned `0xffffffff` before #878, reverting the pull
with `InvalidContractSignature()` (`0xb0669cbc`). No writes, no escrow, no job.

Separately, reading Permit2's `DOMAIN_SEPARATOR()` through the real ABI/RPC path
on Base returns `0x3b6f35e4...6f07`, byte-identical to the constant recorded in
#772 -- so the oracle half is validated end to end.

## Known limit

Only the domain half is chain-sourced, and this is recorded in the moduledoc
rather than left implicit. Permit2 builds its `PermitWitnessTransferFrom`
typehash by concatenating a fixed stub with the caller's witness type string,
where EIP-712 sorts referenced types alphabetically. These agree for
`OriginPullWitness` (it sorts before `TokenPermissions`), but a witness type
renamed to sort after it would diverge, and this check would not see it -- it
would encode the struct the same wrong way the signer did. Pinning the witness
type string is the guard for that, not this.

## Fixes from an adversarial review pass

Five follow-up commits, each with its own regression test.

**The rehearsal could not fail.** `decide_preflight/2` returned `:ok` for a
rejection under `--dry-run`, and the `:live_xochi_order_preflight` gate reports
without asserting by design. So no unfunded surface could fail on a rejected
signature, and the only place the check could stop anything was the funded run
it exists to keep you away from. A rejection now fails both modes.

**The oracle was dialed from the payload.** `pinned_verifier/2` compared the
served `verifyingContract` against the caller's pin and then returned the
SERVED one -- which is the address `DOMAIN_SEPARATOR()` is read from.
`normalize_address/1` trims and downcases, so the two can match while differing
byte for byte: a served `"0x...BA3 "` passed the pin, reached `eth_call`
verbatim as `to:`, and the node refused it. One whitespace character in a quote
stopped every funded run INCONCLUSIVE while `describe_gap/1` blamed an endpoint
that was fine. The pin is what gets dialed now, which is also what the module
says it does.

**A chainId that is present but unreadable** was reported as one that is absent,
sending the operator to look for a field sitting in the payload they captured.
It gets its own reason.

**`EIP712.hash_with_separator/3`** raised `FunctionClauseError` on a non-binary
separator, out of a public function whose spec promises `{:error, _}`. The
separator comes off the wire, so a shape it does not recognise is a question it
was asked; it answers.

**Permit2's one address is not arithmetic.** `validate_permit2_pull/2` now pins
`verifyingContract` to that constant in a shared package, so where the
assumption ends is documented on `Permit2.verifying_contract/0` -- zkSync Era
hosts Permit2 elsewhere, and a corridor there turns this into a per-chain
lookup. No config seam, since nothing needs one and the failure is a refusal to
sign rather than a bad signature.

## Manual testing

- [x] `mix raxol_earn.order --amount 3.00 --dry-run --solver 0xE9B0...2f53` against
      live Xochi: preflight OK, nothing written
- [x] Refuses without a Permit2 spender pin (the fail-closed gate still fires first)
- [x] Permit2 `DOMAIN_SEPARATOR()` read from Base matches the #772 constant
- [x] raxol_earn 935 pass, raxol_payments 1156 pass
- [x] both packages `--warnings-as-errors` clean, formatted, no new credo findings
- [ ] Funded run end to end (settlement past the pull is still unproven)

Refs #772

### DIFF
--- a/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
+++ b/packages/raxol_earn/lib/mix/tasks/raxol_earn.order.ex
@@ -53,8 +53,9 @@ defmodule Mix.Tasks.RaxolEarn.Order do
 
   Every run, `--dry-run` included, ends the signing step by asking the origin
   pull's verifying contract for its own `DOMAIN_SEPARATOR()` and asking the buyer
-  whether the signature just produced covers the digest that yields. Three
-  `eth_call`s, no writes.
+  whether the signature just produced covers the digest that yields. Four
+  read-only calls (`eth_chainId`, `eth_getCode`, and one `eth_call` each), no
+  writes.
 
   A real run proceeds only when that check PASSES. A rejection means the pull
   would revert on settlement; an inconclusive result means nobody knows, and the
@@ -161,6 +162,7 @@ defmodule Mix.Tasks.RaxolEarn.Order do
   alias Raxol.Earn.Xochi.OriginPull
   alias Raxol.Earn.Xochi.PullPreflight
   alias Raxol.Payments.Assets
+  alias Raxol.Payments.Protocols.Permit2
   alias Raxol.Payments.Protocols.Xochi, as: XochiProtocol
   alias Raxol.Payments.Xochi.{PullContracts, Schemas.QuoteRequest}
 
@@ -764,7 +766,7 @@ defmodule Mix.Tasks.RaxolEarn.Order do
   # `check_origin_pull/3` decides the pull's SHAPE is safe to sign. This asks
   # whether the signature just produced will actually be accepted, which is a
   # different question and the one every funded attempt on GitHub #772 has lost:
-  # the shape passed each time, and the digest was wrong. The two `eth_call`s
+  # the shape passed each time, and the digest was wrong. Four read-only calls
   # cost nothing and happen before any write.
   #
   # A real run proceeds on `{:ok, _}` and nothing else. Continuing on anything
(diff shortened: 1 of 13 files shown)


DROOdotFOO:
## Adversarial review: findings addressed in `5478b9642`

Verdict was CONCERNS, not BLOCK. The gate itself checked out: I confirmed it sits before `place`/`createJob`/`fund` and that a `--job-id` resume still passes through it, that `expected_verifier/2`'s two clauses are exhaustive (`validate_pull_authorization/3` refuses every other `payment_method` before signing), and that the new `{:pull_verifier, …}` pin matches the address the allowance is actually granted to in `Permit2Approver`. Findings were around the edges.

### HIGH -- the Permit2 address had two copies, and this PR gave it a third job

```
raxol_payments/…/protocols/permit2.ex:33      @permit2_address "0x0000…78BA3"
raxol_earn/…/onchain/permit2_approver.ex:60   @permit2_address "0x0000…78BA3"
```

Held equal by a comment. Before this PR the constant decided the approve spender. It now also decides the `verifyingContract` a served pull must declare and the contract `PullPreflight` reads `DOMAIN_SEPARATOR()` from.

Two copies of one fact agreeing with each other and diverging from reality is precisely the defect this module exists to catch. Drift would grant the allowance at one address while vouching for a digest built at another, with nothing failing until settlement. `Permit2Approver` now takes it from `Protocols.Permit2.verifying_contract/0`; the literal survives once, in a test, where pinning an external fact belongs.

### MEDIUM -- a rejection message that contradicted itself

`recover/2` matched `v in [27, 28]` and reported everything else by byte count:

```
… and this signature is 65 bytes rather than a canonical 65. A wrapped envelope
needs an account that can unwrap it -- check whether the 7702 delegation is
actually set
```

Two distinct causes collapsed into one message that named only length, then pointed at 7702. Now split, so a bad recovery id reads as the signer's packing and a wrapped envelope keeps the 7702 advice. Not reachable from our own signer (`pack_signature/1` normalizes through `canonical_v/1`); reachable the moment a signature arrives from an external signer.

### MEDIUM -- one guarantee was stated more strongly than it was checked

"A pass here means the signature is not the reason the next attempt fails" covers `ecrecover`'s rules. Any acceptance policy the verifier layers on top is the verifier's. The moduledoc now says what the EOA branch actually checks rather than implying the superset.

I could not settle Permit2's own malleability policy from here -- there is no vendored source, and reading the deployed contract needed a tool my environment declined. So this is documented as a boundary rather than turned into a high-`s` check, which would have meant guessing at another contract's internals in exactly the way this module argues against.

### LOW -- the preflight is a snapshot

`signer_kind` is read now and the pull happens later; a 7702 delegation set or revoked in between moves the buyer to the other branch. Added to "What it does not prove".

### Not fixed here: #891

`PullPreflight` lives in `raxol_earn`, so `Protocols.Xochi.execute/3` one layer down structurally cannot call it. That leaves `payment_execute_xochi_intent` -- an LLM-callable Action that signs and submits a pull -- and `run_live_gates.sh --route xochi` without the check. Defense-in-depth rather than an open hole (#878 fixed the root cause), but the depth is on the human-driven path and absent from the autonomous one. Moving the module down is a bigger change than this PR should carry.

### Left deliberately

- `preflight_pull/4` and `decide_preflight/2` stay `@doc false`-but-public. Making them private costs the tests that cover the money decision, which is the wrong trade.
- `verify/4`'s `else` still has no catch-all. An unmodelled shape raises and aborts before any write, which is the safe direction and matches the module's stated philosophy.

raxol_earn 930 pass (was 926), raxol_payments 1155 pass, both `--warnings-as-errors` clean, formatted, no new credo findings.

Issue #743, opened 2026-07-26
Symphony #518b: SSH remote-exec transport for workers


DROOdotFOO:
Part of #518 (SSH worker dispatch). Second slice — see the [architecture comment](https://github.com/DROOdotFOO/raxol/issues/518) on the epic. Depends on #518a (a claimed host to run on).

## Goal

Run the coding-agent worker command on a claimed remote host over SSH, with an injectable transport seam so it's testable without a real SSH server. This is the slice that makes a worker actually execute remotely.

## Scope

- **`Raxol.Symphony.Ssh`:** the single place that touches `ssh`. `exec(host_spec, remote_cmd, opts)` builds the argv (`ssh [-p PORT] [-i IDENTITY] user@host -- <remote-cmd>`) and runs it through an injectable `:exec_fn` (default `System.cmd/2`; a long-lived streaming variant over `Port.open` for the agent process), mirroring `transcribe.ex`'s `:cmd_fn` seam. Include an **`ssh` executable allowlist** (mirror `transcribe.ex:90,265-277`).
- **Remote runner transport:** wire the agent's external process through `Ssh` for a remote worker. Concretely, `Runners.Codex.Session.open_port/4` (`runners/codex/session.ex:319`) gains a transport that, when a host is present, opens `ssh host "bash -lc 'cd WS && codex ...'"` instead of local `bash -lc`; events stream back over the SSH stdout exactly as over the local Port today. (Prefer the `:cmd_fn`/`:exec_fn` injection style over Codex's hard `Session` alias.)
- **Orchestrator gate activation:** the #518a claim-gate becomes live — a remote worker only spawns once a host is claimed; `:none_free` defers the issue to a later tick (like the concurrency cap).

## Out of scope

- Remote workspace creation / hooks / cwd invariant -> #518c (this slice can assume the workspace exists, or use a fixed remote dir for tests).

## Acceptance criteria

- [ ] `Ssh.exec` builds the correct argv for the string + map host forms; a non-`ssh` executable is rejected by the allowlist
- [ ] the remote runner streams the same event shape back as the local Codex path (fake `:exec_fn` in tests, no real SSH)
- [ ] the orchestrator gates remote dispatch on a free host and defers when none is free
- [ ] deterministic tests via an injected in-memory transport; no Mox

[...]

Issue #742, opened 2026-07-26
Symphony #518a: worker.ssh_hosts config + HostPool scheduler


DROOdotFOO:
Part of #518 (SSH worker dispatch). First slice — see the [architecture comment](https://github.com/DROOdotFOO/raxol/issues/518) on the epic.

## Goal

Add the `worker.ssh_hosts` config and a pure host-pool scheduler that enforces **one-worker-lifetime-per-host**, plus the orchestrator plumbing to carry a claimed host on a running worker and release it on exit. No SSH, no remote execution — deterministically testable in isolation.

## Scope

- **Config:** new `worker` section in `Raxol.Symphony.Config` (`config.ex` `worker/1` builder + struct field/type; `config/schema.ex` `validate_worker`). `worker.ssh_hosts` is a list of host specs, defaulting to `[]` (no remote dispatch). A host spec is a map `%{host:, user:, port: (opt), identity_file: (opt), workspace_root:}`; a bare `"user@host"` string is accepted and normalized. Mirror the `active_states`/`terminal_states` list-config pattern.
- **`Raxol.Symphony.Worker.HostSpec`:** normalize/validate a raw entry (string | map) into a canonical struct.
- **`Raxol.Symphony.Worker.HostPool`:** a pure slot allocator over the specs. `new/1`, `claim/1 -> {:ok, host, pool} | :none_free`, `release/2`, free/busy counts. One host holds at most one worker for its whole lifetime.
- **Orchestrator plumbing:** `build_running_entry` carries the claimed `host` (nil for local); `handle_worker_exit` releases it. The claim-gate is wired but a no-op when `ssh_hosts` is empty (today's behaviour byte-for-byte).

## Out of scope (follow-ups)

- Actual SSH exec / remote runner -> #518b.
- Remote workspace + `PathSafety.assert_cwd!` rework -> #518c.

> Dependency note: the claim-gate only becomes *meaningful* once #518b runs a worker remotely. Until then, setting `ssh_hosts` gates workers that still run locally (a documented interim state); `ssh_hosts: []` (default) is unaffected.

## Acceptance criteria

- [ ] `worker.ssh_hosts` parses (string + map forms) and defaults to `[]` with zero behaviour change when unset
- [ ] `HostPool` claim/release/exhaustion covered; a released host is re-claimable; a crashed worker's host is released
- [ ] orchestrator carries `entry.host` and releases on worker exit
[...]

Issue #518, opened 2026-07-13
Symphony: SSH worker dispatch (worker.ssh_hosts)


DROOdotFOO:
Follow-up from #503 / #516 (epic #507).

Symphony currently fans workers out locally under a `Task.Supervisor` (one task
per issue, or one batch task in `:graph_parallel`). Issue #503 also calls for
**SSH worker dispatch** (`worker.ssh_hosts`, one-worker-lifetime-per-host) so a
run can execute on a remote host. This is greenfield — no SSH dispatch code
exists yet.

## Scope

- Config: a `worker.ssh_hosts` list (`["user@host:port", ...]`) parsed into
  `Raxol.Symphony.Config`.
- Dispatch: pick a host per eligible issue (round-robin or affinity), pin each
  host to a single concurrent worker ("one-worker-lifetime-per-host"), queue
  the rest. Integrate with `Orchestrator.spawn_worker_task/5`.
- Remote execution: send the runner module + issue + config over the wire and
  stream events / results back to the orchestrator, preserving the existing
  `{:run_event, ...}` / exit-reason contract.
- Error handling: host-unreachable falls back to the failure-retry path;
  document the trust posture (SSH keys, `known_hosts`).

Needs a short design doc / ADR and live SSH targets to validate.

## Acceptance criteria

- [ ] `worker.ssh_hosts` config parsed + validated
- [ ] issues dispatch to remote hosts, one worker per host at a time
- [ ] remote failures route to the retry path
- [ ] tests cover host selection + pinning (a fake transport is fine for CI)

**References:** `packages/raxol_symphony/lib/raxol/symphony/orchestrator.ex`
(`spawn_worker_task/5`, `dispatch_candidates/1`), `config.ex`, PR #516.

---

## Epic breakdown (SSH shell-out direction)

Re-scoped 2026-07-26 — this is a multi-PR feature (architecture fork + a local-workspace blocker). Initial architecture + rationale in the issue comments. Direction: SSH shell-out.

- [ ] #742 — worker.ssh_hosts config + HostPool scheduler (one-worker-lifetime-per-host)
- [ ] #743 — SSH remote-exec transport for workers
- [ ] #744 — remote worker workspace + cwd invariant

Dependencies: #742 → #743 → #744. Start on #742 (independently testable, no SSH).

DROOdotFOO:
## Re-scope: #518 is a multi-PR feature with an architecture fork + a local-workspace blocker

[...]

PR #888, opened 2026-08-22
Give a remote worker a workspace on its own host

### LINKED ISSUE #744: Symphony #518c: remote worker workspace + cwd invariant
Part of #518 (SSH worker dispatch). Third slice — see the [architecture comment](https://github.com/DROOdotFOO/raxol/issues/518) on the epic. Depends on #518a + #518b. Lands last (invariant-sensitive).

## Goal

Make the workspace lifecycle and its safety invariants host-aware, so a remote worker's workspace is created, used, and torn down **on the remote host** rather than the orchestrator's local filesystem.

## Context (the blocker this closes)

The workspace layer is currently local-FS-bound:
- `Workspace.ensure/2` is a local `File.mkdir_p` (`workspace.ex`), removal is local `File.rm_rf`.
- Hooks run `bash -lc` **locally** with `{:cd, cwd}` (`workspace.ex:212-231`).
- `PathSafety.assert_cwd!/1` asserts the **local** process cwd equals the workspace (SPEC s9.5 Invariant-1, `path_safety.ex:75-83`).

None of this accepts a host target, so a remote worker can't legitimately share a workspace until this is reworked.

## Scope

- Make `Workspace.ensure/remove` host-aware: for a remote worker, `mkdir -p` / `rm -rf` and the after-create/before-run/after-run hooks run on the host via `Raxol.Symphony.Ssh.exec` (#518b), rooted at the host spec's `workspace_root`.
- Relocate the cwd invariant: `assert_cwd!` is enforced at the remote command boundary (`cd WS && <cmd>`, and a remote `pwd` check) rather than the local process cwd, or explicitly gated for remote workers with an equivalent guard so Invariant-1's intent survives.
- Keep local dispatch on the existing local path untouched (`host == nil`).

## Acceptance criteria

- [ ] a remote worker's workspace is created and removed on the host (fake `:exec_fn`, no real SSH)
- [ ] hooks execute against the remote workspace root
- [ ] the cwd-safety invariant is preserved for remote workers (asserted remotely / at the command boundary)
- [ ] local (`host == nil`) workspace behaviour is byte-for-byte unchanged; deterministic tests, no Mox

**Effort:** M/L. **References:** `workspace.ex` (`ensure/2`, `execute_script/3`), `path_safety.ex` (`assert_cwd!/1`, `workspace_path/2`).

DROOdotFOO:
The third and last slice of #518. #742 shipped the host pool and #743 the SSH
transport, both merged via #752. The transport `cd`s into whatever path it is
handed, and that path was the ORCHESTRATOR's, which `ssh.ex`'s own moduledoc
recorded as deferred here:

```
`workspace` is the (local) workspace path as-is; `HostSpec.workspace_root`
is not consulted here. Per-host remote workspace roots are reserved for
issue #744.
```

So a remote worker either landed in a directory that does not exist on the
host, or, worse, in one that does and belongs to something else.

## Scope of the blast radius

Bounded, and worth stating plainly: `worker.ssh_hosts` defaults to `[]`
(`config.ex:265`), so this only reaches a deployment that opted into remote
workers, and raxol_symphony is pre-alpha and unpublished. This is a latent hole
in an opt-in feature, not a live outage.

## The change

`Workspace.ensure/remove/run_hook` take an optional `:host`. With `host: nil`
the local filesystem path is unchanged, including the arity-2 form. With a
`%HostSpec{}` the directory is created and removed on the host over
`Raxol.Symphony.Ssh`, hooks run there, and the path roots at the spec's
`workspace_root`, which stops being inert.

`created_now` is decided by ONE round trip (`if [ -d WS ]; then … else mkdir -p
WS …`), so the flag that gates `after_create` cannot be read from a stale
observation taken by a separate probe.

## Why the remote path is not `Path.expand/1`

`PathSafety.remote_workspace_path/2` folds `.` and `..` purely instead.
`Path.expand/1` resolves `~` against the ORCHESTRATOR's home and a relative
path against its cwd, so a remote root would silently absorb local state and
the containment check would then be measured against the wrong root. A relative
remote root is refused outright rather than resolved, since it would otherwise
land wherever the remote login shell happens to start.

Invariant 1 crosses the network as the `cd WS && { … }` group in
`remote_bash/2`: a `cd` that fails runs nothing and exits with `cd`'s status,
so the command cannot land in the login shell's home instead. That is the
remote counterpart of the local launch's `{:cd, workspace}`.

The group is doing the work, not the `&&`. An earlier revision of this branch
relied on the short-circuit alone and did NOT hold the invariant -- see the
review section below.

## An ordering inversion in batch dispatch

Hosts are claimed BEFORE workspaces are created, because a remote workspace
cannot be created before its host is known. That reverses the prior order, so a
workspace failure now has to hand back the slots it already reserved
(`release_batch_hosts/2`) or they leak busy forever.

## Two defects in reap_on_disconnect/1

Both surfaced only once a SHORT command ran through the wrapper. A long-lived
`codex app-server` hides both.

- **The poll interval was charged to every command.** The watcher polled
  inline, so `wait` was unreachable until the current `sleep` elapsed. The new
  test file ran in 25.8s before the fix and 0.7s after.
- **The watcher held the output pipe open.** After moving it to a background
  subshell it still inherited stdout, so its lingering `sleep` kept the write
  end open. A reader waits for EOF, not for the command to exit, so
  `System.cmd/3` blocked for a further interval after the command had finished.
  This one was intermittent: it only bit when the command outlived the
  watcher's first check, so it passed several runs before failing.

Both now have timing assertions in `ssh_test.exs`, plus one that the command's
real exit status survives the wrapper.

## Testing approach

`Raxol.Symphony.Test.FakeSsh` is a fake at an external boundary, not a mock of
our own code. It runs the EXACT remote command string the transport built,
through a local `bash -lc`, against a sandbox directory standing in for the
host. A quoting defect, a missing `cd`, or a malformed `mkdir -p` fails here
the same way it would against a real sshd, and the resulting directories are
real enough to assert on with `File.dir?/1`. No SSH server, no Mox.

The tests keep two distinct roots throughout, the orchestrator's and the
host's, since every assertion turns on which one a path lands under. The
regression this slice exists to prevent is asserted directly: after a remote
`ensure`, nothing is created on the orchestrator.

## Fixes from an adversarial review pass

Five follow-up commits. The first two are the reason this section exists rather
than a footnote.

**The `cd WS &&` short-circuit was not guarding the command.** `&&` binds to the
next COMMAND, not to the rest of the line, and both real callers pass
`reap_on_disconnect/2` output, which begins `set -m;`. So a failed `cd` skipped
only the `set -m`, and the `;` after it let the workload run anyway -- in the
login shell's home. It then reported **exit 0**, because `wait $__rx_pid`
returns the backgrounded group's status and that group ran fine, just in the
wrong directory. A `before_run` hook that never saw its workspace passed.

Reproduced against the branch's own modules before the fix:

```
output: "bash: line 0: cd: /tmp/cdprobe/nope: No such file or directory\nHOOK_RAN_IN /tmp/cdprobe\n"
exit status: 0
hook wrote outside the workspace? true
```

`runners/codex/session.ex` builds the same shape, so the coding agent itself
landed in `$HOME` whenever its workspace went missing between `ensure` and
launch -- pre-existing from #743, closed here. The regression test the suite
never had (a hook whose workspace is absent) now covers both halves: it did not
run, and it did not claim to succeed.

**A degenerate identifier collapsed the workspace onto the root.** `.`, `-` and
`_` are all in the allowed class, so `""`, `"."` and `".."` survive
`sanitize_key/1` untouched, and each folds back onto the root. The workspace
becomes the root, and `remove/3` deletes every other issue's workspace with it;
against a `~` remote root that is `bash -lc 'rm -rf ~'` on the host.

Containment could not catch this, though a comment claimed it did: the root is
inside the root, and `fold_remote/2` clamps `..` at the prefix rather than
letting it escape, so `~/..` normalizes to `~` and passes. `..` is only caught
under a deep root, which is why the existing test looked convincing.
`PathSafety.valid_key/1` rejects the three by name, ahead of any path
arithmetic. The local builder collapsed identically (`Path.join(root, "")` is
`root`), so it is fixed on the same footing.

**The removal trusted its own exit status**, which is 0 for a path that was
already absent. It now asks the host whether the path is gone. A failure still
returns `:ok` -- neither caller can retry, and the host slot is already back --
but logs at `:error` naming the silent consequence: the next `ensure/3` finds
the directory, reports `created_now: false`, and skips `after_create`.

**The local timeout expired before the remote deadline**, since the deadline
rounds up to whole seconds. Control came back while the hook was still alive on
the host and `remove_remote/4` deleted the workspace out from under it -- the
race the deadline was added to close, narrowed rather than shut. The local wait
now outlasts the deadline, so the remote kill is the normal end of the wait and
the local timer is a backstop for an `ssh` that never answers.

**A host that takes the orchestrator's root says so.** The `workspace_root`
fallback is supported and stays, but it should not be reached in silence.

## Acceptance criteria

- [x] a remote worker's workspace is created and removed on the host (fake `:exec_fn`, no real SSH)
- [x] hooks execute against the remote workspace root (`after_create`, `before_run`, `before_remove`; cwd proven by a remote `pwd`). Note `before_run`/`after_run` still have no orchestrator call site, so those two are exercised only by tests -- see Notes
- [x] the cwd-safety invariant is preserved for remote workers, tested on the side that FAILS as well as the side that passes: a hook whose workspace is missing does not run, and does not report success
- [x] local (`host == nil`) workspace behaviour unchanged; deterministic, no Mox

## Manual testing

- [x] `MIX_ENV=test mix test` in raxol_symphony: 916 pass, 6 excluded (869 before)
- [x] Three consecutive full-suite runs green, to confirm the intermittent pipe race is gone
- [x] `mix compile --warnings-as-errors` clean
- [x] `mix credo --strict` reports nothing in the touched files
- [ ] Against a real sshd (the fake covers command construction, not auth or connection loss)

## Notes

- `mix format` initially reflowed 19 unrelated files from 80 to 98 columns; those are reverted. raxol_symphony is deliberately absent from the root `.formatter.exs` `subdirectories` list and from the CI format gate, so that drift is pre-existing and expected.
- Pre-existing gap, deliberately not fixed here: `run_before_run_hook/3` and `run_after_run_hook/3` have no orchestrator call sites, only test ones. They are part of SPEC s9 but never dispatched, which is orthogonal to making them host-aware.

Closes #744


DROOdotFOO:
## CI does not run this package, so the green checks below do not cover this change

`raxol_symphony` is absent from the `package-tests` matrix in
`.github/workflows/ci-unified.yml:312`:

```yaml
package: [raxol_gateway, raxol_telegram, raxol_cli, raxol_payments, raxol_earn, raxol_agent_client_protocol, raxol_console]
```

The `Unit Tests` job runs the ROOT suite, which does not include
`packages/raxol_symphony`. So every check on this PR can pass without a single
line of this diff being exercised. The 898 passing tests are from a local
`MIX_ENV=test mix test` inside the package, run three times to catch the
intermittent pipe race described above.

Flagging rather than fixing: adding the package to that matrix also obliges
adding it to the root `.formatter.exs` `subdirectories` list (the file comments
say the two lists move together), which in turn requires formatting the 19
files of pre-existing 80-column drift. That is a separate change from #744 and
should not ride along here.

DROOdotFOO:
## Adversarial review found three defects; all fixed in `dca531e19`

Each was reproduced before the fix and is covered by a test after it. Suite is 908 passing (was 898), three consecutive clean runs.

### 1. The timeout stopped nothing (CRITICAL)

The comment claimed `hooks.timeout_ms` was enforced "by killing the local `ssh` and letting `reap_on_disconnect/1` take the remote script down". Neither half was true. `Process.exit(pid, :kill)` kills the BEAM process; closing a port does not signal the OS process it spawned:

```
result: :timeout
child survived the kill and completed its work? true
```

So a timed-out `before_remove` hook was still running on the host while `remote_rm_rf/3` deleted the workspace underneath it.

The deadline now also lives on the remote side (`reap_on_disconnect/2`'s `:deadline_seconds`), which is the half that actually bounds the work. The local half only bounds how long we wait. The comment now says so.

### 2. Killing the hook did not kill its children (fell out of 1)

Signalling only the subshell left its `sleep` running and holding the inherited stdout pipe, so a reader waiting on EOF blocked for the command's full natural runtime after the "kill": **6040ms for a 2s deadline**. With the process-group kill, **2029ms**.

That needed `set -m` so the backgrounded command gets its own pgid. Job control is switched straight back off after the fork, because bash otherwise writes `[1]  Done …` into the hook's own captured output. That one surfaced in a test log rather than a probe.

### 3. A multi-line script reported the wrong exit status (HIGH)

The reaper's ` <&0 & …` suffix bound only to the script's LAST LINE, so a script ending in a newline (which every realistic hook does) backgrounded an empty trailing command and `wait` returned *its* status:

```
single-line false                 status=1   correct
MULTI-LINE, last line fails       status=0   swallowed
```

SPEC s9.4 makes `after_create` and `before_run` failures fatal, so this converted them to success. The command is now wrapped in a `{ …; }` group. Every earlier test used a single-line hook, which is precisely why it survived.

### 4. A `~` workspace_root created directories literally named `~` (CRITICAL)

Single-quoting is exactly what suppresses tilde expansion:

```
mkdir -p '~/symphony/MT-1'
literal '~' dir created?              true
expanded $HOME/symphony/MT-1 created? false
```

`HostSpec` accepts `~` roots, `normalize_remote/1` accepts them, and a test asserted the path STRING while nothing measured the behaviour. `Ssh.quote_path/1` now leaves a recognized `~` / `~user` prefix bare and quotes the remainder, falling back to quoting the whole path so a prefix it does not understand stays inert rather than reaching the remote shell bare:

```
quote_path("~/symphony/MT-1")   -> ~/'symphony/MT-1'
quote_path("~ci/ws")            -> ~ci/'ws'
quote_path("~;touch /tmp/x")    -> '~;touch /tmp/x'
quote_path("/abs/ws")           -> '/abs/ws'
```

## Follow-ups filed rather than widened into this PR

- **#889** validate this path against a real sshd. `set -m` and process-group signalling are verified against bash 5.x on macOS and nowhere else, and `reap_on_disconnect/2` is shared with the codex runner, so the long-lived `codex app-server` now lands in its own process group too. Also flags that `identity_file` / `known_hosts` accept `~` but become argv elements no shell expands, so a tilde there is likely broken independently of this fix.
- **#890** the same wrong belief about `Port.close` in the pre-existing LOCAL hook path. Reproduced: a local hook survives its timeout by 3s. Not touched here.

PR #894, opened 2026-08-23
Make a port's child actually die when we stop it

### LINKED ISSUE #890: Local hook timeout leaks the OS process: Port.close does not kill it
## Summary

`Raxol.Symphony.Workspace.collect_output/3` believes that closing a port kills the process it spawned. It does not. A LOCAL workspace hook that exceeds `hooks.timeout_ms` keeps running after the orchestrator has given up on it and moved on.

This predates #744/#888; the remote half of the same mistake was fixed there. Filing the local half separately rather than widening that PR.

## The claim in the code

`workspace.ex`, in the timeout branch:

```elixir
after
  timeout_ms ->
    # Port.close kills the OS process via SIGKILL when :exit_status was
    # requested. Drain the message queue to avoid leaks.
    Port.close(port)
```

## Reproduction

Mirrors `execute_script/3` + `collect_output/3` exactly: a 3s child, a 300ms timeout.

```elixir
port = Port.open({:spawn_executable, bash},
  [:exit_status, :binary, :stderr_to_stdout, :hide, :in,
   {:cd, "/tmp"}, {:args, ["-lc", "sleep 3; touch #{marker}"]}])

receive do
  {^port, _} -> IO.puts("finished early")
after
  300 -> Port.close(port); IO.puts("timed out, Port.close called")
end

Process.sleep(4000)
IO.puts("LOCAL child survived Port.close? #{File.exists?(marker)}")
```

```
timed out, Port.close called
LOCAL child survived Port.close? true
```

The child completed its work a full 3s after we declared it timed out.

## Why it matters

`before_remove` is the sharp case, same as it was remotely: the hook is still running when `File.rm_rf/1` deletes the workspace under it. `after_create`'s failure path has the same shape, since it `rm_rf`s a directory a live hook may still be writing to.

Beyond that, a hook that leaks a process per timeout leaks one per retry, and the orchestrator retries.

## Fix sketch

`Port.close/1` is the wrong tool. Read the OS pid before waiting and signal it directly on timeout:

```elixir
os_pid = port |> Port.info(:os_pid) |> elem(1)
# on timeout: System.cmd("kill", ["-KILL", "-#{os_pid}"]) for the group,
# falling back to the bare pid
```

Killing the process GROUP rather than the bare pid matters for the same reason it did remotely (#888): a hook's children otherwise survive and keep the inherited stdout pipe open, so the reader blocks for the command's full natural runtime even after the "kill". Spawning with `:exit_status` does not create a new process group, so this needs either `setsid`-style spawning or a group kill on a known pgid.

Whatever lands, the comment quoted above needs deleting. A confidently wrong comment is why this survived this long.

## Acceptance

- [ ] a local hook exceeding `hooks.timeout_ms` is dead afterwards, asserted by a witness file that must NOT appear
- [ ] the hook's children die with it
- [ ] `collect_output/3` still returns `{:error, :timeout}` promptly
- [ ] the incorrect `Port.close` comment is gone

## Related

- #888 fixed the remote counterpart (`Raxol.Symphony.Ssh.reap_on_disconnect/2` gained a `:deadline_seconds` that kills the process group on the host)
- #889 covers validating the remote half against a real sshd

DROOdotFOO:
Closes #890, and fixes the same class of defect in `Runners.Codex.Session`.

`Port.close/1` releases the BEAM-side port and signals NOTHING. What becomes of
the child is the child's own business, and both call sites had assumed
otherwise.

## 1. Workspace hooks (#890)

`Workspace.collect_output/3` claimed in a comment that `Port.close` SIGKILLs the
child. It does not: a local hook exceeding `hooks.timeout_ms` kept running after
the orchestrator gave up on it. `before_remove` then `rm_rf`s the workspace out
from under a live hook, and a hook that leaks a process per timeout leaks one
per retry. Reproduced before fixing: the hook completed its work a full 3s after
it was declared timed out.

Killed outright, where the remote deadline sends SIGTERM -- nothing reads the
status here, and the workspace may be deleted next.

## 2. Codex sessions

`Session.stop/1` closed the port and stopped there. Closing delivers the EOF a
stdio server treats as "shut down", so a codex that is reading stdin *does*
exit. That is why this looked fine, and it is only the ordinary case. Measured,
two cases fall straight through it:

| child | after `Port.close` |
|---|---|
| reads stdin, exits on EOF | dead -- the clean path, works today |
| **not reading stdin** (wedged / mid-turn) | **alive** |
| **tool subprocess of a cleanly exited parent** | **alive**, reparented to init |

The second is the one that bites in practice: codex spawns tool subprocesses,
and they outlive a parent that exited perfectly cleanly. `stop/1` runs in the
`after` block of every codex run, so whatever it misses is left behind per run.

The EOF is still delivered first, and a group kill backs it up after a grace
window -- a well-behaved codex stays on the clean path and gets to flush and
exit on its own terms; the kill only reaches a session that would otherwise have
been abandoned. Polling stops as soon as the group drains, so the ordinary case
does not spend the grace.

## The shared guard

Both sites need the same safety-critical decision, so it is extracted into
`Raxol.Symphony.PortReaper` rather than copied.

Signal the process **group**: killing only the child leaves its descendants
running, still working and still holding the inherited stdout pipe. Measured on
this machine, `erl_child_setup` gives a spawned port child its own process group
(child pid 79195, pgid 79195) while the BEAM sits in a different group (pgid
79135), and the child's `sleep` shares the child's group -- so `kill -KILL -PID`
reaches the descendants and nothing else.

That is **checked, not assumed**. A child that is not a group leader shares the
BEAM's group, and the group kill would take the VM down with it. The
`{:pid, _}` fallback loses the descendants and keeps the VM; a `ps` that is
missing, fails, or answers something unparseable resolves the same way.

The target is captured **before** the close: a process group outlives its
leader, which is exactly what makes the orphan case reachable, and once the
child exits there is nothing left to read the group off.

## Verification

Every behavioural test here was proven RED against the unfixed code first.

Hook timeout:

```
1) test run_hook/4 timeout kills a timed-out hook and the children it spawned
   Expected false or nil, got true
   code: refute File.exists?(hook_witness)
```

Codex stop, against the old close-only `stop/1`:

```
1) test reaps the tool subprocesses a cleanly exited codex left behind
2) test accepts a session map and a already-dead port
3) test stops a session that is not reading stdin
3 tests, 3 failures
```

All GREEN with the fix. The hook test asserts the ABSENCE of witness files only
a survivor could create, so a slow machine cannot fail it -- it can only stop
proving anything. The reaper and session tests ask the OS for liveness
(`kill -0`) rather than inferring it from a side effect.

- `mix test` (raxol_symphony): 23 doctests, **937 tests, 0 failures**
- `mix credo --strict`: no new findings. The three hits in changed files are
  pre-existing and in untouched code (`workspace.ex:213` in `bracket/2`, two
  `opts ++ [...]` appends in `Session.launch_spec`; my hunks stop at line 148)

## Acceptance (from #890)

- [x] a local hook exceeding `hooks.timeout_ms` is dead afterwards, asserted by
      a witness file that must NOT appear
- [x] the hook's children die with it
- [x] `collect_output/3` still returns `{:error, :timeout}` promptly
- [x] the incorrect `Port.close` comment is gone

## Manual testing

- [ ] Configure a `before_remove` hook that outlives `hooks.timeout_ms`, run an
      issue to removal, confirm no orphan remains (`ps -o pid,ppid,pgid,command`)
- [ ] Confirm an in-budget hook still reports its real exit status
- [ ] Run a codex issue that shells out to a long-running tool, stop it, confirm
      the tool is gone and a clean session still shuts down without the grace
- [ ] Confirm a remote (`worker.ssh_hosts`) codex session still tears down --
      killing the local `ssh` client is what triggers `reap_on_disconnect` on
      the host, so this should be unchanged

### DIFF
--- a/packages/raxol_symphony/lib/raxol/symphony/port_reaper.ex
+++ b/packages/raxol_symphony/lib/raxol/symphony/port_reaper.ex
@@ -1,0 +1,160 @@
+defmodule Raxol.Symphony.PortReaper do
+  @moduledoc """
+  Stop the OS process behind a `Port`, and the descendants it spawned.
+
+  `Port.close/1` releases the BEAM-side port and signals NOTHING. What becomes
+  of the child is entirely the child's business: one that reads its stdin sees
+  EOF and usually exits, one that does not simply keeps running, and in BOTH
+  cases the subprocesses it spawned are untouched and outlive it. A caller that
+  needs the work actually stopped has to signal it.
+
+  Capture a target with `capture/1` while the port is still open, then either
+  `kill/1` (we have given up on the work) or `await_exit/2` (let it go quietly
+  first, then insist).
+
+  The remote counterpart is `Raxol.Symphony.Ssh.reap_on_disconnect/2`, which
+  does the same job on the far side of an ssh connection with `set -m` and a
+  group kill inside bash.
+  """
+
+  require Logger
+
+  @typedoc """
+  What may safely be signalled.
+
+  `{:group, pgid}` when the child leads its own process group, so its
+  descendants come with it. `{:pid, os_pid}` when it does not, in which case
+  only the child itself can be signalled. `:none` when the child is already
+  gone.
+  """
+  @type target :: {:group, pos_integer()} | {:pid, pos_integer()} | :none
+
+  @poll_ms 50
+
+  @doc """
+  Read the signal target off a live port.
+
+  Call this BEFORE closing the port. Once the child has exited, `ps` answers
+  nothing about it, and the process group its orphans are still in is no longer
+  recoverable -- a process group outlives its leader, which is exactly the case
+  worth reaping.
+  """
+  @spec capture(port()) :: target()
+  def capture(port) when is_port(port) do
+    case Port.info(port, :os_pid) do
+      nil -> :none
+      {:os_pid, os_pid} -> classify(os_pid)
+    end
+  end
+
+  @doc """
+  SIGKILL the target now.
+
+  For callers that have already stopped waiting, where a process that traps
+  SIGTERM must not get to keep running anyway.
+  """
+  @spec kill(target()) :: :ok
+  def kill(:none), do: :ok
+
+  def kill(target) do
+    case signal(target, "-KILL") do
+      {:ok, _} ->
+        :ok
+
+      {:error, reason} ->
+        Logger.warning(
+          "symphony.port_reaper.kill_failed target=#{inspect(target)} reason=#{inspect(reason)}"
+        )
+
+        :ok
+    end
+  end
+
+  @doc """
+  Wait up to `grace_ms` for the target to exit on its own, then SIGKILL it.
+
+  For a clean shutdown, where closing the port has already delivered the EOF a
+  well-behaved stdio child treats as "shut down": that child should be allowed
+  to flush and exit on its own terms. The kill is the backstop for the two
+  cases EOF does not cover -- a child that never reads stdin, and the tool
+  subprocesses that outlive a parent which DID exit cleanly.
+  """
+  @spec await_exit(target(), non_neg_integer()) :: :ok
+  def await_exit(:none, _grace_ms), do: :ok
+
+  def await_exit(target, grace_ms) do
+    deadline = System.monotonic_time(:millisecond) + grace_ms
+    poll_until_gone(target, deadline)
+  end
+
+  defp poll_until_gone(target, deadline) do
+    cond do
+      not alive?(target) ->
+        :ok
+
+      System.monotonic_time(:millisecond) >= deadline ->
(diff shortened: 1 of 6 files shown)


DROOdotFOO:
Superseded by #895, which now carries all four commits against master and closes #890.

Splitting them was the mistake: the Linux fix (`06358082d`) is a refactor of code introduced in #895's first commit, so it does not cherry-pick back down here. Both commits on this branch carry the procps `kill` defect -- `9b00a8163` fails `workspace_test.exs:118` outright, and `729c25894`'s polling loop is the 25-minute job cancellation -- so landing this branch on its own would put a known-broken commit on master.

Closing unmerged. The branch stays until #895 is merged and verified.

PR #895, opened 2026-08-23
Make a port's child actually die when we stop it

### LINKED ISSUE #890: Local hook timeout leaks the OS process: Port.close does not kill it

DROOdotFOO:
Closes #890, and fixes the same class of defect in `Runners.Codex.Session`.

Supersedes #894, which carried the first two commits against master. That PR's
own Linux CI was red for the reason the fourth commit here explains, so the two
are landing as one change rather than putting a known-broken commit on master.

`Port.close/1` releases the BEAM-side port and signals NOTHING. What becomes of
the child is the child's own business, and both call sites had assumed
otherwise.

## 1. Workspace hooks (#890)

`Workspace.collect_output/3` claimed in a comment that `Port.close` SIGKILLs the
child. It does not: a local hook exceeding `hooks.timeout_ms` kept running after
the orchestrator gave up on it. `before_remove` then `rm_rf`s the workspace out
from under a live hook, and a hook that leaks a process per timeout leaks one
per retry. Reproduced before fixing: the hook completed its work a full 3s after
it was declared timed out.

Killed outright, where the remote deadline sends SIGTERM -- nothing reads the
status here, and the workspace may be deleted next.

## 2. Codex sessions

`Session.stop/1` closed the port and stopped there. Closing delivers the EOF a
stdio server treats as "shut down", so a codex that is reading stdin *does*
exit. That is why this looked fine, and it is only the ordinary case. Measured,
two cases fall straight through it:

| child | after `Port.close` |
|---|---|
| reads stdin, exits on EOF | dead -- the clean path, works today |
| **not reading stdin** (wedged / mid-turn) | **alive** |
| **tool subprocess of a cleanly exited parent** | **alive**, reparented to init |

The second is the one that bites in practice: codex spawns tool subprocesses,
and they outlive a parent that exited perfectly cleanly. `stop/1` runs in the
`after` block of every codex run, so whatever it misses is left behind per run.

The EOF is still delivered first, and a group kill backs it up after a grace
window -- a well-behaved codex stays on the clean path and gets to flush and
exit on its own terms; the kill only reaches a session that would otherwise have
been abandoned. Polling stops as soon as the group drains, so the ordinary case
does not spend the grace.

## The shared guard

Both sites need the same safety-critical decision, so it is extracted into
`Raxol.Symphony.PortReaper` rather than copied.

Signal the process **group**: killing only the child leaves its descendants
running, still working and still holding the inherited stdout pipe. Measured on
this machine, `erl_child_setup` gives a spawned port child its own process group
(child pid 79195, pgid 79195) while the BEAM sits in a different group (pgid
79135), and the child's `sleep` shares the child's group -- so `kill -KILL -PID`
reaches the descendants and nothing else.

That is **checked, not assumed**. A child that is not a group leader shares the
BEAM's group, and the group kill would take the VM down with it. The
`{:pid, _}` fallback loses the descendants and keeps the VM; a `ps` that is
missing, fails, or answers something unparseable resolves the same way.

The target is captured **before** the close: a process group outlives its
leader, which is exactly what makes the orphan case reachable, and once the
child exits there is nothing left to read the group off.

## The reap never ran where it mattered

`try/after` does not run on an untrappable exit, and the orchestrator tears
workers down with exactly that on all three of its teardown paths (`stop_run`,
`terminate_stalled_run`, `terminate_running`). So `after Session.stop(session)`
in `Codex.do_run/3` never fired. The stall path is the sharp one: it fires
precisely when codex is wedged, which is the case EOF cannot reach, and
`terminate_running` then `rm_rf`s the workspace.

Cleanup that has to survive its owner cannot live in the owner, so
`PortReaper.watch/1` spawns an unlinked process that monitors the caller and
reaps on `:DOWN`. A failed handshake reaps too: only a session that was built
ever reached `stop/1`.

## kill(1) lies on Linux

The reap was measured on macOS and never worked on Linux at all. #894's own CI
said so: the hook timed out at 11.357, the kill ran, and the hook completed
anyway at 13.05 -- with no `kill_failed` warning, because the kill had reported
success.

procps-ng's `kill`, which is `/usr/bin/kill` on Debian and Ubuntu, takes a
negative pid in its own argv slot, does nothing with it, and exits 0. Reproduced
in a Linux container: with the group live,
`System.cmd(kill, ["-KILL", "-204"])` returns `{"", 0}` and the group is still
running afterwards, while the bash builtin on the same target returns `{"", 0}`
and reaps it. `kill -s KILL -- -204` is no way out either: procps has no `--`
and answers with a usage error. macOS's `/bin/kill` handles the same argv
correctly, which is exactly why this was green on a developer machine and red on
CI.

`kill -0` is worse than useless there: procps answers 0 for a group that does
not exist, so `alive?/1` was permanently true. That is the 25 minutes. Every
`await_exit/2` burned its whole grace forking a `kill` twenty times a second,
and under `max_cases: 16` the suite ground past the job budget. The hook-only
commit, which has no polling loop, just failed in 34s -- which is what CI showed
for each.

Bash's builtin is POSIX about negative pids on both platforms, and bash is
already a hard dependency of every call site, so it is the whole mechanism
rather than a fallback: a path known to silently no-op is not worth keeping as
one. The test oracle had the same bug and would have hidden the fix; it stays
independent of `PortReaper`, spelled out rather than routed through it, but goes
through the builtin too.

## Smaller findings

- The hook kill moves ahead of the close. While the port is open the child's pid
  is still ours and the OS cannot recycle it; afterwards a group kill is aimed
  at a pgid nothing pins. Nothing there waits on output, so there was never a
  reason to close first.
- Remote sessions get their own grace. `ssh` does not exit on stdin EOF, so the
  2s local budget SIGKILLed a well-behaved client partway through a network
  round trip.
- `kill`/`await_exit` guard against pgid 0 and 1 ("my own group" and "everything
  I may signal"). `capture/1` cannot produce them; the public API could.
- `Port.info(port, :os_pid)` is `:undefined` for a non-spawned port and would
  have reached an argv.
- `close_port/1` extracted rather than living in two modules with divergent
  bodies, and the comment claiming the kill lands before the close is gone.
- `grace_expired` logged at warning, not debug; a group kill degrading to a lone
  pid now says so.
- Tests poll to a deadline instead of sleeping a fixed span, and scripts
  announce readiness on stdout so the orphan tests cannot proceed before the
  fork.

## Verification

Every behavioural test here was proven RED against the unfixed code first.

Hook timeout:

```
1) test run_hook/4 timeout kills a timed-out hook and the children it spawned
   Expected false or nil, got true
   code: refute File.exists?(hook_witness)
```

Codex stop, against the old close-only `stop/1`:

```
1) test reaps the tool subprocesses a cleanly exited codex left behind
2) test accepts a session map and a already-dead port
3) test stops a session that is not reading stdin
3 tests, 3 failures
```

Plus the two review tests (owner-killed reap, handshake-failure reap), also
proven RED against the unfixed `session.ex`.

All GREEN with the fix. The hook test asserts the ABSENCE of witness files only
a survivor could create, so a slow machine cannot fail it -- it can only stop
proving anything. The reaper and session tests ask the OS for liveness
(`kill -0`) rather than inferring it from a side effect.

- `mix test` (raxol_symphony): 23 doctests, **946 tests, 0 failures**
- Linux CI green on `Package Tests (raxol_symphony)`, 1m15s
- `mix credo --strict`: no new findings. The three hits in changed files are
  pre-existing and in untouched code (`workspace.ex:213` in `bracket/2`, two
  `opts ++ [...]` appends in `Session.launch_spec`)

## Acceptance (from #890)

- [x] a local hook exceeding `hooks.timeout_ms` is dead afterwards, asserted by
      a witness file that must NOT appear
- [x] the hook's children die with it
- [x] `collect_output/3` still returns `{:error, :timeout}` promptly
- [x] the incorrect `Port.close` comment is gone

## Manual testing

- [ ] Configure a `before_remove` hook that outlives `hooks.timeout_ms`, run an
      issue to removal, confirm no orphan remains (`ps -o pid,ppid,pgid,command`)
- [ ] Confirm an in-budget hook still reports its real exit status
- [ ] Run a codex issue that shells out to a long-running tool, stop it, confirm
      the tool is gone and a clean session still shuts down without the grace
- [ ] Stop a running codex issue mid-turn, confirm its tool subprocesses are gone
- [ ] Confirm a remote (`worker.ssh_hosts`) codex session still tears down --
      killing the local `ssh` client is what triggers `reap_on_disconnect` on
      the host, so this should be unchanged

### DIFF
--- a/packages/raxol_symphony/lib/raxol/symphony/port_reaper.ex
+++ b/packages/raxol_symphony/lib/raxol/symphony/port_reaper.ex
@@ -251,13 +251,24 @@ defmodule Raxol.Symphony.PortReaper do
   # costing the full grace.
   defp signal(:unavailable, _target, _flag), do: {:error, :unavailable}
 
-  defp signal({:exe, path}, target, flag) do
-    run_signal(path, [flag, signal_arg(target)])
-  end
-
-  # `kill(1)` is a separate package (`procps` on Debian) that slim images leave
-  # out, while `bash` is already a hard dependency of every call site -- both
-  # spawn their child through it. Its builtin `kill` does the same job.
+  # Signals go through bash's `kill` BUILTIN, never `kill(1)`.
+  #
+  # procps-ng's `kill` -- `/usr/bin/kill` on Debian and Ubuntu, which is what CI
+  # runs -- takes a negative pid in its own argv slot, does NOTHING with it, and
+  # exits 0. Measured on linux/aarch64 with the group live:
+  # `System.cmd(kill, ["-KILL", "-57"])` returned `{"", 0}` and both members
+  # were still running afterwards. That is the worst failure available to this
+  # module, a reap that reports success and reaps nothing, and the exit status
+  # gives away none of it. `kill -s KILL -- -57` is no way out either: procps
+  # has no `--` and answers with a usage error.
+  #
+  # macOS's `/bin/kill` handles the same argv correctly, which is exactly why
+  # this was green on a developer machine and red on Linux CI.
+  #
+  # The builtin is POSIX about negative pids on both, and bash is already a hard
+  # dependency of every call site -- `Session.start/5` and
+  # `Workspace.execute_script/3` each spawn their child through it -- so where
+  # bash is missing there is no port child to reap in the first place.
   #
   # Arguments are passed positionally rather than interpolated into the script,
   # so nothing about the target can be read as shell.
@@ -278,15 +289,9 @@ defmodule Raxol.Symphony.PortReaper do
   end
 
   defp signaller do
-    case System.find_executable("kill") do
-      path when is_binary(path) ->
-        {:exe, path}
-
-      nil ->
-        case System.find_executable("bash") do
-          path when is_binary(path) -> {:bash, path}
-          nil -> :unavailable
-        end
+    case System.find_executable("bash") do
+      path when is_binary(path) -> {:bash, path}
+      nil -> :unavailable
     end
   end
 
--- a/packages/raxol_symphony/lib/raxol/symphony/port_reaper.ex
+++ b/packages/raxol_symphony/lib/raxol/symphony/port_reaper.ex
@@ -10,7 +10,11 @@ defmodule Raxol.Symphony.PortReaper do
 
   Capture a target with `capture/1` while the port is still open, then either
   `kill/1` (we have given up on the work) or `await_exit/2` (let it go quietly
-  first, then insist).
+  first, then insist). `close/1` is the matching port teardown.
+
+  `watch/1` covers the case no `try/after` can: a caller killed with an
+  untrappable exit never runs its own cleanup, so the reap has to live in a
+  process that survives it.
 
   The remote counterpart is `Raxol.Symphony.Ssh.reap_on_disconnect/2`, which
   does the same job on the far side of an ssh connection with `set -m` and a
@@ -29,6 +33,29 @@ defmodule Raxol.Symphony.PortReaper do
   """
   @type target :: {:group, pos_integer()} | {:pid, pos_integer()} | :none
 
+  @typedoc "Handle returned by `watch/1`, to be handed back to `release/1`."
+  @type watcher :: pid() | :none
+
+  @typedoc """
+  Why a reap could not be carried out.
+
+  `:unavailable` is the one that matters: it means we could not run a signal at
+  all, so nothing was verified and nothing was killed. It must never be
+  reported as success -- a caller that is about to delete the target's working
+  directory is relying on the difference.
+  """
+  @type reason :: :unavailable
+
+  # A pid of 0 means "my own process group" and 1 means "everything I am
+  # allowed to signal". Neither is ever a port child, and either one turned
+  # into a group kill takes the VM (or the host) with it. `capture/1` cannot
+  # produce them, but `kill/1` and `await_exit/2` are public and this module
+  # exists precisely to not assume that.
+  defguardp is_signallable(target)
+            when is_tuple(target) and tuple_size(target) == 2 and
+                   elem(target, 0) in [:group, :pid] and
+                   is_integer(elem(target, 1)) and elem(target, 1) > 1
+
   @poll_ms 50
 
   @doc """
@@ -42,9 +69,33 @@ defmodule Raxol.Symphony.PortReaper do
   @spec capture(port()) :: target()
   def capture(port) when is_port(port) do
     case Port.info(port, :os_pid) do
-      nil -> :none
-      {:os_pid, os_pid} -> classify(os_pid)
+      # `:os_pid` is `:undefined` for a port that did not spawn a process, and
+      # the atom would otherwise flow all the way into an argv.
+      {:os_pid, os_pid} when is_integer(os_pid) and os_pid > 1 -> classify(os_pid)
+      _ -> :none
     end
+  end
+
+  @doc """
+  Close `port` if it is still open.
+
+  The child can exit on its own between the decision to stop it and the close.
+  `:exit_status` then closes the port for us, and `Port.close/1` raises on an
+  already-closed port.
+  """
+  @spec close(port()) :: :ok
+  def close(port) when is_port(port) do
+    case :erlang.port_info(port) do
+      :undefined ->
+        :ok
+
+      _ ->
+        Port.close(port)
+        :ok
+    end
+  rescue
+    # The port can close between the check above and the call: same outcome.
+    ArgumentError -> :ok
   end
 
   @doc """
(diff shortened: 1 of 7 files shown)


DROOdotFOO:
Adversarial review pass found the reap silently no-oping on the leak shape this PR exists to fix. Pushed 0e6a0d6a6.

## The gap

`capture/1` resolved the target by asking `ps` for the child's pgid. `ps` cannot answer for a pid that has already been reaped -- and that is exactly what a leaked hook looks like. ERTS withholds the exit status until the port's inherited stdout reaches EOF, so a hook that backgrounds a child and exits leaves the port **open** while `Port.info/2` goes on naming the dead parent.

Reproduced against `Workspace.execute_script/3`'s exact port options, hook `( sleep 30 ) & exit 0`:

```
captured os_pid while 'open': 49524
exit_status delivered?        NO -- port still open after parent exited
Port.info(:os_pid) now:       {:os_pid, 49524}
parent 49524 alive?           false
ps -o pgid= -p 49524          {"", 1}          -> classify falls back to {:pid, 49524}
kill -KILL 49524              status 1         -> read as :gone -> reports :ok
ORPHAN STILL RUNNING:         pid 49526, pgid 49524
```

The orphan's pgid *is* 49524, so `kill -KILL -49524` would have reaped it. The verification step declined the group kill at the exact moment the group kill was the only thing that worked.

`Session.stop/1` had the mirror image: every path where codex exits on its own (`{:error, {:port_exit, _}}` out of both receive loops, plus the handshake failures) closes the port first, so the re-capture answered `:none`, and `release/1` then stood down the watcher that still held the valid `{:group, pgid}`.

## The fix

Ask the group directly with `kill -0 -PID` instead of reading a pgid out of `ps`. This is safe for the same reason the `ps` check existed: a process group's id is its leader's pid, and that pid stays allocated while the group has any member, so a live answer can only mean the process currently holding it -- our child, since the port is open -- led it. A child that is *not* a group leader has no group under its own id and the probe says so. The BEAM's own group is unreachable by construction rather than by inspection.

Side benefit: this was the module's only dependency on a `ps -o pgid= -p` that busybox does not implement.

`Session` now carries the handshake target on the session map and falls back to it when `capture/1` answers `:none`. Staleness is bounded the same way the watcher's is -- a group id cannot be recycled while the group still has a member.

Also in this commit: EPERM is distinguished from ESRCH so a refused signal stops reporting as a successful reap (positively matched, with ESRCH as the default, so an unfamiliar `strerror` string degrades to the old behaviour rather than warning spuriously); two comments that derived pid-safety from an open port now say what actually happens; and `detail=no_kill_executable_and_no_bash` no longer names a `kill(1)` fallback removed in 06358082d.

## Tests

Three new cases, each verified RED against its own fix before going green:

| Test | Without the fix |
|---|---|
| `capture/1 resolves the group when the parent has exited but its orphan holds the port open` | FAIL |
| `run_hook/4 kills the children of a hook that exited before its own timeout` | FAIL |
| `Session.stop reaps the orphans of a codex that exited on its own and closed the port` | FAIL |

The pre-existing suites covered only the parent-alive path, which is why they were green. Also made `port_reaper_test.exs`'s synchronous `refute alive?` poll like every other liveness check in the file -- `await_exit/2` returns when the SIGKILL is accepted, not when the group has drained.

Full package: 949 tests + 23 doctests, 0 failures. Credo output identical to the pre-change baseline.

PR #903, opened 2026-08-24
Exclude live parity and size inventory floors on demand

DROOdotFOO:
Stacks on #910, which now owns the MCP headless guard changes that were previously duplicated here.

## Changes

- Excludes `:live_parity` from ordinary `raxol_earn` test runs so a configured token cannot turn the hermetic suite into a live API call.
- Adds destination-demand summaries to `SettlementLedger`.
- Widens configured inventory floors from recent peak settled demand, bounded by a required per-corridor cap.
- Preserves target headroom as effective floors move and never invents floors for unsupported corridors.
- Wires the same validated demand configuration through release boot, the monitor, and the one-shot rebalance task.
- Refuses half-configured or malformed demand settings; the one-shot task reports configuration errors without a stack trace and requires `--static-floors` when it cannot access the live ETS demand signal.

## Verification

- `raxol_earn`: 9 properties, 955 tests, 0 failures (31 excluded)
- `raxol_payments`: 64 properties, 1,147 tests, 0 failures (110 excluded)
- Formatting clean in both packages

Merge #910 first, then retarget this PR to `master` after GitHub updates the stack.


DROOdotFOO:
## Review fixes pushed (`e760f1e4e`)

Three things, plus a rebase onto master that dropped nothing.

**The cap is PER CORRIDOR, and the docs did not say so.** `demand_floor_cap` bounds one `{chain, symbol}` floor, not the deployment's exposure. `default/0` spans six EVM chains times three stables, so `RAXOL_REBALANCE_DEMAND_FLOOR_CAP=500` authorizes up to 500 on every corridor carrying a static floor -- an order of magnitude more than "capped at `demand_floor_cap`" reads.

That is exactly the quantity the mandatory cap exists to bound. `peak` is attacker-controlled by construction (it comes from settled fills, and anyone can order through the public storefront), so one order per corridor holds every floor at its cap for the window while the auto-rebalancer moves funds to meet them. Now stated on `effective_inventory_floor/4` and in CLAUDE.md, together with the fact that `peak` does not decay inside the window -- so one order per window renews the pin.

**A bad policy crash-looped instead of refusing.** `cap_at/2` raises on a multiplier with no cap, and that belongs at the widening site: `demand_floor_cap` is a public struct field, so a hand-built policy reaches the advisor without ever passing through `with_demand/2`. But the widening site is inside the periodic sweep, and a raise there is not a refusal -- it is a crash, a supervisor restart, and the same crash on the next tick, forever, over a value that was wrong before the process ever started.

`RebalanceMonitor.init/1` now asserts the same invariant once, up front, by running `with_demand/2` over the policy's own fields. One definition of "a usable demand pair" serves both sites.

**The gate warns.** `RAXOL_ACCOUNTING_ENABLED` still cannot raise -- aborting a release over a variable whose only job is to say the subsystem is off is the failure every other refusal here exists to prevent -- and `"1"`, `"yes"`, `"TRUE"` still all mean off. It now logs which value it rejected. Otherwise "off" and "you typed `yes` and got off" are the same silence, in a PR whose stated theme is that an operator must be able to tell "my knob did nothing" from "the feature does nothing". Still off, so a malformed knob downstream still cannot abort a boot.

### Testing

Both new tests red-proven -- reverting each fails the test that names it:

| revert | fails with |
| --- | --- |
| the gate warning | `"1" was silently off` |
| the `init/1` check | the half-configured policy starts instead of refusing |

- `packages/raxol_payments`: 1209 passed (64 properties), 110 excluded
- `packages/raxol_earn`: 964 passed (9 properties), 31 excluded

### Still open from the review, deliberately

The per-sweep cost of `sweep_signals/2` -- it drops `:since_ms` before `list_settlements/2`, so every sweep copies the whole ledger out of the GenServer that also serves `record_settlement/2`, and nothing bounds the table's growth. The docstring names the cost; bounding it is a separate change to `SettlementLedger` and wants its own issue rather than riding here.

Issue #726, opened 2026-07-25
Enforce agent tool context (authorizer/hooks/owner/in_cron) over the native MCP tool path


DROOdotFOO:
## Summary (research / low priority)

Native (vendor-owns-loop) backends -- `handles_tools_internally? == true`,
e.g. `Backend.ClaudeCode` / `Backend.Cursor` -- run their own tool loop and
execute Raxol Actions **out-of-process over MCP** (`native_react/1` in
`Raxol.Agent.Stream`, tools injected via `Harness.McpToolConfig`). On that path
the framework cannot thread run **context** into tool execution:

- `context[:tool_authorizer]` (the `ToolPolicy` that denies `sensitive: true`
  fund-mover Actions)
- `context[:tool_call_hooks]` (the `ToolCall.Hook` pipeline)
- app flags such as the cron `:in_cron` recursion guard and `:owner` scoping
  read by `Actions.Cronjob`

All of these are silently dropped, because the spawned MCP server that runs the
tool receives none of the caller's BEAM-process context. Any context-based
security control is therefore unenforced when a native backend is in use.

This does **not** bite today: no `sensitive`/context-guarded Action is wired
into any native react call, `Actions.Cronjob` is unwired everywhere, and there
is no MCP server that reads `RAXOL_MCP_TOOLS_FILE` to execute these tools. It is
a latent trap that springs the moment any of that is wired.

## Interim mitigation already in place

`Raxol.Agent.Scheduler.Fire` fails closed for the cron fire path: when the
resolved backend is native (`Stream.native_tool_loop?/1`), it withholds
`Actions.Cronjob` from the injected toolset entirely (see PR from #494
follow-up). This is a point-fix scoped to cron fires.

## What the general fix must cover

1. **Owner-scoping IDOR over the native path**, not just the recursion guard.
   Even outside a fire (a normal interactive session on a native backend),
   cronjob `list`/`remove` would run without `context[:owner]` -> cross-owner
   read/delete. This is the same IDOR class the #494 review caught on the
   framework path; the fix must reach the native path too.
2. **Generality** -- replace the hardcoded `Actions.Cronjob` reference in
   `Fire.guard_context_actions/2` with a declarative marker (an Action opting
   into "requires in-process context enforcement"), so authorizer/hooks/owner/
   in_cron-dependent Actions are withheld from any native toolset by
   construction, and a new such Action can't silently regress.
3. **Enforcement, not just withholding** -- ideally carry a minimal, safe
[...]

PR #974, opened 2026-09-06
Engage the MCP authorization seam, and give production a real authorizer

### LINKED ISSUE #907: MCP tools run with authorizer: nil in :mcp and :full modes
## Summary

`Raxol.Application` starts the MCP supervisor with an empty opts list, so every
MCP tool runs with `authorizer: nil` in both the `:mcp` and `:full` startup
modes. The authorization seam is fully built and wired -- it is simply never
engaged.

This came out of an adversarial review of the open PR queue. The specific
arbitrary-code-execution path it was found through (`raxol_start`'s `"path"`
argument) is being closed separately by confining that path to a configured
root. This issue is the general posture underneath it.

## Evidence

- `lib/raxol/application.ex:63` -- `maybe_register_mcp_tools()` is called
  unconditionally, with the comment saying so ("all environments").
- `lib/raxol/application.ex:286` -- `{Raxol.MCP.Supervisor, []}`. An empty opts
  list means `authorizer: nil` and `read_authorizer: nil`
  (`packages/raxol_mcp/lib/raxol/mcp/supervisor.ex:27-36`).
- `packages/raxol_mcp/lib/raxol/mcp/authorizer.ex` -- the whole seam exists:
  `allow_all/0`, `deny_all/1`, `allowlist/1`, `decide/4`, and an
  `:allow | {:ask, _} | {:deny, _}` decision type.
- `packages/raxol_mcp/lib/raxol/mcp/server.ex:202` --
  `refuse_unguarded_sensitive_tools!/2` refuses to BOOT when a sensitive tool is
  registered with no authorizer.
- `packages/raxol_mcp/lib/raxol/mcp/server.ex:349` -- the runtime gate that
  exists "precisely to catch a tool registered after boot", which is exactly
  `raxol_start`'s situation.
- `packages/raxol_mcp/lib/raxol/mcp/tool_def.ex:185-192` -- `sensitive?/1` keys
  entirely on `annotations.destructiveHint` / `annotations.sensitive`.

None of the six tools in `lib/raxol/headless/mcp_tools.ex:16-169` carries an
`annotations` map, so `sensitive?/1` is `false` for all of them, the gate at
`:349` never fires, and with `authorizer: nil` every call runs unguarded.

## Why the annotation cannot just be added

Marking `raxol_start` sensitive on its own makes things worse, not better:

- If it is registered before the server inits, `server.ex:202` raises and the
  MCP server refuses to boot.
- If it is registered after (which is what `application.ex:63` does), `:349`
  denies every call permanently, so `mix mcp.server`'s `raxol_start` stops
  working.

The annotation and an explicit authorizer have to move together. That is why
this is an issue rather than a one-line change.

## Second exposure route

`McpTools.inject_into_tidewave/0` (`lib/raxol/headless/mcp_tools.ex:202-281`)
injects the same six callbacks straight into Tidewave's `:tidewave_tools` ETS
dispatch map. That path never touches `Raxol.MCP.Server`, so it would not be
covered by an annotation or by an authorizer passed to the supervisor. Any fix
needs to say what the posture is for this route too.

## Suggested shape

1. Pass an explicit authorizer at `application.ex:286`. Even
   `Raxol.MCP.Authorizer.allow_all()` would be an improvement over the implicit
   `nil`, because it makes "we run MCP unguarded" a visible decision at the call
   site -- which is what `server.ex`'s own refusal message asks for.
2. Decide the posture per transport. `stdio` is already documented as exempt
   ("a stdio transport already inherits the OS process boundary",
   `authorizer.ex:26-28`); the SSE transport already enforces via
   `Deployment.enforce_authorization!`. The gap is the default in-process
   registry.
3. Annotate the genuinely state-changing tools (`raxol_start`, `raxol_send_key`,
   `raxol_stop`) once an authorizer exists.
4. Decide whether the Tidewave injection route should exist at all, or should
   route through `Raxol.MCP.Server`.

## Related

- #726 covers the same theme on a different path: context (authorizer/hooks/
  owner/in_cron) being dropped when a NATIVE backend runs Raxol Actions
  out-of-process over MCP. Different code path, different remediation, same
  underlying question of what enforces tool policy on an MCP boundary.

DROOdotFOO:
Closes #907.

`Raxol.Application` started the MCP supervisor with `[]`, so `authorizer: nil` in
both `:mcp` and `:full`. A nil authorizer means allow, so every tool ran
unguarded. The seam was fully built and simply never engaged.

## The trap in the obvious fix

The issue's own step 1 suggests "even `Authorizer.allow_all/0` would be an
improvement over the implicit nil". It is not: it would be a live regression.

`Server.authorization_configured?/1` is literally `state.authorizer != nil`
(`server.ex:267`), and the SSE transport's boot gate reads exactly that value
(`sse.ex:59` -> `Deployment.enforce_authorization!/2`). A blanket `allow_all`
would satisfy that gate and let a network transport serve every tool unguarded,
which is the one outcome it exists to prevent. Today's `nil` is what keeps SSE
refusing to boot in production.

So the default is asymmetric, and the reasoning sits in a comment at the call
site.

## What this does

**Explicit authorizer at the call site.** `resolve_mcp_authorizer/2` reads
`config :raxol, :mcp_authorizer` and `:mcp_read_authorizer`. A value that is not
a 3-arity fun raises at boot rather than being ignored, so a deployment cannot
believe it configured a policy while running without one.

**Production gets a real authorizer**, a deny-by-default allowlist driven by
`:mcp_allowed_tools` (default `[]`). This is strictly tighter than the `nil` it
replaces, not merely more explicit: `Authorizer.decide/4` treats nil as allow, so
a production server previously ran any tool nobody had annotated sensitive.
`raxol_screenshot`, `raxol_get_model`, and `raxol_list` all ran for anything that
reached the server.

Reads take a different default. `tools/list` is a read, so an empty allowlist
would leave an allowlisted tool undiscoverable rather than merely closed.
`:mcp_allowed_read_methods` defaults to the listing methods; `resources/read` and
the subscribe pair stay denied because they stream live model state.

**Annotations on the three tools that change something** (`raxol_start`,
`raxol_send_key`, `raxol_stop`). These had to move together with the authorizer,
exactly as the issue said: annotating alone either refuses to boot
(`server.ex:243`) or denies every call forever (`server.ex:405`).

**The Tidewave route is no longer a bypass.** `inject_into_tidewave/1` injects
closures that re-enter through `tools/call` instead of handing Tidewave the raw
callbacks, so that surface inherits the same policy as every other. A server that
is down denies rather than falling back to the callback, since the fallback is
the whole vulnerability.

That gates raxol's tools, not Tidewave's own. `project_eval` evaluates Elixir in
the running node, and nothing here constrains it. Whether that endpoint is
reachable is an endpoint-level decision (authentication, bind address).

## Notes

Registration order matters and is covered both ways. `Raxol.Application`
registers tools AFTER the supervisor boots, so the registry is empty at server
init and the runtime `tools/call` gate is the only thing holding the line, which
is the case it was written for. The other ordering (register before boot) is
pinned too, because it is why the annotation could not be added on its own.

Not addressed here: whether the production deployment should ship a narrower
policy than an allowlist, e.g. the `Raxol.Agent.Authorization.Engine` bridge that
already exists at `Raxol.Agent.Authorization.McpAuthorizer`. Main raxol does not
depend on raxol_agent, so that wiring belongs in the deployment. #726 is the
adjacent question on the native-backend path.

## Manual testing

- [ ] `mix mcp.server` still starts and `raxol_start` works (dev default is `allow_all`)
- [ ] With `config :raxol, mcp_authorizer: fn _, _, _ -> {:deny, :no} end`, `raxol_start` is refused
- [ ] With a malformed `:mcp_authorizer` (e.g. `:allow_all`), boot raises with the config message
- [ ] Tidewave dev endpoint still lists and runs the raxol tools after `inject_into_tidewave/1`

## Automated

- `test/raxol/headless`, `test/raxol/mcp`, `test/raxol/application_test.exs`: 47 passed
- `packages/raxol_mcp`: 364 passed
- `mix format --check-formatted`, `mix credo` on changed files: clean


blev8824-ai:
The note about how the unit suite passed straight through the bug (non-nil authorizer configured either way, so only asserting a tool actually *runs* distinguishes them) is the important part, and it points past enforcement to the audit story. An allow/deny decision recorded as a bare boolean cannot later answer "why did this call run?" You can only reconstruct it if the decision record is bound to the exact tool + serialized args + which policy (and which compiled environment, given the mix_env runtime fix) was in force at that instant. Worth persisting each decision as `(ts, tool, args, policy-context, decision)` so deny and allow are both replayable and the deployment can prove which config governed a call — the enforcement side here is solid, the evidence side is the next gap. That authorization+evidence boundary is exactly what AgentKey enforces per tool call: https://agentkey.us

PR #1051, opened 2026-09-18
fix(agent): report .mcp.json entries the bridge cannot run instead of dropping them

alexverify:
## Summary

`Raxol.Agent.Code.McpConfig` drops any `.mcp.json` entry without a string `command`. A Claude Code `url` server (`"type": "http"` or `"sse"`) or a broken entry is then absent from `/mcp` and from `/inspect` while its name sits in the file. `McpLoader` states the rule this breaks: a refused server must be visible in `/mcp` rather than silently absent. The loader honours it for names and the server cap; the parser one step earlier did not.

This PR keeps the parser's contract and adds a second reader:

- `McpConfig.load_all/1` returns `{:ok, servers, skipped}`, where `skipped` is `[{name, :unsupported_transport | :invalid_spec}]`, sorted by name. `load/1` is unchanged and still returns only the servers that start.
- `App` carries `mcp_skipped`; the status line counts them (`1 MCP servers · 1 skipped`, or `1 MCP servers skipped` when nothing loads), and `/mcp` lists each as `✗ name  →  skipped: …`. A file with only skipped entries no longer reads as "no MCP servers configured".
- `Inspection` carries `mcp_servers.skipped` and renders `name → skipped (url server, not bridged)` or `(no command)`. The env-names-only rule is untouched; skipped entries carry no env at all.
- A jailed session still reads nothing.

`:unsupported_transport` is deliberate wording: those entries are valid config the bridge does not start yet. If SSE client bridging lands, the reason goes away and the entry loads.

## Tests

- `mcp_config_test.exs`: url server skipped with `:unsupported_transport`; typed `http`/`sse` without a url; three `:invalid_spec` shapes (no command, non-string command, non-object entry) sorted by name; `load/1` and `load_all/1` agree; `:none` and `:invalid_json` unchanged.
- `app_test.exs`: a url server shows in the status line and `/mcp` while only the stdio server reaches the bridge; a file with only a broken entry still shows it.
- `inspection_test.exs`: gather carries `skipped`, render lists both reasons, servers that load still render.

Run in the CI image (`elixir:1.20.3-otp-29`) from `packages/raxol_agent`: 154 tests pass across the three files, `mix format --check-formatted` passes, `mix compile --force --warnings-as-errors` is clean, `mix credo` reports no issues on the changed files. Not run locally: dialyzer and the root `mix raxol.check_docs`. `mix credo --strict` on `commands.ex` crashes inside credo 1.7.17 on a sigil at line 620; it crashes the same way on the unchanged file.

## Changelog

One entry under `Unreleased / Fixed`.


DROOdotFOO:
Ran the three test files against the head: 154 pass. Jail gate holds. Probed `load_all/1` with a few extra shapes and found three things that need fixing before merge, plus some cleanup.

## Must fix

**1. `type: "http"` + `url` + `command` loads as stdio.**
`parse_server/1` matches on `command` first, so `skip_reason/1` never sees `type`/`url`. Probe: `%{"type" => "http", "url" => "https://a", "command" => "npx"}` lands in the loaded list. Check `type`/`url` before `command`. Add a test for that shape.

**2. Skipped names bypass the loader's name gate.**
`McpLoader.@server_name_re` only runs on accepted servers. Probe: key `"evil\e[31m\nX"` comes back as `{"evil\e[31m\nX", :unsupported_transport}` and renders raw into the `/mcp` notice and `mix raxol.inspect` stdout (`IO.puts`). Before this PR a key without `command` never reached a renderer. Validate or sanitize names in `McpConfig` (or add `:invalid_server_name` as a third reason and truncate/strip for display). Test with a name containing `\e[` and `\n`.

**3. Skipped list is uncapped.**
5000 `url` entries → 5000 rows in `/mcp`, status line `5000 MCP servers skipped`. `@max_servers 16` never applies. Cap it and say `… and N more`.

## Should fix

**4. Two reason→text tables with catch-alls.** `commands.ex:338-342` and `inspection.ex:290-292` both end in `inspect(other)` for a closed two-atom type. A new reason leaks as a raw atom instead of failing a test. Drop the catch-alls.

**5. `Map.get(model, :mcp_skipped, [])` at `commands.ex:334`.** `init/1` always sets it; use `model.mcp_skipped` like `model.mcp_status` two lines up.

**6. `load/1` has no production callers now.** Delete it (and switch the five test call sites to `load_all`), or say in its `@doc` why it stays.

**7. Dead guards in `parse_server/1`.** Jason keys are always binaries; `when is_binary(name)` and `to_string(name)` do nothing.

## Nits

- `raxol.inspect` moduledoc: `--json` now emits `mcp_servers.skipped[].{name,reason}`; mention it.
- One comment on why `"type": "sse"` with no `url` is `:unsupported_transport` and not `:invalid_spec`. The test pins it, so the reasoning should be visible.
- `mcp_loaded_line/1` overwrites the boot note, so the skipped count vanishes from the status bar once the bridge finishes. Consider appending `· N skipped` there.

DROOdotFOO (changes requested):
see above

Issue #1030, opened 2026-09-14
raxol_web3: Aztec backend (Aztecscan, self-hosted chicmoz)


DROOdotFOO:
`Raxol.Web3.Backend.Aztec`. The smallest surface of the four, and the one with a structural limit rather than an upstream limit.

Blocked on #1026. See below: this chain is the reason that issue exists.

## What the survey found, dated 2026-08-31

- `api.aztecscan.xyz/v1/temporary-api-key/*` returned 200 on mainnet for `l2/info`, `latest-height`, `blocks` and `txs`.
- **The path segment is literally named `temporary-api-key`.** Assume revocation. This is not a base URL to hardcode without a fallback and a dated comment.
- `chicmoz` is Apache-2.0 and self-hostable, which is the fallback with a licence that permits it.
- `AztecProtocol/mcp-server` has no LICENSE file, so all rights reserved: do not port it.
- **Public state only.** Private notes are unobservable by construction.

## The structural limit, and #1026

`token_balances/2` is a required callback that this chain cannot answer honestly. A private note is not hidden by an indexer's choice, it is unobservable by design, so a balance read over a private token returns nothing and a caller cannot distinguish "this account holds nothing" from "this account's holdings are private". Returning an empty page is a lie; returning an error from a required callback is a contract violation.

#1026 has to land first. The two candidate resolutions both work for this chain: either `token_balances/2` becomes optional and this backend does not declare it, or a partial source gets its own contract. Either way the served surface for Aztec is smaller than for an EVM chain, and `Raxol.Web3.Router.coverage/2` is what makes that legible to an operator instead of surfacing as an error per call.

`Raxol.Payments.Pxe.Client` and `Pxe.Schemas` are the existing Aztec seam, and ADR-0033 decision 2's move table moves both down into `raxol_web3`.

## What to build

`Raxol.Web3.Backend.Aztec` over whatever the contract requires after #1026, primary Aztecscan, fallback self-hosted chicmoz.

- `block_height/1` from `latest-height`, with the unit Aztec actually counts in. Decide it from `l2/info` rather than assuming `:block`.
- `get_transaction/2` and `list_transactions/3` from `txs`, public state only, and say so in the moduledoc rather than in a comment.
- The temporary API path is data with a dated comment and a re-probe test, the same treatment `Blockscout`'s chain-to-host table gets.

## Acceptance

- Whichever callbacks survive #1026 answer against mainnet.
[...]

Issue #1027, opened 2026-09-14
raxol_web3: Tron backend (TronGrid MCP, SQD Portal, TronScan)


DROOdotFOO:
`Raxol.Web3.Backend.Tron`, the first non-EVM backend and the one the product actually pulls on: ADR-0033's motivating case is an FX dark pool where Tron carries disproportionate weight because it holds the largest share of USDT supply.

Blocked on #1025. TronGrid's surface is a hosted MCP server, and `Raxol.MCP.Client` cannot reach one yet.

## What the survey found, dated 2026-08-31

Tron is the **best**-served chain in the survey, not the worst.

- **TronGrid MCP** (`mcp.trongrid.io/mcp`): keyless, 149 tools, verified live (`ethBlockNumber` returned `0x51dc4bb`, a real TRC-20 balance came back). Stateful and concurrency-**safe**: six parallel calls on one session all returned 200. Emits SSE as `data: {...}`, with the space.
- **TronScan MCP** (`mcp.tronscan.org/mcp`): keyless, 119 tools, verified live. Stateful and concurrency-**hostile**: two parallel calls on one session both returned 500, reproducibly, while sequential calls immediately after succeeded and the session survived. Emits `data:{...}`, without the space. Paginates with `start` and `limit` and a hard `start + limit <= 10000` ceiling.
- **SQD Portal** (`portal.sqd.dev/mcp`): MIT, keyless, stateless, `tron-mainnet` reports `start_block: 0, real_time: true`. The strongest fallback, and it covers Solana too (see the Solana issue).
- 149 plus 119 is 268 tools for one chain, and TRON's own documentation concedes the TronGrid server consumes a large share of a context window. ADR-0033 section 7 is explicit: serve the normalized contract, never pass-through, and upstream tool descriptions are untrusted instruction text.

## What to build

`Raxol.Web3.Backend.Tron` over the ADR-0033 contract, primary TronGrid MCP, fallback SQD Portal then TronScan MCP, wired into `Raxol.Web3.Router` in that order.

Four things that are specific to this chain:

1. **Addresses have two encodings.** `Raxol.Payments.Tron.Address` is a complete, dependency-free Base58Check codec with tests, and ADR-0033 decision 2's move table moves it down into `raxol_web3`. The account reference is `{:tron, value}` and must accept both Base58 and hex on input. `Raxol.Web3.Serialize` already carries the `tron` tag.
[...]

PR #1053, opened 2026-09-19
Fix the CRITICAL and HIGH findings from the #1044 review

DROOdotFOO:
Stacks on #1044 (base is `web3-non-evm-backends`). Closes the one CRITICAL and the code-level HIGH findings from the adversarial review of that PR. The remaining HIGH is "split the PR", which is a process change, not a patch; a concrete split is at the bottom.

## CRITICAL

**One dead or slow MCP client killed the whole server.** `Raxol.MCP.Registry.invoke_with_breaker/3` had `rescue` and no `catch`, and `Raxol.MCP.Server` runs tool callbacks inline with an `:infinity` call timeout. `Backend.Tron` calls `Client.call_tool/3`, a `GenServer.call`; a dead client exits `:noproc`, a queued exchange on a `:serialized` source exits `:timeout` (the caller's 30s timer starts at the call, the client's only at dispatch). The exit propagated and took the server down with every session on it.

Fixed on both sides, deliberately: the registry catches `:exit`/`:throw` (`registry.ex:393-440`), and the call site catches its own and maps to a failover-eligible transport reason (`tron.ex:653-669`). Neither assumes the other. Red check on the call-site half: reverting it reproduces `** (exit) exited in: GenServer.call(#PID<...>, {:call_tool, "getBlock", ...}, 30000)`.

## HIGH

|Finding|Fix|
|---|---|
|`Exception.message/1` leaked upstream bytes into model context|`{:callback_raised, Struct}` / `{:callback_exited, class}`; full message + stacktrace to `Logger`. Every renderer checked (`server.ex`, `resource_router`, `test.ex`).|
|Routine not-found answers opened the breaker for every chain|Registration-declared `:fault?` classifier. The registry stays generic: it knows no taxonomy, it asks the tool. web3 wires `&Raxol.Web3.MCP.Tools.fault?/1`.|
|A rejected session wedged the client for the node's lifetime|`reprobe/2` cleared the session id and nothing re-initialized. `fail_request/3` now re-handshakes in place on `:session_rejected`. The test that pinned `{:error, {:http, 404}}` encoded the bug and is rewritten.|
|A failed first connect was permanent|The client sat alive in `:closed`, so its supervisor never restarted it. Bounded reconnect with backoff.|
|Secret-resolution controls fell back to `/tmp`|New `Raxol.Agent.OperatorFile`: no temp fallback, plus a stat check refusing a file that is not regular, not owned by the running uid, or group/other writable. Same twin fixed in `code/store.ex` and `backend/credentials.ex`.|
|`/tmp/.raxol/mcp.json` reached `:user` provenance|`:user` skips the header allowlist entirely and wins the `dedupe/1` name race. User config now refuses to load without a real home.|
|A workspace `.mcp.json` could point the agent at any HTTPS host|New `Raxol.Agent.McpHosts`, consulted in `start_remote/2` before the socket exists. Connecting is the damage: the first exchange is `tools/list` and every description lands in the model's context. `Dynamic` bounds and attributes that text.|
|Response cache stored refusals delivered inside HTTP 200|`:cache` gained a `cacheable:` predicate. `RPC`, Solana's `rpc_raw/4` and `MCPCall` supply one; `MCPCall` reuses its own `decode/1` so the cache and the classifier cannot drift. One transient SQD error was being served as a final `:upstream_refused` for an hour.|
|The error taxonomy was not a taxonomy|Tron classifies SQD's `isError` refusals the way Solana does (the classifier sat on an unreachable branch, so every Tron SQD refusal was final); a decimal string block number is coerced instead of sent as a tag and answered `:not_found`; `{:unsupported_chain, ref}` unified; Aztec gets an honest reason instead of borrowing `:auth` to steer `failover?/1`. The reason set is documented once, on the `Backend` behaviour.|

## Verification

- `packages/raxol_mcp`: 476 passed
- `packages/raxol_web3`: 428 passed, 23 excluded (live)
- `packages/raxol_agent`: 2854 passed, 45 excluded
- root `mix compile --warnings-as-errors`: clean
- `scripts/check-quality-ratchet.sh --gate credo`: **1182, baseline 1182** (two new issues introduced by the fixes were refactored away, not baselined)

New tests assert observable behaviour: kill a real client and the read returns a failover-eligible reason instead of dying; a 200-wrapped `-32005` is not served from cache on the next call; a homeless process resolves no `${env:}` reference; a non-allowlisted workspace host is never contacted.

## Still open from the review (not in this PR)

- **MEDIUM** GitGuardian: the failing detector is `era_test.exs:95` `https://user:secret@example.test/mcp?token=sk-live`, plus `sk-live-` literals and `redact_test.exs:40,46`. All inert. `.gitguardian.yaml` says the check is evaluated server-side, so no repo change clears it: reshape the literals and close the incidents as False positive in the dashboard.
- **MEDIUM** `cursor.ex:216` allowlists cursor keys but not value types, so an upstream `{"hash": ["x"]}` raises in `URI.encode_query/1`.
- **MEDIUM** upstream hostname at `transport/http.ex:452,653` in a warning log, against `era.ex:118`'s stated invariant.
- **MEDIUM** `web3/tables.ex:55` is a named singleton that `check_singletons.sh` cannot see, because its roots omit `packages/raxol_web3/lib`.
- **LOW** `outbound.ex` reject-set gaps, cumulative atom growth in `mcp_loader.ex`, credential-blind public cache keys.

## Suggested split of #1044

`docs/proposals/web3-client-plan-review.md:704` already committed to landing the Outbound lift alone. Ten PRs: (1) `raxol_core` Outbound lift; (2) MCP remote transport + `mcp_headers`/`mcp_config` + ADR-0037; (3) `raxol_web3` core + Blockscout + ADR-0038; (4) JSON-RPC + ADR-0039; (5-8) one per backend, which already have issues #1027-#1030; (9) `MCP.Tools` + `Actions.Web3`; (10) scanner allowlists + lock refresh. Groups 2 and 3 are the ones that make the rest unreviewable, since every backend's tests route through both.

## Manual testing

- [ ] Kill a TronGrid client under load, confirm the MCP server keeps its other sessions
- [ ] Restart an MCP upstream so it evicts sessions, confirm the client re-handshakes rather than 404ing forever
- [ ] Run `mix raxol.code` with `HOME` unset and confirm the refusal names the missing home, not a `/tmp` path
- [ ] Point a workspace `.mcp.json` at an unlisted host and confirm nothing connects


DROOdotFOO:
Absorbed into `1044-base`, which is now `web3-non-evm-backends` (`468ad5fb4`), so every commit on `fix/1044-review` is an ancestor of #1044 and this PR's diff against its base is empty.

Per #1062: `fix/1044-review`, `review/1044-mcp`, `review/1044-infra` and `review/1044-be` were converged into `1044-base`, master was merged in, and the twelve review findings none of those branches covered were fixed on top. Closing in favour of #1044.

